PR #4, opened 2020-03-02
Create a flight recorder exporter to compare

Mpdreamz:


Mpdreamz:
tagging @elastic/es-clients this adds an exporter to the flight recorder syntax. Useful to compare the outputs of both programs and as a starting point once we get to start the shadow spec in earnest.

PR #7, opened 2020-04-30
Improve ts output

delvedor:
This pr improves the typescript output to adapt to the js client, it also introduces the `RequestInterface` to better handle path, query, and body later on during the code generation.

I've also updated the linter configuration.


specification/src/domain.ts
```diff
@@ -4,8 +4,11 @@ import _ from 'lodash'
 namespace Domain {
 
   export class Type {
-    constructor (public name: string) {}
-    nullable: boolean;
+    name: string
+    nullable: boolean
```
Mpdreamz, line 8: Is this `nullable` still used?

delvedor: Nope, it's a leftover :)

PR #13, opened 2020-06-30
Update request specs

Mpdreamz:


Mpdreamz:
reviewed this big change with @delvedor over zoom. 

This introduces `RequestInterface` as specialization for `Interface` so we can more distincly document query parameters and expose them on the generator side as well. 

This also allows us to document body's that do not fit the body == requestbase model that NEST follows.

PR #21, opened 2020-07-28
Iterating on the spec.

delvedor:
As titled.

Types tested:
- index
  - [x] request
  - [x] response
- create
  - [x] request
  - [x] response
- delete
  - [x] request
  - [x] response 
- update
  - [x] request (~~the `_source` query parameter does not exist in the input spec~~)
  - [x] response
- get
  - [x] request (~~the `_source` query parameter does not exist in the input spec and the Field type is not correct~~)
  - [x] response 
- bulk
  - [x] request (~~the type def must be rewritten?~~)
  - [ ] response (~~the type def must be rewritten?~~ `respone.items.update.get` does not exist)
- search
  - [ ] request (E_TOO_MANY_ERRORS)
  - [ ] response (E_TOO_MANY_ERRORS)

Notes:
- error responses are not passing (it's expected)


typescript-generator/index.ts
```diff
@@ -101,61 +32,160 @@ const stringTypes = [
...
+    //       otherwise is just a repetition of the type?
+    if (Array.isArray(type.closedGenerics) && type.closedGenerics.length > 1) {
+      return `${type.type.name}<${type.closedGenerics.map(unwrapType).join(', ')}>`
+    }
+    return type.type.name
+  } else {
+    if (Array.isArray(type.closedGenerics) && type.closedGenerics.length > 0) {
+      return `${type.name}<${type.closedGenerics.map(unwrapType).join(', ')}>`
+    }
+    return type.name
```
delvedor, line 165: @Mpdreamz I might need some help here :)

delvedor:
Synced with @Mpdreamz, let's merged as it as and finish the search endpoint in a subsequent pr.

PR #23, opened 2020-08-06
Iterating on the spec (part 2)

delvedor:
Continuation of #21 

Types tested:
- index
  - [x] request
  - [x] response
- create
  - [x] request
  - [x] response
- delete
  - [x] request
  - [x] response 
- update
  - [x] request
  - [x] response
- get
  - [x] request
  - [x] response 
- bulk
  - [x] request
  - [ ] response (`respone.items.update.get` does not exist)
- search
  - [ ] request (working on query dsl and aggregations)
  - [ ] response (E_TOO_MANY_ERRORS)

PR #29, opened 2020-09-08
Iterating on the spec (part 3)

delvedor:
Continuation of #23

Types tested:
- index
  - [x] request
  - [x] response
- create
  - [x] request
  - [x] response
- delete
  - [x] request
  - [x] response 
- update
  - [x] request
  - [x] response
- get
  - [x] request
  - [x] response 
- bulk
  - [x] request
  - [ ] response (`respone.items.update.get` does not exist)
- search
  - [ ] request (almost there!)
  - [ ] response (E_TOO_MANY_ERRORS)


specification/specs/query_dsl/compound/dismax/DisMaxQuery.ts
```diff
@@ -1,4 +1,5 @@
 class DisMaxQuery {
 	queries: QueryContainer[];
 	tie_breaker: double;
+  boost: float;
```
Mpdreamz, line 4: `tab` (do we have `.editorconfig` configured correctly?)

specification/specs/query_dsl/full_text/multi_match/MultiMatchQuery.ts
```diff
@@ -17,4 +17,5 @@ class MultiMatchQuery {
 	type: TextQueryType;
 	use_dis_max: boolean;
 	zero_terms_query: ZeroTermsQuery;
+  boost: float;
```
Mpdreamz, line 20: Do we want to create a base query with `_name` and `boost`?

specification/specs/query_dsl/compound/dismax/DisMaxQuery.ts
delvedor: We do have. it, but the input spec is in the wrong format.
I'll open a separate pr to address this.

Mpdreamz (approved):

specification/specs/query_dsl/Operator.ts
```diff
@@ -1,4 +1,6 @@
 enum Operator {
 	and = 0,
-	or = 1
+	or = 1,
+  AND = 2,
```
Mpdreamz, line 4: Do we need both casings? I think only caps suffices?

delvedor: Both are being used in the flight recorder :S

PR #32, opened 2020-09-14
Update ts output generator

delvedor:
- Remove local install
- Drop enums from output in favor of union types


Mpdreamz (approved):
LGTM 👍

PR #35, opened 2020-09-28
Simplify SearchRequest's sort and _source field

swallez:
Simplification of `sort`: the `string` variant does not exist in the request body, but only as a query parameter

Simplification of `source`:
- rather than considering that there only exists a 2-variant `Union` (aka `Either`), we should let code generators handle unions of arbitrary length (we can have `Union2`, `Union3`, `Union4`, etc in languages missing proper tagged unions).
- Usage of `Fields` instead of `Field | Field[]` may help generators understand that these two variants are related and provide more ergonomic builder methods (although this could also be inferred from the `TypeA | TypeA[]` pattern).

Additional questions:
- `sort` and `_source` are also defined as query parameters in `search.json` (the Rest API spec) but only exist as a body parameters in `SearchRequest.ts`. Is this a general policy to only keep parameters in the body when they exist as both query and body parameters?

  If we do this, this means we close the door to using the TypeScript specs to generate server-side code for existing APIs (code modernization). So maybe we should keep them with a `@server-only` annotation so that they are ignored when generating clients but not forgotten when we want to use them to produce server code?

- if we remove `_source` as a query parameter we should also remove `_source_excludes` and `_source_includes` that go hand in hand with it (and are expressed in the `_source` body parameter).

### DIFF
--- a/specification/specs/common.ts
+++ b/specification/specs/common.ts
@@ -52,7 +52,8 @@ type ScrollIds = string
 type CategoryId = string
 type ActionIds = string
 type Field = string
-type Fields = string
+/** Wildcard (*) pattern or array of patterns containing source fields to return */
+type Fields = Field | Field[]
 type Id = string | number
 type Ids = string | number | string[]
 type IndexName = string
--- a/specification/specs/common.ts
+++ b/specification/specs/common.ts
@@ -52,7 +52,7 @@ type ScrollIds = string
 type CategoryId = string
 type ActionIds = string
 type Field = string
-/** Wildcard (*) pattern or array of patterns containing source fields to return */
+/** Path to field or array of paths. Some API's support wildcards in the path to select multiple fields.  */
 type Fields = Field | Field[]
 type Id = string | number
 type Ids = string | number | string[]
--- a/specification/specs/search/search/SearchRequest.ts
+++ b/specification/specs/search/search/SearchRequest.ts
@@ -62,8 +62,8 @@ class SearchRequest extends RequestBase {
     search_after: Array<integer | string>;
     size: integer;
     slice: SlicedScroll;
-    sort: Array<SingleKeyDictionary<Sort | SortOrder> | string>;
-    _source: Union<Union<boolean, Field>, Union<Field[], SourceFilter>>;
+    sort: SingleKeyDictionary<Sort | SortOrder>[];
+    _source: boolean | Fields | SourceFilter;
     suggest: Dictionary<string, SuggestBucket>;
     terminate_after: long;
     timeout: string;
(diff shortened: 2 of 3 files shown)


Mpdreamz:
> rather than considering that there only exists a 2-variant Union

This was never a conscious restriction. We also recently updated the brain so we can use `typescript piped union` syntax instead of `Union<>`. Unions should have no limit although if we see unions too big we need to question the API 😄 . In the long run we should have no notion of `Union<>` in the input spec at all. 

> Usage of Fields instead of Field | Field[] may h

++ This is why `Fields` exists! Maybe we can open a ticket to fail the brain if we see this pattern instead of using `Fields`. 

>  Is this a general policy to only keep parameters in the body when they exist as both query and body parameters?

Again not a conscious decision but more a by-product of it being an export of the .NET client. Where we do sometime consciously steer the users to the body only. 

> If we do this, this means we close the door to using the TypeScript specs to generate server-side code for existing APIs (code modernization).

In the spec we **always** need to encode both where applicable. I don't think we need annotations. Let generators decide if they only want to expose one. For instance with .NET when it moves to this generator I no longer want to maintain this special edge cases and just generate both.

> if we remove _source as a query parameter 

I don't think we need to remove it. We might want to annotate it somehow that including it might break types. E.g low level clients want and **should** include `_source` but typed high level clients might not want to.

specification/specs/common.ts
```diff
@@ -52,7 +52,8 @@ type ScrollIds = string
 type CategoryId = string
 type ActionIds = string
 type Field = string
-type Fields = string
+/** Wildcard (*) pattern or array of patterns containing source fields to return */
```
Mpdreamz, line 55: I would reword this:
```suggestion
/** Path to field or array of paths. Some API's support wildcards in the path to select multiple fields.  */
```

Mpdreamz (approved):
LGTM

swallez:
> In the spec we always need to encode both where applicable. I don't think we need annotations. Let generators decide if they only want to expose one. For instance with .NET when it moves to this generator I no longer want to maintain this special edge cases and just generate both.

Indeed we can ignore query parameters that have the same name as a body property. There are some edge cases though where there isn't a 1:1 mapping such as with `_source_excludes` and `_source_includes`. We could infer that from the `_source_` prefix, but I don't think this is a robust rule. In that case (and that case only, which should be infrequent) annotations may be useful.

Side note: JSON schema has (draft) support for [conditional sub-schemas](https://json-schema.org/understanding-json-schema/reference/conditionals.html). I really don't think we'll need that level of flexibility, but it's interesting to see how schema languages address this problem. Now this is a slippery slope that leads config languages to become Turing-complete 😛

PR #38, opened 2020-11-25
Add new canonical metamodel and converter from current specification

swallez:
This PR adds a new canonical metamodel for API specifications, and the tools to convert it from the existing specification.

Usage:
```
cd ./specification
npm run generate-schema
...
Schema generated in ../output/schema/schema.json
```

This reads the TypeScript spec definition in `spec` and writes its JSON representation.

Note that not all types are output as the converter starts from the endpoints and recursively traverses type definitions from the endpoint's request and response definitions. There are some type declarations in `spec` that are not in this graph, and they're left out of the generated schema.

Benefits of this new model:
- no redundancy: every piece of information only exists once
- simplified data model: some data structures were flattened, some types removed as they can be expressed by means of other types (e.g. string and number aliases)
- can be serialized as JSON! This means code generators don't have to be written in TypeScript.

Main differences with the existing specification model:
- `InstanceOf` and `ValueOf` have their names swapped to better match the actual semantics of "instance" (instance of a particular type) and "value" (some value that can have many different forms).
- `SingleKeyDictionary` is renamed to `NamedValue` because it's really a `(string, <value>)` tuple and not a dictionary with arbitrary keys constrained to contain only a single entry.
- `StringAlias` and `NumberAlias` are merged within `TypeAlias` which can also be used for arbitrary type aliases.
- `UnionAlias` is renamed to `Union`.
- `UserDefinedValue` is replaced by the internal `object` type, like what is generated for `any`. I'm not sure that we really need `UserDefinedValue` in addition to `any` and generic request/response parameters. If needed, it would be better represented as an additional internal type rather than a first-class metamodel construct.
- descriptions found in the Rest API spec are moved to type and property definitions.


Mpdreamz (approved):
LGTM, `Domain.UserDefinedValue` allows us to annotate in the spec where a request accepts a user defined piece of data and where responses can return these. 


We can't always make this a generic sadly and `object` loses some sementics around how to expose it.

E.g the .NET client uses these points to attach a different serialization routine. E.g our users can inject a serializer specific for these parts (+ TDocument, TVAlue generics) without affecting the overal json serialization of requests/responses. 

We've dubbed this technique donut-hole serialization after donut-hole caching.

swallez:
@Mpdreamz can you point me to examples in the .Net client?

I understand that this is for situations where the user-defined type is in deeply nested structures where bubbling the generic parameter up to the top-level type would be impractical? Or in a dynamic context such as collections where each item can accept a different user-defined type?

Why do we lose semantics compared to `object` (or `any` in TS)? Can't we just say that any field of type `object` should be generated in a form that allows the user to provide either an object and the serializer/deserializer or just raw JSON?

Mpdreamz:
> understand that this is for situations where the user-defined type is in deeply nested structures where bubbling the generic parameter up to the top-level type would be impractical? Or in a dynamic context such as collections where each item can accept a different user-defined type?

Both



> Why do we lose semantics compared to object (or any in TS)? Can't we just say that any field of type object should be generated in a form that allows the user to provide either an object and the serializer/deserializer or just raw JSON?


Sometimes a string is just a string and object is just an object but I think the spec benefits from type aliases to both in most cases. 

I don't want folks reaching for object/any and call it job done.

swallez:
> I don't want folks reaching for object/any and call it job done.

Ok, so the semantics is more about psychology, like "think twice before writing `UserDefinedValue`" compared to just throwing in `any` when you don't want to put more thinking in a non trivial situation, right?

So I'm ok with reintroducing `UserDefinedValue` but then we should forbid `any` in the specs. WDYT?

Mpdreamz:
I'm down with that :) 

Forcing type aliases also promotes reuse of concepts and the types can be documented. 

A property is never **any**thing it's always **some**thing :badadoom:

swallez:
Great, I added back `UserDefinedValue`!

I'd love @delvedor's opinion too 😉

swallez:
I added a commit to rename `_kind` to `kind`. I initially used a `_` prefix to denote a technical attribute since it allows disambiguating the type, but as an interchange format it becomes a first-class citizen of the semantics. Plus the `_` prefix is often used to indicate private fields, which isn't the case.

swallez:
I would like to change the request body definition to something more explicit like this:

```typescript
export class Request extends BaseType {
  // ...
  body?: ValueBody | PropertiesBody;
}

export class ValueBody {
  kind: "value";
  value: ValueOf;
}

export class PropertiesBody {
  kind: "properties";
  properties: Property[];
}
```

When implementing the model in Java, I found out that the current definition as `body?: ValueOf | Property[]` makes serialization and deserialization a bit complex. The Jackson library can distinguish variants based on a property value, but not from the "shape" of the JSON (an object or array) and [I had to write custom (de)serializers for this](https://github.com/elastic/elasticsearch-java/blob/44fd3113550c07a052b8be73e2e7d76d0f3e02a4/generator/src/main/java/co/elastic/clients/elasticsearch/generator/model/TypeDefinition.java#L181-L212).

The same problem will happen in Rust and probably other strongly typed languages.

@delvedor are you ok with this?

swallez:
I've rebased the PR on `master` and updated the body definition.

sethmlarson:
What are everyone's thoughts on having a generated instance of the spec living in the repository rather than only housing the tools to create the JSON spec?

I know it's typically bad form to store artifacts that are generate-able, however if the end goal of this spec is to be taken by many different languages it'd be simpler to grab the spec at a certain URL/path and use rather than running the generator each on our own for each spec update.

To make sure the spec stays up to date could add a unit test that ensures that the committed spec matches what would be generated by the generator.

swallez:
@sethmlarson +1, and that was actually my intention. I did not do it while the format definition was still in flux, but will add it before merging the PR.

+1 for the unit test, and we can also have a commit hook that prevents forgetting to update the generated JSON version.

Having the spec as JSON is indeed important for tools in other repos that use it as a starting point, and can then use the same approach as what we do with the ES Rest specs, i.e. download a specific tag/branch/sha1 from the ES git repo without having to include the translation tool in their own build system.

delvedor (approved):
LGTM

PR #42, opened 2020-12-02
API fixes

delvedor:
- Remove ICatRecord
- Fix bulk apis
- Fix index request

### DIFF
--- a/specification/specs/cat/ICatRecord.ts
+++ b/specification/specs/cat/ICatRecord.ts
@@ -1,2 +1,0 @@
-interface ICatRecord {
-}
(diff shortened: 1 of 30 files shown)


swallez (approved):
LGTM! Let's remove abstractions that are not useful to keep the specs lean and clean!

Mpdreamz (approved):
LGTM, 

`ICatRecord` is not used currently but could serve as a constraint on `TRecord`

```typescript
class CatResponse<TCatRecord> extends ResponseBase {
  records: TCatRecord[];
}
```

Something we currently do not support. So I'm okay to remove now and add back in when the need arrises.

PR #50, opened 2020-12-22
Support optional properties

delvedor:
This is the first pr of a big effort to track optional and required parameters.
Currently, we are marking everything as optional, which is not a sane default, as very likely the response objects will have the described keys.
This pr reverses the approach and marks everything as required unless specified in the spec with the optional syntax:
```ts
class Example {
  thisIsRequired: string
  thisIsNotRequired?: string
}
```

In this pr all the `*Request.ts` files have been converted to optional, we'll fix them, were appropriate while working on the spec.
I'm fairly sure that this change will cause some issues to the Aggregation types, where the "everything required" approach could backfire. But we should start somewhere :D

I discovered which is the field to understand if a property is required with [this tool](https://ts-ast-viewer.com/), I found it very useful, so I'm sharing it here.


specification/src/specification/type-reader.ts
```diff
@@ -163,7 +163,7 @@ class InterfaceVisitor extends Visitor {
     const name = this.symbolName(p.name)
     const returnType = this.visitTypeNode(p.type)
 
-    const prop = new Domain.InterfaceProperty(name, returnType, isQueryParam)
+    const prop = new Domain.InterfaceProperty(name, returnType, isQueryParam, !!p.questionToken)
```
delvedor, line 166: If you are curious, the magic happens here.

delvedor:
> I'm fairly sure that this change will cause some issues to the Aggregation types, where the "everything required" approach could backfire.

See [`8f52f4d` (#50)](https://github.com/elastic/elastic-client-generator/pull/50/commits/8f52f4dd32db416125e465b3692be3951e253a99)

swallez (approved):
LGTM.

I did not check every individual change, but I approve the direction: we should be able to read the TypeScript spec "as is" without having to know about externally-defined conventions such as "all request parameters are optional by default" (until they aren't anymore and we have no way to express it 😉)

Also we shoud clearly distinguish nullable and optional, as null is a value.

PR #57, opened 2021-01-13
Update specification to include optional on responses

Mpdreamz:
- update optional properties in create response
- fix bulk, index, update and delete

### DIFF
--- a/typescript-generator/index.js
+++ b/typescript-generator/index.js
@@ -106,7 +106,8 @@ function createInterface (type) {
   return code
 
   function required (property) {
-    if (isResponse) return ''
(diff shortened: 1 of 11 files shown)


typescript-generator/index.js
```diff
@@ -106,7 +106,8 @@ function createInterface (type) {
   return code
 
   function required (property) {
-    if (isResponse) return ''
+    // response default to required, unless explicitly marked as optional
+    if (isResponse) return property.required === false ? '?' : ''
```
Mpdreamz, line 110: @delvedor had to update this to allow response to mark direct properties as optional

delvedor (approved):
LGTM

PR #59, opened 2021-01-13
Add optionality to SearchRequest

Mpdreamz:
And add missing properties where needed.

Re-introduced TypeQuery, in `NEST` we already removed this obsolete
query construct


delvedor (approved):
LGTM

PR #60, opened 2021-01-13
Added pathParts to input spec

### LINKED ISSUE #54: Path parameters should be represented in the spec
Currently, the path parameters are inferred from the json spec by the compiler and added on the fly.
We should start tracking the path parameters directly in the input spec as well, so we can reuse already present definitions (such as `Indices`) and making it easier to contribute to the spec.

For example:
```diff
 class IndexRequest<TDocument> extends RequestBase {
+  path: {
+    index: IndexName;
+    id?: Id;
+  }
   query_parameters?: {
     if_primary_term?: long;
     if_seq_no?: long;
     op_type?: OpType;
     pipeline?: string;
     refresh?: Refresh;
     routing?: Routing;
     timeout?: Time;
     version?: long;
     version_type?: VersionType;
     wait_for_active_shards?: string;
     require_alias?: boolean;
   }
   body?: TDocument;
 }
```

delvedor:
This pr adds a ~~`pathParts`~~`path_parts` key that contains all the path parameters an API can take, it already specifies which parameters are required.
Currently, it's not using our nice aliases, as it's using the json spec defined types.
We'll update them in the future while working on the spec.

TODO:
- [x] add path parameters to the input spec
- [x] make the compiler aware of the new field

Closes #54


Mpdreamz (approved):
LGTM! :+1:

PR #61, opened 2021-01-14
Update search response with optionalibility

Mpdreamz:
Reduces failures down to 15 locally.

The remainder are aggregations N level deep for which we need to
formulate a battle plan.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -6156,7 +6156,7 @@ export interface Hit<TDocument = unknown> {
   _seq_no?: long
   _primary_term?: long
   _version?: long
-  sort?: Array<number | string>
+  sort?: Array<long | double | string>
 }
 
 export interface HitsMetadata<T = unknown> {
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -19,9 +19,9 @@
 
 import { Readable as ReadableStream } from 'stream'
 
-export type Aggregate = ValueAggregate | AutoDateHistogramAggregate | FiltersAggregate | SignificantTermsAggregate<object> | TermsAggregate<object> | BucketAggregate | CompositeBucketAggregate | MultiBucketAggregate<object> | SingleBucketAggregate | MatrixStatsAggregate | BoxPlotAggregate | ExtendedStatsAggregate | GeoBoundsAggregate | GeoCentroidAggregate | PercentilesAggregate | ScriptedMetricAggregate | StatsAggregate | StringStatsAggregate | TopHitsAggregate | TopMetricsAggregate | KeyedValueAggregate | TDigestPercentilesAggregate | HdrPercentilesAggregate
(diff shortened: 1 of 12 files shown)


specification/specs/search/search/hits/Hit.ts
```diff
@@ -1,21 +1,25 @@
...
-  _version: long;
-  sort: Array<number | string>;
+
+  _score?: double;
+  _type?: TypeName;
+
+  _explanation?: Explanation;
+  fields?: Dictionary<string, LazyDocument>;
+  highlight?: Dictionary<string, string[]>;
+  /** @prop_serializer VerbatimInterfaceReadOnlyDictionaryKeysFormatter`2 */
```
delvedor, line 11: Should we keep this?

specification/specs/search/search/hits/Hit.ts
```diff
@@ -1,21 +1,25 @@
...
+  _ignored?: string[];
+
+  _shard?: string;
+  _node?: string;
+  _routing?: string;
+  _source?: TDocument;
+  _seq_no?: long;
+  _primary_term?: long;
+  _version?: long;
+  sort?: Array<number | string>;
```
delvedor, line 24: We should try to be as much specific as possible, is this a `long` or a `double`?

specification/specs/aggregations/Aggregate.ts
```diff
@@ -67,7 +72,7 @@ class AutoDateHistogramAggregate extends AggregateBase {
 }
 
 class FiltersAggregate extends AggregateBase {
-  buckets:FiltersBucketItem[];
+  buckets:FiltersBucketItem[] | Dictionary<string, FiltersBucketItem>;
```
delvedor, line 75:
```suggestion
  buckets: FiltersBucketItem[] | Dictionary<string, FiltersBucketItem>;
```

output/typescript/types.ts
```diff
@@ -6135,35 +6138,37 @@ export interface HighlightField {
...
+  _nested?: NestedIdentity
+  _ignored?: Array<string>
+  _shard?: string
+  _node?: string
+  _routing?: string
+  _source?: TDocument
+  _seq_no?: long
+  _primary_term?: long
+  _version?: long
+  sort?: Array<number | string>
```
delvedor, line 6159: Same here.

output/typescript/types.ts
```diff
@@ -139,7 +143,7 @@ export interface ExtendedStatsAggregate extends StatsAggregate {
 }
 
 export interface FiltersAggregate extends AggregateBase {
-  buckets: Array<FiltersBucketItem>
+  buckets: Array<FiltersBucketItem> | Record<string, FiltersBucketItem>
```
delvedor, line 146: Should we use `Record` or `Dictionary`?

delvedor (approved):
Overall LGTM, left a couple of nits :)

specification/specs/search/search/hits/Hit.ts
Mpdreamz: For now yeah, these all indicate the .NET client does something special. Want to use these as markers to go over at some point

specification/specs/search/search/hits/Hit.ts
Mpdreamz: both lol should we do `long | double | string` ?

delvedor: Yep! :)

output/typescript/types.ts
Mpdreamz: The input is `Dictionary` but the output is always `Record`

delvedor: My bad, should have looked at the file name!
Some times this output is hidden, others not 😅

delvedor (approved):
LGTM

Issue #58, opened 2021-01-13
Add missing APIs


delvedor:
There are a bunch of endpoints that are not tracked here. Adding them is straightforward, it's enough to add the json spec definition.

PR #62, opened 2021-01-14
Put all json spec files in the same location

delvedor:
As titled.
In addition, this pr updates the metamodel generator so it tracks all endpoints, and leave `null` when the types for a specific endpoint has not been implemented.
The typescript generator has been updated as well.
I've used the 7.10 json spec files from the elasticsearch repo.

Closes: #58


Mpdreamz (approved):
LGTM

delvedor:
🚨 This pr is causing an error while running the generation, I'll fix the validation code as well before merging this.

swallez:
> updates the metamodel generator so it tracks all endpoints, and leave null when the types for a specific endpoint has not been implemented.

Why do we want to output an incomplete model? What is a code generator supposed to do when request or response are missing?

delvedor:
@swallez tt depends on your code generator, the typescript generator will create the method definition and use a ["TODO" type](https://github.com/elastic/elastic-client-generator/blob/f8d40ae081e7619a5ababbeeadf166c42cc389f4/typescript-generator/client.js#L35-L40) to signal that the definition is incomplete, for example:

```ts
search(params: TODO): Promise<TODO>
```

You can easily skip the incomplete definition in your generator by checking the `request` and `response` values.

swallez:
@delvedor my concern here is that we could release incomplete clients in the wild.

So we have to make it clear in `metamodel.ts` and change `request` and `response` to be `TypeName | null` with an explanation of what `null` means, and that it should not be confused with optional.

swallez:
BTW what about using something like `_json` or `_json_specs`  for the JSON files folder, so that it stands out of the crowd in IDEs and can also be easily excluded from directory crawls?

delvedor:
> So we have to make it clear in metamodel.ts and change request and response to be TypeName | null with an explanation of what null means, and that it should not be confused with optional.

Done!

> BTW what about using something like _json or _json_specs for the JSON files folder, so that it stands out of the crowd in IDEs and can also be easily excluded from directory crawls?

Good idea!

PR #66, opened 2021-01-18
make sure lib is not ignored so specification can still be referenced from other node repositories

Mpdreamz:


Mpdreamz:
@swallez @delvedor self-reviewing this one so I can go through the hoops of validating github npm install in flight-recorder

PR #67, opened 2021-01-18
Seperate extends and implements

Mpdreamz:
This PR splits extends and implements to two seperately exposed
components. Previously they were all exposed under `.inherits`.

Interfaces are barely used in the input specification. Where they were
used there was always also a common base class implementation.

I removed all the used interfaces and reintroduced `IDictionary<TKey,
TValue>`

This allows types to declare they implement this dictionary and we can
the presence of this to either generate an appropiate type or know that
we might need to handle serialization explicitly.

```json
{
    "doc_count: 0
    "mykey" { .. }
    "myother_key" : { ... }
}
```

Can therefor now be typed as:

```ts
class ThingWithDocCount implements IDictionary<string, Aggregate> {}
```

In the typescript output we can inspect `interface.implements` to see if
it extends `IDictionary` and know we need to generate:

```ts
type ThingWithDocCount = {
   doc_count: long
} & Record<string, Aggregate>
```

Implements is exposed on all the types to ease generation e.g

```
A implements Y
B extends A
C extends B
```

In the canonical domain when inspecting `C`

`types['C']["implements"]` will hold a reference to `Y`
`types['C']["extends"]` will hold a reference to `B`

Implements exposes `depth` to know if the implements is a direct
constraint or one coming through from a base class.

In `C#` this may be used to know if properties from `Y` need to be
explicitly implemented or if we can assume a baseclass has done so
already.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -1201,7 +1201,6 @@
       },
       "implements": [
         {
-          "depth": 0,
           "type": {
             "namespace": "common",
             "name": "IDictionary"
@@ -1223,6 +1222,9 @@
             }
           ]
         }
+      ],
+      "traits": [
+        "IDictionary"
       ],
       "properties": []
     },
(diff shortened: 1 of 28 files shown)


Mpdreamz:
This PR is now updated to be less impactful to `schema.json`. 

Types now only lists there direct `implements`  no more `depth` and interfaces are not viral by default. 

This frees input spec to use `interface` freely as another abstraction. 

This PR does introduce `trait`: an interface marked with `@trait description` will get exposed e.g

```
/**
 * @trait a description of why this trait is necessary
 */
interface IDictionary<TKey,TValue> {}
```

```json
...
    {
      "kind": "interface",
      "name": {
        "namespace": "aggregations",
        "name": "CompositeBucket"
      },
      "inherits": [
        {
          ...
        }
      ],
      "traits": ["IDictionary"],
      "properties": []
    },
...
```

Only `interfaces` that are a trait end up in `"traits"`.

The compiler has an allow list of known traits that will need manual updating so we can not randomly introduce new traits without documenting them.

delvedor (dismissed):
LGTM

PR #68, opened 2021-01-18
update typescript to generate type alias for implements IDictionary

Mpdreamz:


specification/specs/common.ts
```diff
@@ -103,8 +103,13 @@ class DateField {
   include_unmapped?: boolean;
 }
 
+/** Rather then documenting `object` this specifies places in the response that depend on the mapping and or request
+ * e.g the result of a script can be anything from value types to objects  */
 class UserDefinedValue {}
 
+/** Documents a place that can return ANY document from ANY index that can not be directly related back to a generic */
+class LazyDocument {}
```
delvedor, line 111: Should we use this one for `MultiGet` as well?
Currently isn't a problem, as the `MultiGet` generic defaults to `unknown`.

Mpdreamz: we should :)
https://github.com/elastic/elastic-client-generator/issues/71

Mpdreamz:
This now uses the new TS magic to support the `IDictionary` trait that we translate to a recursive type alias.

Gets the search response failures down to `0` :smile_cat:

delvedor (changes requested):
LGTM with a nit, can you add to the skip list the `AdditionalProperties` interface in the ts generator?

typescript-generator/index.js
```diff
@@ -193,10 +197,43 @@ function isSpecialInterface (type) {
...
+  //find inherits on parent if current type has parent
+  //TODO fix spec so that inherits is no longer an array
+  if (Array.isArray(type.inherits)) return lookupBehaviorImplements(type.inherits[0], name)
+  return null;
+}
+
+function serializeAdditionalPropertiesType(type) {
+  const dictionaryOf = lookupBehaviorImplements(type, 'AdditionalProperties');
+  if (!dictionaryOf) throw new Error(`Unknown implements ${type.name.name}`)
+  //export type SingleBucketAggregate = { doc_count: double } & Record<string, Aggregate>
```
delvedor, line 212: Leftover?

Mpdreamz: Yep! removed

Mpdreamz:
> > can you add to the skip list the AdditionalProperties interface in the ts generator? 
Done!

PR #72, opened 2021-01-20
Update and stabilize 'indices.freeze' API

sethmlarson:
First attempt at fixing an API, is there a way we want to word our commits?


delvedor (approved):
LGTM

We didn't formalize a commit message format :)

PR #73, opened 2021-01-20
Cat Aliases

philkra:
Added request & response without the CSV validation.

### DIFF
--- a/specification/specs/cat/cat_aliases/CatAliasesRequest.ts
+++ b/specification/specs/cat/cat_aliases/CatAliasesRequest.ts
@@ -4,7 +4,7 @@ class CatAliasesRequest extends CatRequestBase {
     name?: Names;
   }
   query_parameters?: {
-    expand_wildcards?: ExpandWildcards | Array<ExpandWildcards>;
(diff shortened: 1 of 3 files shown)


specification/specs/cat/cat_aliases/CatAliasesRequest.ts
```diff
@@ -4,7 +4,7 @@ class CatAliasesRequest extends CatRequestBase {
     name?: Names;
   }
   query_parameters?: {
-    expand_wildcards?: ExpandWildcards;
+    expand_wildcards?: ExpandWildcards | Array<ExpandWildcards>;
```
Mpdreamz, line 7: I think the better fix would be:

```
export enum ExpandWildcardsOptions {
..
}

export type ExpandWildcards = ExpandWildcardsOptions | Array<ExpandWildcardsOptions> 
```

This should fix all usages of `ExpandWildcards`

Mpdreamz (changes requested):

Mpdreamz (dismissed):

philkra:
added suggested improvements

delvedor (approved):
LGTM

PR #77, opened 2021-01-21
Adde Indices.Segment Endpoint + Spec

philkra:


delvedor (approved):
Changes LGTM, but given the tile of the pr I'm not sure you need to do something more.

output/typescript/types.ts
```diff
@@ -3977,7 +3977,7 @@ export interface ShardRecovery {
 }
 
 export interface IndexSegment {
-  shards: Record<string, ShardsSegment>
+  shards: Record<string, ShardsSegment | Array<ShardsSegment>>
```
Mpdreamz, line 3980: Can this ever be just `ShardsSegment` and not `Array<ShardsSegment>` ? 

I think we might also want to introduce a `@cleanup("reason")` reason annotation @delvedor @SylvainJuge to start annotating server responses that we think should be more regular?

delvedor:
Can you rebase on master?

output/typescript/types.ts
philkra: I think you pinged the Sylvain on that comment 😄

delvedor (approved):
LGTM

PR #80, opened 2021-01-25
Indices Analyze

stevejgordon:
Fix request and response for indices analyze.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -753,7 +753,8 @@ export interface ICharFilter {
 
 export interface ITokenFilter {
   type: string
-  version: string
+  version?: string
+  stopwords?: Array<string>
 }
 
 export interface ITokenizer {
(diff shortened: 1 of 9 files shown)


delvedor (approved):
The changes looks good to me!

@Mpdreamz I'm wondering if we should establish a "style guide" on when and how to merge multiple definitions in a single file. This, for example, seems a good case.

Mpdreamz:
> @Mpdreamz I'm wondering if we should establish a "style guide" on when and how to merge multiple definitions in a single file. This, for example, seems a good case.

++ this seems like a good topic for the next types validation sync :)

stevejgordon:
@Mpdreamz Want me to leave this open for the discussion? Or shall I merge as-is for the time being?

output/typescript/types.ts
```diff
@@ -753,7 +753,8 @@ export interface ICharFilter {
 
 export interface ITokenFilter {
   type: string
-  version: string
+  version?: string
+  stopwords?: Array<string>
```
Mpdreamz, line 757: I dont think this should live on `ITokenFilter`. I'm reverting this which might turn indices.analyze red again @stevejgordon

stevejgordon: Ah okay. I'll take another run at it tomorrow.

PR #82, opened 2021-01-25
Add auto label action

delvedor:
As titled. This will automate the label handling and make it easier in the future to filter pull requests.


Mpdreamz (approved):
:+1: :100: LGTM

PR #83, opened 2021-01-25
Cluster Allocation Explain

stevejgordon:
Fix cluster allocation explain response.

I reverse engineered the cluster_info type based on `ClusterInfo.java` in the ES repository.

I've kept the 4 related types in `ClusterInfo.ts` but I can break those out if preferred?


delvedor (approved):
LGTM

> I've kept the 4 related types in ClusterInfo.ts but I can break those out if preferred?

It's perfect like you did, thanks!

PR #84, opened 2021-01-25
Indices Get Settings

stevejgordon:
Fix `IndexState` with nullable properties so that Indices.Get_Settings response tests pass.


delvedor (approved):
LGTM

Issue #74, opened 2021-01-21
HEAD responses should be boolean aliases


delvedor:
As titled.

For example:

```ts
type DocumentExistsResponse = boolean
```

delvedor:
@Mpdreamz suggested top use a base class, as some languages will not return a plain boolean, but a structured object. The clients generators will need to detect this custom class and act accordingly.

```ts
class HeadResponseBase {
  exists: boolean
}

class DocumentExistsResponse extends HeadResponseBase {}
```

Mpdreamz:
```ts
class HeadResponseBase extends ResponseBase {
  exists: boolean
}

class DocumentExistsResponse extends HeadResponseBase {}
```

PR #85, opened 2021-01-26
Introduce HeadResponseBase

delvedor:
As titled.

I've added a new section in the docs as well to track the "special classes" that we are using.
Closes: #74

### DIFF
--- a/docs/special-classes.md
+++ b/docs/special-classes.md
@@ -1,0 +1,42 @@
+# Special classes
+
+Some APIs needs to be handled differenlty based on the output language, while others share many common parameters.
+This document contains the list of this special classes and where those should be used.
+
+## CatRequestBase
+
+This class must be used when implementing any Cat request definition.
+
+For example:
+```ts
+@rest_spec_name("cat.aliases")
+class CatAliasesRequest extends CatRequestBase {
+  path_parts?: {
+    name?: Names;
+  }
+  query_parameters?: {
+    expand_wildcards?: ExpandWildcards;
+  }
+  body?: {
+  }
+}
+```
+
+## CatResponseBase
+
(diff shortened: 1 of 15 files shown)


docs/special-classes.md
```diff
@@ -0,0 +1,42 @@
...
+## CatResponseBase
+
+This class must be used when implementing any Cat response definition.
+
+For example:
+```ts
+class CatAliasesResponse extends CatResponseBase<CatAliasesRecord> {}
+```
+
+## HeadResponseBase
```
Mpdreamz, line 34: I am not feeling the name `HeadResponseBase` since its too tied to the HTTP method e.g why is there no `PostResponseBase` etcetera. How about `EmptyResponseBase` ?

Mpdreamz (approved):
LGTM + :heart: :100: the docs.

I am not somehow not feeling the name so left a suggestion for a new one.

docs/special-classes.md
delvedor: `EmptyResponseBase` could work as well, I liked `HeadResponseBase` because is something that will never leak into language implementations, but it will be used by generators to define the behavior in case of HEAD responses.

Mpdreamz: I am not sure `EmptyResponseBase` won't leak always. I am not sure if we'll expose it in .NET. Currently we already have `e.g` `ExistsResponse` and we might keep that as a parent.

delvedor:
In a subsequent pr I'll move both `CatResponseBase` and `EmptyResponseBase` from extended classes to implements as those are describing behaviors (see https://github.com/elastic/elastic-client-generator/pull/67).

Mpdreamz (approved):

PR #90, opened 2021-01-26
Add new behaviors

delvedor:
Move `CatResponseBase` and `EmptyResponseBase` from classes to behaviors.

### DIFF
--- a/specification/specs/behaviors.ts
+++ b/specification/specs/behaviors.ts
@@ -17,12 +17,12 @@
 interface AdditionalProperties<TKey, TValue> {}
 
 /**
- * A cat response formatted in JSON will be an array of records.
+ * A response formatted as an array of records.
  * Some languages can't represent this easily and need to wrap the
  * array inside an object.
- * @behavior Defines a trait that the response should be an array of records typed to TCatRecord
+ * @behavior Defines a trait that the response should be an array of records typed to TRecord
  */
-interface CatResponseBase<TCatRecord> {}
+interface ArrayResponse<TRecord> {}
 
 /**
  * HEAD APIs can have a different behavior based on the language,
--- a/specification/specs/behaviors.ts
+++ b/specification/specs/behaviors.ts
@@ -15,3 +15,19 @@
  * @behavior Defines a trait that any unknown property for the class should be typed to TValue
  */
 interface AdditionalProperties<TKey, TValue> {}
+
+/**
+ * A cat response formatted in JSON will be an array of records.
+ * Some languages can't represent this easily and need to wrap the
+ * array inside an object.
+ * @behavior Defines a trait that the response should be an array of records typed to TCatRecord
+ */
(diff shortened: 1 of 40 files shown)


specification/specs/behaviors.ts
```diff
@@ -15,3 +15,19 @@
...
  */
 interface AdditionalProperties<TKey, TValue> {}
+
+/**
+ * A cat response formatted in JSON will be an array of records.
+ * Some languages can't represent this easily and need to wrap the
+ * array inside an object.
+ * @behavior Defines a trait that the response should be an array of records typed to TCatRecord
+ */
+interface CatResponseBase<TCatRecord> {}
```
Mpdreamz, line 25: The behavior that deviates is that these API's return an array and not an object on valid responses, but could return an object if an error occurs.  

```suggestion
interface ArrayResponse<TItem> {}
```

That way we can keep `CatResponseBase`

```typescript
class CatResponseBase<TCatRecord> extends ResponseBase implements ArrayResponse<TCatRecord> {
}
```

Reason for enforcing this is that each high level client api method is the same.


```typescript
client.apiCall(TRequestBase r): TResponseBase
```

Therefor: 

```typescript
class PingResponse implements EmptyResponseBase {}
```

should be 

```typescript
class PingResponse extends ResponseBase implements EmptyResponseBase {}
```


We could enforce this, in a follow up commit, by moving `@rest_spec_name()` from `Request` to 

```typescript
@rest_spec_name("")
class SearchRequestEndpointMapping implements EndpointMapping<SearchRequest,SearchResponse> {}

class SearchRequest extends RequestBase {}
```

These types are not emitted by the compiler, but used to replace the string comparison based hack we do now.

or maybe better:

```typescript
@rest_spec_name("")
class SearchRequest extends RequestBase<SearchResponse> {
}

class RequestBase<TResponse extends ResponseBase> extends RequestBase
```

The compiler can enforce only `RequestBase<>` extends `RequestBase` directly

Mpdreamz (changes requested):

specification/specs/behaviors.ts
delvedor: 👍 on renaming `CatResponseBase` to `ArrayResponse`.

> The behavior that deviates is that these API's return an array and not an object on valid responses, but could return an object if an error occurs.

I see your point, but using both `extends` and `implements` (as behavior) here could lead to invalid code in the generators.
For example:
```ts
class CatResponseBase<TCatRecord> extends ResponseBase implements ArrayResponse<TCatRecord> {}
```

In js means that `CatResponseBase` is an array with additional properties coming from `ResponseBase`, which is not true.

delvedor: After a chat with @Mpdreamz we agreed that `ResponseBase` should always be present in response definitions, but `ResponseBase` should not add any additional properties.

Mpdreamz (approved):

PR #92, opened 2021-01-27
Indices Analyze

stevejgordon:
Fixes the request after discussing the approach with @Mpdreamz. Note that `TokenFilterBase` includes a commented out line for the `DictionaryDecompounderTokenFilter` since that causes an exception in the generator. I'll open an issue to track that.


delvedor (approved):
LGTM

I can't understand why only yours pull request are over-labeled by the automation 🤔

stevejgordon:
@delvedor I was also wondering that!

PR #96, opened 2021-01-27
Add linter

delvedor:
This pr adds a linter, [standard](https://github.com/standard/standard) to the typescript generator and [ts-standard](https://github.com/standard/ts-standard) to the compiler, in an effort to improving the code quality and resilience.

Standard is a well known linter in the js community and other than fixing the code style, it also contains a series of gotchas to avoid falling in "brittle code" situations.
The linter will be executed as CI step as well.

This pr is the first step towards refactoring of the compiler.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,30 @@
+name: Linter
+
+on: [pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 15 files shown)


Mpdreamz (approved):
LGTM should we make --fix part of run-validations.sh?

delvedor:
@Mpdreamz I don't think so, as `run-validations.sh` is just for running the validator infrastructure, while the linter is useful only if you are working on the compiler/generator.

PR #99, opened 2021-01-28
Make targets for contrib & check-license

philkra:
* `check-license` target that locally verifies the presence of license headers
* `contrib` target that runs all pre contribution tasks: `check-license`, `format-code` and `generate-spec`


Mpdreamz (approved):
LGTM, we could instead of checking the licenses just add the licenses with the same script we ran for #98 but that can be a follow up PR.

delvedor (approved):
LGTM

PR #101, opened 2021-02-01
New specification compiler

delvedor:
This pr introduces a new specification compiler, it's a complete rewrite of the previous one and has some nice benefits.
- Directly outputs the metamodel without using the intermediate Domain
- Uses [ts-morph](https://ts-morph.com/) instead of [byots](https://www.npmjs.com/package/byots) for navigating the AST. ts-morph is a wrapper of the TypeScript compiler API, which makes it nicer to use.
- The code now is no longer a series of classes with a state, but it is a series of composable stateless functions. This means that we lose some context at the benefit of reusability and readability.
- ts-morph offers a set of nice utilities that could help us adding new features to the compiler in the future. An example that is already here is a check to verify if a declaration is defined but never used.

TODO:
- [x] verify that the output schema is the same (this is a little annoying because the two compilers are traversing the specification differently. which causes a slightly different json output. A stable serializer might solve the problem).
- [x] remove the old compiler and unused dependencies
- [x] remove the currently generated `dump.json` and output the schema in the output folder


Mpdreamz (changes requested):
This is awesome! Thank you for doing the much needed refactoring in order to lift us out of POC phase. Huge!


> verify that the output schema is the same (this is a little annoying because the two compilers are traversing the specification differently. which causes a slightly different json output. A stable serializer might solve the problem).

This should be part of this PR IMO. We need this verification before merging. 
We use an ordered stable json serializer for the `clients-flight-recorder` this seems like a good thing to have on `schema.json` in general. 

> remove the currently generated dump.json and output the schema in the output folder
> remove the old compiler and unused dependencies

++ on doing this in a follow up PR(s) so this rewrite ends up clean in the main line.

specification/compiler/utils.ts
```diff
@@ -380,13 +380,13 @@ export function isKnownBehavior (node: HeritageClause | ExpressionWithTypeArgume
...
     for (const knownBehavior of knownBehaviors) {
-      if (node.getTypeNodes()[0].getText().startsWith(knownBehavior)) {
+      if (node.getTypeNodes()[0].getExpression().getText() === knownBehavior) {
         return true
       }
     }
   } else {
     for (const knownBehavior of knownBehaviors) {
-      if (node.getExpression().getText().startsWith(knownBehavior)) {
+      if (node.getExpression().getText() === knownBehavior) {
```
Mpdreamz, line 389: +++ ty for remembering! :)

delvedor:
> on doing this in a follow up PR(s) so this rewrite ends up clean in the main line.

I prefer to do the swap all in a single pr, so we can detect quickly if something did break and rollback. Furthermore, it should be easier for our users to fetch the updates :)

Mpdreamz:
From sync earlier today:

- description will be added back in a follow up PR
- `@alternate_name` will be added back in in a follow up PR as well.

Mpdreamz (approved):
LGTM

PR #104, opened 2021-02-04
Cluster Health Request

philkra:
as titled.

### DIFF
--- a/specification/specs/common.ts
+++ b/specification/specs/common.ts
@@ -87,24 +87,30 @@ type double = number
 // TODO clean up Id and Name variants and make sure all API's use one purposefully
 // Id includes numeric support, Name does not
 type ScrollId = string
-type ScrollIds = string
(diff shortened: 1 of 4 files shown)


specification/specs/common.ts
```diff
@@ -115,8 +121,11 @@ type MultiTermQueryRewrite = string
 type GeoTilePrecision = number
 type GeoHashPrecision = number
 /** Path to field or array of paths. Some API's support wildcards in the path to select multiple fields.  */
+type Field = string
 type Fields = Field | Field[]
 
+type WaitForActiveShards = byte | WaitForActiveShardOptions
```
delvedor, line 127: @Mpdreamz @swallezthis is a similar case to what we said yesterday what we should support literal types as well.
`WaitForActiveShards` should be `type WaitForActiveShards = 0 | 'all'`

delvedor (approved):
LGTM

PR #105, opened 2021-02-04
Fixed One Error in Cluster.Reroute Request

philkra:
as title.d


specification/specs/cluster/cluster_reroute/ClusterRerouteRequest.ts
```diff
@@ -23,7 +23,7 @@ class ClusterRerouteRequest extends RequestBase {
     dry_run?: boolean
     explain?: boolean
     master_timeout?: Time
-    metric?: string[]
+    metric?: string | string[]
```
delvedor, line 26: We have a `Metrics` alias for this :)
```suggestion
    metric?: Metrics
```

philkra: added

delvedor (approved):
LGTM

PR #113, opened 2021-02-10
Added since decorator

delvedor:
This pr adds a new decorator named `since`. Its value describes when an API has been added to Elasticsearch. The compiler will check if is a valid semver value and add it to the JSON model.

I've used an internal script to generate the first round of `since` values, if I wasn't able to get the value, the default value will be `0.0.0`.

Closes: https://github.com/elastic/elastic-client-generator/issues/8


philkra (approved):
lgtm

PR #120, opened 2021-02-18
Delete by query Req & Res

philkra:
as titled.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -35290,7 +35290,7 @@
       "properties": [
         {
           "name": "batches",
-          "required": true,
+          "required": false,
           "type": {
             "kind": "instance_of",
             "type": {
@@ -35301,7 +35301,7 @@
         },
         {
           "name": "deleted",
-          "required": true,
+          "required": false,
           "type": {
             "kind": "instance_of",
             "type": {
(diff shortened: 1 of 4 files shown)


delvedor (changes requested):
I'm not sure how this API should be defined, as the response type changes based on the request: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-task-api
Might be a good topic for the next types sync.

cc @Mpdreamz @swallez

swallez:
What about representing it as two different API endpoints, each with a fixed value for `wait_for_completion`?

And we can represent the requests with a base `BaseDeleteByQuery` class that contains everything except `wait_for_completion`, and two subclasses `DeleteByQuery` and `DeleteByQueryAsync` that have a fixed value for `wait_for_completion`.

Not sure we support fixed values currently, if not that would be a useful addition for this kind of situation.

Another option is to have the response be a union type and let the user deal with the two variants, but it will be hard to disambiguate.

sethmlarson:
Is there a concept similar to [`typing.overload` in Python](https://docs.python.org/3/library/typing.html#typing.overload) for TypeScript? Where you can define multiple type signatures based on input types

delvedor (approved):
After syncing, we have decided to have all keys as optional, as the improvements should come from the server-side.

PR #122, opened 2021-02-18
Validated ilm.delete_lifecycle & ilm.put_lifecycle Request

philkra:
as titled.

### DIFF
--- a/specification/specs/x_pack/ilm/delete_lifecycle/DeleteLifecycleRequest.ts
+++ b/specification/specs/x_pack/ilm/delete_lifecycle/DeleteLifecycleRequest.ts
@@ -25,7 +25,7 @@
 interface DeleteLifecycleRequest extends RequestBase {
   path_parts?: {
     policy?: Name
-    policy_id: PolicyId
+    policy_id: Id
   }
   query_parameters?: {}
   body?: {}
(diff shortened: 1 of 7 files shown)


specification/specs/x_pack/ilm/delete_lifecycle/DeleteLifecycleRequest.ts
```diff
@@ -24,7 +24,8 @@
  */
 interface DeleteLifecycleRequest extends RequestBase {
   path_parts?: {
-    policy: Name
+    policy?: Name
+    policy_id: PolicyId
```
delvedor, line 28: I see why you created the alias `PolicyId`, but we should use `Id`. The problem is that `Id` can either be a string or a number when it should always be a number, but our validation infrastructure uses numbers as well.
We should fix this there.

philkra: I removed the notion of `PolicyId` and exchanged with `Id`

delvedor (approved):
LGTM

I'll follow up in the flight recorder to fix https://github.com/elastic/elastic-client-generator/pull/122#discussion_r578499558.

PR #131, opened 2021-02-18
Validate cat.templates Response

philkra:
as titled.


delvedor (approved):
LGTM

@Mpdreamz @stevejgordon same question as https://github.com/elastic/elastic-client-generator/pull/132#discussion_r578973327.

PR #132, opened 2021-02-18
Validate cat.snapshots Response

philkra:
as titled.

### DIFF
--- a/specification/specs/cat/cat_snapshots/CatSnapshotsRecord.ts
+++ b/specification/specs/cat/cat_snapshots/CatSnapshotsRecord.ts
@@ -19,20 +19,18 @@
 
 class CatSnapshotsRecord {
   duration: Time
-  /** @prop_serializer StringLongFormatter */
-  end_epoch: long
+  end_epoch: EpochMillis
   end_time: string
   /** @prop_serializer StringLongFormatter */
-  failed_shards: long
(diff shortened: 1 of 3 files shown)


specification/specs/cat/cat_snapshots/CatSnapshotsRecord.ts
```diff
@@ -19,20 +19,18 @@
 
 class CatSnapshotsRecord {
   duration: Time
-  /** @prop_serializer StringLongFormatter */
-  end_epoch: long
+  end_epoch: EpochMillis
   end_time: string
   /** @prop_serializer StringLongFormatter */
```
delvedor, line 24: @Mpdreamz @stevejgordon do you still need the prop serializer?

delvedor (approved):
LGTM

specification/specs/cat/cat_snapshots/CatSnapshotsRecord.ts
stevejgordon: @delvedor Sorry, totally blanked on this! I don't think we do. That formatter knows to try either reading from a string or a long value and I think we can apply similar behaviour in our generated code based on seeing anything that is `EpochMillis`. @Mpdreamz Sound reasonable?

PR #134, opened 2021-02-19
Fix indices.flush_synced response

stevejgordon:

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -100608,8 +100608,24 @@
     {
       "inherits": [
         {
+          "generics": [
+            {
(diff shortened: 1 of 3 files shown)


philkra (approved):
LGTM

delvedor (commented):
The fix looks ok! I want to do a double check on `DictionaryResponseBase`.
Holding the merge for now :)

stevejgordon:
@delvedor Yeah I basically followed the approach we use in NEST to produce that type, but if it needs tweaking, happy to do so :-)

delvedor (approved):
LGTM

PR #135, opened 2021-02-19
Fix license.post_start_basic response

stevejgordon:

### DIFF
--- a/specification/specs/x_pack/license/start_basic_license/StartBasicLicenseResponse.ts
+++ b/specification/specs/x_pack/license/start_basic_license/StartBasicLicenseResponse.ts
@@ -18,7 +18,7 @@
  */
 
 class StartBasicLicenseResponse extends AcknowledgedResponseBase {
-  acknowledge: Dictionary<string, Union<string, string[]>>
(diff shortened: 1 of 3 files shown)


philkra (dismissed):
LGTM

specification/specs/x_pack/license/start_basic_license/StartBasicLicenseResponse.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 class StartBasicLicenseResponse extends AcknowledgedResponseBase {
-  acknowledge: Dictionary<string, string[]>
+  acknowledge: Dictionary<string, Union<string, string[]>>
```
delvedor, line 21: Please avoid using the `Union` type and use the union syntax :)
```suggestion
  acknowledge: Dictionary<string, string | string[]>
```

stevejgordon: My bad! Be a week or so since I touched this and that dropped out of my head already! Will commit a fix and push regen shortly.

delvedor (approved):
LGTM

PR #139, opened 2021-02-22
Fix field_caps Request

philkra:
as titled.


output/typescript/types.ts
```diff
@@ -3631,11 +3632,19 @@ export interface FieldCapabilitiesBodyIndexFilterRangeTimestamp {
   lt?: integer
 }
 
+export interface FieldCapabilitiesBodyIndexFilterTerm {
+  versionControl: FieldCapabilitiesBodyIndexFilterTermVersionControl
```
Mpdreamz, line 3636: is this really `versionControl`? Might be worth opening a ticket on elasticsearch if so should be `version_control`.

Mpdreamz (approved):
LGTM

PR #140, opened 2021-02-22
Cleanup of Type & Types

philkra:
The type `Type` and `Types` (array of Type) were ambiguous with `TypeName` and `TypeNames`. This PR removes the latter and replaces all occurrences with the former.

### DIFF
--- a/specification/specs/common.ts
+++ b/specification/specs/common.ts
@@ -104,7 +104,8 @@ type Indices = string | string[]
 
 type TypeName = string
 type TypeNames = string | string[]
-type Types = string | string[]
+type Type = string
+type Types = Type | Type[]
 
 type Routing = string | number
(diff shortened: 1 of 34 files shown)


specification/specs/common.ts
```diff
@@ -102,9 +102,10 @@ type Ids = string | number | string[]
 type IndexName = string
 type Indices = string | string[]
 
-type TypeName = string
-type TypeNames = string | string[]
-type Types = string | string[]
+type TypeName = string // TODO remove and use Type
+type TypeNames = string | string[] // TODO remove and use Types
```
delvedor, line 106: You can safely remove these two, as they both ended up in the handling report.

delvedor (approved):
LGTM

PR #142, opened 2021-02-22
Cleanup the usage of `Fields`

philkra:
The `Fields` type was expressed ambiguously with the union `Field | Field[]`. This PR removes all union occurrences and replaces them with an explicit type `Fields`.


delvedor (approved):
LGTM

PR #144, opened 2021-02-22
Fix ml.start_datafeed response

stevejgordon:
Dangling.csv got generated but I'm not entirely sure if it was due to my change, included it for the time being.


philkra (approved):
LGTM

PR #148, opened 2021-02-22
Fix cat.health request and response

stevejgordon:


output/typescript/types.ts
```diff
@@ -1017,7 +1017,7 @@ export type CatFielddataResponse = CatFielddataRecord[]
 
 export interface CatHealthRecord {
   cluster: string
-  epoch: string
+  epoch?: string
```
philkra, line 1020: please use type `EpochMillis`

philkra (changes requested):

output/typescript/types.ts
stevejgordon: Done 😄

stevejgordon:
Just untangling these conflicts!

stevejgordon:
Oh! I see now, @philkra - You stole this one from my assigned APIs and it's already fixed up! 😄

PR #149, opened 2021-02-22
Initial cat nodes columns

stevejgordon:
This is incomplete but adds all possible colums to the `CatNodesRecord` based on https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/rest/action/cat/RestNodesAction.java

Up for discussion is how to handle aliases as some of the tests use different names for the same field:

```
{
    "disk": "226.3gb",
    "dt": "250.9gb",
    "du": "24.6gb",
    "dup": "9.82"
  }
```

and 

```
 {
    "diskAvail": "226.3gb",
    "diskTotal": "250.9gb",
    "diskUsed": "24.6gb",
    "diskUsedPercent": "9.82"
  }
```

`disk` and `diskAvail` and `disk.avail` are used but refer to the same column.


delvedor (commented):
As per discussion, we should track every type alias (as optional).

Mpdreamz:
superseded by #185 which adds `@aliases` on all `_cat` endpoints.

PR #150, opened 2021-02-22
Fix cat.thread_pool Request

philkra:
as titled.


Mpdreamz (approved):
Interesting that `size` takes `boolean` https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-thread-pool.html makes no mention of it. Nice find :)

delvedor (approved):
LGTM

PR #151, opened 2021-02-23
Add support for generics on type aliases

Mpdreamz:


delvedor (dismissed):
Great work!
This should be applied also in https://github.com/elastic/elastic-client-generator/pull/134 isn't it?

Mpdreamz:
@delvedor I don't think so. Not sure I see the connection to that PR :)

delvedor (approved):
LGTM

PR #153, opened 2021-02-24
Tackle bulk of search request failures

Mpdreamz:
- fix aggregations
- stage query fixes
- fix remaining query failures

This brings search request failures down to 9. 

The remainder is in sort, functionscore queries and runtime mapping which require some more 
mapping work by pulling bits over from NEST.

Will follow up trying to get to 0


delvedor (approved):
LGTM

PR #155, opened 2021-03-01
Validate several Watcher endpoints

philkra:
endpoints added:
* response for `watcher.ack_watch` 
* response for `watcher.activate_watch` 
* response for `watcher.deactivate_watch` 
* response for `watcher.get_watch` 

Improved validation for `watcher.execute_watch`, close but not cigar!


delvedor (approved):
LGTM

PR #163, opened 2021-03-03
Typescript generator lookupBehavior behavior

Mpdreamz:
This ensures the typescript generator traverses the types to find the
behavior and removes the assumption the behavior is on the type itself.

It also makes sure `RequestBase` and `ResponseBase` make it to the to
output as these have the behaviors attached to them.

When generating a type because the output is all interfaces this now
collects all extends/implements and behaviors not on the skiplist and
ensures the generated interfaces extends from all.

### DIFF
--- a/typescript-generator/index.js
+++ b/typescript-generator/index.js
@@ -225,8 +225,9 @@ function cleanPropertyName (name) {
 function isSpecialInterface (type) {
   if (Array.isArray(type.attachedBehaviors)) {
     // assume types ending with Base are abstract and are not the ones doing the implements
-    if (type.name.name.endsWith('Base'))
+    if (type.name.name.endsWith('Base')) {
       return false
+    }
     return type.attachedBehaviors.length > 0
   }
   switch (type.name.name) {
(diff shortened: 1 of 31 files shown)


typescript-generator/index.js
```diff
@@ -237,6 +224,9 @@ function cleanPropertyName (name) {
 
 function isSpecialInterface (type) {
   if (Array.isArray(type.attachedBehaviors)) {
+    // assume types ending with Base are abstract and are not the ones doing the implements
+    if (type.name.name.endsWith('Base'))
```
Mpdreamz, line 228: @swallez @delvedor Thoughts on making this explicit in the model? e.g:

Any type that ends with `Base` gets `abstract: true` on type unless `@not_abstract` is specified on the type for edgecases?

Also happy to flip that and explicitly mark classes with `@abstract` in the input spec.

delvedor (changes requested):
LGTM

Can you also update the [docs](https://github.com/elastic/elastic-client-generator/blob/master/docs/behaviors.md) where appropriate?

typescript-generator/index.js
swallez: I'd rather be explicit in the input spec rather than rely on implicit naming conventions.

Mpdreamz: ++ aggreed, this should be easier once https://github.com/elastic/elastic-client-generator/pull/1634lands

PR #164, opened 2021-03-03
Removes annotations and parses them explicitly

Mpdreamz:
Exposing them on the actual types that they are defined on.

Removes `@types_stability`


specification/compiler/model/build-model.ts
```diff
@@ -48,34 +46,51 @@ import {
...
+        return {
+          path: path.path,
+          methods: path.methods,
+          ...(path.deprecated != null && { deprecation: path.deprecated })
+        }
+      })
+    }
+    map[api] = endpoint;
+
+    //model.endpoints.push(endpoint)
```
delvedor, line 87: Leftover?

specification/compiler/model/utils.ts
```diff
@@ -422,13 +384,137 @@ export function modelProperty (declaration: PropertySignature | PropertyDeclarat
...
+  }
+}
+
+/** Lifts jsDoc type annotations to request properties */
+export function hoistRequestAnnotations(
+  request: model.Request, jsDocs: JSDoc[], mappings: Record<string, model.Endpoint>, response: model.TypeName | null
+) {
+  const knownRequestAnnotations = [
+    'since', 'rest_spec_name', 'stability', 'visibility', 'behavior', 'class_serializer'
+  ]
```
delvedor, line 452: I think this list should live as top-level declaration, so it will be easier to find and update in the future if necessary.

specification/compiler/model/utils.ts
```diff
@@ -422,13 +384,137 @@ export function modelProperty (declaration: PropertySignature | PropertyDeclarat
...
+    if (tags[tag]) {
+      setter(tags, tag, tags[tag])
+    }
+  }
+
+  function getName(type) {
+    if (type instanceof model.BaseType) return type.name
+    if (type instanceof model.Property) return type.name
+    if (type instanceof model.EnumMember) return type.name
+    return typeof(type)
```
delvedor, line 442: Probably I'm missing something, what is the `typeof` function?

delvedor (changes requested):
LGTM

Left few comments. Can you run the linter as well?

specification/compiler/model/utils.ts
Mpdreamz: leftover :) it always returns `[Object]`

specification/compiler/model/utils.ts
Mpdreamz: It lives in the only function that use it. I quite like to keep this local until we have a need to expose it.

specification/compiler/model/build-model.ts
Mpdreamz: yep!

specification/compiler/model/metamodel.ts
```diff
@@ -120,21 +120,39 @@ export class Property {
...
 }
 
 // ------------------------------------------------------------------------------------------------
 // Type definitions
 
 export type TypeDefinition = Interface | Request | Union | Enum | TypeAlias
 
+export type Depreciable = TypeDefinition | EnumMember
+export const isDepreciable = (type: object): type is Depreciable => {
+  return (type as Depreciable).name !== undefined;
```
swallez, line 139: Isn't `name` catching a too big nest of objects that aren't Depreciable (and same for `BehaviorAttachable`? Also we should avoid adding code in `metamodel.ts`.

specification/compiler/model/metamodel.ts
```diff
@@ -189,6 +207,11 @@ export class Request extends BaseType {
   body?: ValueBody | PropertiesBody
   behaviors?: Implements[]
   attachedBehaviors?: string[]
+
+  since?: string
+  restSpecName?: string;
```
swallez, line 212: This should be `endpointName`. But I don't think this information is useful since we already have the request type identified in endpoints.

specification/compiler/model/metamodel.ts
```diff
@@ -251,8 +277,13 @@ export class TypeAlias extends BaseType {
...
   beta = 'beta',
-  experimental = 'experimental'
-  // NOTE: do we filter out "private"?
+  experimental = 'experimental',
+  TODO = 'TODO'
+}
+export enum Visibility {
+  public = 'public',
+  beta = 'feature_flag',
+  experimental = 'private'
```
swallez, line 286: Failed search/replace? Should the identifiers be `feature_flag` and `private`?

swallez (changes requested):

specification/compiler/model/metamodel.ts
```diff
@@ -120,21 +120,39 @@ export class Property {
   type: ValueOf
   required: boolean
   description?: string
-  annotations?: Record<string, string>
+  /**
+   * If specified takes precedence over `name` when generating code. `name` is always the value
+   * to be sent over the wire
+   */
+  alternateName?: string[]
```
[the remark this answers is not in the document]
Mpdreamz: Good catch! I like `identifier` but would like to keep `aliases` so `alternateName` does not get a new meaning and was never a fan of the name to begin with. 

So I settled on

* `name: string:` name to use for serialization and as "current" json name for deserialization (others being deprecated)
* `aliases?: string[]:` deprecated names that should be accepted for deserialization
* `identifier?: string:` identifier to be used for this property in code generation, defaulting to name if not specified.

specification/compiler/model/metamodel.ts
Mpdreamz: It is, all these typeguards are not even needed. Have updated the code to remove these

specification/compiler/model/metamodel.ts
Mpdreamz: yep, updated!

specification/compiler/model/metamodel.ts
Mpdreamz: I've moved all properties to the endpoint now and throw if they differ from the spec

Mpdreamz:
@swallez @delvedor this is ready for another round

`Check output freshness / build (pull_request)` is failing but if I run the worflow commands verbatim locally my git is clean, since @swallez PR is also failing on this I think this check is broken?

delvedor (dismissed):
LGTM

specification/compiler/model/metamodel.ts
```diff
@@ -120,21 +120,39 @@ export class Property {
...
+  /** An optional set of aliases for `name` */
+  aliases?: string[]
```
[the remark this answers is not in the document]
Mpdreamz: we should, added this and that also removes the need for `Depreciable` :+1:

specification/compiler/model/metamodel.ts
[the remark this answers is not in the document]
swallez: sounds good!

swallez (approved):
LGTM! I'll update https://github.com/elastic/elastic-client-generator/pull/169 once this is merged.

PR #165, opened 2021-03-04
Address Kibana feedback

delvedor:
As titled.

Ref: https://github.com/elastic/kibana/pull/83808


swallez:
Cannot really comment on the details of API changes without diving into the ES specs/code, but I have a few questions on the some similar/redundant constructs.

What is the difference between those and when should we use which?
- `object`, `any` and `UserDefinedValue` (changed in `Aggregate`)
- `Union<A, B>` and `A | B` (changed in `PutMappingRequest`)

delvedor:
> object, any and UserDefinedValue (changed in Aggregate)

`any` should not be used in the input spec, if it's there is a leftover.

> Union<A, B> and A | B (changed in PutMappingRequest)

The `Union` utility was the old way of expressing unions, now the compiler can use the native typescript union type (`|`), so we should update it where we are still using the old one :)

PR #169, opened 2021-03-04
Add model validator

swallez:
This PR adds two steps to the model compilation pipeline:
- removal of all incomplete endpoints that are missing either a request or response type
- validation of the internal consistency of the model (without looking at the REST spec)

Model validation can work in "just log errors" or "fail hard" modes. In this first iteration the model has some inconsistencies so we do not fail hard.

The validations are too numerous to detail here, the code has lots of comments detailing them.


specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,520 @@
...
+    try {
+      new URL(value) // eslint-disable-line no-new
+      return true
+    } catch (error) {
+      return false
+    }
+  },
+
+  // FIXME: Difference with stability?
+  type_stability: (value) => {
```
Mpdreamz, line 39: This can be removed once #164 lands

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,520 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as model from '../model/metamodel'
+
+const annotationValidators: Record<string, (value: string) => boolean> = {
+
+  stability: (value) => {
+    return ['stable', 'TODO'].includes(value)
```
Mpdreamz, line 25:
```suggestion
    return ['stable', 'beta', 'experimental', 'TODO'].includes(value)
```

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,520 @@
...
+      if (values.has(lcName)) {
+        modelError(`Duplicate lowercase enum name: ${member.name}`)
+      }
+      values.add(lcName)
+    }
+  }
+
+  function validateTypeAlias (typeDef: model.TypeAlias): void {
+    validateAnnotations(typeDef.annotations)
+    // FIXME: do we really need generics on a type alias? An alias is semantically similar to a 1-member union
```
Mpdreamz, line 370: We do, this came through as part of: https://github.com/elastic/elastic-client-generator/pull/151/files

This means generics from types can be passed down to type aliases on properties without losing content. 


Union itself can not be generic but the items it contain can

Mpdreamz (approved):
LGTM!, I added some notes that might need addressing after rebasing.

swallez:
I updated the PR after the removal of annotations.

I also added a validation that checks that a type that children (i.e. used in a `implements` or `inherits`) is not used as proprety type, because in that case we don't know what concrete type to instantiate for deserialization. Fixing it will require either refactoring the model or adding additional annotation to identify abstract classes. This happens e.g. with `XPackUsage` and `ErrorCause`.

Validations currently produces 119 errors. Some are definitely legit, some I will have to look more closely to understand if the validation is overly zealous (or blatantly wrong).

Mpdreamz:
> I also added a validation that checks that a type that has children (i.e. used in a implements or inherits) is not used as proprety type, because in that case we don't know what concrete type to instantiate for deserialization. Fixing it will require either refactoring the model or adding additional annotation to identify abstract classes. This happens e.g. with XPackUsage and ErrorCause.


Not all `base` classes are `abstract` so e.g `ErrorCause` should be able to used on its own. Can we update this check to mandate an `@abstract true|false` annotation on classes that are used as parents in the spec?

swallez:
I've updated the validator and now agree with its output. There are more validation to add, but they will come in subsequent PRs.

There are 14 errors:

```
enum definition query_dsl:Operator - Duplicate enum member identifier 'AND'
enum definition query_dsl:Operator - Duplicate enum member identifier 'OR'
Endpoint ilm.delete_lifecycle / Request - Request path property 'policy_id' doesn't exist in endpoint URL templates
Endpoint ilm.put_lifecycle / Request - Request path property 'policy_id' doesn't exist in endpoint URL templates
enum definition ingest.processors:DateRounding - Duplicate enum member identifier 'M'
Endpoint mget / Response - Type document.multiple.multi_get.request:MultiGetRequest not found
Endpoint ml.delete_expired_data / Request - Url path property 'job_id' is missing in request definition
type_alias definition x_pack.info.x_pack_usage:UrlConfig / union_of / instance_of - Non-leaf type cannot be used here: 'x_pack.info.x_pack_usage:BaseUrlConfig'
request definition search.multi_search:MultiSearchRequest / body / Property 'operations' / dictionary_of / instance_of - 'search.search:SearchRequest' is a request and must only be used in endpoints
Endpoint render_search_template / Request - Url path property 'id' is missing in request definition
request definition modules.snapshot_and_restore.restore:RestoreRequest / body / Property 'index_settings' / instance_of - 'indices.index_settings.update_index_settings:UpdateIndexSettingsRequest' is a request and must only be used in endpoints
interface definition x_pack.sql.translate_sql:TranslateSqlResponse / Property 'result' / instance_of - 'search.search:SearchRequest' is a request and must only be used in endpoints
interface definition x_pack.watcher.input:SearchInputRequestDefinition / Property 'body' / instance_of - 'search.search:SearchRequest' is a request and must only be used in endpoints
interface definition x_pack.watcher.input:SearchInputRequestDefinition / Property 'template' / instance_of - 'search.search_template:SearchTemplateRequest' is a request and must only be used in endpoints
Model validation: 14 errors. 1681 types visited, 179 dangling types.
```

The "Auto label" task fails with `path must not be empty` - is this a known issue? How can we debug/fix that?

swallez:
@Mpdreamz 

> Not all base classes are abstract so e.g ErrorCause should be able to used on its own. Can we update this check to mandate an @abstract true|false annotation on classes that are used as parents in the spec?

Indeed, I figured that out. I've identified different types of acceptable use of non-leaf types as property values. The are all listed explicitely in `validateIsLeafType()`:

- usage of a non-leaf type and its subtypes happens in distinct contexts. This is indeed the case of `ErrorCause` and its `MainError` child, and a few others like `XPackUsage` or `SearchResponse`.
- inheritance trees that aren't modeled as such and use an [externally tagged representation](https://serde.rs/enum-representations.html#externally-tagged). This is [`QueryContainer`](https://github.com/elastic/elastic-client-generator/blob/master/specification/specs/query_dsl/abstractions/container/QueryContainer.ts). I'm not sure there are others in the ES spec, but this approach is pretty common and could be indicated with an annotation on `QueryContainer`.
- inheritance trees that use an [internally tagged](https://serde.rs/enum-representations.html#internally-tagged) representation (with a `type` property) such as `PropertyBase` and `TokenFilterBase`. The model currently doesn't provide values for this `type` property that allow identifying the variant. This information could be captured using annotations, or (better IMHO) using constant string values in the type definitions like we do in the schema metamodel.
- variations on internally tagged using [`typed_keys`](https://www.elastic.co/guide/en/elasticsearch/reference/7.11/search-aggregations.html#return-agg-type). This is only used in aggregations. Unfortunately this means we can't represent this as constant values for a `type` property, so that will be annotations for this one.

Other uses of a non-leaf type in a property's type are forbidden as they don't provide any means to identify the variant to choose for deserialization. This is the case of `BaseUrlConfig` listed in the errors above.

**Discussion for futher metamodel updates:**
In terms of json model representation, the various options above could be represented as an additional optional property on interfaces: `variant_strategy: 'distinct_contexts' | 'external_tag' | 'internal_tag' | 'typed_key'`:
- if that property isn't present on a non-leaf type, then its usage as a field type is forbidden. 
- `distinct_contexts` means that the type is not an abstract type, and should be identified as such with an annotation in the input spec (i.e. it's opt-in, whereas an `abstract` annotation would be opt-out and prone to errors).
- `external_tag`
- `internal_tag` would normally require an additional property identifying the tag property, but since AFAIK only `type` is used, we can use it as the implicit tag name
- `type_keys` is our snowball...

And child types of a `'external_tag' | 'internal_tag' | 'typed_keys'` parent should have a `variant_id` string property that uniquely identifies them among their siblings.

specification/compiler/index.ts
```diff
@@ -18,19 +18,21 @@
...
 compiler
   .generateModel()
   .step(addInfo)
-  // validateRestSpec is not enabled until the RequestBase
-  // properties gets moved to a custom behavior
-  // .step(validateRestSpec)
+  .step(validateRestSpec)
+  .step(removeIncompleteEndpoints)
   .step(addDescription)
+  .step(async (model, spec) => validateModel(model, false))
```
delvedor, line 35: I'm wondering if we can keep the step declaration like the others, and choose to `failHard` based on an env variable or command line parameter.

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,594 @@
...
+  // Type definitions by name
+  const typeDefByName = new Map<string, model.TypeDefinition>()
+
+  for (const type of apiModel.types) {
+    typeDefByName.set(fqn(type.name), type)
+  }
+
+  // Register builtin types
+  for (const name of [
+    'string', 'boolean', 'number', 'object', 'null'
```
delvedor, line 69: If we have any definition that uses `object`, it should be updated to use `UserDefinedValue`, as it has no sense to exists.

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,594 @@
...
+        break
+
+      case 'named_value_of':
+        validateValueOf(valueOf.value, openGenerics)
+        break
+
+      case 'user_defined_value':
+        // Nothing to validate
+        break
+    }
```
delvedor, line 591: Let's make this future proof :)

```suggestion
      case 'user_defined_value':
        // Nothing to validate
        break
    
      default:
        throw new Error(`Unknown kind: ${valueOf.kind}`)
    }
```

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,594 @@
...
+/**
+ * Validates the internal consistency of the model (doesn't check the json spec)
+ *
+ * Any inconsistency is logged as an error.
+ *
+ * Missing validations:
+ * - verify uniqueness of property names in the inheritance chain
+ * - verify that request parents don't define properties (would they be path/request/body properties?)
+ */
+export default function validateModel (apiModel: model.Model, failHard: boolean): model.Model {
```
delvedor, line 36: As said above I would keep the export function as a step declaration, and read the `failHard` from env or cli args.

specification/compiler/steps/validate-model.ts
```diff
@@ -0,0 +1,594 @@
...
+    validateProperties(typeDef.properties, openGenerics)
+  }
+
+  function validateEnum (typeDef: model.Enum): void {
+    const allIdentifiers = new Set<string>()
+    const allNames = new Set<string>()
+
+    for (const item of typeDef.members) {
+      // Identifier must be unique among all items (case insensitive)
+      const identifier = (item.identifier ?? item.name).toLowerCase()
```
delvedor, line 374: I fear there are some cases in the type validator that require different casing, as the Elasticsearch API can accept multiple variations of the type.

delvedor (commented):
Amazing work!!

Mpdreamz (dismissed):
LGTM, nothing to add beyond @delvedor's comments

specification/compiler/steps/validate-model.ts
swallez: This is an interesting one... Since all `kind` are tested, this fails with `error TS2339: Property 'kind' does not exist on type 'never'.`

And omiting a `kind` does not cause a compilation error like it would do if we returned a value in the various `case` branches (like in shown in [this post](https://blog.logrocket.com/pattern-matching-and-type-safety-in-typescript-1da1231a2e34/))

This is the "correct" way to address this:

```
      default:
        // @ts-expect-error
        throw new Error(`Unknown kind: ${valueOf.kind}`)
```

Forgetting one of the `kind` values then causes the `ts-expect-error` to fail 😆

specification/compiler/index.ts
```diff
@@ -18,19 +18,21 @@
...
 const compiler = new Compiler()
 
 compiler
   .generateModel()
   .step(addInfo)
-  // validateRestSpec is not enabled until the RequestBase
-  // properties gets moved to a custom behavior
-  // .step(validateRestSpec)
+  .step(validateRestSpec)
+  .step(removeIncompleteEndpoints)
```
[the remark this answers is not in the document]
swallez: Removed!

specification/compiler/index.ts
swallez: I initially went this way so that validation could easily be extracted in a different context.

But since we've decided that it will stay there and ultimately output validation errors in an additional file that will consume these errors, it doesn't make sense anymore. I changed it to be consistent with other functions and have a nice-looking pipeline ;-)

specification/compiler/steps/validate-model.ts
swallez: Removed `object`, it will now be reported as an error!

specification/compiler/steps/validate-model.ts
swallez: Done. It now checks if the `FAIL_HARD` env variable is set.

swallez:
@delvedor I addressed your comments. Can you cross-check?

specification/compiler/steps/validate-model.ts
delvedor: Perfect!

delvedor (approved):
LGTM

Mpdreamz (approved):

PR #172, opened 2021-03-05
update git clean check and print the diff if its not

Mpdreamz:
Switch to using `--porcelain` and now emits `git diff` if it fails so we can debug why the generation is not fresh


delvedor (approved):
LGTM

PR #173, opened 2021-03-08
Updated ML definitions

delvedor:
As titled.

All `Id` issues will be fixed in a separate pr.

Related: https://github.com/elastic/kibana/pull/83808#issuecomment-791479571
/cc @jgowdyelastic

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -315,8 +315,8 @@ export interface AnalysisConfig {
   bucket_span: TimeSpan
   categorization_field_name?: Field
   categorization_filters?: Array<string>
-  detectors?: Array<Detector>
-  influencers?: Fields
+  detectors: Array<Detector>
+  influencers: Fields
   latency?: Time
   multivariate_by_fields?: boolean
   per_partition_categorization?: PerPartitionCategorization
@@ -2676,10 +2676,6 @@ export interface DataDescription {
   time_format?: string
 }
 
-export interface DataFeed {
-  count: long
-}
-
 export interface DataPathStats {
   available: string
   available_in_bytes: long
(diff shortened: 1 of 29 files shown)


output/typescript/types.ts
```diff
@@ -6461,6 +6499,8 @@ export type ModelMemoryStatus = 'ok' | 'soft_limit' | 'hard_limit'
 
 export interface ModelPlotConfig {
   terms: Fields
```
jgowdyelastic, line 6501: should be `string`

output/typescript/types.ts
```diff
@@ -7697,6 +7742,7 @@ export interface PutDatafeedRequest extends RequestBase {
     chunking_config?: ChunkingConfig
     frequency?: Time
     indices?: Indices
+    indexes?: Indices
```
jgowdyelastic, line 7745: `indices` and `indexes` should be `string[]`

output/typescript/types.ts
```diff
@@ -11305,6 +11351,7 @@ export interface UpdateJobRequest extends RequestBase {
     model_snapshot_retention_days?: long
     renormalization_window_days?: long
     results_retention_days?: long
+    groups?: Array<string>
```
jgowdyelastic, line 11354: `groups` doesn't belong here

jgowdyelastic:
will it also be possible to update the version of es these types are generated against so `findFileStructure` is removed from `kibana.d.ts`?

delvedor:
> will it also be possible to update the version of es these types are generated against so findFileStructure is removed from kibana.d.ts?

These types are not generated, but created and maintained by hand, if a type is wrong it should be updated directly here :)

delvedor:
I've added the `text_structure.find_structure` as well, in a subsequent or I'll update all of our json spec files, which are out of date.

specification/specs/x_pack/text_structure/FindStructure.ts
```diff
@@ -0,0 +1,83 @@
...
+interface FindStructureRequest<TBody> {
+  query_parameters?: {
+    charset?: string
+    column_names?: string
+    delimiter?: string
+    explain?: boolean
+    format?: string
+    grok_pattern?: string
+    has_header_row?: boolean
+    lines_to_sample?: uint
```
delvedor, line 34: @philkra 💪

jgowdyelastic (approved):
LGTM!

PR #175, opened 2021-03-08
Validated all cat.* Responses

philkra:
added the missing responses for:
fixed cat.aliases response
* `cat.allocations`
* `cat.indices`
* `cat.nodes`
* `cat.plugins`
* `cat.recovery`
* `cat.segments`
* `cat.shards`
* `cat.threaded_pool`


delvedor (approved):
LGTM

PR #176, opened 2021-03-08
Id should be a string

delvedor:
Finally, the `Id` type is now correct thanks to https://github.com/elastic/clients-flight-recorder/pull/655 🎉


Mpdreamz (approved):
🥇 💯

PR #185, opened 2021-03-09
fix/main/cat continued

Mpdreamz:
- add fields and aliases from server
- update duplicate aliases
- 2nd pass over all the aliases
- clean up quotes on aliases
- revalidated cat namespace
- format fix


philkra (approved):
LGTM

PR #188, opened 2021-03-09
Fix ssl.certificates response

philkra:
as titled.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -27815,7 +27815,7 @@
       "properties": [
         {
           "name": "alias",
-          "required": true,
+          "required": false,
           "type": {
             "kind": "instance_of",
             "type": {
@@ -51886,9 +51886,18 @@
     {
       "inherits": [
         {
+          "generics": [
+            {
+              "kind": "instance_of",
+              "type": {
+                "name": "ClusterCertificateInformation",
+                "namespace": "x_pack.ssl.get_certificates"
+              }
+            }
+          ],
           "type": {
-            "name": "ResponseBase",
-            "namespace": "common_abstractions.response"
+            "name": "Array",
+            "namespace": ".Users.pkrauss.github.com.elastic-client-generator.specification.node_modules.typescript.lib"
           }
         }
       ],
@@ -51897,22 +51906,7 @@
         "name": "GetCertificatesResponse",
         "namespace": "x_pack.ssl.get_certificates"
       },
-      "properties": [
-        {
-          "name": "certificates",
-          "required": true,
-          "type": {
-            "kind": "array_of",
-            "value": {
-              "kind": "instance_of",
-              "type": {
-                "name": "ClusterCertificateInformation",
-                "namespace": "x_pack.ssl.get_certificates"
-              }
-            }
-          }
-        }
-      ]
+      "properties": []
     },
     {
       "attachedBehaviors": [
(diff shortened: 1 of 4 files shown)


output/schema/schema.json
```diff
@@ -51886,9 +51886,18 @@
...
+                "name": "ClusterCertificateInformation",
+                "namespace": "x_pack.ssl.get_certificates"
+              }
+            }
+          ],
           "type": {
-            "name": "ResponseBase",
-            "namespace": "common_abstractions.response"
+            "name": "Array",
+            "namespace": ".Users.pkrauss.github.com.elastic-client-generator.specification.node_modules.typescript.lib"
```
Mpdreamz, line 51900: What a freaky namespace?! cc @swallez @delvedor any clue how this one made it in ?

Mpdreamz (changes requested):

output/schema/schema.json
delvedor: It's "correct", as `Array` is a native typescript type. I think @philkra wanted to use [`ArrayResponseBase`](https://github.com/elastic/elastic-client-generator/blob/master/docs/behaviors.md#arrayresponsebase).

output/typescript/types.ts
```diff
@@ -4677,8 +4677,7 @@ export interface GetCategoriesResponse extends ResponseBase {
 export interface GetCertificatesRequest extends RequestBase {
 }
 
-export interface GetCertificatesResponse extends ResponseBase {
-  certificates: Array<ClusterCertificateInformation>
+export interface GetCertificatesResponse extends Array<ClusterCertificateInformation> {
```
delvedor, line 4680:
```suggestion
export interface GetCertificatesResponse
  extends ResponseBase
  implements ArrayResponseBase<ClusterCertificateInformation> {
```

output/schema/schema.json
philkra: Fixed, @delvedor was on point

specification/specs/x_pack/ssl/get_certificates/GetCertificatesResponse.ts
```diff
@@ -17,7 +17,5 @@
  * under the License.
  */
 
-@class_serializer('GetCertificatesResponseFormatter')
-class GetCertificatesResponse extends ResponseBase {
-  certificates: ClusterCertificateInformation[]
-}
+class GetCertificatesResponse
+  implements ArrayResponseBase<ClusterCertificateInformation> {}
```
delvedor, line 21: I think this should extend from `ResponseBase` as well (cc @Mpdreamz)

Mpdreamz: ++, it should. Another possible validation @swallez :)

swallez: What is the effect of not extending `ResponseBase`? Is this marker interface used to identify endpoint responses?

Mpdreamz: Yeah, we want to be in a position where each api invocation can be thought of as 

`client.api(RequestBase): ResponseBase`

That way clients have a path for abstraction to

`transport.request(....) ...`

PR #196, opened 2021-03-11
Add error response back

delvedor:
As titled, the fix is ugly, but it's needed.

### DIFF
--- a/specification/compiler/steps/validate-model.ts
+++ b/specification/compiler/steps/validate-model.ts
@@ -142,13 +142,7 @@ export default async function validateModel (apiModel: model.Model): Promise<mod
   // The ErrorResponse is not referenced anywhere, but we need to export it
   // as any API could return it if an error happens. It's widely used in
   // the type test as well.
-  let ErrorResponseType: model.Interface | undefined
-  for (const type of apiModel.types) {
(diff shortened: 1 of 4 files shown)


specification/compiler/steps/validate-model.ts
```diff
@@ -139,6 +139,18 @@ export default async function validateModel (apiModel: model.Model): Promise<mod
...
+  // as any API could return it if an error happens. It's widely used in
+  // the type test as well.
+  let ErrorResponseType: model.Interface | undefined
+  for (const type of apiModel.types) {
+    if (type.name.name === 'ErrorResponse') {
+      ErrorResponseType = type as model.Interface
+      break
+    }
+  }
+
```
swallez, line 152: You can replace all this (and the `push` below) with:

```ts
validateTypeRef({namespace: "common_abstractions.response", name: "ErrorResponse"}, undefined, new Set());
```

This will have the added benefits of validating this type definition instead of just adding it to the list of visisted types.

swallez (changes requested):

swallez (approved):
LGTM. And now we can remove "it's ugly" from the description as it's not that bad 😉

PR #199, opened 2021-03-11
Fix Async.* endpoints

philkra:
* `async_search.submit` request + response
* `async_search.get` request

### DIFF
--- a/specification/specs/x_pack/async_search/submit/AsyncSearchSubmitRequest.ts
+++ b/specification/specs/x_pack/async_search/submit/AsyncSearchSubmitRequest.ts
@@ -25,6 +25,10 @@
 interface AsyncSearchSubmitRequest extends RequestBase {
   path_parts?: {
     index?: Indices
+    batched_reduce_size?: long
+    wait_for_completion_timeout?: Time
+    keep_on_completion?: boolean
+    typed_keys?: boolean
   }
   query_parameters?: {}
   body?: {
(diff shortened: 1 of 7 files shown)


specification/specs/x_pack/async_search/submit/AsyncSearchSubmitRequest.ts
```diff
@@ -25,6 +25,10 @@
 interface AsyncSearchSubmitRequest extends RequestBase {
   path_parts?: {
     index?: Indices
+    batched_reduce_size?: long
+    wait_for_completion_timeout?: Time
+    keep_on_completion?: boolean
+    typed_keys?: boolean
```
delvedor, line 31: Are you sure about this? I think this should go in `query_parameters`.

philkra: Now that you mention it! Yes, they should be in `query_parameters`! The tests did pass though 😉

philkra: feedback addressed

delvedor: > The tests did pass though 

Yes, unfortunately, we can't check if a parameter should go in the URL or in the query :S

delvedor (approved):
LGTM

PR #202, opened 2021-03-15
Write validation errors to a json file

swallez:
This PR collects errors of the different compilation/validation pipeline steps and outputs them as JSON in `output/schema/validation-errors.json` (and also in the console, sorted by endpoint name).

The file is structured by endpoint/request & response with also a section for general errors (e.g. on builtin types) which aren't associated to an endpoint.

This file should be used to show model validation errors in [`clients-flight-recorder/recordings/types-validation/types-validation.md`](https://github.com/elastic/clients-flight-recorder/blob/dev/recordings/types-validation/types-validation.md) to provide a complete picture of input specifications "readiness". I'll prepare a PR in that repo once this one is merged.

### DIFF
--- a/specification/compiler/index.ts
+++ b/specification/compiler/index.ts
@@ -23,6 +23,7 @@ import addInfo from './steps/add-info'
 import addDescription from './steps/add-description'
 import validateModel from './steps/validate-model'
 import addContentType from './steps/add-content-type'
+import * as errors from './validation-errors'
 
 const compiler = new Compiler()
 
@@ -35,6 +36,7 @@ compiler
   .step(validateModel)
   .write()
   .then(() => {
+    errors.logErrors()
     console.log('Done')
   })
   .catch(err => {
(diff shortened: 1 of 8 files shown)


specification/compiler/index.ts
```diff
@@ -35,6 +36,7 @@ compiler
   .step(validateModel)
```
Mpdreamz, line 36: LGTM small nitpick: it'd be nicer if `logErrors()` is included in a new method on compiler: `writeErrors()` so that the emit of errors is all one in place.

Mpdreamz (dismissed):

delvedor (approved):
I don't like that much that the error utility is a singleton, but unless we change the generator API to allow returning the model and an array of errors, this is the only way.
It might be a good follow-up for the future.

specification/compiler/index.ts
delvedor: Yes! More or less the same I said in https://github.com/elastic/elastic-client-generator/pull/202#pullrequestreview-612341955 :)

swallez:
I've updated error management so that it's now an additional parameter to the `step` functions.

Can you cross-check?

Mpdreamz (approved):

PR #210, opened 2021-03-15
Fix search request & response

Mpdreamz:
- validate response
- Add new aggregations and fix new tests failures for request


specification/specs/common.ts
```diff
@@ -99,8 +101,8 @@ type ScrollIds = string // TODO: array of ScrollIds
 type CategoryId = string
 type ActionIds = string // TODO: check if this should be an array of ActionId
 
-type Id = string
-type Ids = string | string[]
+type Id = string | long
```
delvedor, line 104: This does not feel right, can you point me to the yaml test that is failing?

Mpdreamz: the `ids_query` accepts both

delvedor: Then we should add this edge case only there, every other API needs a string and never a number :)

philkra: I actually would vote to not have a union for `Id` it's a string everywhere else

Mpdreamz: updated, it was in two places (pinned_query and LikeDocument).

philkra (approved):
lgtm

PR #215, opened 2021-03-18
Point in Time Endpoints

philkra:
added the request & response for:
* `open_point_in_time`
* `close_point_in_time`

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -2918,6 +2918,15 @@ export interface CloseJobResponse extends ResponseBase {
   closed: boolean
 }
 
+export interface ClosePointInTimeRequest extends RequestBase {
+  body?: {
+    id: Id
+  }
+}
+
+export interface ClosePointInTimeResponse extends ResponseBase {
+}
+
 export interface CloseShardResult {
   failures: Array<ShardFailure>
 }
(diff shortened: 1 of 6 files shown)


sethmlarson (dismissed):
LGTM

output/typescript/types.ts
```diff
@@ -7855,6 +7866,14 @@ export interface OpenJobResponse extends ResponseBase {
   opened: boolean
 }
 
+export interface OpenPointInTimeRequest extends RequestBase {
+  index: IndexName
```
stevejgordon, line 7870: Should these be here? I believe the request for this is empty and the index and keep_alive come from the path and query?

output/typescript/types.ts
```diff
@@ -7855,6 +7866,14 @@ export interface OpenJobResponse extends ResponseBase {
...
+  keep_alive?: Time
+}
+
+export interface OpenPointInTimeResponse extends ResponseBase {
+}
```
stevejgordon, line 7875: Should this not include the returned ID?

output/typescript/types.ts
philkra: `keep_alive` is a query param and `index` is in the url path according to the docs: https://www.elastic.co/guide/en/elasticsearch/reference/master/point-in-time-api.html#point-in-time-api-example

stevejgordon: Sorry, didn't realise I was looking at the output!

output/typescript/types.ts
philkra: done with https://github.com/elastic/elastic-client-generator/pull/215/commits/c0a2c666de5c642c73b33602abf3d896bc598a65 - great catch, thanks!

sethmlarson: Shouldn't this have been caught via validator? Or are there no tests for PIT APIs?

sethmlarson (approved):

output/typescript/types.ts
philkra: I had the same train of thought, but thank god we have @stevejgordon :)

stevejgordon: Haha! Thank God I'd recently been poking around those APIs! 😄

PR #223, opened 2021-03-22
Validate EQL endpoints

philkra:
Validate all eql.* endpoints.

### DIFF
--- a/specification/specs/x_pack/eql/EqlHits.ts
+++ b/specification/specs/x_pack/eql/EqlHits.ts
@@ -17,8 +17,8 @@
  * under the License.
  */
 
-class EqlHits {
+class EqlHits<TEvent> {
   total?: TotalHits
-  events?: string[]
-  sequences?: string[]
+  events?: EqlHitsEvent<TEvent>[]
+  sequences?: EqlHitsSequence<TEvent>[]
 }
--- a/specification/specs/x_pack/eql/EqlHits.ts
+++ b/specification/specs/x_pack/eql/EqlHits.ts
@@ -17,4 +17,8 @@
  * under the License.
  */
 
-class EqlSearchResponse extends EqlSearchResponseBase {}
+class EqlHits {
+  total?: TotalHits
+  events?: string[]
+  sequences?: string[]
+}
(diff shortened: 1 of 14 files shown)


specification/specs/x_pack/eql/EqlHits.ts
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+class EqlHits<TEvent> {
+  total?: TotalHits
+  events?: EqlHitsEvent<TEvent>[]
+  sequences?: EqlHitsSequence<TEvent>[]
```
delvedor, line 23: I think these properties are always present, so there is no need to add the `?`.

specification/specs/x_pack/eql/EqlHitsEvent.ts
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+class EqlHitsEvent<TEvent> {
+  _index?: IndexName
+  _id?: Id
+  _source?: TEvent
+  fields?: Dictionary<Field, UserDefinedValue[]>
```
delvedor, line 24: Same here.

specification/specs/x_pack/eql/EqlHits.ts
philkra: `sequences` are conditional to the `sequence` keyword in a EQL query and if `sequences` is an available property, the `events` is not 😄 
This tests shows that `total` is also conditional: `xpack-Execute_EQL_events_query_with_filter_path_2_response.json.test-d.ts`

In other words all 3 properties must stay optional.

specification/specs/x_pack/eql/EqlHitsEvent.ts
philkra: No everything is optional as well, look at this test: `xpack-Execute_EQL_sequence_with_filter_path_2_response.json.test-d.ts` e.g. where only `fields` is available.

delvedor: The reason is that the request is using `filter_path`, but in reality, those properties should be required.
This is something we should fix on the validator level.

delvedor (approved):
This is now good to go!

PR #229, opened 2021-03-26
Update ClearApiKeyCacheRequest

stevejgordon:
During .NET code-gen I noticed that the path parameter should be `Ids`.


philkra (approved):
LGTM

PR #230, opened 2021-03-26
Remove filter_path from EQL Search

stevejgordon:
`filter_path` is specified in the `CommonQueryParameters` behaviour already.


philkra (approved):
LGTM

PR #232, opened 2021-03-26
Address feedback from Kibana

delvedor:
Related: https://github.com/elastic/elastic-client-generator/issues/197


specification/specs/indices/mapping_management/put_mapping/PutMappingRequest.ts
```diff
@@ -47,7 +48,7 @@ interface PutMappingRequest extends RequestBase {
     index_field?: IndexField
     meta?: Dictionary<string, UserDefinedValue>
     numeric_detection?: boolean
-    properties?: Dictionary<PropertyName, PropertyBase>
+    properties?: Dictionary<PropertyName, Property>
```
delvedor, line 51: Fixing the properties types is a beast. @swallez and me are working on the `variants` feature of the compiler to improve the definitions here. As is, this API can't be fixed until we finish the `variants` feature.

Mpdreamz (approved):
Lgtm. We should open a ticket for model validation that ensures the  unions are exhaustive.

PR #233, opened 2021-03-26
Switch to aliases

stevejgordon:
Any reason not to use aliases on these? Today .NET has a custom formatter to handle the oddities of the naming here. I think the approach I've applied on Cat*Records for alias generation should avoid that custom handling.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -49470,43 +49470,24 @@
           }
         },
         {
+          "aliases": [
+            "resource.id"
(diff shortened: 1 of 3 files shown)


stevejgordon:
Ah! Perhaps not then as the formatter is unhappy with this.

stevejgordon:
Doh! One extra space has been dealt with!

delvedor (approved):
LGTM

PR #234, opened 2021-03-26
Added variants handling

delvedor:
As titled.

External variant example:
```ts
/** @variants external */
type FooContainer = Faz | Bar

/** @variant name='faz' */
class Faz {
  faz: string
}

/** @variant name='bar' */
class Bar {
  bar: string
}
```

Internal variant example:
```ts
class TestPolymorphicBase {
  baz: boolean
}

class TestPolymorphic extends TestPolymorphicBase {
  type: 'heya'
}

enum Foo {
  hello = 1,
  ciao = 2
}

class TestPolymorphic2 extends TestPolymorphicBase {
  type: Foo
}

/** @variants internal tag='type' */
type TestPolymorphicType = TestPolymorphic | TestPolymorphic2
```

TODO:
- [x] Docs!
- [x] Update `Property` ~~and `TokenFilter`~~ to use the new variant utilities. (depends on https://github.com/elastic/elastic-client-generator/pull/232)


Mpdreamz (dismissed):

specification/compiler/model/metamodel.ts
```diff
@@ -179,6 +196,8 @@ export class Interface extends BaseType {
    */
   attachedBehaviors?: string[]
   properties: Property[]
+  /** Identify containers */
+  variants?: ExternalTag
```
swallez, line 200: Variants on interfaces can now only be used for containers. Even if from the json perspective the tag is _external_ to the data structure, it's a bit weird to have `@variants external` on a type that lists all variants (properties are _internal_ to the type).

So what about a more explicit syntax in the spec:
```ts
/** @variants container */
class QueryContainer {
  bool?: BoolQuery
  boosting?: BoostingQuery
  common?: SingleKeyDictionary<Union<CommonTermsQuery, string>>
  constant_score?: ConstantScoreQuery
  ...
```

And then in the metamodel:

```ts
type Variants = ExternalTag | InternalTag | Container

class Container {
  kind: 'container'
}

export class Interface extends BaseType {
  /** Identify containers */
  variants?: Container 
  ...

export class Union extends BaseType {
  /** Identify typed_key unions (external) and variant inventories (internal) */
  variants?: ExternalTag | InternalTag
  ...
```

This could also be a simple boolean flag, but making it part of `Variant` ensures this type lists all possible variations of the `@variants` annotation.

And ultimately that will allow us to drop the technical `Container` suffix that is now redundant and have a nice `Query` type 😉

specification/compiler/model/metamodel.ts
```diff
@@ -267,6 +274,8 @@ export class TypeAlias extends BaseType {
   type: ValueOf
   /** generic parameters: either concrete types or open parameters from the enclosing type */
   generics?: ValueOf[]
+  /** Identify typed_key unions (external) and variant inventories (internal) */
+  variants?: Variants
```
swallez, line 278: With the `container` variant, this becomes:

```ts
variants?: ExternalTag | InternalTag
```

specification/compiler/model/utils.ts
```diff
@@ -634,6 +653,50 @@ export function parseJsDocTags (jsDoc: JSDoc[]): Record<string, string> {
...
+  const tags = parseJsDocTags(jsDoc)
+  if (tags.variants == null) {
+    return undefined
+  }
+
+  const [type, value] = tags.variants.split(' ')
+  if (type === 'external') {
+    return { kind: 'external_tag' }
+  }
+
```
swallez, line 670: Add:

```ts
  if (type === 'container') {
    return { kind: 'container' }
  }
```

swallez (changes requested):
Loving how we end up with something simple yet capturing all that is needed for strongly typed languages!

Added some comments to add a 3rd variant kind `container`.

specification/compiler/model/metamodel.ts
Mpdreamz: > And ultimately that will allow us to drop the technical Container suffix that is now redundant and have a nice Query type 😉

All for removing [smurfisms](https://devcards.io/smurf-naming-convention) but the container is not the query. It's the thing carrying the query. 

```
query
-- terms: Terms query
```

Is also confusing I think. 

Gets even more ambiguous with aggregations and nesting.

Mpdreamz: +1 on explicitly tagging it as variant though

delvedor: Agreed!

specification/compiler/model/metamodel.ts
```diff
@@ -179,6 +200,8 @@ export class Interface extends BaseType {
    */
   attachedBehaviors?: string[]
   properties: Property[]
+  /** Identify containers */
+  variants?: ExternalTag
```
swallez, line 204: This should be:

```suggestion
  variants?: Container
```

swallez (approved):
LGTM once https://github.com/elastic/elastic-client-generator/pull/234/files#r602496498 has been fixed. This is beautiful!

delvedor:
This is good to go 🎉 
I've fixed all `*Property` issues, there are still a few errors in `indices.put_mapping`, but not related to this.

specification/compiler/model/build-model.ts
```diff
@@ -257,6 +259,17 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
...
     hoistTypeAnnotations(type, declaration.getJsDocs())
 
+    const variant = parseVariantNameTag(declaration.getJsDocs())
+    if (typeof variant === 'string') {
+      type.variantName = variant
+    }
+
+    const variants = parseVariantsTag(declaration.getJsDocs())
+    if (variants != null) {
+      assert(variants.kind === 'container', 'Interfaces can only use `container` vairant kind')
```
swallez, line 269: Typo.

```suggestion
      assert(variants.kind === 'container', 'Interfaces can only use `container` variant kind')
```

swallez (dismissed):
LGTM (still found a typo)

Awesome stuff, I'll add the `@variants` annotations and additional model validations in a follow-up PR.

PR #235, opened 2021-03-26
Validate eql.search Response

philkra:
as titled.

Needed `EmptyObject` union to `hits` property to resolve the no hits found case.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -49288,7 +49288,7 @@
       "properties": [
         {
           "name": "total",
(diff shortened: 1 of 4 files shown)


delvedor (changes requested):
I don't think this is the right change, how about marking as optional `EqlHits.total`?
It's unfortunate that this does not behave like the search API.

philkra:
@delvedor - feedback addressed

delvedor (approved):
LGTM

PR #237, opened 2021-03-26
Add missing searchable snapshot Endpoints

philkra:
added request & response for:
* searchable_snapshots.clear_cache
* searchable_snapshots.mount
* searchable_snapshots.repository_stats
* searchable_snapshots.stats

the endpoints don't validate (yet) just the classes added ..


delvedor (changes requested):
You should add the stubs in requests definitions as well, otherwise, the typescript client output will generate bad types that should not be there.

philkra:
@delvedor - added the stubbing to the requests and validated the mount endpoint.

delvedor (approved):
LGTM

PR #240, opened 2021-03-26
Validated transform.get_transform Request

philkra:
as titled.

### DIFF
--- a/specification/specs/x_pack/transform/get_transform/GetTransformRequest.ts
+++ b/specification/specs/x_pack/transform/get_transform/GetTransformRequest.ts
@@ -25,12 +25,12 @@
 interface GetTransformRequest extends RequestBase {
   path_parts?: {
     transform_id?: Name
-    exclude_generated?: boolean
   }
   query_parameters?: {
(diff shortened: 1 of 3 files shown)


specification/specs/x_pack/transform/get_transform/GetTransformRequest.ts
```diff
@@ -25,6 +25,7 @@
 interface GetTransformRequest extends RequestBase {
   path_parts?: {
     transform_id?: Name
+    exclude_generated?: boolean
```
delvedor, line 28: This should go in the [querystring](https://www.elastic.co/guide/en/elasticsearch/reference/current/get-transform.html#get-transform-query-parms).

delvedor (changes requested):

specification/specs/x_pack/transform/get_transform/GetTransformRequest.ts
philkra: Yes, sorry about that and great catch!

delvedor (approved):
LGTM

PR #242, opened 2021-03-26
Validated watcher.stats Response

philkra:
as titled.

@delvedor, I wanted to validated the request as well but ran into a type casting issue:
```
expectAssignable<T.WatcherStatsRequest>({
  "metric": "all",
  "emit_stacktraces": "true"
})
```

and `emit_stacktraces` is of type `boolean` in `WatcherStatsRequest.ts` 🤷

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -124486,7 +124486,7 @@
       "properties": [
         {
           "name": "current_watches",
-          "required": true,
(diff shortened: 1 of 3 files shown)


delvedor (approved):
Very likely the yaml test is using `"true"` instead of `true`. It should be fixed there.

PR #260, opened 2021-03-29
Update the metamodel with cleanups and consistency improvements

swallez:
This PR updates the metamodel with a few cleanups and consistency improvements:

- Remove the `internal.Array` type: `Array<SomeType>` now compiles to a `array_of` value.

- Rename `BaseType.url` to `BaseType.docUrl` to be consistent with `Endpoint.docUrl` and add `Property.docUrl`. Endpoints, types and properties now all have `description`, `docUrl` and `deprecation`.

- Remove `Implements` and use `Inherits` everywhere.

- Change open generic parameters from `string[]` to `TypeName[]`. The namespace is arbitrary (the fact that it's the enclosing type's namespace is an implementation detail). This fully qualified type name is the one used to reference that open generic in property types. No more guess-based-on-convention to resolve type names.

### DIFF
--- a/specification/compiler/model/utils.ts
+++ b/specification/compiler/model/utils.ts
@@ -194,6 +194,15 @@ export function modelType (node: Node): model.ValueOf {
 
       const name = identifier.compilerNode.escapedText as string
       switch (name) {
+        case 'Array':
+          assert(node.getTypeArguments().length === 1, 'An array must have one argument')
+          const [value] = node.getTypeArguments().map(node => modelType(node))
+          const type: model.ArrayOf = {
+            kind: 'array_of',
+            value
+          }
+          return type
+
         case 'Dictionary':
         case 'AdditionalProperties': {
           assert(node.getTypeArguments().length === 2, 'A Dictionary must have two arguments')
@@ -276,7 +285,7 @@ export function modelInherits (node: HeritageClause): model.Inherits[] {
  * A class could implement from multiple classes, which are
  * defined inside the node typeArguments.
  */
-export function modelImplements (node: ExpressionWithTypeArguments): model.Implements {
+export function modelImplements (node: ExpressionWithTypeArguments): model.Inherits {
   const generics = node.getTypeArguments().map(node => modelType(node))
   return {
     type: {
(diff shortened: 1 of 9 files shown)


specification/compiler/model/utils.ts
```diff
@@ -489,29 +499,31 @@ export function hoistRequestAnnotations (
...
+    } else if (tag === 'doc_url') {
+      type.docUrl = value
     } else throw new Error(`Unhandled tag: '${tag}' with value: '${value}' on type ${type.name.name}`)
   })
 }
 
 /** Lifts jsDoc type annotations to fixed properties on Property */
 function hoistPropertyAnnotations (property: model.Property, jsDocs: JSDoc[]): void {
-  const validTags = ['stability', 'prop_serializer', 'url', 'aliases', 'identifier']
+  const validTags = ['stability', 'prop_serializer', 'doc_url', 'aliases', 'identifier', 'url']
```
delvedor, line 517: I think you can remove `url` from here.

```suggestion
  const validTags = ['stability', 'prop_serializer', 'doc_url', 'aliases', 'identifier']
```

typescript-generator/index.js
```diff
@@ -211,8 +211,8 @@ function buildGenerics (type, noDefault = false) {
 
   // generics can either be a value/instance_of or a named generics
   function buildGeneric (type) {
-    return typeof type === 'string'
-      ? (noDefault ? type : `${type} = unknown`)
+    return typeof type.namespace === 'string'
```
delvedor, line 214: Not sure why you are checking `namespace` here:
```suggestion
    return typeof type.name === 'string'
```

delvedor (changes requested):
Great work! The new schema looks a lot nicer!
Left a couple of notes :)

typescript-generator/index.js
swallez: Execution was failing because the `string` is changed to a `TypeName`. Since it's JS and not TS I'm not sure what other types can reach this function, so my thinking is that the `name` property was more likely to appear on other types than `namespace` that is very specific to `TypeName`.

Checking for the existence of `kind` would actually be more correct IHMO since this is what is used in `buildType`. So what about this:

```js
  function buildGeneric (type) {
    return typeof type.kind === 'string'
      ? buildType(type)
      : (noDefault ? type.name : `${type.name} = unknown`)
  }
```

delvedor: Go for it!

swallez: Done. Please cross-check!

delvedor (approved):
LGTM

PR #263, opened 2021-03-30
Remove duplicate properties which appear on base types

stevejgordon:
The following were flagged in compiler warnings in the .NET generated code because the types included properties which duplicates properties from types they extend.


delvedor (approved):
LGTM

PR #265, opened 2021-03-30
watcher.stats Request

philkra:
The request is failing to validate based on the casting of `true` a string `"true"`. The [docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.12/watcher-api-stats.html) state that `emit_stacktraces` is of type `boolean` but the compiler presents a string:
```
expectAssignable<T.WatcherStatsRequest>({
  "metric": "all",
  "emit_stacktraces": "true"
})
```

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -11223,7 +11223,7 @@
         "name": "WatcherStatsResponse",
         "namespace": "x_pack.watcher.watcher_stats"
       },
-      "since": "0.0.0",
+      "since": "5.5.0",
       "stability": "stable",
       "urls": [
         {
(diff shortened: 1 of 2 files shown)


delvedor (commented):
Unfortunately, this is an issue in the yaml test, it should be fixed there.

philkra:
Opened PR to assert type sanity in the upstream project: https://github.com/elastic/elasticsearch/pull/71361

philkra:
With https://github.com/elastic/elasticsearch/pull/71361 merged, I assume this can be merged as it's only adjusting the version string.

delvedor (approved):

PR #268, opened 2021-03-31
Introduce SequenceNumber type

philkra:
add type `SequenceNumber` of type `integer`, that represents all `*seq_no: [int,long]` occurrences.


delvedor (approved):
LGTM

PR #281, opened 2021-03-31
Fix mget Request

philkra:
as titled.


delvedor (approved):
According to the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html) the id should be a string. I guess you added the `| integer` because in the yaml test they are using numbers as well.
Let's keep it as is for know, but probably we'll need to revisit.

PR #289, opened 2021-04-01
Validated indices.get_data_stream Response

philkra:
as titled.


specification/specs/x_pack/data_streams/get_data_stream/IndicesGetDataStreamResponse.ts
```diff
@@ -18,6 +18,26 @@
...
+}
+
+class IndicesGetDataStreamItem {
+  name: DataStreamName
+  timestamp_field: IndicesGetDataStreamItemTimestampField
+  indices: IndicesGetDataStreamItemIndex[]
+  generation: integer
+  template: Name
+  hidden: boolean
+  status: string // TODO: use type Health --> must be case insensitive
```
delvedor, line 31: You can define the enum like we are doing in the query dsl:

https://github.com/elastic/elastic-client-generator/blob/174b5bd7c8be68b78e64f8f38efcabaef78a6119/specification/specs/query_dsl/Operator.ts#L20-L25

philkra: Added `DataStreamHealthStatus` with the aliases

delvedor (changes requested):
LGTM once the conflicts are resolved! :)

PR #291, opened 2021-04-01
Improve indices.validate_query

philkra:
Response is validating. Request is still failing but eliminated 7 failures.


delvedor (approved):
LGTM

PR #292, opened 2021-04-01
Annotate `data_frozen` with the according version

philkra:
as titled.


delvedor (commented):
The version tag should be `@since`, but we haven't formalized it yet.

philkra:
The compiler refuses to accept the `@since` annotation for class members. Left "it" as comment for later

delvedor (approved):
LGTM

PR #295, opened 2021-04-01
Validate ml.post_data endpoint

philkra:
as titled.


specification/specs/x_pack/machine_learning/post_job_data/PostJobDataRequest.ts
```diff
@@ -31,7 +31,6 @@ interface PostJobDataRequest extends RequestBase {
     reset_end?: DateString
     reset_start?: DateString
   }
-  body?: {
-    data?: UserDefinedValue[]
-  }
+  // TODO: should be name JSON Document
+  body: UserDefinedValue
```
delvedor, line 35: According to the doc and the rest-api-spec, this API expects a bulk body, we should express it in the same way of the [BulkRequest.ts](https://github.com/elastic/elastic-client-generator/blob/master/specification/specs/document/multiple/bulk/BulkRequest.ts).

```ts
interface PostJobDataRequest<TSource> extends RequestBase {
  path_parts?: {
    job_id: Id
  }
  query_parameters?: {
    reset_end?: DateString
    reset_start?: DateString
  }
  body?: TSource[]
}
```

Also, we are not signaling in any way in the input spec that an API requires a ndjson body, we should do it. I'll open a separate issue for this.

delvedor: https://github.com/elastic/elastic-client-generator/issues/297

philkra: Added a generic as body, but it must be a json not ndjson, see [docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/ml-post-data.html#_request_body_23)

delvedor: From the docs:

> A sequence of one or more JSON documents containing the data to be analyzed. Only whitespace characters are permitted in between the documents.

The rest-api-spec for this endpoint:

```jsoni
    "body":{
      "description":"The data to process",
      "required":true,
      "serialize":"bulk"
    }
```

philkra:
@delvedor , as discussed - I assigned to the PR to you to look why `body: TJsonDocument[] | TJsonDocument` is failing to generate.

sethmlarson:
I think this is superseded by https://github.com/elastic/elasticsearch-specification/pull/515

PR #296, opened 2021-04-01
Validate ml.stop_datafeed Request

philkra:
as titled.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -115703,7 +115703,7 @@
           "type": {
             "kind": "instance_of",
             "type": {
-              "name": "Id",
+              "name": "Ids",
               "namespace": "internal"
             }
           }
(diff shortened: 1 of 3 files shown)


delvedor (changes requested):
The documentation looks different 🤔 
https://www.elastic.co/guide/en/elasticsearch//reference/7.12/ml-stop-datafeed.html

`allow_no_datafeeds` should be `allow_no_match` and `force` should only be present in the body.

philkra:
good catch with `allow_no_datafeeds`, removed it and added `allow_no_match` - I assume the tests don't cover this property.

`force` must be also present in the query params as one of the tests requests it to be in query:
```
  workbench/xpack-Test_force_stop_datafeed_6_request.json.test-d.ts:5:2
  ✖  5:2  Argument of type { "datafeed_id": string; "force": boolean; } is not assignable to parameter of type StopDatafeedRequest.
  Object literal may only specify known properties, and "force" does not exist in type StopDatafeedRequest.  
```

delvedor (approved):
LGTM

PR #307, opened 2021-04-02
Validate ML DataCounts and Datafeed type

philkra:
addressed feedback from https://github.com/elastic/elastic-client-generator/issues/252

Added missing properties and validated endpoints:
* `ml.get_datafeeds`
* `ml.get_job_stats`


delvedor (approved):
LGTM

jgowdyelastic:
@delvedor thank you for fixing these types, will they automatically appear in the next version of the elastic client on master?

PR #314, opened 2021-04-05
Fix license.delete Response

philkra:
as titled. Validated based on [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/delete-license.html) will add tests to recorder separately.


delvedor (approved):
LGTM

Issue #267, opened 2021-03-31
msearch api body asking for object but examples are using an array


dhurley14:
## 💬 Questions and Help

The [type defined](https://github.com/elastic/elasticsearch-js/blob/master/api/types.d.ts#L7718-L7720) for the `msearch` api request is a `body` property typed as an object with nested `operations` but the [example docs](https://github.com/elastic/elasticsearch-js/blob/master/docs/examples/msearch.asciidoc) show the operations passed into the body property as an array of objects.

Curious which form is the expected format for the request? Do the docs need to be updated? Is the request typed incorrectly?

Thanks!

### Please note that this issue tracker is not a help forum and this issue may be closed.

It's not uncommon that somebody already opened an issue or in the best case it's already fixed but not merged. That's the reason why you should [search](https://github.com/elastic/elasticsearch-js/issues) at first before submitting a new one.

delvedor:
Hello! The types are wrong, it should be a ndjson body :)
Thank you for reporting!

dhurley14:
Great thanks!

delvedor:
Closed in #318.

PR #318, opened 2021-04-07
Fix msearch and indices.put_mapping

### LINKED ISSUE #258: Interface for RuntimeField missing string option
## 🐛 Wrong type

From the documentation, it seems like the script field in runtime mapping can be [a string](https://www.elastic.co/guide/en/elasticsearch/reference/master/runtime-examples.html#runtime-examples-grok) or [a script object](https://www.elastic.co/guide/en/elasticsearch/reference/master/runtime-retrieving-fields.html#runtime-define-field-dayofweek). So the follow two examples should be both valid:

```javascript
PUT my-index/_mappings
{
  "runtime": {
    "http.clientip": {
      "type": "ip",
      "script": """
        String clientip=grok('%{COMMONAPACHELOG}').extract(doc["message"].value)?.clientip;
        if (clientip != null) emit(clientip); 
      """
    }
  }
}
```

```javascript
  "runtime_mappings": {
    "day_of_week": {
      "type": "keyword",
      "script": {
        "source": "emit(doc['@timestamp'].value.dayOfWeekEnum.getDisplayName(TextStyle.FULL, Locale.ROOT))"
      }
    }
  },
```

The `type: FieldType` is also incorrect here because currently, [runtime supports the following types](https://www.elastic.co/guide/en/elasticsearch/reference/7.12/runtime-mapping-fields.html#runtime-mapping-fields):
```
boolean
date
double
geo_point
ip
keyword
long
```

## Definition

If possible provide a snippet with the fix.

```diff
export type RuntimeType = 'keyword' | 'long' | 'double' | 'date' | 'ip' | 'boolean'

export interface RuntimeField {
  format?: string
  script?: string | StoredScript
  type: RuntimeType
}
```

delvedor:
As titled.

Closes #258 #267


philkra (approved):
LGTM

PR #323, opened 2021-04-08
Improve error UX

delvedor:
With this pr assertions will no longer show which part of the compiler throws the error, but which file and line(s) of the input spec caused such error.
Furthermore, some error messages have been improved.


philkra (approved):
LGTM

PR #324, opened 2021-04-08
Introduce since, description and default tags

### LINKED ISSUE #251: Default, Version & Description Parameter Annotations
## 🚀 Feature Proposal

Enable the enrichment of request parameters via annotation with
* a default value
* version introduced 
* parameter description
All values should be optional and this must been seen as a long term goal to provide this data with collaboration of other teams.

## Motivation

1. Developer experience, by getting a hint on the default value of a parameter and a proper description.
2. The docs team could provide more elaborate generated documentation.
3. The UI could add more context to UI fields and generate larger parts of forms.

## Example

Example definition of a param within a request class:
```
/**
 * @default 30s
 * @description This is the description of a parameter that was introduced in 7.8.0 with the default value 30s
 * @version 7.8.0
 */
a_parameter: string
```

json output:

```
{
  "name": "a_parameter",
  "required": true,
  "default": "30s",
  "description": "This is the description of a parameter that was introduced in 7.8.0 with the default value 30s",
  "version": "7.8.0",
  "type": {
    "kind": "instance_of",
    "type": {
      "name": "SearchableSnapshotsUsage",
      "namespace": "x_pack.info.x_pack_usage"
    }
  }
}
```

delvedor:
Closes #251

### DIFF
--- a/docs/modeling-guide.md
+++ b/docs/modeling-guide.md
@@ -279,3 +279,66 @@ class FooContainer {
   faz: FazDefinition
 }
 ```
+
+### Additional information
+
+If needed, you can specify additional information on each type with the approariate JSDoc tag.
+Following you can find a list of the supported tags:
+
+#### `@since`
+
+Every API already has a `@since` tag, which describes when an API has been added.
+You can specify an additional `@since` tag for every parameter that has been added afterwards.
+If the tag is not defined, it's assumed that the parameter has been added with the API the first time
+
+```ts
+/**
+ * @since 7.10.0
+ */
+class FooRequest {
+  bar: string
+  /** @since 7.11.0 */
+  baz: string
+  faz: string
+}
+```
+
+#### `@description`
+
+A custom description of the property.
+
+```ts
+class Foo {
+  bar: string
+  /** @description You can baz! */
+  baz: string
+  faz: string
+}
+```
+
+#### `@default`
+
+The default value if the property is not specified.
+Default values can only be specified on optional properties.
+
+```ts
+class Foo {
+  bar: string
+  /** @default hello */
+  baz?: string
+  faz: string
+}
+```
+
+#### `@doc_url`
+
+The documentation url for the parameter.
+
+```ts
+class Foo {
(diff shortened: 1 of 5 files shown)


philkra (approved):
LGTM

docs/modeling-guide.md
```diff
@@ -278,4 +278,67 @@ class FooContainer {
...
+
+### Additional information
+
+If needed, you can specify additional information on each type with the approariate JSDoc tag.
+Following you can find a list of the supported tags:
+
+#### `@since`
+
+Every API already has a `@since` tag, which describes when an API has been added.
+You can specify an additional `@since` tag for every parameter that has been added afterwards.
```
swallez, line 291: "property" would be more suited than "parameter" IMHO as this applies to path & query parameters and json properties, which are all represented as TS structure properties

docs/modeling-guide.md
```diff
@@ -278,4 +278,67 @@ class FooContainer {
...
+```
+
+#### `@description`
+
+A custom description of the property.
+
+```ts
+class Foo {
+  bar: string
+  /** @description You can baz! */
```
swallez, line 313: Do we need an annotation for this ? This should be the jsdoc comment text:

```ts
/**
 * You can baz!
 */
```

And BTW we also should define the markup langage in descriptions. My suggestion is to use [CommonMark](https://commonmark.org/) that is implemented in all languages and that code generators can translate to whatever markup language is used in their respective documentation (some HTML with custom markers in JavaDocs, reStructuredText in Python, etc)

swallez (changes requested):

docs/modeling-guide.md
delvedor: Good suggestion, let me see what I can do about it.
Regarding the markup, I like the idea, but let's discuss it in a separate issue.

swallez (approved):
LGTM

philkra (approved):
LGTM - I like `server_default` much better!

PR #325, opened 2021-04-09
Validate Cluster.health Response

philkra:
* validated and annotated `cluster.health` response
* introduced new type `HttpHeaders` for, well HTTP headers :)


delvedor (approved):
LGTM

PR #326, opened 2021-04-09
Add visibility to Endpoint definition

### LINKED ISSUE #228: Add `visibility` property to the canonical spec
## 🚀 Feature Proposal

The `visibility` property has landed in the rest-api-spec, [bulk](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/bulk.json#L8) for example and generates value for every consumer of the canonical spec. Please add that property to the endpoint definition, next to `stability` in the tree.

## Motivation

as previously described.

## Example

```
{
      ...
      "since": "7.7.0",
      "stability": "stable",
      "visbility": "public",
      "urls": [
        ...
      ]
    },
```

delvedor:
Closes #228.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -33,7 +33,8 @@
           ],
           "path": "/_async_search/{id}"
         }
-      ]
+      ],
+      "visibility": "public"
     },
     {
       "accept": [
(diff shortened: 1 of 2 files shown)


philkra (approved):
LGTM

PR #330, opened 2021-04-11
Address Indices.get settings Response

philkra:
Mostly cosmetics with annotations and descriptions.

The primary cause of failing tests is based on upstream tests using string instead of int, e.g.
```
"hidden": "true",
"number_of_shards": "1",
"number_of_replicas": "1",
```

should be 
```
"hidden": true,
"number_of_shards": 1,
"number_of_replicas": 1,
```

Those required changed in the integration tests will addressed next.

this PR can be merged as it might uncover some cross-class failures because  refactored `IndexSettings`.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -72170,6 +72170,7 @@
           ],
           "name": "number_of_shards",
           "required": false,
+          "serverDefault": "1",
           "type": {
             "kind": "instance_of",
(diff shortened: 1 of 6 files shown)


philkra:
Opened https://github.com/elastic/elasticsearch/pull/71554 to address the required changes.

PR #333, opened 2021-04-12
Tag polymorphic types. Part 1: containers

swallez:
Add `variants` tag to polymorphic types.

This first part is about container, union types will come in a follow-up PR, as their current model (in particular agregations) requires some more in-depth refactoring to represent typed keys.

A new annotation is introduced for properties, `@variant container_property` to identify properties that belong to the container and aren't part of the variants. This is the case of `aggs` and `meta` on aggregations.

This PR also includes a "json ambiguity check" in the model validation, that verifies that the json structure of the model can be analyzed unambiguously with a streaming JSON parser, which is what strongly typed languages generally use to avoid creating an intermediate in-memory JSON AST.

This validation is currently disabled as polymorphic unions cause a lot of errors, which will go away once they are modelled and tagged correctly.

### DIFF
--- a/specification/compiler/model/utils.ts
+++ b/specification/compiler/model/utils.ts
@@ -552,11 +552,8 @@ function hoistPropertyAnnotations (property: model.Property, jsDocs: JSDoc[]): v
     } else if (tag === 'doc_url') {
       property.docUrl = value
     } else if (tag === 'variant') {
-      if (value === 'container_property') {
-        property.container_property = true
-      } else {
-        throw new Error(`Unknown 'variant' value '${value}' on property ${property.name}`)
-      }
+      assert(jsDocs, value === 'container_property', `Unknown 'variant' value '${value}' on property ${property.name}`)
+      property.container_property = true
     } else if (tag === 'since') {
       assert(jsDocs, semver.valid(value), `${property.name}'s @since is not valid semver: ${value}`)
       property.since = value
--- a/specification/compiler/model/utils.ts
+++ b/specification/compiler/model/utils.ts
@@ -47,7 +47,8 @@ export const knownBehaviors = [
   'ArrayResponseBase',
   'EmptyResponseBase',
   'CommonQueryParameters',
-  'CommonCatQueryParameters'
+  'CommonCatQueryParameters',
+  'TypedKeysContainer'
 ]
 
 /**
@@ -449,14 +450,9 @@ function setTags<TType extends model.BaseType | model.Property | model.EnumMembe
     }
   }
 
-  function getName (type: TType): string {
-    if (typeof type.name === 'string') {
-      return type.name
-    } else if (typeof type.name.name === 'string') {
-      return type.name.name
-    } else {
-      return 'unknown'
-    }
+  function getName (type): string {
+    if (type.name?.name != null) return type.name.name
+    return type.name
   }
 }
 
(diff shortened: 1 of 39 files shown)


specification/compiler/model/utils.ts
```diff
@@ -449,14 +450,9 @@ function setTags<TType extends model.BaseType | model.Property | model.EnumMembe
...
-    if (typeof type.name === 'string') {
-      return type.name
-    } else if (typeof type.name.name === 'string') {
-      return type.name.name
-    } else {
-      return 'unknown'
-    }
+  function getName (type): string {
+    if (type.name?.name != null) return type.name.name
+    return type.name
```
delvedor, line 455: Are we sure that `type.name` always exists?
```suggestion
    return type.name ?? 'unknown'
```

specification/compiler/model/utils.ts
```diff
@@ -555,6 +551,12 @@ function hoistPropertyAnnotations (property: model.Property, jsDocs: JSDoc[]): v
       property.identifier = value
     } else if (tag === 'doc_url') {
       property.docUrl = value
+    } else if (tag === 'variant') {
+      if (value === 'container_property') {
+        property.container_property = true
+      } else {
+        throw new Error(`Unknown 'variant' value '${value}' on property ${property.name}`)
+      }
```
delvedor, line 559: Please use the internal assertion tool, which will show to the user which part of the input spec caused the error :)
```suggestion
      assert(jsDocs, value === 'container_property', `Unknown 'variant' value '${value}' on property ${property.name}`)
      property.container_property = true
```

delvedor (changes requested):
Looks good, left a couple of comments.

I think we should try to move the variant validation directly in the compiler, so we can show to the user which part of the spec caused the error. Do you think is possible?

specification/compiler/model/utils.ts
swallez: Yes, `type` is `model.BaseType | model.Property | model.EnumMembe` that all have a `type.name` property that is either a `string` or a `TypeName`. Otherwise the TS compiler and linter would complain 😉

swallez:
> I think we should try to move the variant validation directly in the compiler, so we can show to the user which part of the spec caused the error.

That may be difficult as the validation works on the high level representation of the model, and not the AST because it has to traverse a lot of definitions. Now what we can do (and would also be useful in other contexts) is to add a `sourceLocation: string` property to `BaseType` and `Property` in the model, so that we can know where they come from. That way even the most complex validations will be able to point to the relevant source code location.

specification/specs/x_pack/async_search/get/AsyncSearchGetRequest.ts
```diff
@@ -25,7 +25,6 @@
 interface AsyncSearchGetRequest extends RequestBase {
   path_parts?: {
     id: Id
-    typed_keys?: boolean
```
delvedor, line 28: This should be moved to `query_parameters`, with `wait_for_completion_timeout` and `keep_alive`.

specification/specs/aggregations/bucket/date_range/DateRangeExpression.ts
```diff
@@ -19,9 +19,5 @@
 
 class DateRangeExpression {
   from?: DateMath | float
-  from_as_string?: string
-  to_as_string?: string
-  key?: string
   to?: DateMath | float
-  doc_count?: long
```
delvedor, line 26: Those should not be removed :)

https://www.elastic.co/guide/en/elasticsearch/reference/7.12/search-aggregations-bucket-daterange-aggregation.html

specification/specs/x_pack/async_search/get/AsyncSearchGetRequest.ts
swallez: This has been there for a while and there are many similar validation errors. Let's address all of them in a separate PR.

specification/specs/aggregations/bucket/date_range/DateRangeExpression.ts
swallez: Yes they can be removed. `DateRangeExpression` is only used in `DateRangeAggregation` (the query part) that doesn't have these fields.

The `*_as_string` fields exist in the aggregation result buckets. They are currently not explicitely defined in `Aggregate` and are considered as valid because of how buckets are defined: `class KeyedBucket<TKey>
  implements AdditionalProperties<AggregateName, Aggregate>`.

I have some pending changes on aggregations that explicitely adds those `*_as_tring` fields (among other things) and will be part of the follow-up PR mentioned in the description.

delvedor (approved):
LGTM

PR #334, opened 2021-04-12
Validate rank_eval API

philkra:
as titled.


philkra:
all conflicts resolved and ready for new schema.

delvedor (changes requested):
The changes look ok, but you should add `export ` to all class definitions.

philkra:
added exports.

delvedor (approved):
LGTM

PR #343, opened 2021-04-19
Add annotations and validations for polymorphic types

swallez:
Backport of #333 with additional validations and fewer spec changes.

This adds `variants` tags to polymorphic types (container classes and variant unions) and additional related model validations.

The `Aggregate` container class has been left out for now as it requires extensive refactoring, which will be done in a separate PR.

A new annotation is introduced for properties, `@variant container_property` to identify properties that belong to the container and aren't part of the variants. This is the case of `aggs` and `meta` on aggregations.

This PR also includes a "json ambiguity check" in the model validation, that verifies that the json structure of the model can be analyzed unambiguously with a streaming JSON parser, which is what strongly typed languages generally use to avoid creating an intermediate in-memory JSON AST.

This validation is currently disabled as polymorphic unions cause a lot of errors, which will go away once they are modelled and tagged correctly.

### DIFF
--- a/compiler/model/metamodel.ts
+++ b/compiler/model/metamodel.ts
@@ -140,6 +140,8 @@ export class Property {
   identifier?: string
   /** An optional set of aliases for `name` */
   aliases?: string[]
+  /** If the enclosing class is a variants container, is this a property of the container and not a variant? */
+  container_property?: boolean
 }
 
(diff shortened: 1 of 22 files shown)


delvedor (approved):
LGTM

PR #349, opened 2021-04-20
Only output one validation message for "todo" endpoints, fix non-todo endpoints

swallez:
The input specification has a lot of endpoints that are still a work in progress. They have either `@stability TODO` or do not have associated requests and responses. Validating these endpoints creates a lot of validation errors (we had 600+), making the validation report practically useless.

This PR makes sure we only output one single validation error for these work-in-progress endpoints, so that work to be done is tracked, but we're not flooded with non-actionable errors.

Note that model validation still happens, which allows these endpoints and their transitive dependent types to be output in `schema.json`, even if incomplete/incorrect.

After this fitlering phase, some non-TODO endpoints showed some validation errors that this PR also fixes.

### DIFF
--- a/compiler/compiler.ts
+++ b/compiler/compiler.ts
@@ -20,8 +20,8 @@
 import { writeFile } from 'fs/promises'
 import { join } from 'path'
 import stringify from 'safe-stable-stringify'
-import { Model } from './model/metamodel'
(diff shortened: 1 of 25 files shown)


delvedor (dismissed):
LGTM

PR #352, opened 2021-04-20
Fix ml.info

philkra:
during the namespace exercise, the response class was rename incorrectly and hence was not picked up by the validator.


delvedor (approved):
Great catch!

PR #353, opened 2021-04-20
Validate search_shards response & consolidate shared classes

philkra:
as titled.


specification/transform/get_transform_stats/TransformStats.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { Id } from '@common/common'
-import { NodeAttributes } from './NodeAttributes'
+import { NodeAttributes } from '../../nodes/NodeAttributes'
```
delvedor, line 21: This should be an absolute import, I'll fix it in a separate pr.

delvedor (approved):
LGTM

PR #354, opened 2021-04-21
Refactor common types

delvedor:
As titled.

There are still three folders that need to be reorganized, still trying to figure out where those should go.

- [ ] `__common/analysis`
- [ ] `__common/index`
- [ ]  `__common/mapping`

Currently, `__common/common.ts` contains a lot of unrelated definitions, we could reorganize them better in the future.


swallez:
+1 for moving away lots of things out of commons and aggregating into larger files.

Regarding `__commons/mapping`: there are lot of nested folders that don't bring additional value as they only contain one type definition that also contains the name of their parent folder.

For example, `__commongs/mapping/meta_fields/all/AllFields.ts` should end up in `__commongs/mapping/meta_fields/AllFields.ts` (remove the intermediate `all` directory that is redudant)

swallez (commented):
As discussed, I think we should keep additional nesting in `_common` types related to a particular domain, and only keep the very general ones at the root.

Domain-related types:
- Geo
- Scripting
- Similarity (and should it be in `query_dsl`?)
- Stats
- Time

swallez (approved):
LGTM!

philkra (approved):
lgtm

PR #358, opened 2021-04-22
Validate get_source & get response

philkra:
* annotated `get` request properties
* extending `get_source` from `get` request as they are equal
* validated `get` response
* temporary fix for `get_source` response


delvedor (approved):
LGTM

Issue #359, opened 2021-04-22
get_source response must return generic


philkra:
The `get_source` endpoint solely responds with the generic content. We currently don't have a way of representing that behaviour. See: https://github.com/elastic/elastic-client-generator/pull/358/files#diff-251b022526f297715eae9846f8d24cf550bb634ab05d3b786002583fa83e7297R24

PR #361, opened 2021-04-22
Introduce GenericResponseBody behaviour

delvedor:
Some APIs, `get_source` for example, do not have a structured body, they return something we have no knowledge of.
This pr introduces a new behavior to handle these cases.

Fixes: #359


delvedor:
Closed by #391.

PR #365, opened 2021-04-22
Use a single value for inherits in metamodel

swallez:
This updates the metamodel to have a single optional inherits clause instead of an array.

This has no impact on the spec as TypeScript doesn't allow extending more than one class, and we had a model validation ensuring that `inherits` had zero or one parent class.


delvedor (approved):
LGTM

PR #366, opened 2021-04-22
Add a key parameter to SingleKeyDictionary

swallez:
Semantic typing of strings is important to precisely capture the nature of fields, which is useful for very strongly typed languages and tools like editors where types can drive autocompletion and validation.

The `SingleKeyDictionary` behavior assumes that its keys are `strings` and doesn't allow expressing the real nature of these keys.

This PR changes `SingleKeyDictionary` to have a `Key` generic parameter, just like `Dictionary`, and updates the specification accordingly (using `Field` in `QueryContainer`, for example).

With `SingleKeyDictionary` also having a `key`, there's no real point for it to be a dedicated abstration in the metamodel. It is now represented as a `isSingleKey` boolean value on the `Dictionary` model class. This simplifies the model, and also code generators for languages that can't make a difference between single and multiple key dictionaries (see changes in the TS generator).

### DIFF
--- a/compiler/model/metamodel.ts
+++ b/compiler/model/metamodel.ts
@@ -49,7 +49,7 @@ export class TypeName {
 /**
  * Type of a value. Used both for property types and nested type definitions.
  */
-export type ValueOf = InstanceOf | ArrayOf | UnionOf | DictionaryOf | NamedValueOf | UserDefinedValue | LiteralValue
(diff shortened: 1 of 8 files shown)


delvedor (approved):
LGTM

delvedor (approved):

PR #367, opened 2021-04-22
Nodes.info

philkra:
Validated the response down to 1 error which is based on the `flat_settings` property. Aliasing everything will not do the trick we need a different approach to flatten everything within an object.

test: free-node_info_test_flat_settings_2_response.json.test-d.ts

```
"settings": {
        "action.destructive_requires_name": "false",
        "bootstrap.memory_lock": "true",
        "client.type": "node",
        "cluster.election.strategy": "supports_voting_only",
        "cluster.initial_master_nodes": "instance",
        "cluster.name": "elasticsearch-8-0-0-SNAPSHOT-rest-test",
        "cluster.routing.allocation.disk.threshold_enabled": "false",
...
```


specification/nodes/nodes_info/NodeInfo.ts
```diff
@@ -34,22 +36,226 @@ import { NodeThreadPoolInfo } from './NodeThreadPoolInfo'
...
+  bootstrap?: NodeInfoBootstrap
+  transport: NodeInfoSettingsTransport
+  network?: NodeInfoSettingsNetwork
+  xpack?: NodeInfoXpack
+  script?: NodeInfoScript
+  search?: NodeInfoSearch
+}
+
+export class NodeInfoSettingsCluster {
+  // @aliases cluster.name */
```
delvedor, line 92: This is not a valid JSDoc comment :)

```suggestion
  /** @aliases cluster.name */
```

Same below.

delvedor (approved):
LGTM

PR #372, opened 2021-04-23
Update inherits resolver

delvedor:
This pr improves the inherit resolving, now we can also import an alias, and the generated schema will use the original declaration with the correct namespace.


swallez (approved):
LGTM!

PR #373, opened 2021-04-23
Introduce HttpStatusCode Response type

philkra:
Several API's respond with only a status code, this PR introduces a dedicated identification for these API's.

Furthermore, does this PR validate the endpoints: `security.enable_user` and `security.disable_user`

The next phase for this action is to be able to register, `success` and `error` properties in the class that are driven by the status codes returned.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -62190,6 +62190,32 @@
       ]
     },
     {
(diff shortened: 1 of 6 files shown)


philkra:
obsolete because of https://github.com/elastic/elastic-client-generator/pull/391

PR #374, opened 2021-04-23
Support import alias

delvedor:
This change will allow us to fully (🤞) support import aliases.

For example:
```ts
// ns/foo.ts
export type Hello = string


// ns/bar.ts
import { Hello as Greeting } from './foo'

export class Bar {
  greeting: Greeting
}
``` 

In the schema, the `greeting` property of `Bar` will be:
```json
{
  "name": "Hello",
  "namespace": "ns"
}
```


swallez (approved):
LGTM

PR #376, opened 2021-04-23
Validate searchable_snapshots.clear_cache fix

philkra:
as titled, but has no upstream tests.

added `human` and `pretty` based on https://github.com/elastic/support-diagnostics/pull/482/files


delvedor (approved):
LGTM

PR #378, opened 2021-04-23
Fix Nodes.stats Request

philkra:
fixed the request and will address based on the high amount of shared classes the response alongside nodes.info


delvedor (approved):
LGTM

PR #386, opened 2021-04-27
Fix ml.get trained models Endpoints

philkra:
validated request & response for:
* `ml.get_trained_models`
* `ml.get_trained_models_stats`


delvedor (approved):
Great work!

PR #391, opened 2021-04-28
New response type

delvedor:
This pr introduces the new Reponse type as outlined [here](https://gist.github.com/delvedor/e388ca6b079fc2cb9bbb607fef3ced53).

To keep the specification consistent and actionable it also changes the following:
- Removes the `EmptyResponseBase` and `ArrayResponseBase` behaviors.
- Adds support for the Void custom type
- Adds the support to have response body as generic
- Removes the legacy ResponseBase class

Closes: #359 #361


compiler/model/metamodel.ts
```diff
@@ -240,6 +247,19 @@ export class Request extends BaseType {
   attachedBehaviors?: string[]
 }
 
+/**
+ * A response type
+ */
+export class Response extends BaseType {
```
philkra, line 253: the properties:
* `statusCode: number | number[]`
* `headers: string[]`

are missing based on https://gist.github.com/delvedor/e388ca6b079fc2cb9bbb607fef3ced53#what-about-the-schema

delvedor: That was an example, we aren't tracking those yet as we should discuss more what to store there.
For example,  `statusCode` should be a simple number or a more complex structure that shows which type will be used based on the status?
As soon as we define what we want to have there, we can add new properties.

philkra: 👍

philkra (approved):
LGTM

sethmlarson (approved):
The changes look good from looking over all the TypeScript changes :)

compiler/model/metamodel.ts
```diff
@@ -116,6 +116,13 @@ export class LiteralValue {
   value: string | number | boolean
 }
 
+/**
+ * The absence of any type. This is commonly used in APIs that returns an empty body.
+ */
+export class VoidValue {
+  kind: 'void_value'
+}
+
```
swallez, line 125: Late review of this PR... IMHO we should remove this newly introduced `VoidValue`.

`void` is not a value, it's the absence of value (which is different from `null` and `undefined` that _are_ values). As such a property cannot be of type void, but adding it to the `ValueOf` enumeration makes it possible.

We already have a simpler means to represent requests and responses with an empty body (or more exactly no body, as an empty body could be `{}`): the `body` property is optional: `body?: ValueBody | PropertiesBody`. If it's not specified, then the request or response has no body.

Or is was this introduced so that body-less requests and responses are explicitly represented in the spec? In that case we should make the body property a required field and add a 3rd alternative: `body: ValueBody | PropertiesBody | VoidBody`

PR #395, opened 2021-04-29
Fix ml.get_model_snapshots API

philkra:
as titled.

Also contains a fix for https://github.com/elastic/elastic-client-generator/pull/391


delvedor (approved):
Great catch!

PR #396, opened 2021-04-29
Fix ml.get_overall_buckets API

philkra:
as tiled.

One remaining test in the response is failing but that is due a bug in Elasticsearch: https://github.com/elastic/elasticsearch/issues/72478


delvedor (approved):
LGTM

PR #401, opened 2021-04-29
Improve indices.get response

philkra:
as titled.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -3660,8 +3660,8 @@ export interface IndexIndexSettings {
   'index.translog.durability'?: string
   'query_string.lenient'?: boolean | string
   'index.query_string.lenient'?: boolean | string
-  priority?: integer
-  'index.priority'?: integer
+  priority?: integer | string
+  'index.priority'?: integer | string
   top_metrics_max_size?: integer
   analysis?: IndexIndexSettingsAnalysis
 }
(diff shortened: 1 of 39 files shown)


delvedor (changes requested):
I've investigated this issue, and the error in the validator is happening because `type: 'object'` is the default value if the `type` property is not provided.
This pr should be updated to mark again every `type` as required but the `type: 'object'` property, which should be marked as optional.

https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html

output/typescript/types.ts
```diff
@@ -3818,26 +3818,26 @@ export interface MappingTypesComplexNestedNestedProperty extends MappingTypesCor
...
-  type: 'nested'
+  type?: 'nested'
 }
 
 export interface MappingTypesComplexObjectObjectProperty extends MappingTypesCorePropertyBase {
   dynamic?: boolean | MappingDynamicMapping
   enabled?: boolean
   properties?: Record<PropertyName, MappingTypesProperty>
-  type: 'object'
+  type?: 'object'
```
swallez, line 3828: As discussed with @delvedor, we only need this single line. This is because field mapping uses `object` as the default type (see [bottom of this page](https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html)). In all other cases (other variants and containers that don't have a default type) `type` should be require.

specification/_types/mapping/types/complex/object/ObjectProperty.ts
```diff
@@ -28,5 +28,5 @@ export class ObjectProperty extends CorePropertyBase {
   dynamic?: boolean | DynamicMapping
   enabled?: boolean
   properties?: Dictionary<PropertyName, Property>
-  type: 'object'
+  type?: 'object'
```
swallez, line 31: This is the right `object`. Re adding previous comment:

> As discussed with @delvedor, we only need this single line. This is because field mapping uses object as the default type (see bottom of this page). In all other cases (other variants and containers that don't have a default type) type should be require.

delvedor (approved):
LGTM

PR #402, opened 2021-04-30
Fix security.create_api_key Request + Metadata

philkra:
validated `security.create_api_key` Request and introduced alongside the type `Metadata` for (surprise 😄 ) metadata.


delvedor (approved):
LGTM

PR #408, opened 2021-04-30
Add missing Classes

philkra:
added the classes:
* `indices.simulate_template`
* `ml.find_file_structure`
* `indices.exists_index_template`
* `indices.get_upgrade`
* `ml.start_data_frame_analytics`
* `ml.upgrade_job_snapshot`


specification/indices/exists_index_template/IndicesExistsIndexTemplateResponse.ts
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.
+ */
+
+/**
+ * Response codes:
+ * 200: Indicates all specified index templates exist.
+ * 404: Indicates one or more specified index templates do not exist.
+ */
+export class Response {
+  body: {}
```
delvedor, line 26: This is an `HEAD` API, and there is no body, you should use [`Void`](https://github.com/elastic/elastic-client-generator/blob/master/docs/modeling-guide.md#void-value) :)

specification/ml/find_file_structure/MlFindFileStructureRequest.ts
```diff
@@ -0,0 +1,34 @@
...
+ * under the License.
+ */
+
+import { RequestBase } from '@_types/Base'
+
+/**
+ * Depreacted in 7.12
+ * see: https://www.elastic.co/guide/en/elasticsearch/reference/7.12/ml-find-file-structure.html
+ *
+ * @rest_spec_name ml.find_file_structure
```
delvedor, line 26: Should we track a deprecated API?
It was never in GA.

delvedor (changes requested):

specification/indices/exists_index_template/IndicesExistsIndexTemplateResponse.ts
delvedor: See [here](https://github.com/elastic/elastic-client-generator/blob/master/specification/_global/exists/DocumentExistsResponse.ts).

philkra: replaced all occurrences of `body: {}` with https://github.com/elastic/elastic-client-generator/pull/408/commits/ce6a15b9528c9521700ba04ce07add2f67248c10

specification/ml/find_file_structure/MlFindFileStructureRequest.ts
philkra: left it as comment, we need to adjust these deprecations upstream in the spec. But, having the information is necessary for 7.x clients using the spec.

delvedor (approved):
LGTM once merged on master and `make contrib` :)

Issue #409, opened 2021-05-03
Specification re-organization


delvedor:
Now that we have adopted the new specification folder structure that follows the rest-api-spec and migrated the request and response definitions, the next step is to reorganize every other definition to ensure it's stored in the right place.

For example, the `cat.aliases` API has 3 definitions, request, response, and record. Request and response definitions should always go inside the API namespace, so in this case `/cat/aliases`, while every other type can go either inside the API folder or inside the namespace's `_types` folder. A good rule of thumb is that if a definition it's used across different APIs within a namespace, then it should go in `_types`, otherwise inside the API folder. For instance, the cat aliases record definition should go inside `/cat/aliases`, while the cat base definition should go inside `/cat/_types`.

Furthermore, many folders will need to be updated according to the API name. For example, the `cat.aliases` API should be defined in `/cat/aliases` and not `/cat/cat_aliases`.

Finally, some types will need to be renamed. A language generator can [create a type name by combining the namespace and the type name](https://gist.github.com/delvedor/2dd236902df373bf0bdcd95238d81cab) (the file name is not important), which means that every type should have a meaningful name based on where it's defined.
Let's take the `cat.aliases` API again. Other than the `Request` and `Response` definition, it defines the alias record as well. The record should be named `AliasRecord ` and not `CatAliasesRecord`. In this way, language generators will have a nice and easy to discover the type name.

Remember to run `make contrib` before opening a pr!

### FAQ:

> Where should I start?

If you never worked on the specification, start [here](https://github.com/elastic/elastic-client-generator#prepare-the-environment) and [here](https://github.com/elastic/elastic-client-generator#how-to-validate-the-specification).

There is no specific order to follow, but it's recommended to work on one namespace at a time.
There is no need to fork the project, you can create a branch directly here, I recommend naming the branch ` {username)/fix-409-{namespace}` to smooth things :)

> Which editor should I use?

See [here](https://github.com/elastic/elastic-client-generator#which-editor-should-i-use).
Pro tip: if you move files inside the tree view of VSCode, it will automatically fix the imports!

> While renaming and moving type definitions, I broke too many imports, what should I do?

While it's recommended to fix the imports while you are moving the types, you can run `make spec-imports-fix`, which will fix most of the imports (import aliases won't be updated). Be aware that if you rename a type from `FooBar` to `Foo`, you should also update all its references before running the spec imports fix command.

> How can I be sure that I didn't forget anything behind?

[...]

PR #411, opened 2021-05-03
[409] get_script and more

philkra:
_global/get_script
 _global/get_script_context
 _global/get_script_languages 

FYI: @sethmlarson - ref: #409


delvedor (approved):
LGTM

PR #413, opened 2021-05-03
Refactor xpack

philkra:
as titled. ref: #409


specification/xpack/usage/types.ts
```diff
@@ -0,0 +1,428 @@
...
+export class AlertingExecution {
+  actions: Dictionary<string, ExecutionAction>
+}
+
+export class AlertingInput {
+  input: Dictionary<string, UsageCount>
+  trigger: Dictionary<string, UsageCount>
+}
+
+export class AnalyticsStatsUsage {
```
delvedor, line 45: There is no need to name all of this types `*Usage`, as those are already defined inside the `usage` folder.
You should either drop the `Usage` from the name, or move all of these definitions to the `xpack/_types` folder.

delvedor (changes requested):

specification/xpack/usage/types.ts
philkra: Right! Will rename

specification/xpack/usage/types.ts
```diff
@@ -0,0 +1,429 @@
...
+ */
+
+import { Phases } from '@ilm/_types/Phases'
+import { Dictionary } from '@spec_utils/Dictionary'
+import { ByteSize, EmptyObject, Field, Name } from '@_types/common'
+import { Job, JobStatistics } from '@_types/ml/Job'
+import { integer, long, uint, ulong } from '@_types/Numeric'
+import { Statistics } from '@_types/slm/Statistics'
+
+export class Usage {
```
delvedor, line 27: Let's rename this type to `Base`, so the type name will be `XPack.Usage.Base` :)

```suggestion
export class Base {
```

philkra: done

philkra:
addressed feedback.

delvedor (approved):
LGTM once the conflicts are resolved.

philkra:
@delvedor - conflicts resolved

delvedor (approved):

PR #429, opened 2021-05-04
Reorg types/query_dsl

delvedor:
Given that the global `_types` folder is big and used everywhere, I'll split the reorg into multiple pull requests, hopefully, this would not cause too many conflicts.

Related: #409


philkra (approved):
LGTM

PR #430, opened 2021-05-04
Reorg EQL

philkra:
rel: #409


specification/eql/_types/EqlHits.ts
```diff
@@ -36,3 +37,23 @@ export class EqlHits<TEvent> {
    */
   sequences?: EqlHitsSequence<TEvent>[]
 }
+
+export class EqlHitsEvent<TEvent> {
```
delvedor, line 41: This type and the following one can just be called `HitsEvent` and `HitsSequence`.
```suggestion
export class HitsEvent<TEvent> {
```

delvedor (approved):
LGTM

PR #432, opened 2021-05-04
Reorg SQL

philkra:
rel: #409


delvedor (changes requested):
I think you can rename `SqlColumn` to just `Column` and move `SqlRow` from the `_types` namespace to `sql`.

philkra:
great catch @delvedor ! - done

delvedor (approved):
LGTM

PR #440, opened 2021-05-05
Reorg mtermvectors

stevejgordon:
Contributes to #409 

@delvedor - I've moved some of the types from `termvectors/types.ts` into `_types/termvectors/TermVector.ts` since those are reused across `mtermvectors` and `termvectors` APIs. Is that the correct approach here?


delvedor (changes requested):
I would follow the same approach as `mget` and `msearch` and leave the base types in `termvectors/types.ts` and reused them in `mtermvectors`. The rest of the changes looks good!

stevejgordon:
@delvedor Done! I've moved `TermVectorsResult` under mtermvectors as it's only used there.

delvedor (approved):
LGTM

PR #444, opened 2021-05-05
Changes "void_value" into "no_body" for req/responses with no body

swallez:
Follow-up to https://github.com/elastic/elastic-client-generator/pull/391#discussion_r626069633

This PR changes the way we represent no-body responses (and requests) in `schema.json`:
- the `body` property is now required
- it can now be a `NoBody` in addition to the existing `PropertyBody | ValueBody`
- the `void_value` value kind has been removed (it doesn't make sense for a property to be of type void)

We also formalize the the close relation between body type and inheritance: when a request or response extends a class, the parent class defines body properties inherited by the child class. The child class therefore has to have a `PropertyBody`.

The body is required in the schema but is still optional in the spec (ultimately we should enforce it), and a simple heuristic is used to find the correct body type when it's not present in the TS spec:
- if the request extends a parent class, the body is set to a `PropertyBody` with an emptu list of properties
- otherwise it is set to `NoBody`

New validations verify this.

Aside:
> The only exception to the above is `RequestBase` which is actually a marker class that also links to the `CommonRequestParameter` behavior. IMHO we should get rid of `RequestBase` entirely and have requests directly implement the behavior. This is out of scope of this PR.

There are changes in the schema with the new body type, but the TS output has no change except some formatting that is a side effect of `body` now being required.

### DIFF
--- a/compiler/model/build-model.ts
+++ b/compiler/model/build-model.ts
@@ -179,6 +179,8 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
     )
   }
 
+  let bodyWasSet = false
+
   // Request and Response definitions needs to be handled
   // differently from normal classes
   if (name === 'Request' || name === 'Response') {
@@ -188,7 +190,8 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
         kind: 'request',
         name: { name, namespace: getNameSpace(declaration) },
         path: new Array<model.Property>(),
-        query: new Array<model.Property>()
+        query: new Array<model.Property>(),
+        body: { kind: 'no_body' }
       }
 
       const response: model.TypeName = {
@@ -210,20 +213,27 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
           type.path = property.properties
         } else if (property.name === 'query_parameters') {
           type.query = property.properties
-        } else {
-          // the body can either by a value (eg Array<string> or an object with properties)
+        } else if (property.name === 'body') {
+          bodyWasSet = true
+          // the body can either be a value (eg Array<string> or an object with properties)
           if (property.valueOf != null) {
-            type.body = { kind: 'value', value: property.valueOf }
+            if (property.valueOf.kind === 'instance_of' && property.valueOf.type.name === 'Void') {
+              type.body = { kind: 'no_body' }
+            } else {
+              type.body = { kind: 'value', value: property.valueOf }
+            }
           } else if (property.properties.length > 0) {
             type.body = { kind: 'properties', properties: property.properties }
           }
+        } else {
+          assert(member, false, 'Unknown request property ' + property.name)
         }
       }
     } else {
       type = {
         kind: 'response',
         name: { name, namespace: getNameSpace(declaration) },
-        body: undefined
+        body: { kind: 'no_body' }
       }
 
       for (const member of declaration.getMembers()) {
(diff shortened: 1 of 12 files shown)


compiler/model/build-model.ts
```diff
@@ -188,7 +190,8 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
         kind: 'request',
         name: { name, namespace: getNameSpace(declaration) },
         path: new Array<model.Property>(),
-        query: new Array<model.Property>()
+        query: new Array<model.Property>(),
+        body: { kind: 'no_body' }
```
delvedor, line 194: Having `no_body` as the default value could cause some side effects at the moment. For example, any API that we still need to define will be generated as a boolean. This could be fine tho, so we can use this "bug" to identify bad definitions.

compiler/model/build-model.ts
```diff
@@ -279,6 +294,15 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
...
     }
 
+    // If the body wasn't set and we have a parent class, then it's a property body with no additional properties
+    if (!bodyWasSet && type.inherits != null) {
+      const parent = type.inherits.type
+      // RequestBase is special as it's a "marker" base class that doesn't imply a property body type. We should get rid of it.
+      if (!(parent.name === 'RequestBase' && parent.namespace === '_types')) {
+        type.body = { kind: 'properties', properties: new Array<model.Property>() }
+      }
+    }
```
delvedor, line 304: Should we check the parent body kind? It might happen that the parent has a `value` body instead of  `properties`.

typescript-generator/index.ts
```diff
@@ -254,13 +252,13 @@ function buildRequest (type: M.Request): string {
     if (pathPropertiesNames.includes(property.name)) continue
     code += `  ${cleanPropertyName(property.name)}${property.required ? '' : '?'}: ${buildValue(property.type, openGenerics)}\n`
   }
-  if ((type.body != null) && type.body.kind === 'properties') {
+  if (type.body.kind === 'properties') {
```
delvedor, line 255: You should also check if `type.body.properties` contains something, otherwise you can generate definitions like:
```ts
export interface FooRequest {
  body?: {
  }
}
```

```suggestion
  if (type.body.kind === 'properties' && type.body.properties.length > 0) {
```

delvedor (changes requested):
Great work! I've left a couple of notes.

philkra (changes requested):
Can we also directly add a specific `Response` class to extend from in _no body_ cases? Thanks in advance

compiler/model/build-model.ts
swallez: Indeed. I defaulted to `no_body` for now as we have many requests and responses that miss a body definition in the spec and wanted to minimize changes to the spec in this PR.

Ultimately we should remove this default. But if we decide that now is the time, I can add the missing `body` as part of this PR.

compiler/model/build-model.ts
swallez: This is checked in `validate-model`: https://github.com/elastic/elastic-client-generator/blob/ce700b315a9ba2838beec1fddc2e0cac1ff06abd/compiler/steps/validate-model.ts#L431-L435

typescript-generator/index.ts
swallez: If we enforce the existence of `body` in the spec, then requests/responses that extend a parent class with no additional properties need to have this kind of declaration. We have a few responses like this, e.g. children of `WriteResponseBase`.

And if an empty property body is used without extending a parent class, then it's an error in the spec that we can be checked in model validation, so that we don't have to check it in the code generation.

swallez:
@philkra 
> Can we also directly add a specific Response class to extend from in no body cases? Thanks in advance

Unfortunately no. Parent classes in responses are regular interfaces and not responses (i.e. they don't have this nested structure with `body`) and so can only be used to defined body properties that are shared among responses extending them.

compiler/model/build-model.ts
delvedor: No, let's do it in a follow-up.

typescript-generator/index.ts
delvedor: Enforcing the existence of `body` in the response is ok, but it's not in the request.
Please add that check in the generator, we can remove it later, but for now it avoids generating ugly/useless definitions :)

swallez: Got it and applied your suggestion. No more ugly definitions :-)

delvedor (approved):
LGTM

PR #450, opened 2021-05-06
Reorg and complete security

stevejgordon:
A bit of a beast but this includes renaming/reorg of the security namespace. In addition, I've implemented the missing APIs and got the tests all passing.

@delvedor As discussed, after a rebase against master, previously passing (empty) responses are failing tests for `security.change_password`, `security.disable_user` and `security.enable_user`.

Contributes to #409


delvedor (approved):
LGTM

PR #451, opened 2021-05-06
Allow empty object responses

delvedor:
#444 surfaced a bug, where we aren't allowing empty object responses, which some security APIs are using.
It would be nice to see those APIs migrate to the `AcknowledgedResponseBase` tho.

Related: #450


philkra (approved):
LGTM

PR #453, opened 2021-05-10
Reorg bulk

delvedor:
Related: #409.


specification/_global/bulk/types.ts
```diff
@@ -56,3 +63,27 @@ export class BulkCreateResponseItem extends BulkResponseItemBase {}
...
+export class BulkOperation {
+  _id: Id
+  _index: IndexName
+  retry_on_conflict: integer
+  routing: Routing
+  version: VersionNumber
+  version_type: VersionType
+}
+
+export class BulkOperationContainer {
```
philkra, line 76: don't we need to annotate this with `/** @variants container */` ?

philkra (approved):
one question, otherwise LGTM

specification/_global/bulk/types.ts
delvedor: Great catch!

philkra (approved):
LGTM

PR #458, opened 2021-05-10
Reorg indices

delvedor:
Related: #409.


specification/cluster/_types/ClusterStateBlocksIndex.ts
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { IndicesRolloverConditions } from '@indices/rollover/IndicesRolloverConditions'
+import { RolloverConditions } from '@indices/rollover/types'
```
philkra, line 20: shouldn't `RolloverConditions` being in `indices/_types` if it's used in more than one endpoint?

specification/nodes/_types/Stats.ts
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { IndexStats } from '@indices/stats/IndexStats'
+import { IndexStats } from '@indices/stats/types'
```
philkra, line 20: same here, especially if used from a different namespace, should go into `indices/_types`

philkra (changes requested):

specification/nodes/_types/Stats.ts
delvedor: It's used only here :)

specification/cluster/_types/ClusterStateBlocksIndex.ts
delvedor: It's used only there (and cluster) :)

philkra (approved):
LGTM

PR #474, opened 2021-05-24
Remove duplicated properties

stevejgordon:
During .NET code gen I identified a few types which duplicate properties defined on types they derive from. This causes compilation errors in the C# code. This PR removes the duplicated properties.


stevejgordon:
Darn! After resetting WSL I mistyped my email address for git config so this commit is not picking up the CLA! Will close and resubmit with a fixed commit!

PR #475, opened 2021-05-24
Remove duplicated properties

stevejgordon:
During .NET code gen I identified a few types which duplicate properties defined on types they derive from. This causes compilation errors in the C# code. This PR removes the duplicated properties.


philkra (approved):
LGTM

PR #478, opened 2021-06-03
Fix ML definitions

delvedor:
#225 introduced some issues, as xPack endpoints have most of the properties marked as optional, while the MlJob definition does not.

This pr fixes that, I've also left a TODO to remember to fix the xPack endpoint, which is kind of a snowflake.


jgowdyelastic (approved):
LGTM

PR #483, opened 2021-06-28
Fix "container_property" naming discrepancy

swallez:
`container_property` in the metamodel should be `containerProperty` to be consistent with the camelCase convention used in that file.


stevejgordon (approved):
LGTM!

PR #484, opened 2021-06-29
[WIP] Add "base_class" variants attribute

swallez:
Adds a `base_class` attribute to the `@variants` jsdoc tag. It is a hint for object-oriented language generators indicating the base class that is common to all variants. This attribute applies to all kinds of variants (containers, internal & external).

**[WIP]:** this PR doesn't correctly resolve the namespace of the `base_class` value (need help from @delvedor): it uses the namespace from the _current_ sourcefile, which is likely to be ok in many cases, but will fail on more complex module structures.


swallez:
Closing abandoned proposal.

PR #489, opened 2021-07-09
Add @identifier annotation to _nodes properties

swallez:
Some properties in the ES API have a leading underscore, e.g. `_id`, `_source`, `_nodes`, etc.

In languages where JSON properties can be renamed, it's nicer from a DX perspective to remove these leading underscores from the generated property identifiers. Leading underscores also have special meanings in some languages like private in Python or unused in Rust.

Removing leading underscores from property names however is causing issues in classes that have a `_nodes: NodeStatistics` property that also contain a `nodes` property.

This PR adds `@identifier node_stats` to all `_nodes: NodeStatistics` properties.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -12833,9 +12833,9 @@ export interface SecurityClearCachedRealmsRequest extends RequestBase {
 }
 
 export interface SecurityClearCachedRealmsResponse {
+  _nodes: NodeStatistics
   cluster_name: Name
   nodes: Record<string, SecurityClusterNode>
-  _nodes: NodeStatistics
 }
 
 export interface SecurityClearCachedRolesRequest extends RequestBase {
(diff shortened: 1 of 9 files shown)


output/typescript/types.ts
```diff
@@ -12833,9 +12833,9 @@ export interface SecurityClearCachedRealmsRequest extends RequestBase {
 }
 
 export interface SecurityClearCachedRealmsResponse {
+  _nodes: NodeStatistics
```
delvedor, line 12836: Missing `@identififer`?

delvedor (approved):
LGTM

output/typescript/types.ts
swallez: No, this is the TypeScript generator output. This diff is caused by the reordering of some `_nodes` properties that were last in the list to the first position and which I moved at the top so that all uses are consistent.

I was also confused by this while reading the diff before creating the PR :-)

sethmlarson (approved):

PR #495, opened 2021-07-19
Remove error from WriteResponseBase

stevejgordon:
Remove the error from `WriteResponseBase` which may conflict with error handling defined in individual clients (as it did with .NET).


swallez (approved):
LGTM

Issue #499, opened 2021-07-22
Implement Versioned Branching Strategy


philkra:
Currently we are not distinguishing between Elasticsearch's `master` and `7.x` branches. The goal is to follow the same branching strategy.

## Blocking Tasks
- ~~Merge all open [PRs](https://github.com/elastic/elasticsearch-specification/pulls)~~

## New Branches
- [x] Create `7.x` branch from `main`
- [x] Create `7.14` branch from `7.x`

## Tooling & Infra
- [x] Investigate & update the scripts to be version aware - https://github.com/elastic/elasticsearch-specification/pull/503
- [x] Add the backporting bot and ensure only compiler changes are backport-able - https://github.com/elastic/elasticsearch-specification/pull/502
- [x] Read the version number & build hash from the spec artefact and write these values to the [schema.json](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json) `_info` object - https://github.com/elastic/elasticsearch-specification/pull/501
- [x] [This action](https://github.com/elastic/elasticsearch-specification/blob/f7bb07134d24fe95ba69032c8e3b7d3f99ba5b50/.github/workflows/update-rest-api-json.yml#L28) should be updated as well.

delvedor:
We can merge the other prs later on.

PR #501, opened 2021-07-26
Generate version and hash fields in schema output

delvedor:
This pr introduces a new github action that is executed every time some code is pushed in `main` or `7.*` branches.
It also updates the info step of the compiler to update the `version` and `hash` fields to use the current git values if the compiler is being run from a base branch, leave those fields as is otherwise.

NOTE: It will require a test run in the main branch to be sure that everything works as expected.

Related: #499

### DIFF
--- a/compiler/model/metamodel.ts
+++ b/compiler/model/metamodel.ts
@@ -376,6 +376,7 @@ export class UrlTemplate {
 export class Model {
   _info?: {
     version: string
+    hash: string
     title: string
     license: {
       name: string
--- a/compiler/steps/add-info.ts
+++ b/compiler/steps/add-info.ts
@@ -19,18 +19,44 @@
 
 import * as model from '../model/metamodel'
 import { JsonSpec } from '../model/json-spec'
+import { execSync } from 'child_process'
+import { readFileSync } from 'fs'
+import { join } from 'path'
 
 /**
- * Adds the `_info` field to the JSON model
+ * Adds the `_info` field to the JSON model.
+ * If this code is being run in the base branch, it will update the version and hash
+ * fields with the value read from git, otherwise it will reuse the values
+ * currently stored in the output schema.
  */
 export default async function addInfo (model: model.Model, jsonSpec: Map<string, JsonSpec>): Promise<model.Model> {
-  model._info = {
-    version: '7.13.0',
-    title: 'Elasticsearch Request & Response Specification',
-    license: {
-      name: 'Apache 2.0',
-      url: 'https://github.com/elastic/elasticsearch-specification/blob/master/LICENSE'
+  const branch = execSync('git branch --show-current').toString().trim()
+  const isBaseBranch = branch === 'main' || branch.startsWith('7.')
+
+  if (isBaseBranch) {
+    model._info = {
+      version: branch,
+      hash: execSync('git rev-parse --short HEAD').toString().trim(),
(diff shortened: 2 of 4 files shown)


compiler/model/metamodel.ts
```diff
@@ -376,6 +376,7 @@ export class UrlTemplate {
 export class Model {
   _info?: {
     version: string
+    hash: string
```
swallez, line 379: We're doing something similar in the [`download-schema` task](https://github.com/elastic/elastic-client-generator-java/blob/38da213659c72b1af58cd308aa0609c837070590/generator/build.gradle.kts#L65) in the Java client generator, but also add the branch name, which could be useful in some contexts.

The resulting metadata would be:

```typescript
export class Model {
  _info?: {
    version: string
    git: {
        revision: string
        branch: string
    }
    title: string
    ...
```

swallez (changes requested):

compiler/model/metamodel.ts
delvedor: In this case, version and branch are the same thing.

swallez: > In this case, version and branch are the same thing.

Hmm... It should be different for `main` and maybe also for `7.x`. Here's what I have as a result of the Java generator's `download-schema`:

```json
  "_info" : {
    "license" : {
      "name" : "Apache 2.0",
      "url" : "https://github.com/elastic/elasticsearch-specification/blob/master/LICENSE"
    },
    "title" : "Elasticsearch Request & Response Specification",
    "version" : "7.13.0",
    "git" : {
      "branch" : "main",
      "revision" : "d59679ae28f894729f11a206042ba24533a5624c"
    }
  },
```

swallez: In terms of use cases, having the branch name can be useful for example for tools that would compare an existing schema with the current revision from the same branch (API change log, backwards compatibility check, etc)

delvedor: Yes, but in our case, we are not tracking patch numbers, and the branch name is the "version" that we are actually tracking.

Eg, in the `main` branch we are developing the `main` version of the specification, the same goes for `7.x` and `7.14`.

compiler/steps/add-info.ts
```diff
@@ -19,18 +19,44 @@
...
+import { execSync } from 'child_process'
+import { readFileSync } from 'fs'
+import { join } from 'path'
 
 /**
- * Adds the `_info` field to the JSON model
+ * Adds the `_info` field to the JSON model.
+ * If this code is being run in the base branch, it will update the version and hash
+ * fields with the value read from git, otherwise it will reuse the values
+ * currently stored in the output schema.
```
swallez, line 30: I'm actually confused by `reuse the values currently stored in the output schema`: why don't we update the hash in that case?

It seems to me that the purpose of adding the hash to `schema.json` automatically after a change is to provide strong traceability information from downstream artefacts (e.g. generated code) to the spec that was used to produce them.

Am I missing something?

delvedor: We generate a new one if the generation starts from a base branch (main, 7.x, 7.n), while if you are generating the schema from a feature branch we don't care about the hash, so we reused the old one to have a nicer diff.

swallez: Ah right, I was actually missing something and that makes sense!

swallez (approved):

PR #503, opened 2021-07-26
Require the use of STACK_VERSION when running validations

delvedor:
As titled. This will tell the flight recorder which branch to use.

Related: #499

### DIFF
--- a/run-validations.sh
+++ b/run-validations.sh
@@ -1,4 +1,14 @@
 #!/usr/bin/env bash
+
+function require_stack_version() {
+  if [[ -z $STACK_VERSION ]]; then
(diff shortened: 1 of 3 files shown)


run-validations.sh
```diff
@@ -19,7 +29,7 @@ fi
 pushd "${recorderFolder}"
 function finish { popd; }
 trap finish EXIT
-./run-validations.sh $@
+STACK_VERSION=${STACK_VERSION} ./run-validations.sh $@
```
swallez, line 32: We should not need to re-export the variable as it's already in the environment and therefore inherited by child processes.

swallez (approved):
LGTM with a minor shell nitpick

PR #505, opened 2021-07-28
Fix IndicesDataStreamsStatsRequest duplicate property

stevejgordon:
Remove human property which comes from CommonQueryParameters. I discovered this conflict when generating the .NET class for this type.


swallez (approved):
LGTM!

Cood catch! I haven't encountered this issue in Java as I simply ignore this parameter that IHMO doesn't really make sense in the context of strongly typed data structures.

PR #506, opened 2021-07-29
Fix MainError duplicated property

stevejgordon:
`MainError` includes a `root_cause` property which is a duplicate of the property from `ErrorCause` which it extends.


swallez (approved):
LGTM

PR #507, opened 2021-07-29
Add a model validation that detects duplicate properties on ancestor classes

swallez:
Follow-up to PR #506 and #505

This PR adds a model validation that checks that the properties of an interface/request/response are not already defined in a parent class or behavior.

There are still some occurences of this issue but they are on `@stability TODO` types and are therefore not reported in validation errors. Be prepared though to see them show up in the future 😉


delvedor (approved):
LGTM

stevejgordon (approved):
LGTM too

PR #508, opened 2021-07-29
Add 'error?' to MlCloseJobResponse

sethmlarson:


stevejgordon:
I've actually removed error property from at least one response (#495) due to conflicts. Discussed with @swallez that clients need to define the appropriate behaviour for errors. In the .NET client, we have a bespoke `RequestBase` that has properties such as error that could appear on responses from any failed (4xx) requests. 

I have run into one API so far (indices.validate_query.Response) where an error field is expected as part of the successful payload in a different format to the actual failed request case. It's a string in the expected payload. Not yet sure how that one should be handled.

swallez:
What @stevejgordon said. Response bodies should describe the nominal response. Clients should expect an `ErrorCause` if the status code is different from 200 (or other nominal status code). If we ever have other error types, then we'll add them as a (yet to be defined) `errorBody` property to the response definition.

That being said, [looking at the code](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/rest/job/RestDeleteJobAction.java) the response body seems pretty different from what we have in the spec and should be:
```ts
export class Response {
  body: {
    task?: TaskId // if wait_for_completion is false
    acknowledged?: boolean // if wait_for_completion is true or unspecified
}
```

delvedor (changes requested):
Can we update this pr with the format introduced in https://github.com/elastic/elasticsearch-specification/pull/1468 ?

sethmlarson:
Going to close this as it's not in the right format and will require some reworking.

PR #515, opened 2021-07-29
Validate 'ml.post_data' API

sethmlarson:
- Receiving an error regarding `@timestamp` not working with `UserDefinedValue`? Perhaps some JavaScript weirdness happening here? cc @delvedor
- The Request/Response weren't named to exactly match the API, although I think the API would have been better named `ml.post_job_data`... not sure what we should do here?

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -7489,15 +7489,15 @@
       "name": "ml.post_data",
       "request": {
         "name": "Request",
-        "namespace": "ml.post_job_data"
+        "namespace": "ml.post_data"
       },
       "requestBodyRequired": true,
       "response": {
         "name": "Response",
-        "namespace": "ml.post_job_data"
+        "namespace": "ml.post_data"
       },
       "since": "5.4.0",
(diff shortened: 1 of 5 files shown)


delvedor (commented):
TypeScript will only show the very first encountered error, in this case, is the `@timestamp` key, if you comment it out in the workbench, then the error is the next property.
The issue here is that the request type is completely wrong.

> The Request/Response weren't named to exactly match the API, although I think the API would have been better named ml.post_job_data... not sure what we should do here?

The name comes from the rest-api-spec, so there isn't much we can do other than ping someone from ML and offer a feedback :)

sethmlarson:
Added new types to support the multiple ways of inputting data to `ml.post_data`.

> The name comes from the rest-api-spec, so there isn't much we can do other than ping someone from ML and offer a feedback

Yeah it's fine as-is, was just highlighting the change of the namespace/request object to match the `post_data` name.

delvedor (approved):

PR #518, opened 2021-07-29
Validate 'ml.put_datafeed' API

sethmlarson:


specification/ml/put_datafeed/MlPutDatafeedRequest.ts
```diff
@@ -48,12 +48,13 @@ export interface Request extends RequestBase {
...
   body?: {
+    aggs?: Dictionary<string, AggregationContainer>
     aggregations?: Dictionary<string, AggregationContainer>
     chunking_config?: ChunkingConfig
     delayed_data_check_config?: DelayedDataCheckConfig
     frequency?: Time
-    indices?: string[]
-    indexes?: string[]
+    indices?: Indices
+    indexes?: Indices
```
stevejgordon, line 57: This is unfortunate!

stevejgordon (approved):
LGTM. A shame "indexes" exists. One could argue that it's dropped from the spec and swept under the carpet but they may impact others.

delvedor (approved):

specification/ml/put_datafeed/MlPutDatafeedRequest.ts
sethmlarson: Something to drop in 8.0 for sure!

swallez:
Looking at [the code](https://github.com/elastic/elasticsearch/blob/1f94aaf28031a7abbb111ac3e8150c7bfdd962c4/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/datafeed/DatafeedUpdate.java#L61) `indexes` is supported when parsing the structure and goes to the same property as `indices`. On serialization only `indices` is used.

So we should use `@aliases indexes` on the `indices` property so that clients expose a single property. I'll open a follow-up PR.

PR #520, opened 2021-07-29
Validate 'ml.revert_model_snapshot' API

sethmlarson:


stevejgordon (approved):
LGTM

delvedor (approved):
According to the [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-revert-snapshot.html) there should not be a query parameter here, but very likely the API is accepting both. Let's merge it and revisit it if needed.

PR #521, opened 2021-07-29
Validate 'ml.start_data_frame_analytics' API

sethmlarson:
There doesn't look to be any tests for the Request portion of this API?


delvedor (approved):
There might be an issue in our test infra, I'll investigate.
Let's merge this as-is for now.

Issue #494, opened 2021-07-19
ML job reset endpoint


jgowdyelastic:
A new endpoint has been added for reseting anomaly detection jobs
https://github.com/elastic/elasticsearch/pull/73908

```
POST _ml/anomaly_detectors/<job_id>/_reset
```

I imagine the request type will looks something like this
```
export interface MlResetJobRequest extends RequestBase {
  job_id: Id
  wait_for_completion?: boolean
}
```

`MlJob` will also need updating with new properties added with this reset work

```diff
 class MlJob {
...
+ blocked?: MlJobBlocked
 }
```

where `MlJobBlocked` should look like this:
```
MlJobBlocked {
  reason: 'delete' | 'revert' | 'reset';
  task_id?: string;
}
```

PR #524, opened 2021-07-30
Add 'ml.reset_job' API and 'MlJob.blocked'

sethmlarson:
Closes https://github.com/elastic/elasticsearch-specification/issues/494

@jgowdyelastic can you take a look and verify this matches what you expect?
There don't seem to be any YAML tests for the `blocked` attribute (can we get one or more of these?)

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -10719,7 +10719,7 @@ export interface MlJob {
 
 export interface MlJobBlocked {
   reason: MlJobBlockedReason
-  task_id: TaskId
+  task_id?: TaskId
 }
 
 export type MlJobBlockedReason = 'delete' | 'reset' | 'revert'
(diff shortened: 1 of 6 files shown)


output/typescript/types.ts
```diff
@@ -10716,6 +10717,13 @@ export interface MlJob {
   system_annotations_retention_days?: long
 }
 
+export interface MlJobBlocked {
+  reason: MlJobBlockedReason
+  task_id: TaskId
```
jgowdyelastic, line 10722: `task_id` should be optional

sethmlarson: Done!

delvedor (approved):
LGTM

jgowdyelastic (approved):
LGTM

PR #525, opened 2021-07-30
Add identifier for `_source` property in `source` type.

stevejgordon:
In .NET properties of a type cannot be named the same as the enclosing class. The `source` type includes an `_source`  property which currently conflicts. This could be handled in the .NET generator but perhaps this property can be more clearly named.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -20425,6 +20425,7 @@
           }
         },
         {
+          "identifier": "source_fields",
           "name": "_source",
           "required": false,
           "type": {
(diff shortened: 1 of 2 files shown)


delvedor (approved):
LGTM

swallez:
@stevejgordon instead of renaming the property, can we instead rename the enclosing class?

The class name can be whatever we want since it never appears on the wire. Property names, on the other hand, should IMHO follow as closely as possible the names found in the JSON structures.

So what about `ReindexSource` or even `LocalSource` since we also have `RemoteSource`?

swallez:
Also, since this is an issue for .Net only, can it be handled once for all in the code generator by checking if the class name conflicts with one of its properties and then using a specific naming strategy either for the class name or the property name?

PR #529, opened 2021-07-30
Validate 'update' API

sethmlarson:
Only needed `Routing` to be `string | number`, might fix a few other APIs too?

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -1,12 +1,12 @@
 {
   "_info": {
-    "hash": "#",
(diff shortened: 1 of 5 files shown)


delvedor (commented):
I think this is one of the cases where our testing fails. `routing` should always be a number, but given that it will go in the querystring, the yaml tests are lazily using numbers.
It's better to update the testing infra to address this.

delvedor (commented):
Thanks to https://github.com/elastic/elasticsearch/pull/77144 this can now be closed.

PR #534, opened 2021-08-05
Remove ML system_annotations_retention_days setting

droberts195:
This is a reversion of #522.

We decided not to add a new setting after all, but to
use results_retention_days for both results and system
annotations (on the basis that system annotations are
a type of result).

This setting has been added and removed within the 7.15
development cycle so there are no implications for
anyone using released product.

Relates to elastic/elasticsearch#76096

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -105393,17 +105393,6 @@
               "namespace": "_types"
             }
           }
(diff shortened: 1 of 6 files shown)


sethmlarson (approved):
LGTM

PR #536, opened 2021-08-06
Make search_after type the same as Hit.sort

sethmlarson:
It's common to do `search_after = hits[-1].sort` so type-wise we should ensure this works out.


philkra (approved):
LGTM

delvedor (approved):

PR #543, opened 2021-08-16
Script lang can be a custom string

delvedor:
As titled, see https://github.com/elastic/elasticsearch-js/issues/1518.

I'm not sure defining the type like this makes sense. It might be worth just define `lang` as `string`.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -33153,11 +33153,23 @@
           "name": "lang",
           "required": false,
           "type": {
-            "kind": "instance_of",
-            "type": {
-              "name": "ScriptLanguage",
-              "namespace": "_types"
-            }
+            "items": [
+              {
+                "kind": "instance_of",
+                "type": {
+                  "name": "ScriptLanguage",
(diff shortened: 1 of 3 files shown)


philkra (approved):
LGTM

swallez:
If any `string` is a valid `script`, then what is the need for `ScriptLanguage` at all? We can just use `lang?: string` and remove the union.

swallez:
Or if we somehow want to keep a list of built-in languages around, then `ScriptLanguage` should be defined as:

```ts
export enum ScriptLanguage {
  painless = 0,
  expression = 1,
  mustache = 2
  // java removed as it's just an example in the docs.
}

export type ScriptLanguage = BuiltinScriptLanguage | string
```

and then keep the fields as `lang?: ScriptLanguage`.

This is BTW a general rule about field types: if we have to write the same union more than once, then it's more than likely that the type should be the union. This will not only make the spec cleaner, but also help code generators that need to implement special handling of some known unions.

PR #544, opened 2021-08-16
Add back ML type fixes

delvedor:
This reverts commit d2680ae3b660ef6b7a5c7e799f13e232183ea157.
Most of these changes are ok, but there are some of them which are not, please take a look and test in the Kibana codebase to validate if the generated types are what you expect, thanks!

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -6629,7 +6629,7 @@
         "namespace": "ml.estimate_model_memory"
       },
       "since": "7.7.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7034,7 +7034,7 @@
         "namespace": "ml.get_datafeed_stats"
       },
       "since": "5.5.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7168,7 +7168,7 @@
         "namespace": "ml.get_job_stats"
       },
       "since": "5.5.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7202,7 +7202,7 @@
         "namespace": "ml.get_jobs"
       },
       "since": "5.5.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7239,7 +7239,7 @@
         "namespace": "ml.get_model_snapshots"
       },
       "since": "5.4.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7435,7 +7435,7 @@
         "namespace": "ml.open_job"
       },
       "since": "5.4.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
@@ -7490,15 +7490,15 @@
       "name": "ml.post_data",
       "request": {
         "name": "Request",
-        "namespace": "ml.post_job_data"
+        "namespace": "ml.post_data"
       },
       "requestBodyRequired": true,
       "response": {
         "name": "Response",
-        "namespace": "ml.post_job_data"
+        "namespace": "ml.post_data"
       },
       "since": "5.4.0",
-      "stability": "TODO",
+      "stability": "stable",
       "urls": [
         {
           "methods": [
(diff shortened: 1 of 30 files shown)


output/schema/schema.json
```diff
@@ -100034,7 +100034,7 @@
         {
           "description": "A comma separated list of influencer field names. Typically these can be the by, over, or partition fields that are used in the detector configuration. You might also want to use a field name that is not specifically named in a detector, but is available as part of the input data. When you use multiple detectors, the use of influencers is recommended as it aggregates results for each influencer entity.",
           "name": "influencers",
-          "required": true,
+          "required": false,
```
droberts195, line 100037: `false` is correct if this interface is only using when putting jobs.  But if it's also used when getting jobs then it needs to be split, because `influencers` will always be present in `analysis_config` when getting jobs.

output/schema/schema.json
```diff
@@ -104554,7 +104613,7 @@
         {
           "description": "The analysis function that is used. For example, `count`, `rare`, `mean`, `min`, `max`, or `sum`.",
           "name": "function",
-          "required": true,
+          "required": false,
```
droberts195, line 104616: `false` is correct here if this interface is only used when creating a job for the first time.  But if this interface is also used to generate code for getting jobs then it needs to be split because the `detectors` in a job retrieved from the get jobs API will always have a `function` set for every detector.

output/typescript/types.ts
```diff
@@ -10626,10 +10634,11 @@ export interface MlDetector {
   detector_index?: integer
   exclude_frequent?: MlExcludeFrequent
   field_name?: Field
-  function: string
+  function?: string
```
droberts195, line 10637: This interface needs splitting into 2.  When creating a job for the first time `function` is not compulsory, but when getting the definition back from the get jobs API `function` will be present for every detector.

output/typescript/types.ts
```diff
@@ -10701,8 +10710,8 @@ export interface MlJob {
   finished_time?: integer
   groups?: string[]
   job_id: Id
-  job_type: string
-  job_version: VersionString
+  job_type?: string
```
droberts195, line 10713: Similar to the other fields I've mentioned, `job_type` would not be set when putting a job, but would be present when getting it back.

specification/ml/_types/Detector.ts
```diff
@@ -49,7 +49,7 @@ export class Detector {
   /**
    * The analysis function that is used. For example, `count`, `rare`, `mean`, `min`, `max`, or `sum`.
    */
-  function: string
+  function?: string
```
droberts195, line 52: As before, this is only optional when putting a job, not when getting it back.

specification/ml/_types/Analysis.ts
```diff
@@ -50,7 +50,7 @@ export class AnalysisConfig {
   /**
    * A comma separated list of influencer field names. Typically these can be the by, over, or partition fields that are used in the detector configuration. You might also want to use a field name that is not specifically named in a detector, but is available as part of the input data. When you use multiple detectors, the use of influencers is recommended as it aggregates results for each influencer entity.
    */
-  influencers: Field[]
+  influencers?: Field[]
```
droberts195, line 53: This will always be present when getting the job definitions from the get jobs endpoint.  It is not compulsory when creating a job.

droberts195 (commented):
Thanks for pinging the ML team on this one.

I've commented on the problems I can see.  @jgowdyelastic please let us know if you can see any other problems.

jgowdyelastic:
It might be good to think up a naming convention for pre and post saved jobs and their related types.
We have `request` and `response` versions for some of the types, e.g. [MlPutJobRequest](https://github.com/elastic/elasticsearch-specification/blob/f38b6e172fd6782a136de566a8c5b8345c563a6f/output/typescript/types.ts#L11726)  which is very similar to [MlJob](https://github.com/elastic/elasticsearch-specification/blob/f38b6e172fd6782a136de566a8c5b8345c563a6f/output/typescript/types.ts#L10692). but both contain the same [MlAnalysisConfig](https://github.com/elastic/elasticsearch-specification/blob/f38b6e172fd6782a136de566a8c5b8345c563a6f/output/typescript/types.ts#L10136) 

Adding `request` to the name of the version of `MlAnalysisConfig` that is used in `MlPutJobRequest` doesn't seem right as it isn't used on its own as a request payload. 

I'm not suggesting replacing all `request` and `response` types. But in special cases like `MlPutJobRequest` the body could be something like `MlJobPreSave` (terrible name) which contains `MlAnalysisConfigPreSave` and the response `MlJobPostSave` which contains `MlAnalysisConfigPostSave`

Also, I just notices that `MlPutJobResponse` has its own type which is separate from `MLJob`. These should be the same.

PR #546, opened 2021-08-16
Throw if the path_parts do not use the same name as rest-api-spec

delvedor:
As titled.

This pr also renames few folders that weren't strictly following the rest-api-spec naming and removes useless path parts.


philkra (approved):
LGTM

sethmlarson (approved):
LGTM

PR #548, opened 2021-08-17
[ML] Changes model_prune_window description

szabosteve:
## Overview

This PR changes the `model_prune_window` parameter description to be in line with the published ML API docs in the ES guide.


delvedor (approved):
LGTM

PR #550, opened 2021-08-17
[DOCS] Update put role API params

lockewritesdocs:
Updates parameter descriptions for the PUT roles security API.

### DIFF
--- a/specification/security/put_role/SecurityPutRoleRequest.ts
+++ b/specification/security/put_role/SecurityPutRoleRequest.ts
@@ -43,14 +43,11 @@ export interface Request extends RequestBase {
      */
     applications?: ApplicationPrivileges[]
     /**
-     * A list of cluster privileges. These privileges define the cluster-level
-     * actions for users with this role.
+     * A list of cluster privileges. These privileges define the cluster-level actions for users with this role.
      */
     cluster?: string[]
     /**
-     * An object defining global privileges. A global privilege is a form of
-     * cluster privilege that is request-aware. Support for global privileges
-     * is currently limited to the management of application privileges.
+     * An object defining global privileges. A global privilege is a form of cluster privilege that is request-aware. Support for global privileges is currently limited to the management of application privileges.
      */
     global?: Dictionary<string, UserDefinedValue>
     /**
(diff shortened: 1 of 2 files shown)


specification/security/put_role/SecurityPutRoleRequest.ts
```diff
@@ -38,12 +38,44 @@ export interface Request extends RequestBase {
     refresh?: Refresh
   }
   body?: {
+    /**
+     * A list of application privilege entries.
+     */
     applications?: ApplicationPrivileges[]
+    /**
+     * A list of cluster privileges. These privileges define the cluster-level
+     * actions for users with this role.
```
jrodewig, line 47: Drive-by comment:
Multi-line comments insert escaped line breaks (`\n`) into the generated `schema.json`. I've been avoiding them, but it may not matter. @delvedor or @sethmlarson may be able to say.

We probably wouldn't want to insert line breaks into any docs we generate from this tho.

lockewritesdocs: Thanks @jrodewig, that's good feedback. I inspected the generated `schema.json` and see the line breaks. I removed them in 869a5de53ed9c8c48a17073a1ee2b7d77c6e8642

specification/security/put_role/SecurityPutRoleRequest.ts
```diff
@@ -38,12 +38,34 @@ export interface Request extends RequestBase {
...
+     * A list of indices permissions entries.
+     */
     indices?: IndicesPrivileges[]
+    /**
+     * Optional metadata. Within the metadata object, keys that begin with an underscore (`_`) are reserved for system use.
+     */
     metadata?: Metadata
+    /**
+     * A list of users that the owners of this role can impersonate.
+     * @doc_url https://www.elastic.co/guide/en/elasticsearch/reference/current/run-as-privilege.html[Submit requests on behalf of other users]
```
lockewritesdocs, line 63: @delvedor, is this formatting correct for `@doc_url`? I see in the output how the URL renders, but I'm curious how the explanatory text in brackets actually displays.

>"docUrl": "https://www.elastic.co/guide/en/elasticsearch/reference/current/run-as-privilege.html[Submit requests on behalf of other users]",

delvedor: The `doc_url` should just be a plain URL, no markdown/asciidoc format :)

specification/security/put_role/SecurityPutRoleRequest.ts
delvedor: I think this should not be happening, we should be able to write multiline descriptions, but the resulting JSON should not contain line breaks. I'll send a pr to address this :)

delvedor (approved):

PR #553, opened 2021-08-18
Avoid line breaks in descriptions and validate doc_url

delvedor:
As titled.

Related: https://github.com/elastic/elasticsearch-specification/pull/550

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -19703,7 +19703,7 @@
       },
       "path": [
         {
-          "description": "Comma-separated list of data streams, indices, and index aliases used to limit the request. Wildcard (*) expressions are supported.\nTo target all data streams and indices in a cluster, omit this parameter or use _all or *.",
+          "description": "Comma-separated list of data streams, indices, and index aliases used to limit the request. Wildcard (*) expressions are supported. To target all data streams and indices in a cluster, omit this parameter or use _all or *.",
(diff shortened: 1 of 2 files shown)


jrodewig (approved):
👍  LGTM. Thanks!

sethmlarson (approved):
LGTM!

lockewritesdocs (approved):
LGTM 👍 Thanks @delvedor!

output/schema/schema.json
```diff
@@ -111810,7 +111810,7 @@
       ],
       "query": [
         {
-          "description": "Specifies what to do when the request:\n- Contains wildcard expressions and there are no models that match.\n- Contains the _all string or no identifiers and there are no matches.\n- Contains wildcard expressions and there are only partial matches.",
+          "description": "Specifies what to do when the request: - Contains wildcard expressions and there are no models that match. - Contains the _all string or no identifiers and there are no matches. - Contains wildcard expressions and there are only partial matches.",
```
swallez, line 111813: Why don't we allow line breaks in descriptions? The above was a list and will be output as a weird single line with dashes.

We should settle on a markup language for descriptions that allows formatting for long descriptions. Considering the use of lists and `` `code` `` snippets, I would suggest CommonMark that has implementations in most languages and can thus be used in code generators.

PR #554, opened 2021-08-18
[DOCS] Document parameters for the 'search' API

sethmlarson:
Wanted to fill in a bunch of descriptions for the `search` API.

Also `@server_default` doesn't work great for `union_of` types which is unfortunate because there are a couple cases where we can't use `@server_default` with these parameters like `_source`.


philkra (approved):
LGTM

PR #559, opened 2021-08-19
[DOCS] Update security indices

lockewritesdocs:
Applies changes from #551, which were merged into a GitHub void and utterly consumed by its crushing power.


delvedor (approved):
Thanks!

PR #567, opened 2021-08-23
Fix Security Responses

philkra:
* `security.get_service_accounts` response
* `security.get_service_credentials` response


delvedor (approved):
LGTM

PR #569, opened 2021-08-23
[ON HOLD] Fix snapshot.get Response

philkra:
as titled.


philkra:
need to discuss the handling of outdated test artefacts.

PR #571, opened 2021-08-24
Improve Watcher Requests

philkra:
improved the following requests but not resolved them:

* `watcher.stats`
* `watcher.put_watch`


delvedor (approved):
LGTM

PR #581, opened 2021-08-24
Fix cluster.delete_voting_config_exclusions API

sethmlarson:
Still wasn't able to get the validation error for "no body" on the Request. Want feedback on where I should go to fix that issue.


delvedor (approved):
LGTM

Let's merge it for now, I think this should be fixed in the validator step (cc @swallez)

PR #589, opened 2021-08-24
Unstub the Stubbed Req & Resp Stubs

philkra:
* indices.get_upgrade
* indices.simulate_template
* indices.upgrade
* logstash.delete_pipeline
* logstash.get_pipeline
* logstash.put_pipeline
* ml.find_file_structure
* ml.put_trained_model
* monitoring.bulk
* rollup.rollup

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -89037,19 +89037,7 @@
         "name": "Request",
         "namespace": "indices.get_upgrade"
       },
-      "path": [
-        {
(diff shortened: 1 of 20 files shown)


delvedor (changes requested):
Heya, the goal is not only to remove stubbed properties but also to add the correct definition :)

delvedor:
Other prs will address this :)

PR #592, opened 2021-08-24
Add shutdown API specs

swallez:
As titled

### DIFF
--- a/specification/shutdown/delete_node/ShutdownDeleteNodeResponse.ts
+++ b/specification/shutdown/delete_node/ShutdownDeleteNodeResponse.ts
@@ -17,8 +17,6 @@
  * under the License.
  */
 
-export class Response {
-  body: {
-    acknowledged: boolean
-  }
-}
+import { AcknowledgedResponseBase } from '@_types/Base'
+
+export class Response extends AcknowledgedResponseBase {}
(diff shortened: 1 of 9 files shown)


delvedor (changes requested):
LGTM, please run `make contrib`

swallez:
Ran `make contrib`. Please cross-check!

specification/shutdown/delete_node/ShutdownDeleteNodeResponse.ts
```diff
@@ -18,5 +18,7 @@
  */
 
 export class Response {
```
delvedor, line 20: You should use `AcknowledgedResponseBase` :)
```suggestion
export class Response extends AcknowledgedResponseBase {}
```

specification/shutdown/put_node/ShutdownPutNodeResponse.ts
```diff
@@ -18,5 +18,7 @@
  */
 
 export class Response {
```
delvedor, line 20: Same here.

delvedor (changes requested):
Just a nit.

swallez:
Fixed to use `AcknowledgedResponseBase` (this actually more than an nit!)

delvedor (approved):

PR #599, opened 2021-08-25
Add support for unions in server_default

delvedor:
As titled 🚀


sethmlarson (approved):
LGTM

swallez (changes requested):
Looks like all the use cases in this PR for unions with server defaults are for properties declared as

```ts
some_prop: integer | string // TODO: should be int only
```

So do we really need unions for server defaults when it seems that we should just fix the spec and remove these unions?

delvedor:
@swallez it might be that the yaml test are wrong for this case, but in anyhow this seems a legit feature to me.

swallez (approved):
I have the feeling that we aren't going to need it, but we'll see...

PR #615, opened 2021-08-30
Fix required properties

delvedor:
This pr fixes a bunch of `path_parts` that should be required.

This change will break the validation for [`security.clear_api_key_cache`](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-clear-api-key-cache.html), our test are sending requests without `ids`, but in both docs and rest-api-spec `ids` is required.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -901,6 +901,7 @@ export interface ReindexRethrottleResponse {
 }
 
 export interface RenderSearchTemplateRequest extends RequestBase {
+  id?: Id
   body?: {
     file?: string
     params?: Record<string, any>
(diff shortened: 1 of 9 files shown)


sethmlarson (approved):
LGTM

output/typescript/types.ts
```diff
@@ -901,6 +901,7 @@ export interface ReindexRethrottleResponse {
 }
 
 export interface RenderSearchTemplateRequest extends RequestBase {
+  id?: Id
```
swallez, line 904: Shouldn't this be in nested in a `path_parts`? It's confusing to see a `body` property and free-floating properties (same comment for other similar additions).

PR #618, opened 2021-08-30
Validate the 'update' API

sethmlarson:
Two test cases use strings for the `routing` parameter, will have to fix that in rest-api-spec.


philkra (approved):
LGTM

Issue #580, opened 2021-08-24
Add compile check that `body: Void` isn't used on Request types


sethmlarson:
We use `body: Void` for Responses to signal that there isn't a response body however on Requests this almost works but adds a validation error for that API which says "API shouldn't have a body" despite having `kind: no_body` in the `schema.json`.

Instead we want to not have a `body` field at all in the Request types. We should add a compile error that disallows this from almost working.

PR #620, opened 2021-08-31
Validate the 'cluster.remote_info' API

sethmlarson:
As titled.

### DIFF
--- a/specification/cluster/remote_info/ClusterRemoteInfoRequest.ts
+++ b/specification/cluster/remote_info/ClusterRemoteInfoRequest.ts
@@ -29,5 +29,4 @@ import { Void } from '@spec_utils/VoidValue'
  * @stability stable
  */
 export interface Request extends RequestBase {
-  body: Void
 }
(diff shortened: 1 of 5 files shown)


philkra (approved):
LGTM

philkra (approved):
LGTM

delvedor:
> Where should this response go to make sure we're actually testing proper data?

You can create a custom test in the flight recorder.

specification/cluster/remote_info/ClusterRemoteInfoRequest.ts
```diff
@@ -18,14 +18,16 @@
...
  * @rest_spec_name cluster.remote_info
  * @since 6.1.0
- * @stability TODO
+ * @stability stable
  */
 export interface Request extends RequestBase {
-  body?: {
-    stub: string
-  }
+  body: Void
```
delvedor, line 32: There is no need to use `body: Void`, if the API has no body at all, all you need to do is to not define the body :)
```suggestion
```

delvedor (changes requested):

specification/cluster/remote_info/ClusterRemoteInfoRequest.ts
sethmlarson: Sure, either way I get a validation error unfortunately.

delvedor: Related: https://github.com/elastic/elasticsearch-specification/issues/580

delvedor (approved):
LGTM

PR #624, opened 2021-08-31
Fix stubbed ml apis

delvedor:
As titled.

I've removed `ml.find_file_structure` as is no longer present in ES.


specification/ml/put_trained_model/MlPutTrainedModelRequest.ts
```diff
@@ -17,21 +17,36 @@
...
-    stub?: string
+    /** The unique identifier of the trained model. */
+    model_id: Id
   }
   body?: {
-    stub?: string
+    /** The compressed (GZipped and Base64 encoded) inference definition of the model. If compressed_definition is specified, then definition cannot be specified. */
+    compressed_definition?: string
+    /** The inference definition for the model. If definition is specified, then compressed_definition cannot be specified. */
+    definition?: Definition
```
darnautov, line 40: Can we introduce a union type to handle these mutually exclusive properties?

delvedor: Since it's already explained in the property descriptions for now we prefer to keep it this way. We still need to figure out an elegant way to address this case that works well in every downstream client.

darnautov (approved):

PR #627, opened 2021-09-01
Fix addDescription step

delvedor:
This pr updates the `addDescription` step of the compiler which wasn't updated with the last iteration of the spec.
Now, if a path or query parameter does not contain a description in the input specification, the description will be pulled from the rest-api-spec.


philkra (approved):
LGTM

ezimuel:
@delvedor we also discussed collecting the missing description in a log file to forward to the docs team. I think we should open an issue to track this improvement, WDYT?

delvedor:
I'm not against creating such file, but I think is useless, at least for now. We are missing the descriptions for **so many** parameters that it would be a massive logfile that no one will read.
Eventually, we will have more descriptions, while keeping the spec updated.

PR #633, opened 2021-09-01
Bulk Operation properties should be optional

delvedor:
As titled, see [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html).


philkra (approved):
LGTM

sethmlarson (approved):

PR #634, opened 2021-09-01
Fix shutdown get node API

stevejgordon:
- Updates NodeIds to be a union of `string` and `string[]` to align with other types.
- Updates Shutdown Get Node to use NodeIds rather than `NodeId[]` for the path part.

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -40,7 +40,7 @@ export type ActionIds = string // TODO: check if this should be an array of Acti
 export type Id = string
 export type Ids = Id | Id[]
 export type NodeId = string
-export type NodeIds = string | string[]
+export type NodeIds = NodeId | NodeId[]
 
 export type IndexName = string
 export type Indices = string | string[]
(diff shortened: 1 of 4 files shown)


specification/_types/common.ts
```diff
@@ -40,6 +40,7 @@ export type ActionIds = string // TODO: check if this should be an array of Acti
 export type Id = string
 export type Ids = Id | Id[]
 export type NodeId = string
+export type NodeIds = string | string[]
```
philkra, line 43: nit, can you please change to this:
`export type NodeIds = NodeId | NodeId[]`

philkra (changes requested):

specification/_types/common.ts
stevejgordon: Good catch! Fixed.

delvedor (approved):
LGTM


I think we should get rid of `NodeId` and `NodeIds` and just use `Id` and `Ids`.

PR #635, opened 2021-09-01
Fix delete autoscaling policy request

stevejgordon:
Use the Name type for the name path part rather than string.


philkra (approved):
LGTM

PR #638, opened 2021-09-01
Add documentation on spec documentation

swallez:
Adds a doc explaining how to write and structure JSDoc documentation in the specification.

@goodroot I added you as a reviewer, feel free to add more people from the docs team to raise awareness and provide feedback.

### DIFF
--- a/docs/doc-comments-guide.md
+++ b/docs/doc-comments-guide.md
@@ -1,6 +1,6 @@
 # Documenting the API specification
 
-A specification is not only about formalizing data structures, it's also about explaining what these structures and their properties mean are used for.
+A specification is not only about formalizing data structures, it's also about explaining what these structures and their properties are used for.
 
 Documentation of the TypeScript specification is made using [JSDoc](https://jsdoc.app/) comments, i.e. multiline comments starting with `/**` above a type or field declaration.
 
--- a/docs/doc-comments-guide.md
+++ b/docs/doc-comments-guide.md
@@ -1,0 +1,90 @@
+# Documenting the API specification
+
+A specification is not only about formalizing data structures, it's also about explaining what these structures and their properties mean are used for.
+
+Documentation of the TypeScript specification is made using [JSDoc](https://jsdoc.app/) comments, i.e. multiline comments starting with `/**` above a type or field declaration.
+
+Additional lines start with a `*` followed by a space. Long lines are allowed but it's better if text is formatted to a maximum of 120 characters per line.
+
+## Example
+
+```ts
+/**
+ * @rest_spec_name rank_eval
+ * @since 6.2.0
+ */
+export interface Request extends RequestBase {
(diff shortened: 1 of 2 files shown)


docs/doc-comments-guide.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Terseness
+
+**Doc comments are reference material**: they should be as succinct as possible while capturing all the necessary information to use the elements they're documenting. Remember that they will often show up in small IDE autocompletion popups!
+
+In particular, doc comments are not the right place for tutorials or examples, which should be in dedicated documentation pages. These pages can of course be linked from the doc comments.
+
+API endpoints will also have a `@doc_url` JSDoc tag that links to that API's detailed documentation page.
+
+### Multi-paragraph doc comments
```
philkra, line 86: Can we merge the content with [this section](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#description)?

goodroot (commented):
Thanks @swallez 👍 

Appreciate the documentation. The writing advice re: how to structure annotations to be succinct, contain  a descriptive line before substantial paragraphs, is also appreciated. It is a step forward to a standard we can adopt across many contexts when generating API docs. 

Pinging writers from Elasticsearch: @debadair @jrodewig @lockewritesdocs 

And writers with a general interest in API docs: @KOTungseth @lcawl @kilfoyle 

Folks: this PR documents how the API specifications now support GitHub Flavoured Markdown. These annotated TypeScript API specifications render down into JSON which an eventual "next docs" generator will turn into MDX doc pages.  Happy to unpack in greater detail and no action is required, but feel free to leave feedback.

delvedor (commented):
I'm wondering if we should also add a check in the compiler to verify if the `description` can be parsed with GFM.

docs/doc-comments-guide.md
swallez: Added a [commit](https://github.com/elastic/elasticsearch-specification/pull/638/commits/8483d278157c687fc00f1224237f5d83bd9cbd72) that that adds a link to this document in that section.

docs/doc-comments-guide.md
```diff
@@ -0,0 +1,90 @@
+# Documenting the API specification
+
+A specification is not only about formalizing data structures, it's also about explaining what these structures and their properties mean are used for.
```
gtback, line 3:
```suggestion
A specification is not only about formalizing data structures, it's also about explaining what these structures and their properties are used for.
```

I think maybe an extra word here?

gtback (approved):
❤️ documentation about how to write documentation ✍🏻 .

docs/doc-comments-guide.md
swallez: Indeed, good catch!

PR #642, opened 2021-09-02
Fix ML put trained model alias request

stevejgordon:
Use `Name` rather than `string` for the `model_alias` path part.


delvedor (approved):
LGTM

PR #649, opened 2021-09-02
Fix snapshot.create Response

philkra:
as titled.


delvedor (commented):
LGTM, are we sure that both properties in the response are optional?

philkra:
@delvedor , good call. `snapshot` is mandatory!

PR #650, opened 2021-09-02
Fix Snapshot.get

philkra:
https://github.com/elastic/elasticsearch-specification/pull/649 will add remaining validation for the response.


sethmlarson (approved):
LGTM

PR #652, opened 2021-09-02
Fix slm.put_lifecycle Request

philkra:
as titled.

Problem was that `CronExpression` was a class extending `ScheduleBase` which is an empty object, hence the correct type string could not be casted.

Commented out obsolete classes.

Will also fix, `slm.get_lifecycle` response

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -130630,8 +130630,10 @@
       },
       "properties": [
         {
+          "description": "If false, the snapshot fails if any data stream or index in indices is missing or closed. If true, the snapshot ignores missing or closed data streams and indices.",
           "name": "ignore_unavailable",
           "required": false,
+          "serverDefault": false,
           "type": {
             "kind": "instance_of",
             "type": {
@@ -130641,17 +130643,7 @@
           }
         },
         {
-          "name": "include_global_state",
-          "required": false,
-          "type": {
-            "kind": "instance_of",
-            "type": {
-              "name": "boolean",
-              "namespace": "internal"
-            }
-          }
-        },
-        {
+          "description": "A comma-separated list of data streams and indices to include in the snapshot. Multi-index syntax is supported. By default, a snapshot includes all data streams and indices in the cluster. If this argument is provided, the snapshot only includes the specified data streams and clusters.",
           "name": "indices",
           "required": true,
           "type": {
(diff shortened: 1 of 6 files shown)


delvedor (approved):
LGTM, @swallez can you double-check if removing `Schedule` is an issue for you? This looks like the same case of `RankFeatureFunction`.

swallez:
**TL;DR:** we can remove them as from a spec perspective they don't add value, but we have to devise something for object-oriented languages like Java and C#.

**Discussion:**

These empty base classes seem unnecessary in languages that are not strictly object-oriented. However having a base class (or a marker interface) is necessary to produce idiomatic code in strongly typed object-oriented languages.

Let me expand on this. `ScheduleContainer` describes the various kinds of `Schedule` and from an OOP perspective it is actually a polymorphic type definition that describes all possible schedules.

So in essence `ScheduleContainer` tells us that there is a `Schedule` base type, and its members exhaustively define the implementations of this base type that ES understands. So in languages like Java (or C#), a `schedule: ScheduleContainer` property should actually be represented as `schedule: Schedule`, i.e. a polymorphic type that allows users to use any of the implementing subclasses.

Now the (not so) fun part is things like `monthly?: TimeOfMonth[]`: it's an array and not an object that can extend a base class. However [if we look at the ES code base](https://github.com/elastic/elasticsearch/blob/2d1e8b3abd57d0ecc831d5460b42365117f2aea7/x-pack/plugin/watcher/src/main/java/org/elasticsearch/xpack/watcher/trigger/schedule/MonthlySchedule.java#L26), the corresponding data structure is a wrapper around an array that extends the base class.

So we have here a discrepancy between the _intention_ of the API (polymorphic types) and their _actual representation_ (types that cannot have a common parent class because of the array). Since the API follows the JSON representation, having base types that are used only by a subset of the container's members definitely feels awkward.

So in the end, I think we should get rid of these empty base classes for polymorphic types, and have OOP language generators automatically create a base class and produce wrappers for arrays like in the example above.

I'm currently working on this for the Java generator, so we should have a final solution (with actual working client code) soon for this hairy topic.

ezimuel:
@swallez I understood your point and I agree. I would implement this empty class as an interface (eg. in PHP), just to have a common marker to be able to inject a specific class implementation (kind of polymorphic type).
Maybe, we can reference this empty classes as `"kind": "empty-interface"` instead of `"interface"`.

swallez:
@ezimuel thanks! I wasn't sure about PHP, but I've seen that it makes the same distinction between interfaces and classes than Java.

Actually I'm suggesting that we remove these empty classes entirely from the spec since they aren't necessary to define a correct API spec. And also because these array properties in container classes mean that some members of the containers will not extend these empty classes anyway.

What I'm suggesting is that producing these marker interfaces should be the responsibility or language generators that need them to produce idiomatic polymorphic code.

If a class has a `@variants container` annotation, then a marker interface should be generated, and the _generated type_ of each of its members must implement this marker interface. "Generated type" meaning that when member types are classes then we just have to add an `implements` clause to their definition, and when they're not classes (e.g. arrays) then the generator has to create a wrapper class that implements the marker interface. Pretty much like what can be found in the ES server code.

Also, since we have the `@variants container` annotation, I would remove the `-Container` suffix from these classes, i.e. `SchedulerContainer` should be named `Scheduler`. No need to pollute the names with this technicality that is already captured by the annotation.

swallez (approved):
LGTM, you can remove the empty base classes.

PR #653, opened 2021-09-02
Add 'search_mvt' API

sethmlarson:
This API is different than others because it's response body is a binary format. Wasn't sure how else to represent this except for `any`? Currently there aren't any test cases for this API but I plan on adding some.

cc @iverase

### DIFF
--- a/specification/_global/search_mvt/SearchMvtResponse.ts
+++ b/specification/_global/search_mvt/SearchMvtResponse.ts
@@ -1,0 +1,22 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export class Response {
+  body: any
+}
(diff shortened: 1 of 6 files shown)


specification/_global/search_mvt/SearchMvtResponse.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export class Response {
```
philkra, line 20: Would it make sense to create a dedicated `MapboxResponse` or for the `body: Mapbox`? IMO, it makes sense to indicate a differentiation from a JSON response.

@delvedor, I remember some precedent in an ML API. As for `_cat`, JSON is enforced

delvedor: By default, cat apis returns a string, but we are running the recording with `format: 'json'` to be able to test our types.

I think we can create a `Binary` spec util to handle this case.

delvedor (changes requested):
You should not add `specification/_json_spec/search_mvt.json`, it should be handled by our automation 🤔 
Let me double-check why it's not already there.

Edit: #654 and #655 should fix this.

specification/_global/search_mvt/SearchMvtResponse.ts
swallez: Since this isn't some arbitrary binary data but a well-defined one, what about:

```ts
// Arbitrary binary data
type Binary = ArrayBuffer

// Vector tile response
type MapboxVectorTiles = Binary
```

That way we capture both that it's a binary format through the type alias and that more specifically it's MVT data.

Code generators can then either provide the MVT data type of a geo library or just plain binary data that the user will have to handle by themselves.

delvedor: Let's open a separate issue for this, @sethmlarson for now use `UserDefinedValue`, in a follow-up or we'll use a more appropriate type.

sethmlarson:
Thanks for all the reviews everyone!

delvedor (approved):
LGTM

PR #654, opened 2021-09-03
Updates for v8

delvedor:
This pr removes all the `type` properties from path_parts where is no longer supported in v8.
It also removes apis that are no longer present in ES v8.
Furthermore, it unblocks the rest-api-spec action which was failing because of this.

**Do not backport to v7.x!**


philkra (approved):
LGTM

PR #655, opened 2021-09-03
Update rest-api-spec action

delvedor:
As titled. This should fix the current issue once merged with https://github.com/elastic/elasticsearch-specification/pull/654. Furthermore, if it fails it will now open an issue.

### DIFF
--- a/.github/workflows/update-rest-api-json.yml
+++ b/.github/workflows/update-rest-api-json.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        branch: [main, 7.x, 7.14]
+        branch: ['main', '7.x', '7.14', '7.15']
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/update-rest-api-json.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        branch: [main, 7.x, 7.14]
+        branch: ['main', '7.x', '7.14', '7.15']
```
philkra, line 15: can you change this to a pattern, otherwise we'll have to remember this for every release?

for example: 
```
 branches:
      - main
      - master
      - '[0-9]+.[0-9]+'
      - '[0-9]+.x'
```

^^ copied from https://github.com/elastic/elasticsearch-net/blob/master/.github/workflows/test-jobs.yml#L13

philkra (changes requested):

.github/workflows/update-rest-api-json.yml
delvedor: I'll open a follw-up for this, unfortunately, this is not github actions syntax, is the input for the nodejs script that downloads the artifacts.

PR #674, opened 2021-09-03
Type aliases for multiple items use correct alias

stevejgordon:
`Indices` and `Names` should use the related type alias rather than `string`.

Note: I've left `Metrics` for the time being as this is a little more nuanced. We likely want the individual metrics to be constrained (enum?) to the correct values. Right now, string is more permissive until we agree a design for this.


delvedor (approved):
LGTM

PR #541, opened 2021-08-13
Revert ML types changes

delvedor:
This pr reverts every pr we did to the ML types in the past two weeks but https://github.com/elastic/elasticsearch-specification/pull/524, to be able to unblock https://github.com/elastic/kibana/pull/107536.

PR #681, opened 2021-09-03
Fix ml.estimate_model_memory Request

philkra:
as titled.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -10154,7 +10154,7 @@ export interface MlAnalysisConfig {
   categorization_field_name?: Field
   categorization_filters?: string[]
   detectors: MlDetector[]
-  influencers: Field[]
+  influencers?: Field[]
   model_prune_window?: Time
   latency?: Time
   multivariate_by_fields?: boolean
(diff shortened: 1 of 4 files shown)


delvedor (commented):
@elastic/ml-core @elastic/ml-ui can you take a look?

output/typescript/types.ts
```diff
@@ -10154,7 +10154,7 @@ export interface MlAnalysisConfig {
   categorization_field_name?: Field
   categorization_filters?: string[]
   detectors: MlDetector[]
-  influencers: Field[]
+  influencers?: Field[]
```
droberts195, line 10157: This is the same problem as https://github.com/elastic/elasticsearch-specification/pull/544/files#r689565963

I believe `influencers` being made optional was the main change that caused #541 to have to be merged in a hurry 3 weeks ago.

It seems that a single `MlAnalysisConfig` cannot be shared between responses and requests.  Is there a precedent for how to name these?

Then for getting job configs the version without optional `influencers` should be used, and for requests (like estimate model memory, but also put job, update job, etc.) the version _with_ optional `influencers` should be used.

The only other alternative would be to make it optional everywhere but change the ML UI code to treat `influencers` not present on get job as being the same as an empty list of influencers _even though that situation will never be seen in practice_.

/cc @jgowdyelastic

delvedor:
Closing in favor of #544.

PR #692, opened 2021-09-06
Spec cleanup

delvedor:
There is no need to have `path_parts`, `query_parameters`, and `body` marked as optional in requests definitions, they should be there or not. We had the optional syntax for some arcane historical reason that no longer exists.


swallez (approved):
LGTM! Let's kill useless cargo culting :-)

PR #648, opened 2021-09-02
Fixes and descriptions for basic indices APIs

sethmlarson:
Adds descriptions and tweaks the types of many common `indices.*` APIs.

Noticed a large number of parameters either not needed or not named properly for `indices.put_mapping` so corrected these.

PR #693, opened 2021-09-06
Remove index from PropertyBase

stevejgordon:
#648 added `index?: boolean` to `PropertyBase` which seems incorrect as not all derived properties (fields) support this option. It also causes the .NET generated code not to compile since a derived type cannot define a property with the same name as one of the types it inherits from. 

This PR returns `PropertyBase` to its previous state.

@sethmlarson Do let me know if there was a specific reason for adding this we need to consider? We can also discuss on the type sync tomorrow.


stevejgordon:
@delvedor As discussed today, happy for me to merge this?

delvedor:
@stevejgordon go for it!

PR #630, opened 2021-09-01
Fix stubbed rollup API

stevejgordon:
As this API is not documented, quite complex, and experimental, we have agreed to treat the request and response body as `UserDefinedValue` for now. I'm awaiting some feedback from the Elasticsearch engineers to understand if this API is valid still.

PR #700, opened 2021-09-07
Add indices.disk_usage Enpoint

philkra:
added the request but I need some help from @delvedor & @swallez how to model the reponse.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -87454,25 +87454,9 @@
     },
     {
       "body": {
-        "kind": "properties",
-        "properties": [
-          {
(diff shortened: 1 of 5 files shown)


delvedor (commented):
Given it's still an experimental API, I would define the entire response as `UserDefinedValue` and revisit later, similar to what we did in https://github.com/elastic/elasticsearch-specification/pull/630.

philkra:
adjusted to @delvedor https://github.com/elastic/elasticsearch-specification/pull/700#pullrequestreview-748786846

delvedor (approved):

PR #702, opened 2021-09-08
Add support for OverloadOf behavior

delvedor:
As titled! 🚀 

This will solve most of the issues identified in #544.
cc @droberts195


delvedor:
I’m not sure if `ReadOf` is the best name, we can generalize it further and call it `CopyOf` or `SameAs` or something along those lines. What do you think?

compiler/steps/read-definition-validation.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    for (const parent of model.types) {
+      if (parent.name.name === readBehavior.generics[0].type.name && parent.name.namespace === readBehavior.generics[0].type.namespace) {
+        assert(parent.kind === 'interface')
+        const readProperties = type.properties.map(p => p.name)
+        for (const property of parent.properties) {
+          // have we defined the same properties?
+          if (!readProperties.includes(property.name)) {
+            console.log(chalk.red`The property '${property.name}' is present in ${parent.name.namespace}.${parent.name.name} but not in ${type.name.namespace}.${type.name.name}`)
+            process.exit(1)
```
swallez, line 46: An idea for later: we don't have location information here as we're not in the compiler. What about adding a `sourceLocation` property on type definition and property definitions that would allow us to provide context in this new step and also in the model validation?

swallez (approved):
LGTM with the new `OverloadOf` name we just discussed

compiler/steps/read-definition-validation.ts
delvedor: Good idea, we can even pass the ts compiler instance around, which should provide even more context. I'll think about it.

Issue #656, opened 2021-09-03
Add support for Binary data type


delvedor:
Some APIs, such as `search_mvt` returns binary data instead of JSON, we should have a custom type to represent this so clients can generate their code accordingly.

@swallez suggested in #653 to have a basic `Binary` type, which can be used as alias for describing custom responses:

```ts
// Arbitrary binary data
export class Binary {}

// Vector tile response
type MapboxVectorTiles = Binary


// search_mvt response definition
export class Response {
  body: MapboxVectorTiles
}
```

We can even create custom behavior for this:

```ts
/**
 * @behavior Arbitrary binary data
 */
export interface Binary {}

// Vector tile response
export class MapboxVectorTiles implements Binary {}

// search_mvt response definition
export class Response {
  body: MapboxVectorTiles
}
```

delvedor:
We agreed on the type alias version. An implementation will arrive soon.

PR #703, opened 2021-09-08
Add Binary data type

delvedor:
As titled.

Closes: #656.


delvedor:
I've changed our initial idea of using a marker `Binary` class as it was confusing in the resulting json output. I've updated the compiler to accept the native typescript `ArrayBuffer` type and output an `internal` custom binary type.

swallez (approved):
LGTM!

PR #711, opened 2021-09-08
Remove unused NEST options from APIs

sethmlarson:
Noticed these options on multiple APIs which looks like an artifact from the past? Removed as it didn't seem to be used/documented on the Elasticsearch side.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -20882,19 +20882,6 @@
                 "namespace": "_types"
               }
             }
-          },
-          {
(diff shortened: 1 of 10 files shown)


russcam:
If it helps, this represents the `q=<query>` query string parameter. It's named as `QueryOnQuerystring` in NEST as it would be ideal to name it `Query` to make it  less opaque, but this collides with the existing `Query` property in request types that sets the query in the body.

delvedor (approved):
LGTM

Issue #713, opened 2021-09-09
API security prerequisites 


lcawl:
## 🚀 Feature Proposal

Optional list of security privileges (and any other prerequisites?) for an API.

## Motivation

When security is enabled in Elasticsearch, a user must have specific [security privileges](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-privileges.html) for many of our APIs to succeed.

## Example

We currently document this type of information in the API reference docs like this, for example: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-prereqs

However, I think we're open to other formats. Since other Elastic products use OpenAPI specs (e.g. Cloud APIs), it might be useful to consider aligning this with the "security" field in https://swagger.io/specification/#openapi-object

PR #717, opened 2021-09-10
Add security prerequisites support

delvedor:
As titled.

Closes: #713.


compiler/model/utils.ts
```diff
@@ -547,6 +547,31 @@ export function hoistRequestAnnotations (
     } else if (tag === 'since') {
       assert(jsDocs, semver.valid(value), `Request ${request.name.name}'s @since is not valid semver: ${value}`)
       endpoint.since = value
+    } else if (tag === 'security_prerequisites_index') {
+      const privileges = [
```
philkra, line 551: can we reuse the values from this: https://github.com/elastic/elasticsearch-specification/pull/718 -- seems like an anti pattern to maintain the same values in two places, same for cluster below

docs/modeling-guide.md
```diff
@@ -469,3 +469,35 @@ export interface Request<TDocument> extends RequestBase {
...
+#### `@security_prerequisites_index`
+
+If an endpoint has some index security prerequisites to satisfy, you can specify them here.
+
+```ts
+/**
+ * @rest_spec_name indices.create
+ * @since 0.0.0
+ * @stability stable
+ * @security_prerequisites_index create_index manage
```
philkra, line 482: Please use as list separator the commata `,`. That overlaps how `@aliases` is used.

philkra (changes requested):

docs/modeling-guide.md
delvedor: Good catch!

compiler/model/utils.ts
delvedor: Unfortunately not, as the `specification` and the `compiler` are two separate typescript projects, one cannot import the other, as the `tsconfig.json` is very different.

philkra: that's unfortunate!

philkra (approved):
LGTM

lcawl:
In machine learning APIs, we have some that involve "source" and "destination" indices, with different index privileges for each. For example: https://www.elastic.co/guide/en/elasticsearch/reference/current/put-dfanalytics.html, https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html.  Would it be possible to accommodate that here?

delvedor:
@lcawl I would say that this should be done via the description, since it's an edge case, and the edge case differs based on the API, I would write all the prerequisites in the tag (both source/destination and leader/follower) and add a note about it in the API description.

lcawl:
That seems reasonable, thanks!

PR #718, opened 2021-09-10
Add more fine grained security privileges definition

delvedor:
As titled.


philkra (approved):
LGTM, but wouldn't it makes sense to have `ClusterPrivilege` and `IndexPrivilege` in one file called `Privileges` ?

delvedor:
Good call @philkra, done.

philkra (approved):
LGTM

PR #719, opened 2021-09-10
Add support for endpoint description

delvedor:
For some reason, we are not outputting the endpoint descriptions in the json, this pr fixes that.
If an endpoint definition is not present in the input spec, the rest-api-spec will be used, in the same way we are doing for the path/query parameters.


philkra (approved):
LGTM

lcawl (approved):
Yay, thanks!

PR #725, opened 2021-09-10
Fix indices.stats response

philkra:
as titled.

### DIFF
--- a/specification/indices/stats/types.ts
+++ b/specification/indices/stats/types.ts
@@ -74,6 +74,8 @@ export class IndexStats {
   /** Contains statistics about index warming operations for the node. */
   warmer?: WarmerStats
   bulk?: BulkStats
+  /** @since 7.15.0 */
+  shards?: ShardsTotalCount
 }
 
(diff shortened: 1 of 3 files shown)


specification/indices/stats/types.ts
```diff
@@ -176,4 +178,10 @@ export class ShardStats {
   translog: TranslogStats
   warmer: WarmerStats
   bulk?: BulkStats
+  /** @since 7.15.0 */
+  shards?: ShardsTotalCount
+}
+
+export class ShardsTotalCount {
```
sethmlarson, line 185: In the interest of future additions should we name this differently? Maybe `ShardsTotalStats`/`ShardsAllStats`?

Also is this field optional in 7.15+ or is it always there?

sethmlarson (commented):
Looks good, two small comments:

specification/indices/stats/types.ts
delvedor: @sethmlarson I agree, can you finish up this pr? Thanks!

PR #726, opened 2021-09-10
Fix update_by_query request

philkra:
as titled.


delvedor (commented):
According to the [doc](https://www.elastic.co/guide/en/elasticsearch/reference/7.15/query-dsl-range-query.html), the range query should be an object and not an array.
@swallez can you double check the ES source code to understand what's going on here?

swallez:
Confirmed after looking at the ES code: there's no support for arrays there.

philkra:
Ack, I'll dig how to resolve this.

philkra:
Opened https://github.com/elastic/elasticsearch/pull/77700 to fix the test upstream.

Issue #532, opened 2021-08-02
Missing TS types for hidden setting and data stream property on Get indices response


cjcenizal:
## 🐛 Wrong type

The `indices.get` response's indices are missing two properties:
* `settings.index.hidden`
* `data_stream`

## Definition

If I attempt to access the hidden property of an index's settings, e.g. `index.settings.index.hidden`, TS complains:

```
Property 'index' does not exist on type 'IndicesIndexSettings | IndicesIndexStatePrefixedSettings'.
  Property 'index' does not exist on type 'IndicesIndexSettings'. ts(2339)
```

However, this property exists and is accessible in the response. It looks like it's defined as an alias here: https://github.com/elastic/elasticsearch-specification/blob/main/specification/indices/_types/IndexSettings.ts#L74. My TS fu is not strong enough to suggest a fix for this.

If I attempt to access the data_stream property of an index, e.g. `index.data_stream`, TS complains:

```
Property 'data_stream' does not exist on type 'IndicesIndexState'.ts(2339)
```

I believe we need to update `IndexState` to have this property, defined here: https://github.com/elastic/elasticsearch-specification/blob/main/specification/indices/_types/IndexState.ts.

PR #727, opened 2021-09-10
Add variants to analysis classes, Add IndexState.data_stream

sethmlarson:
Most of this is adding `@variant` to analyzers, tokenizers, token filters and char filters.

Added `IndexState.data_stream?` to close https://github.com/elastic/elasticsearch-specification/issues/532

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -43787,6 +43787,106 @@
       ]
     },
     {
+      "kind": "type_alias",
+      "name": {
+        "name": "Analyzer",
+        "namespace": "_types.analysis"
+      },
+      "type": {
+        "items": [
+          {
+            "kind": "instance_of",
+            "type": {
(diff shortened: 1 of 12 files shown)


delvedor (approved):
Great work! ❤️

sebelga:
@sethmlarson You closed #532 for the `IndexState.data_stream` but I am not seeing the missing `settings.index.hidden` as part of this PR. Is this correct?

sethmlarson:
@sebelga Hmm, I saw [`@aliases index.hidden` on the `hidden` property](https://github.com/elastic/elasticsearch-specification/blob/main/specification/indices/_types/IndexSettings.ts#L77) and figured that solved the issue. @delvedor is that sufficient for this case or does more need to be done to make this value accessible?

sebelga:
Oh I see it got added in a previous PR. Thanks @delvedor

PR #735, opened 2021-09-14
Fix cluster apis

delvedor:
As titled.

`cluster.reroute` and `cluster.state`, according to the ES team, should not be typed, as they return the internal cluster state which isn't considered reliable.


specification/cluster/reroute/ClusterRerouteResponse.ts
```diff
@@ -17,12 +17,13 @@
  * under the License.
  */
 
-import { AcknowledgedResponseBase } from '@_types/Base'
-import { RerouteExplanation, RerouteState } from './types'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 
-export class Response extends AcknowledgedResponseBase {
-  body: {
-    explanations?: RerouteExplanation[]
```
Mpdreamz, line 25: Explanations should be stable? Understood stood state is untyped.

philkra (approved):
LGTM

specification/cluster/reroute/ClusterRerouteResponse.ts
delvedor: According to the es-distrib team, both this APIs don't have any stability guarantee.

Mpdreamz: kk I would challenge the `explanations` portion of this API. Now that we have a spec we can say this API is stable but the cluster state it exposes is just a map/object/dict/whatever with no guarantees.

PR #732, opened 2021-09-14
Add support for doc_id

delvedor:
As titled.

Closes: #714.

PR #746, opened 2021-09-15
Add doc_id; remove out-dated links

lcawl:
Related to https://github.com/elastic/elasticsearch-specification/pull/732

This PR tests the new [@doc_id ](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#doc_id) functionality. The ml.customUrls keyword in this case should resolve to the same type of URL as exists in https://github.com/elastic/kibana/blob/master/src/core/public/doc_links/doc_links_service.ts

I also removed some links to out-dated URLs that are no longer valid or necessary.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -119727,8 +119727,7 @@
             }
           },
           {
-            "description": "Advanced configuration option. The time between each periodic persistence of the model. See Job resources.",
-            "docUrl": "https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-job-resource.html",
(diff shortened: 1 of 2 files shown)


delvedor (changes requested):
LGTM

Can you run `make contrib`?

lcawl:
> Can you run `make contrib`?

I get this error: `Error: 'custom_settings' has the following unknown annotations: doc_id`. Is there something else I need to do to make it aware of this option?

delvedor (commented):
It looks like that `doc_id` can only live in request definitions, I'll update the compiler!

delvedor:
Oks, I've merged https://github.com/elastic/elasticsearch-specification/pull/755, please sync with the main branch and run `make contrib` again, sorry for the confusion!

delvedor (approved):

PR #750, opened 2021-09-16
Fix enumerations and add more meaningful identifiers

swallez:
This fixes a few enumerations (e.g. `GREEN / green` or `z-score`) and adds some meaningful `snake_case` identifiers that translate well using the various casing conventions used in generated code (e.g. `UpperCamelCase` in Java).


philkra (approved):
LGTM, awesome!

PR #755, opened 2021-09-20
Add support for doc_id in properties

delvedor:
As titled, see https://github.com/elastic/elasticsearch-specification/pull/746


swallez (approved):
LGTM

PR #756, opened 2021-09-20
Expand body identifier use

delvedor:
A body identifier should be configured for every `value` body.

### DIFF
--- a/specification/indices/put_settings/IndicesPutSettingsRequest.ts
+++ b/specification/indices/put_settings/IndicesPutSettingsRequest.ts
@@ -41,9 +41,5 @@ export interface Request extends RequestBase {
     timeout?: Time
   }
   /** @identifier settings */
-  body: IndexSettingsBody
-}
-
-export class IndexSettingsBody extends IndexSettings {
-  settings?: IndexSettings
+  body: IndexSettings
 }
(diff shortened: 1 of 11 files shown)


specification/rollup/rollup/RollupRequest.ts
```diff
@@ -31,5 +31,6 @@ export interface Request extends RequestBase {
     index: IndexName
     rollup_index: IndexName
   }
+  /** @identifier rollup */
```
swallez, line 34: Should be named `config`. There's no doc for it (hence the `UserDefinedValue`), but the json spec says that this is "The rollup configuration".

specification/indices/put_settings/IndicesPutSettingsRequest.ts
```diff
@@ -40,6 +40,7 @@ export interface Request extends RequestBase {
     preserve_existing?: boolean
     timeout?: Time
   }
+  /** @identifier settings */
```
swallez, line 43: As discussed, this should be `body: IndexSettings` and we should remove `IndexSettingsBody`. The nesting in a `settings` property is a legacy representation that ES server eliminates if present before processing the request further.

swallez (changes requested):

swallez (approved):
LGTM!

PR #765, opened 2021-09-21
Fix enrich.stats response

delvedor:
As titled.

### DIFF
--- a/specification/enrich/stats/EnrichStatsResponse.ts
+++ b/specification/enrich/stats/EnrichStatsResponse.ts
@@ -17,11 +17,12 @@
  * under the License.
  */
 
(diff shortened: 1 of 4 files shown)


specification/enrich/stats/EnrichStatsResponse.ts
```diff
@@ -17,11 +17,12 @@
...
  */
 
-import { ExecutingPolicy, CoordinatorStats } from './types'
+import { ExecutingPolicy, CoordinatorStats, CacheStats } from './types'
 
 export class Response {
   body: {
     coordinator_stats: CoordinatorStats[]
     executing_policies: ExecutingPolicy[]
+    cache_stats?: CacheStats[]
```
sethmlarson, line 26: `@since 7.16.0`?

PR #769, opened 2021-09-21
Fix Security.Get_User_Privileges

stevejgordon:
In the response at least, some `IndicesPrivileges` properties are returned as arrays, potentially containing a single item. Since this type is used for requests also where single items can be sent, I've made these a union.


delvedor (changes requested):
LGTM

Can you fix the conflicts?

PR #772, opened 2021-09-21
Fix search request

delvedor:
Both request and response types have been validated, there are still a few errors in the request, but some are caused by ids being numbers instead of string which will be fixed upstream, and others are caused by experimental aggregations that we are not tracking yet.

### DIFF
--- a/specification/_types/aggregations/pipeline.ts
+++ b/specification/_types/aggregations/pipeline.ts
@@ -18,13 +18,14 @@
  */
 
 import { Sort } from '@global/search/_types/sort'
+import { Dictionary } from '@spec_utils/Dictionary'
 import { Name, Field } from '@_types/common'
 import { integer, double, float } from '@_types/Numeric'
 import { Script } from '@_types/Scripting'
 import { Aggregation } from './Aggregation'
 
 export class PipelineAggregationBase extends Aggregation {
-  buckets_path?: BucketsPath
+  buckets_path?: string | Dictionary<string, string>
   format?: string
   gap_policy?: GapPolicy
 }
@@ -33,8 +34,6 @@ export enum GapPolicy {
   skip = 0,
   insert_zeros = 1
 }
-
-export class BucketsPath {}
 
 export class AverageBucketAggregation extends PipelineAggregationBase {}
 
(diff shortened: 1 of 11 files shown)


specification/_types/query_dsl/specialized.ts
```diff
@@ -121,9 +121,19 @@ export class PercolateQuery extends QueryBase {
...
+/**
+ * @variants container
+ */
 export class PinnedQuery extends QueryBase {
-  ids: Id[]
+  /** @variant container_property */
   organic: QueryContainer
+  ids?: Id[]
+  docs?: PinnedDoc[]
+}
```
delvedor, line 132: @swallez is this valid? Or should I manually declare the `QueryBase` properties as `container_property` here?

swallez: `PinnedQuery` isn't a container as it doesn't have variants. Its `organic` property is just a `QueryContainer` (which is indeed a container).

So no, it's not valid, these jstags should not be there and aren't needed.

Here's a breakdown of the structure:
```json5
{
  "query": {                       // query: QueryContainer
    "pinned": {                    //   variant identifier (PinnedQuery)
      "ids": [ "1", "4", "100" ],  //     PinnedQuery field (Id[])
      "organic": {                 //     PinnedQuery field (QueryContainer)
        "match": {                 //       variant identifier (SingleKeyDictionary<Field, MatchQuery>)
          "description": "iphone"  //         SingleKeyDictionary<Field, MatchQuery> entry
        }
      }
    }
  }
}
```

specification/_types/aggregations/pipeline.ts
```diff
@@ -18,13 +18,14 @@
...
 import { Sort } from '@global/search/_types/sort'
+import { Dictionary } from '@spec_utils/Dictionary'
 import { Name, Field } from '@_types/common'
 import { integer, double, float } from '@_types/Numeric'
 import { Script } from '@_types/Scripting'
 import { Aggregation } from './Aggregation'
 
 export class PipelineAggregationBase extends Aggregation {
-  buckets_path?: BucketsPath
+  buckets_path?: string | Dictionary<string, string>
```
swallez, line 28: From what I can see in the ES code base, the type of `bucket_path` heavily depends on the enclosing aggregation type: it can be a simple `string`, a simple `Dictionary<string, string>` or also a `string[]`. So it will have to be defined for each individual pipeline aggregation.

This some significant work, so will be revisited in a different PR.

In the meantime, can you also add the `string[]` union member and leave a TODO comment that this need to be revisited and specialized for each pipeline aggregation?

swallez (approved):
LGTM!

PR #773, opened 2021-09-21
obsolete/deprecation/g

delvedor:
As titled.

### DIFF
--- a/docs/modeling-guide.md
+++ b/docs/modeling-guide.md
@@ -516,3 +516,29 @@ export interface Request extends RequestBase {
  ...
 }
 ```
+
+#### `@deprecated`
+
+Use if an endpoint or property is deprecated, you should add the version as well.
+
+```ts
+class Foo {
+  bar: string
+  /** @deprecated 7.0.0 */
+  baz?: string
+  faz: string
+}
+```
+
+#### `@deprecated_description`
+
+Use if an endpoint or property is deprecated and you need to add a description.
+
+```ts
+class Foo {
+  bar: string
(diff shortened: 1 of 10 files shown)


docs/modeling-guide.md
```diff
@@ -516,3 +516,29 @@ export interface Request extends RequestBase {
...
+```
+
+#### `@deprecated_description`
+
+Use if an endpoint or property is deprecated and you need to add a description.
+
+```ts
+class Foo {
+  bar: string
+  /** @deprecated_description 7.0.0 */
```
sethmlarson, line 540: Should we document `@deprecated` and `@deprecated_description` together? Also should use prose in the `@deprecated_description` example?

specification/_types/query_dsl/geo.ts
```diff
@@ -77,7 +77,7 @@ export class GeoPolygonPoints {
   points: GeoLocation[]
 }
 
-/** @obsolete 7.12.0 Use geo-shape instead. */
+/** @deprecated 7.12.0 Use geo-shape instead. */
```
sethmlarson, line 80: Expand this to `@deprecated_description`

sethmlarson (changes requested):
Awesome! Thanks for this. One thought, can we get a validator that the `@deprecated` tag matches a version number if given? Also a couple other todos below:

delvedor:
@sethmlarson I've updated the `@deprecation` tag to accept both the version and an optional description, there is no need to have two distinct tags for this.

sethmlarson (approved):
LGTM

PR #831, opened 2021-09-29
Throw if the api can't have a body

delvedor:
As titled.

Closes: https://github.com/elastic/elasticsearch-specification/pull/830


swallez (approved):
LGTM.

We do have endpoints that have a body and accept `GET` (like `_search`), but as they also accept `POST`, they will pass this validation.

PR #844, opened 2021-10-05
Fix nodes.stats

Anaethelion:
Incomplete but most of the heavy lifting seems ok.
Will do a second part for the special cases left!


swallez (commented):
Some fields are still required. Should we make them all optional considering that filtering is a first class feature of this API?

specification/nodes/_types/Stats.ts
```diff
@@ -65,22 +142,52 @@ export class KeyedProcessor {
 }
 
 export class AdaptiveSelection {
-  avg_queue_size: long
-  avg_response_time: long
+  avg_queue_size?: long
+  avg_response_time?: long
   avg_response_time_ns: long
```
swallez, line 147: This field (and some others in other classes) is left required. Is this intentional?

delvedor (approved):
LGTM

Anaethelion:
I'll rebase the main branch to resolve conflicts!

PR #852, opened 2021-10-07
Adds missing doc details for ML get calendar APIs

lcawl:
This PR updates [get_calendar_events](https://github.com/elastic/elasticsearch-specification/tree/main/specification/ml/get_calendar_events) and [get_calendars](https://github.com/elastic/elasticsearch-specification/tree/main/specification/ml/get_calendars) with info from the documentation (https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-get-calendar-event.html, https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-get-calendar.html).

NOTE: The request body is omitted from the get calendar events API, per https://github.com/elastic/elasticsearch/issues/79271

### DIFF
--- a/specification/ml/get_calendars/MlGetCalendarsRequest.ts
+++ b/specification/ml/get_calendars/MlGetCalendarsRequest.ts
@@ -35,16 +35,17 @@ export interface Request extends RequestBase {
     calendar_id?: Id
   }
   query_parameters: {
-    /** Skips the specified number of calendars.
+    /** Skips the specified number of calendars. This parameter is supported only when you omit the calendar identifier.
      * @server_default 0
     */
     from?: integer
-    /** Specifies the maximum number of calendars to obtain. 
+    /** Specifies the maximum number of calendars to obtain. This parameter is supported only when you omit the calendar identifier.
      * @server_default 10000
     */
     size?: integer
   }
   body: {
+    /** This object is supported only when you omit the calendar identifier. */
     page?: Page
   }
 }
--- a/specification/ml/get_calendars/MlGetCalendarsRequest.ts
+++ b/specification/ml/get_calendars/MlGetCalendarsRequest.ts
@@ -37,11 +37,11 @@ export interface Request extends RequestBase {
   query_parameters: {
     /** Skips the specified number of calendars. This parameter is supported only when you omit the calendar identifier.
      * @server_default 0
-    */
+     */
     from?: integer
     /** Specifies the maximum number of calendars to obtain. This parameter is supported only when you omit the calendar identifier.
      * @server_default 10000
-    */
+     */
     size?: integer
   }
   body: {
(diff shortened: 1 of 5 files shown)


sethmlarson:
This is blocked on https://github.com/elastic/elasticsearch/issues/79271

specification/ml/get_calendars/MlGetCalendarsRequest.ts
```diff
@@ -23,22 +23,29 @@ import { Id } from '@_types/common'
...
+     * @server_default 0
+     */
     from?: integer
-    /** Specifies the maximum number of calendars to obtain. */
+    /** Specifies the maximum number of calendars to obtain. This parameter is supported only when you omit the calendar identifier.
+     * @server_default 10000
+     */
     size?: integer
   }
   body: {
```
sethmlarson, line 47: Given the decision in https://github.com/elastic/elasticsearch/issues/79271 should we omit the `body` in this API as well?

specification/ml/get_calendars/MlGetCalendarsRequest.ts
```diff
@@ -23,22 +23,29 @@ import { Id } from '@_types/common'
...
  * @rest_spec_name ml.get_calendars
  * @since 6.2.0
  * @stability stable
+ * @cluster_privileges monitor_ml
  */
 export interface Request extends RequestBase {
   path_parts: {
-    /** A string that uniquely identifies a calendar. */
+    /** A string that uniquely identifies a calendar. You can get information for multiple calendars by using a comma-separated list of ids or a wildcard expression. You can get information for all calendars by using `_all` or `*` or by omitting the calendar identifier.*/
     calendar_id?: Id
```
sethmlarson, line 35: Should this parameter be required if we omit `body`?

sethmlarson (changes requested):
Left two comments about the request body of `ml.get_calendars`

specification/ml/get_calendars/MlGetCalendarsRequest.ts
lcawl: This particular API didn't fail when I ran the `make contrib`, so I thought it was not an exceptional one.  It has both "GET" and "POST" in its spec here: https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ml.get_calendars.json

sethmlarson: @lcawl You're right, I must have been looking at the wrong spec.

sethmlarson (approved):
LGTM

specification/ml/get_calendars/MlGetCalendarsRequest.ts
lcawl: I don't think it's required when you omit the body. For example, these work in 7.16:

> GET _ml/calendars/
> GET _ml/calendars?from=0

However, if you try to specify both a calendar ID and a query parameter, you get an error. For example:

> GET _ml/calendars/_all?from=0
... returns the following error:
> "Validation Failed: 1: Params [from, size] are incompatible with [calendar_id].

Issue #854, opened 2021-10-07
"Update Model Info" GitHub Action force-pushing which allows for merged commits to be deleted


sethmlarson:
After merging multiple PRs within a shorter timespan I discovered the commit from https://github.com/elastic/elasticsearch-specification/pull/850 wasn't in `main`, it had been overwritten by the Update Model Info GitHub Action started by the previous squash commit.

I believe the problem lies in the use of `--force-with-lease` option. Perhaps we should allow our Update Model Info action to fail if upstream has changed because the new commit will handle updating the model info?

Anaethelion:
This seems bound to happen again, I'd suggest we remove the `--force-with-lease` and update the action to be a time based one.
Updating the schema once or twice a day could be enough.

sethmlarson:
Having removed the data loss and not seeing any downside yet I'm going to close this issue.

PR #856, opened 2021-10-07
Removing force push option from the update model info action

Anaethelion:
Prevents dataloss for now until https://github.com/elastic/elasticsearch-specification/issues/854 has been discussed and a solution found.


Anaethelion:
Yup we do need a backport, thanks !

PR #868, opened 2021-10-12
Simplify ErrorCause definition

swallez:
This PR simplifies the definition of `ErrorCause` to include only the minimal set of fields that are common to all error types (even if some are optional). Depending on the error type, additional details may be provided in the `AdditionalProperties` spill-over. These details can have any shape, even if most of the time they will be strings or string arrays.

### DIFF
--- a/specification/_types/Errors.ts
+++ b/specification/_types/Errors.ts
@@ -39,7 +39,7 @@ export class ErrorCause
    */
   reason: string
   /**
-   * The server stack trace. Present only f the `error_trace=true` parameter was sent with the request.
+   * The server stack trace. Present only if the `error_trace=true` parameter was sent with the request.
    */
   stack_trace?: string
 
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -31053,7 +31053,7 @@
           }
         },
         {
-          "description": "The server stack trace. Present only f the `error_trace=true` parameter was sent with the request.",
+          "description": "The server stack trace. Present only if the `error_trace=true` parameter was sent with the request.",
           "name": "stack_trace",
           "required": false,
           "type": {


specification/_types/Errors.ts
```diff
@@ -17,56 +17,35 @@
  * under the License.
  */
 
-import { PainlessExecutionPosition } from '@global/scripts_painless_execute/types'
-import { Dictionary } from '@spec_utils/Dictionary'
-import { HttpHeaders, Id, Ids, IndexName, Uuid } from './common'
+import { Id, IndexName } from './common'
 import { integer, long } from './Numeric'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+import { AdditionalProperties } from '@spec_utils/behaviors'
```
technige, line 23: Is `AdditionalProperties` used for anything else? The name feels cumbersome in this context, and I'm wondering if we can pick something shorter, and possibly avoid the word "additional".

swallez: Additional properties is a "behavior", interfaces that capture things that can't be fully expressed in TypeScript.

That one is used to indicate that a class has a set of well-known properties with each their specific types (those described in the class) but also accepts arbitrary properties of a given type. It's heavily used in aggregations.

In this context the type of these additional properties is anything (`UserDefinedValue`), but in other contexts the type is well-defined (see [Aggregate.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/aggregations/Aggregate.ts))

Behaviors also trigger specialized code generation schemes, so a renaming is possible (naming is hard) but would have to be coordinated with updates in the code generators.

More details at https://github.com/elastic/elasticsearch-specification/blob/main/docs/behaviors.md

specification/_types/Errors.ts
```diff
@@ -17,56 +17,35 @@
...
+ */
+export class ErrorCause
+  implements AdditionalProperties<string, UserDefinedValue> {
   /**
-   * resource id
-   * @aliases resource.id
+   * The type of error
    */
-  resource_id?: Ids
+  type?: string
```
swallez, line 36: Note: `type` is optional as not all errors return it:
```
GET /_cat/foobar
{
  "error" : "Incorrect HTTP method for uri [/_cat/foobar?pretty=true] and method [GET], allowed: [POST]",
  "status" : 405
}
```

This format is captured by the `@shortcut_property reason`: if `error` is a string, then it's equivalent to an `ErrorCause` with a `reason` only.

swallez:
Added a commit to make `ErrorResponseBase.status` required after extensive inspection of the ES code base.

PR #869, opened 2021-10-12
Use aliases for GeoOrientation synonyms

swallez:
Uses aliases for `GeoOrientation` synonyms, and removes the duplicate `Orientation` enum.

Also moves `ShapeProperty` to `mapping/geo.ts` where it belongs.


Anaethelion (approved):
I was about to make a comment on enum, since it's already fixed, LGTM!

PR #877, opened 2021-10-13
Fix normalizer definition in index settings

swallez:
This fixes the normalizers definition for index settings. There's one builtin `lowercase` normalizer, and the `custom` normalizer defined as a filter pipeline.


sethmlarson (approved):
LGTM

PR #878, opened 2021-10-13
Fix TermsSetQuery and GeoDistanceSort

swallez:
Fixes two issues:
* `TermSetQuery` is defined as a `terms_set?: SingleKeyDictionary<Field, TermsSetQuery>` and doesn't have an additional nesting level
* `GeoDistanceSort` only accepts a single field's location

These issues were not caught by the flight recorder because the definitions were too lazy.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -23785,7 +23785,7 @@
     },
     {
       "attachedBehaviors": [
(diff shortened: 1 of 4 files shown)


philkra (approved):
LGTM

PR #894, opened 2021-10-15
Fix number properties definition

swallez:
This PR fixes the number field mapping definitions. Instead of having one single class with a `type` discriminant of an enum type, we now have a hierarchy of classes that takes into account the specificities of each type.


delvedor (approved):
LGTM

PR #906, opened 2021-10-19
Add property duplicates check

delvedor:
If you are generating clients with a body-less API, it might happen that you don't know what to do if a parameter is present in the path, query, and body. This pr enforces the use of an identifier in such cases, so we can have a consistent solution across all clients.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -84619,6 +84619,7 @@
       "path": [
         {
           "description": "The name of the index lifecycle policy",
+          "identifier": "name",
           "name": "policy",
           "required": true,
           "type": {
@@ -85096,6 +85097,7 @@
       "path": [
         {
           "description": "The name of the index lifecycle policy",
+          "identifier": "name",
           "name": "policy",
           "required": false,
           "type": {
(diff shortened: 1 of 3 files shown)


swallez:
Quick analysis or path duplications.

**Path and body duplication**:
- `_global.clear_scroll.Request: [scroll_id]` - path and body parameters are the same thing. The path parameter is [deprecated since 7.0.0](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/clear-scroll-api.html) in favor of the body
- `ilm.put_lifecycle.Request: [policy]` - path and body parameters are different things. Path parameter should be renamed `name` with an `@identifier`.
- `security.put_user.Request: [username]` - `username` in the body, if present, is checked against the one in the path. It therefore doesn't hurt if the same value is sent twice, in the path and in the body.
- `snapshot.create_repository.Request: [repository]` - path and body parameters are different things. Path parameter should be renamed `name` with an `@identifier`.

When duplicates are not the same thing: `@identifier` should be used on path parameters (rename path parameters as never appear on the wire contrarily to query parameters).

When duplicates are the same thing, we should not add an `@identifier`. Model validation should ensure they have the same type and code generators can choose to use either path or body param (driven by which one is required).

**Path and query duplication**:
- `_global.bulk.Request: [type]` - path and query parameters are the same thing.
- `watcher.stats.Request: [metric]` - path and query parameters are the same thing.

When a path and query duplication exist, model validation should ensure they have the same type. Clients should choose to send that parameter either as a path or query param. We **must not** add an `@identifier` to distinguish them, since they're really the same thing.

PR #908, opened 2021-10-19
Fine tuning and minor fixes

swallez:
Fixes a few tiny issues on some unrelated types:
- `Error.root_cause` should be optional
- `FieldMapping.mapping` maps fields
- `ExtendedMemoryStats` has some duplicate fields with the parent `ExtendedMemory`, but with different tyles (long vs int)


Anaethelion:
LGTM

PR #909, opened 2021-10-19
[ML] Updates specs for GET job, job stats, model snapshots, and overall buckets APIs

szabosteve:
## Overview

This PR adds descriptions to the properties of the GET job, job stats, model snapshots, and overall buckets APIs.

### DIFF
--- a/specification/ml/get_job_stats/MlGetJobStatsRequest.ts
+++ b/specification/ml/get_job_stats/MlGetJobStatsRequest.ts
@@ -22,10 +22,6 @@ import { Id } from '@_types/common'
 
 /**
  * Retrieves usage information for anomaly detection jobs.
- * You can get statistics for multiple anomaly detection jobs in a single API
- * request by using a group name, a comma-separated list of jobs, or a wildcard
- * expression. You can get statistics for all anomaly detection jobs by using
- * `_all`, by specifying `*` as the `<job_id>`, or by omitting the `<job_id>`.
  * @rest_spec_name ml.get_job_stats
  * @since 5.5.0
  * @stability stable
@@ -35,8 +31,9 @@ export interface Request extends RequestBase {
   path_parts: {
     /**
      * Identifier for the anomaly detection job. It can be a job identifier, a
-     * group name, or a wildcard expression. If you do not specify one of these
-     * options, the API returns information for all anomaly detection jobs.
+     * group name, a comma-separated list of jobs, or a wildcard expression. If
+     * you do not specify one of these options, the API returns information for
+     * all anomaly detection jobs.
      */
     job_id?: Id
   }
--- a/specification/ml/get_jobs/MlGetJobsRequest.ts
+++ b/specification/ml/get_jobs/MlGetJobsRequest.ts
@@ -55,7 +55,9 @@ export interface Request extends RequestBase {
      * @server_default true
      */
     allow_no_match?: boolean
-    /** @server_default true */
+    /**
+     * @deprecated 7.10.0
+     */
     allow_no_jobs?: boolean
     /**
      * Indicates if certain fields should be removed from the configuration on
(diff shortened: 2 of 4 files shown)


specification/ml/get_job_stats/MlGetJobStatsRequest.ts
```diff
@@ -21,15 +21,39 @@ import { RequestBase } from '@_types/Base'
 import { Id } from '@_types/common'
 
 /**
+ * Retrieves usage information for anomaly detection jobs.
+ * You can get statistics for multiple anomaly detection jobs in a single API
```
lcawl, line 25: > You can get statistics for multiple anomaly detection jobs in a single API...

IMO this detail doesn't need too much prominence. In https://github.com/elastic/elasticsearch-specification/pull/852 I therefore moved that info into the appropriate parameter description:
https://github.com/elastic/elasticsearch-specification/blob/3149ef4eade38ebcd953e3c74b5bdbf6854be711/specification/ml/get_calendars/MlGetCalendarsRequest.ts#L34

If you agree, we should do the same for all these type of APIs.

specification/ml/get_jobs/MlGetJobsRequest.ts
```diff
@@ -21,20 +21,48 @@ import { RequestBase } from '@_types/Base'
...
+     *
+     * The default value is `true`, which returns an empty `jobs` array when
+     * there are no matches and the subset of results when there are partial
+     * matches. If this parameter is `false`, the request returns a `404` status
+     * code when there are no matches or only partial matches.
+     * @server_default true
+     */
     allow_no_match?: boolean
     /** @server_default true */
     allow_no_jobs?: boolean
```
lcawl, line 59: >    allow_no_jobs?: boolean

I think this option is deprecated per https://github.com/elastic/elasticsearch/pull/60601, so this might be a good place to add the `@deprecated` too:
https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#deprecated

specification/ml/get_overall_buckets/MlGetOverallBucketsRequest.ts
```diff
@@ -23,31 +23,87 @@ import { double, integer } from '@_types/Numeric'
...
     end?: Time
-    /** Returns overall buckets with timestamps after this time. */
+    /**
+     * Returns overall buckets with timestamps after this time.
+     */
     start?: Time
-    /**  If true, the output excludes interim results. By default, interim results are included. */
+    /**
+     * If true, the output excludes interim results. By default, interim results
+     * are included.
```
lcawl, line 92: > By default, interim results are included...

I think instead of writing it out, we should just add `@server_default false`

specification/ml/get_job_stats/MlGetJobStatsRequest.ts
szabosteve: Addressed in 8d49755.

specification/ml/get_jobs/MlGetJobsRequest.ts
szabosteve: Addressed in 8d49755.

specification/ml/get_overall_buckets/MlGetOverallBucketsRequest.ts
```diff
@@ -23,31 +23,87 @@ import { double, integer } from '@_types/Numeric'
...
+     * expression.
+     *
+     * You can summarize the bucket results for all anomaly detection jobs by
+     * using `_all` or by specifying `*` as the `<job_id>`.
+     *
+     * By default, an overall bucket has a span equal to the largest bucket span
+     * of the specified anomaly detection jobs. To override that behavior, use
+     * the optional `bucket_span` parameter.
+     *
+     * The `overall_score` is calculated by combining the scores of all the
```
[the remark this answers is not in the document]
szabosteve: 8d49755 relocates this piece of text to the API description.

Issue #917, opened 2021-10-19
Handling of `array_of` type in `server_default`


lcawl:
## 🚀 Feature Proposal

Relates to https://github.com/elastic/elasticsearch-specification/pull/904

`make contrib` fails with the following error:

> Error: Default values can only be configured for instance_of or union_of types, you are using array_of
> elasticsearch-specification/specification/ml/_types/Rule.ts:26:3
> 
> 24  
> 25  export class DetectionRule {
> 26    /**
> 27     * The set of actions to be triggered when the rule applies. If more than one action is specified the effects of all actions are combined.
> 28     * @server_default skip_result
> 29     */
> 30    actions?: RuleAction[]
> 31    /**

## Motivation

The `actions` array in the [create anomaly detection jobs API](https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-put-job.html) lets you specify `skip_result`, `skip_model_update`, or both values. It defaults to just `skip_result`. This doesn't seem possible in the spec at the moment (per error above).

PR #921, opened 2021-10-20
Add support for arrays server default values

delvedor:
As titled.

Closes: https://github.com/elastic/elasticsearch-specification/issues/917


swallez (approved):
LGTM, but I'm wondering if we should just make it `any`

PR #923, opened 2021-10-20
Update ErrorResponseBase

delvedor:
As titled, see the code comments to know more details.

### DIFF
--- a/specification/_types/Base.ts
+++ b/specification/_types/Base.ts
@@ -71,7 +71,7 @@ export class ElasticsearchVersionInfo {
  * The response returned by Elasticsearch when request execution did not succeed.
  */
 export class ErrorResponseBase {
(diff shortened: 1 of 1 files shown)


swallez (approved):
LGTM! Hopefully this iteration is the last one 😅

specification/_types/Base.ts
```diff
@@ -71,17 +71,13 @@ export class ElasticsearchVersionInfo {
  * The response returned by Elasticsearch when request execution did not succeed.
  */
 export class ErrorResponseBase {
-  error: TopLevelError
+  // In some edge cases `error` can be a string, for example if you call `GET _cat/foo`.
```
swallez, line 74:
```suggestion
  // In some edge cases `error` can be a string that is a shortcut to `error.reason`, for example if you call `GET _cat/foo`.
```

Issue #845, opened 2021-10-05
Add new url parameters to Get snapshot endpoint and new properties to the response body


yuliacech:
The endpoint [Get snapshot](https://www.elastic.co/guide/en/elasticsearch/reference/7.14/get-snapshot-api.html) received a lot of updates starting from 7.14 to improve scalability of this endpoint when working with a lot of snapshots. 
Here is a suggestion of how the new url parameters could be added to the [SnapshotGetRequest.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/snapshot/get/SnapshotGetRequest.ts) file.

```
{
    /**
     * Allows setting a sort order for the result. Defaults to start_time, i.e. sorting by snapshot start time stamp.
     * @since 7.14.0
     * @server_default 'start_time'
     */
    sort?: 'start_time' | 'duration' | 'name' | 'index_count' 
    // these properties since 7.16.0
    | 'repository' | 'shard_count' | 'failed_shard_count';

    /**
     * Maximum number of snapshots to return. Defaults to 0 which means return all that match the request without limit.
     * @since 7.14.0
     * @server_default 0
     */
    size?: integer;

    /**
     * Sort order. Valid values are asc for ascending and desc for descending order. Defaults to asc, meaning ascending order.
     * @since 7.14.0
     * @server_default 'asc'
     */
    order?: 'asc' | 'desc';

    /**
     * Offset identifier to start pagination from as returned by the next field in the response body.
     * @since 7.14.0
     */
    after?: string;

    /**
     * Numeric offset to start pagination from based on the snapshots matching this request. Using a non-zero value for this parameter is mutually exclusive with using the after parameter. Defaults to 0.
     * @since 7.15.0
     * @server_default 0
     */
    offset?: integer;

    /**
     * Value of the current sort column at which to start retrieval. Can either be a string snapshot- or repository name when sorting by snapshot or repository name, a millisecond time value or a number when sorting by index- or shard count.
     * @since 7.16.0
     */
    from_sort_value?: string;

    /**
     * Filter snapshots by a comma-separated list of SLM policy names that snapshots belong to. Also accepts wildcards (*) and combinations of wildcards followed by exclude patterns starting with -. To include snapshots not created by an SLM policy you can use the special pattern _none that will match all snapshots without an SLM policy.
     * @since 7.16.0
     */
    slm_policy_filter?: Name;

}

```

There are also new properties added to the response body that is returned from the endpoint, for example 
```

    /**
[...]

PR #926, opened 2021-10-21
Update snapshot.get endpoint

delvedor:
As titled.

@yuliacech it looks like the response type is already correct, for instance, the `failures` array is present in the `SnapshotInfo` type.

Closes: https://github.com/elastic/elasticsearch-specification/issues/845


specification/snapshot/get/SnapshotGetRequest.ts
```diff
@@ -63,5 +66,44 @@ export interface Request extends RequestBase {
...
+     * Numeric offset to start pagination from based on the snapshots matching this request. Using a non-zero value for this parameter is mutually exclusive with using the after parameter. Defaults to 0.
+     * @since 7.15.0
+     * @server_default 0
+     */
+    offset?: integer
+    /**
+     * Value of the current sort column at which to start retrieval. Can either be a string snapshot- or repository name when sorting by snapshot or repository name, a millisecond time value or a number when sorting by index- or shard count.
+     * @since 7.16.0
+     */
+    from_sort_value?: string
```
sethmlarson, line 102: By the description should this be `string | long`? Might need to be an alias?

sethmlarson (changes requested):
Two comments:

Do we need to add `next` to the SnapshotGetResponse?

PR #930, opened 2021-10-21
Use codegen_name instead of identifier

delvedor:
To make it more clear what this tag is about.


compiler/model/metamodel.ts
```diff
@@ -133,7 +133,7 @@ export class Property {
    * If specified takes precedence over `name` when generating code. `name` is always the value
    * to be sent over the wire
    */
-  identifier?: string
+  codegen_name?: string
```
swallez, line 136: This should be `codegenName` (we use camelCase in the metamodel), here and in 2 other places in this file.

compiler/model/utils.ts
```diff
@@ -292,7 +292,7 @@ export function modelType (node: Node): model.ValueOf {
         default: {
           const generics = node.getTypeArguments().map(node => modelType(node))
           const identifier = node.getTypeName()
-          assert(node, Node.isIdentifier(identifier), 'Not an identifier')
+          assert(node, Node.isIdentifier(identifier), 'Not an codegen_name')
```
swallez, line 295: "Not a valid identifier for codegen_name" would me more explicit. Compiler messages are sometimes so succinct that they're hard to understand 😉

swallez (changes requested):
Looks good overall, except casing in `metamodel.ts`

swallez (approved):
LGTM!

PR #932, opened 2021-10-21
Add types to 'knn_search' API

sethmlarson:
Adds the [`knn_search` API](https://github.com/elastic/elasticsearch/pull/79013)


specification/_global/knn_search/KnnSearchRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @since 7.16.0
+ * @stability experimental
+ */
+export interface Request extends RequestBase {
+  path_parts: {
+    /**
+     * A comma-separated list of index names to search;
+     * use `_all` or empty string to perform the operation on all indices
+     */
+    index: Indices
```
delvedor, line 35: I guess this should be optional?
```suggestion
    index?: Indices
```

jtibshirani: This is a mistake in the REST spec description -- I'll make a note to correct it. Only `/<index-pattern>/_knn_search` is supported.

output/schema/schema.json
```diff
@@ -17080,6 +17087,256 @@
         "namespace": "_global.info"
       }
     },
+    {
+      "attachedBehaviors": [
+        "CommonQueryParameters"
```
jtibshirani, line 17092: For my knowledge, what does it mean to attach "CommonQueryParameters"?

output/schema/schema.json
```diff
@@ -17080,6 +17087,256 @@
...
+      ],
+      "body": {
+        "kind": "properties",
+        "properties": [
+          {
+            "description": "Array of wildcard (*) patterns. The request returns values for field names\nmatching these patterns in the hits.fields property of the response.",
```
jtibshirani, line 17098: Tiny comment, presenting `fields` as primarily an "array of wildcard patterns" is a bit surprising, since by far the most common use case is to provide field names with no wildcards. It's a fairly minor detail that it also accepts field wildcard patterns.

output/schema/schema.json
```diff
@@ -17080,6 +17087,256 @@
...
+            "type": {
+              "kind": "instance_of",
+              "type": {
+                "name": "Fields",
+                "namespace": "_types"
+              }
+            }
+          },
+          {
+            "description": "KNN query to execute",
```
jtibshirani, line 17110: Small comment: we're trying to consistently style it as kNN in docs to match the community's terminology.

output/typescript/types.ts
```diff
@@ -479,6 +479,33 @@ export interface InfoResponse {
...
+export interface KnnSearchResponse<TDocument = unknown> {
+  took: long
+  timed_out: boolean
+  _shards: ShardStatistics
+  hits: SearchHitsMetadata<TDocument>
+  fields?: Record<string, any>
+  max_score?: double
+}
+
+export interface KnnSearchKnnQuery {
```
jtibshirani, line 500: This naming is a little surprising since Knn appears twice (but not sure it will actually be presented to users?)

specification/_global/knn_search/_types/Knn.ts
```diff
@@ -0,0 +1,30 @@
...
+ * under the License.
+ */
+
+import { Field } from '@_types/common'
+import { long, double } from '@_types/Numeric'
+
+export type KnnQueryVector = double[]
+
+export interface KnnQuery {
+  field: Field
```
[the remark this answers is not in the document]
jtibshirani: Certainly, here are some descriptions:
* field: the name of the vector field to search against
* query_vector: the query vector
* k: the final number of nearest neighbors to return as top hits
* num_candidates: the number of nearest neighbor candidates to consider per shard

PR #970, opened 2021-11-02
Add property alias support in requests and responses & improve error message

swallez:
Adds support for `@aliases` to request and response properties (it was only implemented on classes/interfaces)

Also improves error message for the expected `@variant` syntax

PR #971, opened 2021-11-02
Dedup aggregrations/aggs fields

swallez:
Depends on PR #970

Uses `@aliases` to deduplicate `aggs` and `aggregations` properties in search requests and a few other places. `aggregations` is chosen as the canonical name and `aggs` as an alias as ES uses `aggregations` when serializing objects (and it's also more meaningful).

`SearchMvtRequest` is the only exception, since it only accepts `aggs` with no alias.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -18013,29 +18013,10 @@
       },
       "properties": [
         {
+          "aliases": [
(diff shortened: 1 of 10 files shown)


delvedor (approved):
LGTM

PR #972, opened 2021-11-02
Major rewrite of aggregation results

swallez:
Depends on #970 and #971

Major rewrite of aggregation results, using the ES source code as a reference. Also adds `@variant` information to the `Aggregate` union and its members.

### DIFF
--- a/specification/_types/Geo.ts
+++ b/specification/_types/Geo.ts
@@ -56,6 +56,14 @@ export enum GeoDistanceType {
 /** A GeoJson shape, that can also use Elasticsearch's `envelope` extension. */
 export type GeoShape = UserDefinedValue
 
+/** A GeoJson GeoLine. */
+export class GeoLine {
+  /** Always `"LineString"` */
+  type: string
+  /** Array of `[lon, lat]` coordinates */
+  coordinates: Array<Array<double>>
+}
+
 export enum GeoShapeRelation {
   intersects = 0,
   disjoint = 1,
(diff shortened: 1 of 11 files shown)


swallez:
Ready for review.

Running `make setup-env` rewrite the `package-lock.json` files, I'm not sure why. Versions used:
- `node --version`: `v16.13.0`
- `npm --version`: `8.1.0`

specification/_types/Geo.ts
```diff
@@ -56,6 +56,14 @@ export enum GeoDistanceType {
 /** A GeoJson shape, that can also use Elasticsearch's `envelope` extension. */
 export type GeoShape = UserDefinedValue
 
+/** A GeoJson GeoLine. */
+export class GeoLine {
+  /** Always `"LineString"` */
+  type: string
```
delvedor, line 62: If it's always `LineString`, why not define it as is? Or use a single element enum?

delvedor (approved):
Great work!

swallez:
Validation results:
- `search` response: 0 errors
- `search` response: from 16 to 14 errors. Errors solved:
    `2472:12  Type alias AggregationsAggregate circularly references itself.`
    `3228:12  Type alias AggregationsSingleBucketAggregate circularly references itself.`

specification/_types/Geo.ts
swallez: In a previous iteration I defined it as a literal `type: 'LineString'`, but this has some implications on code generators that would now have to implement code generation for arbitrary constant values.

Our other uses of literal values is for union tags (`@variants internal tag=xxx`), which are handled in a very specific way.

Since this type is used as part of a response in the `geo_line` aggregation, defining it as `string` should be fine since users normally never have to write it in their applications.

PR #977, opened 2021-11-03
Fix create index request

stevejgordon:
This PR updates the create index request type so that it uses just a mapping instance rather than a union, and uses the `IndexSettings` type for settings. This required many tokenizers, char filters, token filters and analyzers to be fixed up due to failing tests.

The remaining failures are due to the flattened, vs object style settings and we will deal with those separately as discussed with @delvedor.

This will require a manual cherry-pick backport once merged since some changes fix additions for the new time series properties which won't be needed for 7.x.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -44808,6 +44808,20 @@
               "name": "KuromojiAnalyzer",
               "namespace": "_types.analysis"
             }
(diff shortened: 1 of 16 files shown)


delvedor (approved):
LGTM

PR #980, opened 2021-11-03
Fix duplication of alias name in IndexSettings

stevejgordon:
Fixes the name in the type.


philkra (approved):
LGTM - does this change need to be backported to 7.16?

stevejgordon:
@philkra No, this need a manual backport as part of another branch I have in flight

PR #983, opened 2021-11-03
Adds descriptions and missing parameters for start and stop datafeed APIs

lcawl:
Relates to https://github.com/elastic/stack-docs/issues/1845

Uses descriptions from https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-start-datafeed.html and https://www.elastic.co/guide/en/elasticsearch/reference/master/ml-stop-datafeed.html

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -7839,6 +7839,11 @@
       "description": "Starts one or more datafeeds.",
       "docUrl": "https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-start-datafeed.html",
       "name": "ml.start_datafeed",
+      "privileges": {
+        "cluster": [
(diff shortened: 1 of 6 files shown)


delvedor (approved):
LGTM

Should this be backported to 7.16?

lcawl:
@delvedor These two APIs contain sets of properties that are repeated identically in their request body and query parameters. To avoid having to repeat the descriptions, I've just written "Refer to the description for the xxx query parameter".  I'd be interested to know if you think there's a better way to annotate these duplicated properties.

delvedor:
@lcawl we don't have anything in place for "duplicate" descriptions.If "Refer to the description for the xxx query parameter" works for you I'm completely fine with it :)

PR #989, opened 2021-11-04
Fix ml.update_model_snapshot response

philkra:
as titled.


delvedor (approved):
LGTM

PR #991, opened 2021-11-04
Document CommonQueryParameters

sethmlarson:
Documented all the parameters defined in `CommonQueryParameters` and `CommonCatQueryParameters`.

I also found `source_query_string` which isn't an Elasticsearch parameter but is instead likely inherited from NEST and would've been translated into `?source=...` as a means of smuggling small request bodies for GET into the query string. I've removed this parameter in this PR.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -153634,8 +153634,10 @@
       },
       "properties": [
         {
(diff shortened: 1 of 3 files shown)


delvedor (approved):
LGTM

sethmlarson:
Got confirmation from @Mpdreamz about `source_query_string`, merging now!

PR #993, opened 2021-11-04
Add task to clean npm dependencies

philkra:
make task for https://github.com/elastic/elasticsearch-specification/pull/990

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,12 @@ setup-env:	## Install dependencies for contrib target
 	@npm install --prefix compiler
 	@npm install --prefix typescript-generator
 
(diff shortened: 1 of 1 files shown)


sethmlarson:
Alternate suggestion: should setup-env automatically do this since they're no-ops on the initial setup?

delvedor (changes requested):
There is no need to clean the lockfile now, it's committed to the repo so we'll always have the same dependencies installed :)

philkra:
@sethmlarson good call, I added the task to `setup-env`.

@delvedor , removed the deletion of lockfiles

delvedor (approved):
LGTM

PR #1003, opened 2021-11-05
Add codegen name for seq_no in GetResponse

stevejgordon:
As titled.


stevejgordon:
Closing as agreed during the type spec meeting.

PR #1008, opened 2021-11-05
Mark 'keep_alive' for 'open_point_in_time' API as required

sethmlarson:
The `keep_alive` parameter isn't documented as required/options in the docs, however trying to use the API without it [always produces an error](https://github.com/elastic/elasticsearch/blob/26c86871fc091900952e88e252c36fbfedf8d5fa/server/src/main/java/org/elasticsearch/action/search/OpenPointInTimeRequest.java#L104). I'm not sure if there are other query parameters that are required, if not this may be the first of it's kind!

Corresponding PR to the other API spec: https://github.com/elastic/elasticsearch/pull/80449

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -19396,7 +19396,7 @@
         {
           "description": "Specific the time to live for the point in time",
           "name": "keep_alive",
-          "required": false,
+          "required": true,
           "type": {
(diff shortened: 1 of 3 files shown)


philkra (approved):
LGTM

PR #1022, opened 2021-11-11
Add 'fleet.global_checkpoints' API

sethmlarson:
Adds the `fleet.global_checkpoints` API which unfortunately doesn't have any recorded tests?

Request was built from the documentation, an example response for this API looks like:

```json
{"timed_out":false,"global_checkpoints":[2]}
```


delvedor (approved):
LGTM

PR #1027, opened 2021-11-17
Name unions

swallez:
Most strongly typed languages need each member of a union type to have an identifier. In Java we generate custom wrappers for unions, Rust needs names for enum variants (Rust unifies enums and tagged unions).

This PR updates the spec to change inline unions used in property declarations to a distinct type with a new `@codegen_names` (plural) annotation that names the union members.

```ts
class Foo {
  field: Bar | Baz
}
```
becomes
```ts
class Foo {
  field: SomeBarBaz
}

/** @codegen_names bar, baz */
type SomeBarBaz = Bar | Baz
```

Some inline unions are left in the spec, that represent the leniency rules that ES implements. These are notably `SomeType | SomeType[]` (read a single object as a 1-sized array) and strings that accept any primitive type.

These leniency rules are currently implemented in the Java generator and will later be formalized as new model validation rules that will report inline unions that require a separate type definition.

------

Along with refactoring enumerations to enum types, this PR also fixes a number of definitions that were ambiguous or actually wrong. API validation has been run on most of these changes to verify that there were no regressions.

Validation of the `search` API has 3 errors less, and the remaing ones are because of the use of numbers for ids (they're not quoted in the YAML tests) and experimental `bucket_correlation` and `bucket_count_ks_test` aggregations that have not been represented in the spec.

### DIFF
--- a/specification/_global/search/_types/sort.ts
+++ b/specification/_global/search/_types/sort.ts
@@ -20,7 +20,7 @@
 import { AdditionalProperty } from '@spec_utils/behaviors'
 import { Missing } from '@_types/aggregations/AggregationContainer'
 import { Field } from '@_types/common'
-import {DistanceUnit, GeoDistanceType, GeoLocation} from '@_types/Geo'
+import { DistanceUnit, GeoDistanceType, GeoLocation } from '@_types/Geo'
 import { FieldType } from '@_types/mapping/Property'
 import { double, integer, long } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
@@ -59,6 +59,7 @@ export class GeoDistanceSort
 {
   mode?: SortMode
   distance_type?: GeoDistanceType
+  ignore_unmapped?: boolean
   order?: SortOrder
   unit?: DistanceUnit
 }
@@ -80,8 +81,7 @@ export enum ScriptSortType {
  * @doc_url https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html
  * @variants container
  */
-export class SortOptions implements AdditionalProperty<Field, FieldSort>
-{
+export class SortOptions implements AdditionalProperty<Field, FieldSort> {
   _score?: ScoreSort
   _doc?: ScoreSort
   _geo_distance?: GeoDistanceSort
@@ -110,4 +110,3 @@ export enum SortOrder {
   asc = 0,
   desc = 1
 }
-
--- a/specification/_global/search/_types/sort.ts
+++ b/specification/_global/search/_types/sort.ts
@@ -17,42 +17,41 @@
  * under the License.
  */
 
-import { AdditionalProperties, AdditionalProperty } from '@spec_utils/behaviors'
+import { AdditionalProperty } from '@spec_utils/behaviors'
 import { Missing } from '@_types/aggregations/AggregationContainer'
 import { Field } from '@_types/common'
-import { DistanceUnit, GeoDistanceType } from '@_types/Geo'
+import {DistanceUnit, GeoDistanceType, GeoLocation} from '@_types/Geo'
 import { FieldType } from '@_types/mapping/Property'
 import { double, integer, long } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
-import { GeoLocation } from '@_types/query_dsl/geo'
 import { Script } from '@_types/Scripting'
 
 export class NestedSortValue {
   filter?: QueryContainer
   max_children?: integer
+  nested?: NestedSortValue
   path: Field
-  // nested: NestedSortValue
 }
 
-// export type NestedSort = Dictionary<Field, NestedSortKey>
-
-export enum NumericType {
+export enum FieldSortNumericType {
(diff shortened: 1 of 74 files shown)


specification/_types/Geo.ts
```diff
@@ -82,3 +82,60 @@ export class LatLon {
...
 }
+
+/**
+ * A latitude/longitude as a 2 dimensional point. It can be represented in various ways:
+ * - as a `{lat, long}` object
+ * - as a geo hash value
+ * - as a `[lon, lat]` array
+ * - as a string in `"<lat>, <lon>"` or WKT point formats
+ *
+ * @codegen_names latlon, geohash, coords, text
```
philkra, line 93: minor nit: can we use `coordinates` instead of `coords`?

specification/_types/Geo.ts
```diff
@@ -82,3 +82,60 @@ export class LatLon {
...
+}
+
+/**
+ * A geo bounding box. It can be represented in various ways:
+ * - as 4 top/bottom/left/right coordinates
+ * - as 2 top_left / bottom_right points
+ * - as 2 top_right / bottom_left points
+ * - as a WKT bounding box
+ *
+ * @codegen_names coords, tlbr, trbl, wkt
```
philkra, line 114: same nit: can we use coordinates instead of coords?

specification/_global/search/_types/suggester.ts
```diff
@@ -57,6 +56,7 @@ export class SuggesterBase {
   size?: integer
 }
 
+/** @codegen_names completion, phrase, term */
```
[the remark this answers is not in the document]
swallez: There are three types if you click on the little "Expand down" icon to see more text 😆

specification/_types/Geo.ts
swallez: As discussed, `coords` is heavily used in the geo domain, so it keeps things short without hurting understandability.

specification/_global/search/_types/sort.ts
```diff
@@ -104,12 +108,5 @@ export enum SortMode {
...
 export enum SortOrder {
   asc = 0,
-  desc = 1,
-  /** @codegen_name Document */
-  _doc = 2
-}
-
-export enum SortSpecialField {
-  _score = 0,
-  _doc = 1
```
swallez, line 114: This enum is not needed. If users want to use the default order for `_score` and `_doc` then they should use the `Field` member of `SortCombinations`. And otherwise use the `_score` and `_doc` variants of `SortOptions` whose properties are different from that of a `FieldSort`.

delvedor (changes requested):
LGTM

I would add a compiler check that verifies that if `codegen_names` is present, then the number of identifiers should be the same as the union members.

swallez:
@delvedor check added in 48d3699 (it's already checked in the Java generator, but should indeed definitely be checked at spec compile time).

Also added a unification of redundant `DocValueField` and `FieldAndFormat` in f88c637

specification/_types/common.ts
```diff
@@ -238,9 +228,13 @@ export enum OpType {
   create = 1
 }
 
-export type Refresh = boolean | RefreshOptions
-export enum RefreshOptions {
-  wait_for = 1
+// ES uses a string but accepts a json boolean for strings
+export type Refresh = boolean | RefreshValues
```
philkra, line 232: Isn't the type definition of `Refresh` redundant, because boolean values are implied by `RefreshValues`? Which means we could just define `Refresh = RefreshValues` or directly as enum.

specification/_types/mapping/dynamic-template.ts
```diff
@@ -34,9 +34,12 @@ export enum MatchType {
   regex = 1
 }
 
-export enum DynamicMapping {
-  strict = 0,
-  runtime = 1,
-  true = 2,
-  false = 3
+// ES uses a string but accepts a json boolean for strings
+export type DynamicMapping = boolean | DynamicMappingValues
```
philkra, line 38: same question as above on `export type Refresh = boolean | RefreshValues`

philkra (approved):
LGTM, left two questions, not sure if they're applicable, though!

specification/_types/common.ts
swallez: Very good point. The "problem" is that in dynamically typed languages people may expect to just set this to `true` (the boolean value) instead of `"true"` (the string). Elasticsearch uses the string version, but its lenient JSON parsing transparently converts a JSON boolean to a string.

In Java this union makes things unnecessarily complex, so in the model preparation phase we rename the `RefreshValues` type to `Refresh` and only expose the enum, and same for `DynamicMapping`.

Note though that we could also handle this the other way around: if an enum contains `true` and `false`, then dynamically typed languages could generate the union by themselves. I'll do this in the TS generator and simplify the spec.

swallez: Done in a62542c! And it also caught the same pattern for `IndicesIndexCheckOnStartup`. Code generation FTW 😄

philkra: Awesome! Thanks for the swift turnaround and explanation.

delvedor (approved):
LGTM

PR #1032, opened 2021-11-17
Add 'model_type' to 'TrainedModelConfig'

sethmlarson:
Adds `model_type` enum to `TrainedModelConfig`.

### DIFF
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -57,6 +57,8 @@ export class TrainedModelInferenceStats {
 export class TrainedModelConfig {
   /** Idetifier for the trained model. */
   model_id: Id
+  /** The model type */
+  model_type: TrainedModelType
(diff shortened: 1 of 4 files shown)


specification/ml/_types/TrainedModel.ts
```diff
@@ -139,3 +141,19 @@ export class TotalFeatureImportanceStatistics {
   /** The minimum importance value across all the training data for this feature. */
   min: integer
 }
+
+export enum TrainedModelType {
+  /**
+   * The model definition is an ensemble model of decision trees.
+   */
+  tree_ensemble = 0,
```
delvedor, line 149: There is no need to add the `= 0` (and so on), I'll do a pr soon to cleanup this usage pattern that we inherited from .NET.

sethmlarson: Done!

delvedor (approved):

PR #1034, opened 2021-11-18
Refactor and fix multi* APIs

swallez:
Refactor and fix the `Multi{Search|Get|TemplateSearch}` APIs:
* fix some definitions that were not correct
* introduce unions to distiguish the header and body parts of ndjson requests
* reuse types between APIs when possible.

### DIFF
--- a/specification/_global/get/GetResponse.ts
+++ b/specification/_global/get/GetResponse.ts
@@ -19,6 +19,6 @@
 
 import { GetResult } from '@global/get/types'
 
-export class Response<TDocument> extends GetResult<TDocument> {
-  body: {}
+export class Response<TDocument> {
+  body: GetResult<TDocument>
 }
--- a/specification/_global/get/GetResponse.ts
+++ b/specification/_global/get/GetResponse.ts
@@ -17,29 +17,8 @@
  * under the License.
  */
 
-import { Dictionary } from '@spec_utils/Dictionary'
-import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
-import {
-  Id,
-  IndexName,
-  SequenceNumber,
-  Type,
-  VersionNumber
-} from '@_types/common'
-import { long } from '@_types/Numeric'
+import { GetResult } from '@global/get/types'
 
-export class Response<TDocument> {
-  body: {
-    _index: IndexName
-    fields?: Dictionary<string, UserDefinedValue>
-    found: boolean
-    _id: Id
-    _primary_term?: long
-    _routing?: string
-    _seq_no?: SequenceNumber
-    _source?: TDocument
-    /** @deprecated 7.0.0 */
-    _type?: Type
-    _version?: VersionNumber
-  }
+export class Response<TDocument> extends GetResult<TDocument> {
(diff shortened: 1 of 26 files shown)


specification/_global/get/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  _index: IndexName
+  fields?: Dictionary<string, UserDefinedValue>
+  found: boolean
+  _id: Id
+  _primary_term?: long
+  _routing?: string
+  _seq_no?: SequenceNumber
+  _source?: TDocument
+  /** @deprecated 7.0.0 */
+  _type?: Type
```
philkra, line 41: Can you keep track, to ensure we remove `_type` for 8.0 ?

philkra (approved):
LGTM - left one note

specification/_global/get/types.ts
swallez: I opened https://github.com/elastic/elasticsearch-specification/issues/1035 to clean things up globally

philkra: Perfect, thank you!

specification/_global/get/GetResponse.ts
```diff
@@ -17,29 +17,8 @@
...
-    _id: Id
-    _primary_term?: long
-    _routing?: string
-    _seq_no?: SequenceNumber
-    _source?: TDocument
-    /** @deprecated 7.0.0 */
-    _type?: Type
-    _version?: VersionNumber
-  }
+export class Response<TDocument> extends GetResult<TDocument> {
```
delvedor, line 22: There is no compiler check yet, but we should not extend Response classes, as if you define a property in the extending class it will override the one defined in the extended class.

specification/_global/msearch/MultiSearchResponse.ts
```diff
@@ -17,13 +17,6 @@
...
-import { SearchResult } from './types'
+import { MultisearchResponseBase } from '@global/msearch/types'
 
-export class Response<TDocument> {
-  body: {
-    took: long
-    responses: Array<SearchResult<TDocument> | ErrorResponseBase>
-  }
-}
+export class Response<TDocument> extends MultisearchResponseBase<TDocument> {}
```
delvedor, line 22: Same here.

delvedor (changes requested):
The changes look good, just a nit.

specification/_global/get/GetResponse.ts
swallez: I don't understand _"if you define a property in the extending class it will override the one defined in the extended class"_. Can you show a simple example?

From a typed language perspective, we really need to be able to use a single type to represent the result of a `get` and the hits of an `mget`. So the idea here is to use `GetResult` as the base class of the `get` response and use it directly in `mget` hits.

swallez: I've replaced the `extends` with a value body in 011e936 as this is really what we want here: be able to use the same structure that is used for `get` responses as `mget` response items. WDYT?

delvedor (approved):
LGTM

PR #1036, opened 2021-11-18
Generate Id as string|number in TypeScript

swallez:
This PR updates the TypeScript generator to produce `type Id = string | number`. Although document ids are strings, a lot of yaml tests use numbers as test document identifiers, which yaml returns as... well... numbers, and cause a number of API validations to fail.

With this change numbers are accepted for document id, which allows many more validation tests to pass.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -2063,7 +2063,7 @@ export type Host = string
 
 export type HttpHeaders = Record<string, string | string[]>
 
-export type Id = string
+export type Id = string | number
 
 export type Ids = Id | Id[]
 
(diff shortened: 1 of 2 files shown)


delvedor (commented):
I don't really like this, but I see that we don't have my other options at the moment.

swallez:
Another option would be to fix all the offending tests, but new ones with id-as-integers can pop at any moment because of YAML where people are used to write unquoted strings.

So I consider it as an acceptable compromise for API validation, so that we don't have to remember "yeah, 5 of these 11 errors on `search-request` are because of id-as-number" and finally aim towards zero everywhere.

Note that his is for API validation and should be considered as form of processing-time [data cleansing](https://en.wikipedia.org/wiki/Data_cleansing). This doesn't mean it has to exist also in the TypeScript client, where we can keep `Id` as a pure string.

philkra:
I don't believe we can fix all nest & upstream tests. It would make more sense to make the compiler accept `string | integer`

delvedor (approved):
LGTM

PR #1046, opened 2021-11-22
Unify all health status enums as a single type

swallez:
Unifies a number of green/yellow/red enums as a single `HealthStatus` type.

Also adds aliases with upper-case versions of the enum, since some APIs send it in that format (which is the internal default format), as reported in https://github.com/elastic/elasticsearch-java/issues/43

Also renames:
- `common.Health` to `HealhStatus`, the terminology used everywhere these values are mentioned
- `indices.shard_stores.ShardStatus` to `ShardStoresStatus`, as it's very specific to the shard_stores API


sethmlarson (approved):

philkra (approved):
LGTM

delvedor (approved):
LGTM

PR #1052, opened 2021-11-23
Unify ScoreMode, ChildScoreMode & move sort to common types

swallez:
Removing some duplicate enums, and move sort-related types to commons where they're already used.


delvedor (changes requested):
Overall LGTM, but I think the sort definition should go in the global `_types` folder instead of `_types/query_dsl`, as it's used outside of the context of a query.

delvedor (approved):
LGTM

PR #1060, opened 2021-11-23
Adds ML infer trained model deployment API

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/73660

This PR adds a specification for the ML [infer trained model deployment API](https://www.elastic.co/guide/en/elasticsearch/reference/master/infer-trained-model-deployment.html) per https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ml.infer_trained_model_deployment.json

NOTE: I have omitted the security details for now, since they're missing from the documentation but we'll want to fix that in both places.

### DIFF
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -247,3 +247,5 @@ export class TopClassEntry {
   class_probability: double
   class_score: double
 }
+
+export type PredictedValue = string | double
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -239,6 +239,6 @@ export class TrainedModelEntities {
   class_name: string
   class_probability: double
   entity: string
-  start_pos: number
-  end_pos: number
+  start_pos: integer
+  end_pos: integer
 }
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -235,3 +235,10 @@ export class TrainedModelAllocation {
   start_time: DateString
   task_parameters: TrainedModelAllocationTaskParameters
 }
+export class TrainedModelEntities {
+  class_name: string
+  class_probability: double
+  entity: string
+  start_pos: number
+  end_pos: number
+}
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -242,3 +242,8 @@ export class TrainedModelEntities {
   start_pos: integer
   end_pos: integer
 }
+export class TopClassEntry {
+  class_name: string
+  class_probability: double
+  class_score: double
+}
--- a/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
+++ b/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
@@ -20,7 +20,7 @@
 import { RequestBase } from '@_types/Base'
 import { Id } from '@_types/common'
 import { Time } from '@_types/Time'
-import { SingleKeyDictionary } from '@spec_utils/Dictionary'
+import { Dictionary } from '@spec_utils/Dictionary'
 
 /**
  * Evaluates a trained model.
@@ -47,6 +47,6 @@ export interface Request extends RequestBase {
      * An array of objects to pass to the model for inference. The objects should contain a field matching your
      * configured trained model input. Typically, the field name is `text_field`. Currently, only a single value is allowed.
      */
-    docs: SingleKeyDictionary<string, string>[]
+    docs: Dictionary<string, string>[]
   }
 }
--- a/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
+++ b/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
@@ -20,6 +20,7 @@
 import { RequestBase } from '@_types/Base'
 import { Id } from '@_types/common'
 import { Time } from '@_types/Time'
+import { SingleKeyDictionary } from '@spec_utils/Dictionary'
 
 /**
  * Evaluates a trained model.
@@ -46,6 +47,6 @@ export interface Request extends RequestBase {
      * An array of objects to pass to the model for inference. The objects should contain a field matching your
      * configured trained model input. Typically, the field name is `text_field`. Currently, only a single value is allowed.
      */
-    docs: string[]
+    docs: SingleKeyDictionary<string, string>[]
   }
 }
--- a/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
+++ b/specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
@@ -1,0 +1,51 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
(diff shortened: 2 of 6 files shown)


specification/ml/_types/TrainedModel.ts
```diff
@@ -157,3 +157,11 @@ export enum TrainedModelType {
    */
   pytorch
 }
+
+export class TrainedModelEntities {
+  class_name: string
+  class_probability: double
+  entity: string
+  start_pos: number
```
sethmlarson, line 165: Use `integer`/`long`/`float`/`double` instead of `number` so strongly typed languages don't have to guess. Might be something to add to the best practices document.

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
```diff
@@ -0,0 +1,51 @@
...
+     * @server_default 10s
+     */
+    timeout?: Time
+  }
+  body: {
+    /**
+     * An array of objects to pass to the model for inference. The objects should contain a field matching your
+     * configured trained model input. Typically, the field name is `text_field`. Currently, only a single value is allowed.
+     */
+    docs: string[]
```
sethmlarson, line 49: Shouldn't this be an array of objects instead of `string`? Documentation has:

```json
{
	"docs": [{"text_field": "The movie was awesome!!"}]
}
```

sethmlarson (changes requested):
Thanks for this! Here's a couple comments:

specification/ml/_types/TrainedModel.ts
lcawl: Thanks! I've changed it to integers in aec74e54f04c763e7e37c72c109098c54fd5fe7b

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
lcawl: Ah, you're right. Do you have advice for how to code this, given that the field name in each object is variable?

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentResponse.ts
```diff
@@ -0,0 +1,39 @@
...
+  body: {
+    /**
+     * If the model is trained for named entity recognition (NER) tasks, the response contains the recognized entities.
+     */
+    entities?: TrainedModelEntities[]
+    /**
+     * If the model is trained for a text classification task, the response is a predicted label.
+     * For named entity recognition (NER) tasks, the response contains the annotated text output.
+     */
+    predicted_value?: string
```
davidkyle, line 33: This is a tricky one. All task types will have `prediction_probability` and `predicted_value` fields (if the request was successful) but the actual contents of `predicted_value` depend on the task type

### NER
The comment is correct, you get the annotated text in `predicted_value` and a list of `entities`

### Fill Mask
**predicted_value: string**: The top prediction for replacing the mask token. e.g. "Paris is the capital of [MASK]" will return "France" here

**predicted_value_sequence: string**
The input text sequence with the `[MASK]` token replaced by the predicted_value

**top_classes**: List of TopClassEntry where TopClassEntry is 
```
export class TopClassEntry {
  class_name: string
  class_probability: double
  class_score: double
}
```

### Text Classification
**predicted_value: string**: The predicted class

**top_classes**: List of TopClassEntry 

### Text Embedding
**predicted_value:  array of double** :  The raw text embedding values

### Zero Shot Classification
**predicted_value: string**: The predicted class

**top_classes**: List of TopClassEntry 


### Warning
**warning : string** : If the request failed the reason for the failure

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentResponse.ts
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * If the model is trained for a text classification task, the response is a confidence score.
+     */
+    prediction_probability?: double
+      /**
+       * True if the input text was truncated to meet the model's maximum sequence length limit. is_truncated is present only when it is true.
```
lcawl, line 39:
```suggestion
       * Indicates whether the input text was truncated to meet the model's maximum sequence length limit. This property is present only when it is true.
```

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentResponse.ts
lcawl: Thanks!  I've tried to incorporate all that information in f40c639874ef56f8d061cd6888b887d1b9fcb3fb

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
sethmlarson: You can use `Dictionary<KeyType, ValueType>` or if it's always a single key in the dictionary you can use `SingleKeyDictionary<KeyType, ValueType>` for this scenario.

lcawl: Thanks, I've updated that type in 3365acded91537bbcddf0dd4da3ea503cedccb8f

davidkyle (approved):
LGTM

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentResponse.ts
```diff
@@ -0,0 +1,61 @@
...
+     */
+    is_truncated?: boolean
+    /**
+     * If the model is trained for a text classification or zero shot classification task, the response is the
+     * predicted class.
+     * For named entity recognition (NER) tasks, it contains the annotated text output.
+     * For fill mask tasks, it contains the top prediction for replacing the mask token.
+     * For text embedding tasks, it contains the raw numerical text embedding values.
+     */
+    predicted_value?: string | double[]
```
sethmlarson, line 41: We'll need an alias for this, maybe `export type PredictedValue = string | double`

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+     * @server_default 10s
+     */
+    timeout?: Time
+  }
+  body: {
+    /**
+     * An array of objects to pass to the model for inference. The objects should contain a field matching your
+     * configured trained model input. Typically, the field name is `text_field`. Currently, only a single value is allowed.
+     */
+    docs: SingleKeyDictionary<string, string>[]
```
sethmlarson, line 50: Not to be forward thinking, but if there are plans to allow multiple keys in this dictionary eventually it's fine to make this a normal `Dictionary<string, string>[]`

sethmlarson (approved):
One question and one comment, otherwise this LGTM! :tada:

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentResponse.ts
lcawl: Thanks, I added that to the TrainedModel types in fa0bad14c7570d345ad3075c1908aeb8dd9a6a33

specification/ml/infer_trained_model_deployment/MlInferTrainedModelDeploymentRequest.ts
lcawl: Done in fa0bad14c7570d345ad3075c1908aeb8dd9a6a33, thanks!

sethmlarson (approved):

PR #1061, opened 2021-11-24
Adds ML start and stop trained model deployment specifications

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/73660

Adds specifications for the [start trained model deployment API](https://www.elastic.co/guide/en/elasticsearch/reference/master/start-trained-model-deployment.html) and [stop trained model deployment APII](https://www.elastic.co/guide/en/elasticsearch/reference/master/stop-trained-model-deployment.html) based on the documentation and specs ([ml.stop_trained_model_deployment.json](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ml.stop_trained_model_deployment.json), [ml.start_trained_model_deployment](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ml.start_trained_model_deployment.json))

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -124730,7 +124730,19 @@
     {
       "body": {
         "kind": "properties",
-        "properties": []
+        "properties": [
+          {
+            "name": "stopped",
+            "required": true,
+            "type": {
+              "kind": "instance_of",
+              "type": {
+                "name": "boolean",
+                "namespace": "internal"
+              }
+            }
+          }
+        ]
       },
       "kind": "response",
       "name": {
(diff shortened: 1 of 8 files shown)


lcawl:
@sethmlarson I'm unsure how to represent the `routing_table` object in the start trained model deployment specification. Per [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/start-trained-model-deployment.html#start-trained-model-deployment-example), it looks like it contains properties with names that differ depending on the nodes involved. There's a `routing_table` in the get cluster state API too, so perhaps it will be something common we can use here.

PR #1064, opened 2021-11-25
Adds reset transform API specification

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/79828


sethmlarson:
`main` is now targeting 8.1 so this PR can continue.

PR #1068, opened 2021-11-29
Remove NESTism documents property

stevejgordon:
Remove a property which is a carry over from a NEST shortcut to the hits. This is not on the response.


sethmlarson (approved):
@stevejgordon can you rebase+regen and then merge this?

PR #1079, opened 2021-11-30
Fix async search apis

delvedor:
As titled.


philkra (approved):
LGTM

specification/_global/search/_types/hits.ts
```diff
@@ -40,9 +40,7 @@ import { Sort, SortResults } from '@_types/sort'
 export class Hit<TDocument> {
   _index: IndexName
   _id: Id
-
-  _score?: double
-
+  _score?: double | null
```
swallez, line 43: Is it optional, null, or both? When do we have null values and when do we have no value?

swallez: Answering my own question after checking the source code: `_score` and `max_score` are always there (i.e. not optional) but can indeed be `null` if no score was calculated because the query was a non-scoring query.

I'll open a follow-up PR to fix this.

PR #1088, opened 2021-11-30
Remove deprecated 'type' properties

### LINKED ISSUE #1035: Remove `_type` and other deprecated properties from the main branch
We have a lot of `@deprecated 7.0.0` properties in the spec, which should be removed from the `main` (8.0) branch.

They're still supported by ES 8.0 in compatibility mode, but should not be present in 8.0 clients.

sethmlarson:
Looks like there are still `type` path parameters on multiple APIs but when running the API validator it doesn't fail. Maybe these parameters are supposed to be removed upstream?

Upstream PR here: https://github.com/elastic/elasticsearch/pull/81179

Closes #1035


philkra (approved):
LGTM and 👋  to good'ol types in 8.0

sethmlarson:
Waiting for the results of https://github.com/elastic/elasticsearch/pull/81179 to hit the artifacts API, then will rebase.

swallez (approved):
LGTM, but we should backport the removal of the two `size` parameters to 7.x

sethmlarson:
@swallez I'm not sure we should backport the removal of `size` to 7.16, I tested the parameter and it appears to work even if not documented on elastic.co. Still defined in the rest-api-spec.

PR #1091, opened 2021-12-01
Remove source parameters from update_by_query/delete_by_query

sethmlarson:
Follows from https://github.com/elastic/elasticsearch/pull/81131, these parameters were always no-ops as both `update_by_query` and `delete_by_query` don't return source information.


philkra (approved):
LGTM

PR #1095, opened 2021-12-01
Switch to using concrete types for ml.put_trained_model_definition_part

sethmlarson:
`number` converts to `float` in Python which isn't correct, should be an integer/long.


philkra (approved):
LGTM

lcawl:
Thanks for the fix!

PR #1126, opened 2021-12-16
Remove _type which is no longer returned on bulk response items

stevejgordon:
As per title


sethmlarson (approved):

sethmlarson:
Backported to 8.0: https://github.com/elastic/elasticsearch-specification/pull/1128

PR #1132, opened 2021-12-16
Add missing "wrapper" query type

swallez:
The "wrapper" query type is missing in `QueryContainer`.

Reported in https://github.com/elastic/elasticsearch-java/issues/67

PR #1136, opened 2021-12-16
query.wrapper should be optional

delvedor:
As titled.

Follow-up of #1132


swallez (approved):
LGTM. And sorry for having overlooked that in the previous PR :-/

PR #1144, opened 2021-12-16
max_score and _score are required (and can be null)

swallez:
Follow-up to https://github.com/elastic/elasticsearch-specification/pull/1079

The `_score` and `max_score` can be null, but are always present. This PR makes them required.


sethmlarson (approved):

delvedor (approved):

delvedor (changes requested):
Blocking as we have discovered that many yaml tests suggest that this property should be optional.

swallez:
It's weird to not see these fields in the yaml tests as they are always sent by the ES server.

This could be related to some lenient behavior of the yaml test engine. Putting this PR on hold for now.

pquentin (approved):
Nearly three years later, YAML test validation no longer suggests those properties should be optional, as 2041/2089 is the current response score.

pquentin (approved):

swallez:
`max_score` and `_score` can be `null` (see [here](https://github.com/elastic/elasticsearch/blob/3566ee9a0129394cb23cc40c6af1f19a7fa68193/server/src/main/java/org/elasticsearch/search/SearchHits.java#L301) and [here](https://github.com/elastic/elasticsearch/blob/3566ee9a0129394cb23cc40c6af1f19a7fa68193/server/src/main/java/org/elasticsearch/search/SearchHit.java#L837)) when the score is `NaN`. This is semantically equivalent to a missing value (there's no known score) and is not a "meaningful null", and so should be represented as an optional value in strongly typed languages.

@Anaethelion @l-trotta @flobernd can you make sure your respective code generators correctly interpret a required `double | null` as an optional `double`?

flobernd:
@swallez I can confirm that the .NET generator converts `X | null` unions to nullable (optional) values (`X?`).

pquentin:
There are few existing mandatory `| null` in the specification, one example being `IndicatorNode` in the Health Report API.

 * Java uses the `Nullable` annotation: https://github.com/elastic/elasticsearch-java/blob/71d6401c7fb0eb1b80399dd9320e250368c5eff5/java-client/src/main/java/co/elastic/clients/elasticsearch/core/health_report/IndicatorNode.java#L62-L66
 * .NET treats them as Nullable<T>, as mentioned above: https://github.com/elastic/elasticsearch-net/blob/b5fe64620459767f9b2ef751223509c4b45e925e/src/Elastic.Clients.Elasticsearch.Serverless/_Generated/Types/Core/HealthReport/IndicatorNode.g.cs#L32-L35
 * Go uses `omitempty`, which does not appear to use `nil` as I would have naively expected: https://github.com/elastic/go-elasticsearch/blob/3571fa5dd10f462d5902fcc344bc8932c033f734/typedapi/types/indicatornode.go#L36-L37

In any case, since this already exists, I would be in favor of merging anyway.

l-trotta:
that's a no for java, the result here is just `double`, which is not nullable

l-trotta:
if java is the only problem I'll assign this one to myself and merge it once I have a fix for it

Anaethelion:
Go is in the same state as Java, for this particular case it translates to a `Float64`.

pquentin:
I believe Java is the only language left here since https://github.com/elastic/elastic-client-generator-go/pull/39 was merged.

l-trotta:
I am still working on this yes

l-trotta:
sorry I forgot, this was fixed in version 9 of the java client! we can finally merge this

PR #1146, opened 2021-12-17
Adds missing model_size_bytes parameter to ML put trained model API

dimitris-athanasiou:


specification/ml/put_trained_model/MlPutTrainedModelRequest.ts
```diff
@@ -73,6 +74,11 @@ export interface Request extends RequestBase {
      * @server_default tree_ensemble
      */
     model_type?: TrainedModelType
+    /**
+     * The estimated memory usage in bytes to keep the trained model in memory.
+     * Provide this when the model definition is not supplied.
```
sethmlarson, line 79: From the docs:

```
This property is supported only if defer_definition_decompression is true or the model definition is not supplied. 
```

dimitris-athanasiou: Fixed!

sethmlarson (approved):
LGTM

PR #1158, opened 2021-12-21
Validate ilm.delete_lifecycle

philkra:
added more descriptions and fixed upstream tests in https://github.com/elastic/clients-flight-recorder/pull/839


delvedor (approved):
LGTM

PR #1174, opened 2021-12-21
Improve type validation DX

delvedor:
Other than having nicer logs, this new script reduces the build time form ~40s to ~20s on my machine 🚀 

Example command:
```sh
make validate api=xpack.info type=request stack-version=8.1.0-SNAPSHOT
```

Once merged, you should run `make setup-env`.


Makefile
```diff
@@ -1,24 +1,10 @@
 SHELL := /bin/bash
 
-validation-all:  ## Run Validation on all Endpoints
-	@echo ">> running validations for all endpoints .."
-	./run-validations.sh
+validate: ## Validate a given endpoint request or response
```
philkra, line 3: is stack version mandatory? If so, I'd prefer to add a test like [here](https://github.com/elastic/elasticsearch-specification/pull/1174/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L12)

philkra (approved):
One question, otherwise - LGTM

Makefile
delvedor: Yes, it is. We **could** fallback top a default value, but then we would need to update it every minor release.

philkra: Yea, that is not very convenient. Let me make a suggestion

Makefile
```diff
@@ -1,24 +1,10 @@
...
+	@node compiler/run-validations.js --api $(api) --type $(type) --stack-version $(stack-version)
```
philkra, line 4:
```suggestion
  test -n "$(stack-version)" # missing stack-version param
	@node compiler/run-validations.js --api $(api) --type $(type) --stack-version $(stack-version)
```

philkra (changes requested):

Makefile
philkra: downstream check makes this irrelevant

philkra (approved):
LGTM

Makefile
delvedor: The script runs the validation of the parameters.

Anaethelion (approved):
Kudos for speed and better logs!

LGTM

PR #1176, opened 2021-12-22
Fix multiple ml endpoints

philkra:
* `ml.delete_trained_model`
* `ml.flush_jobs`
* `ml.validate`


specification/ml/evaluate_data_frame/MlEvaluateDataFrameRequest.ts
```diff
@@ -37,16 +37,15 @@ export interface Request extends RequestBase {
...
     evaluation: DataframeEvaluationContainer
     /**
-     * Defines the index in which the evaluation will be performed.
+     * Defines the `index` in which the evaluation will be performed.
      */
     index: IndexName
     /**
-     * A query clause that retrieves a subset of data from the source index.
-     * @doc_url https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html
+     * A query clause that retrieves a subset of data from the source index. See [QueryDSL](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html).
```
delvedor, line 48: @lcawl do you prefer the link as markdown link or with `@doc_url`?
If provided with `@doc_url`, it will be easier to access in the resulting JSON, while if added in the description you can't easily access it if needed.

specification/ml/delete_trained_model/MlDeleteTrainedModelRequest.ts
```diff
@@ -35,4 +35,7 @@ export interface Request extends RequestBase {
      */
     model_id: Id
   }
+  query_parameters: {
+    force?: boolean
```
lcawl, line 39:
```suggestion
    /** Forcefully deletes a trained model that is referenced by ingest pipelines or has a started deployment. **/
    force?: boolean
```

specification/ml/evaluate_data_frame/MlEvaluateDataFrameRequest.ts
```diff
@@ -37,16 +37,15 @@ export interface Request extends RequestBase {
   body: {
     /**
      * Defines the type of evaluation you want to perform.
-     * @doc_url https://www.elastic.co/guide/en/elasticsearch/reference/current/evaluate-dfanalytics.html#ml-evaluate-dfanalytics-resources
+     * See [Data frame analytics evaluation resources](https://www.elastic.co/guide/en/elasticsearch/reference/current/evaluate-dfanalytics.html#ml-evaluate-dfanalytics-resources).
```
lcawl, line 40: This is just a link to a lower section of the API page, so IMO it's not useful here
```suggestion
```

specification/ml/evaluate_data_frame/MlEvaluateDataFrameRequest.ts
lcawl: My preference is [doc_id](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#doc_id), so we don't have to clean up these URLs in multiple places when we move to the new docs system. Is there somewhere we can start recording those ID<->URL mappings?

delvedor: Unfortunately we don’t know how to find the appropriate doc_id, where can we find it?

lcawl: Could it just be a flat file that contains keywords and URLs, which is checked when we run make contrib (i.e. substitute the keywords for the URLs in the output)? The goal would be to just isolate all the URLs in one place, so they're easier to maintain. The way they accomplished this in Kibana is per this file, but we could probably be even simpler in this case: https://github.com/elastic/kibana/blob/main/src/core/public/doc_links/doc_links_service.ts

philkra:
Merging this PR as the doc Id conversation can be redirected towards this draft PR: https://github.com/elastic/elasticsearch-specification/pull/1231

PR #1177, opened 2021-12-22
Use Metadata type consistently

philkra:
replaced every occurrence of `meta?: Dictionairy<string, UserDefinedValue>` with the type `Metadata`.


specification/_types/aggregations/AggregationContainer.ts
```diff
@@ -110,7 +111,7 @@ export class AggregationContainer {
   /**
    * @variant container_property
    */
-  meta?: Dictionary<string, UserDefinedValue>
+  meta?: Metadata
```
philkra, line 114: @swallez can you confirm that this change has no impact on the variant annotation?

specification/security/get_service_accounts/types.ts
```diff
@@ -37,5 +36,5 @@ export class RoleDescriptor {
   applications?: ApplicationPrivileges[]
   metadata?: Metadata
   run_as?: string[]
-  transient_metadata?: Dictionary<string, UserDefinedValue>
+  transient_metadata?: TransientMetadata
```
swallez, line 39: `TransientMetadata` is actually badly named and should be renamed to something like `TransientMetadataConfig` since its definition is

```ts
export class TransientMetadata {
  enabled: boolean
}
```

specification/_types/aggregations/AggregationContainer.ts
philkra: Thank you double checking!

specification/_types/mapping/Property.ts
```diff
@@ -41,7 +41,7 @@ import {
 
 export class PropertyBase {
   local_metadata?: Metadata
-  meta?: Dictionary<string, string>
+  meta?: Metadata
```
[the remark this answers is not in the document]
philkra: Reverted back to `Dictionary<string, string>`

specification/security/get_service_accounts/types.ts
philkra: Moved to `_types` as it was used 3 times and renamed to `TransientMetadataConfig `

swallez (approved):
LGTM!

PR #1178, opened 2021-12-22
Fix msearch_template Response

philkra:
as titled.

### DIFF
--- a/specification/_global/msearch/types.ts
+++ b/specification/_global/msearch/types.ts
@@ -65,9 +65,5 @@ export class MultiSearchResult<TDocument> {
 
 /** @codegen_names result, failure */
 export type ResponseItem<TDocument> =
-  | MultiSearchItem<TDocument>
+  | SearchResponse<TDocument>
   | ErrorResponseBase
-
-export class MultiSearchItem<TDocument> extends SearchResponse<TDocument> {
-  status: integer
-}
--- a/specification/_types/Base.ts
+++ b/specification/_types/Base.ts
@@ -73,7 +73,7 @@ export class ErrorResponseBase {
   // If the error is a string, it means that it was not caused by an exception on ES side, but on the HTTP routing layer.
   // This should never happen in clients, because we assume we will never send malformed request.
   error: ErrorCause
-  status: integer
+  status?: integer
 }
 
 export class IndicesResponseBase extends AcknowledgedResponseBase {
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1472,12 +1472,6 @@
       ],
       "response": []
     },
-    "search": {
-      "request": [],
-      "response": [
-        "Non-leaf type cannot be used here: '_global.search:Response'"
-      ]
-    },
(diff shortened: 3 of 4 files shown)


delvedor (commented):
@swallez can you take a look at this one? You added this change in https://github.com/elastic/elasticsearch-specification/pull/1034.

specification/_global/msearch/types.ts
```diff
@@ -65,9 +65,5 @@ export class MultiSearchResult<TDocument> {
 
 /** @codegen_names result, failure */
 export type ResponseItem<TDocument> =
-  | MultiSearchItem<TDocument>
+  | SearchResponse<TDocument>
```
swallez, line 68: multisearch response items have a `status` field that indicates the result of each search request. This is present in the [ES source code](https://github.com/elastic/elasticsearch/blob/8754c17a2d6218bce2041cddc572d2a58cef895f/server/src/main/java/org/elasticsearch/action/search/MultiSearchResponse.java#L164) and [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html#search-multi-search-api-response-body).

swallez, line 68: We need `MultiSearchItem` as multisearch response items have a `status` field that indicates the result of each search request. This is present in the [ES source code](https://github.com/elastic/elasticsearch/blob/8754c17a2d6218bce2041cddc572d2a58cef895f/server/src/main/java/org/elasticsearch/action/search/MultiSearchResponse.java#L164) and explained in [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html#search-multi-search-api-response-body).

specification/_types/Base.ts
```diff
@@ -73,7 +73,7 @@ export class ErrorResponseBase {
   // If the error is a string, it means that it was not caused by an exception on ES side, but on the HTTP routing layer.
   // This should never happen in clients, because we assume we will never send malformed request.
   error: ErrorCause
-  status: integer
+  status?: integer
```
swallez, line 76: Status is always present in error responses, including in multisearch responses. See [ES code](https://github.com/elastic/elasticsearch/blob/8754c17a2d6218bce2041cddc572d2a58cef895f/server/src/main/java/org/elasticsearch/action/search/MultiSearchResponse.java#L161) and [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html#search-multi-search-api-response-body) "an object with error message and corresponding status code will be returned in place of the actual search response".

Keeping `error` and `status` required also plays an important role (in the Java client) to disambiguate some non-200 results where ES will return either a valid response or an error depending on the parameters (e.g. delete document that returns 404+error if the index doesn't exist but 404+valid response if the index exist but not the document).

swallez (changes requested):
Looking at the reference material (code + docs), this change seems wrong. What are the recordings that led to it?

philkra:
Closing this in favour of https://github.com/elastic/elasticsearch-specification/pull/1320

PR #1187, opened 2021-12-23
Validate Security.* endpoints

philkra:
* `security.authenticate`
* `security.get_role`
* `security.get_role_mappings`
* `security.get_service_accounts`
* `security.get_token`
* `security.get_user_privileges`
* `security.put_role`


specification/security/get_role_mapping/SecurityGetRoleMappingRequest.ts
```diff
@@ -18,15 +18,19 @@
...
  * @stability stable
+ * @cluster_privileges manage_security
  */
 export interface Request extends RequestBase {
   path_parts: {
-    name?: Name
+    /**
+     * The distinct name that identifies the role mapping. The name is used solely as an identifier to facilitate interaction via the API; it does not affect the behavior of the mapping in any way. You can specify multiple mapping names as a comma-separated list. If you do not specify this parameter, the API returns information about all role mappings.
+     */
+    name?: Names | null
```
delvedor, line 34: I don't think `null` should be here, it will be serialized as `'null'` in the path.
```suggestion
    name?: Names
```

philkra: Without the test:
```
// Test file: /test/platinum/role_mapping/10_basic.yml
// Test name: Test put role_mapping api
```

is failing with the payload:
```{
  "name": null
}```

philkra: @delvedor  we need to skip the test then?

delvedor: We can either skip it or update the upstream yaml to not pass the `name` at all.
Very likely the test has been coded in that way because the Java runner doesn't send anything if it sees `null`.

philkra: I commited the suggestion, let's address this later. The test will be failing for now, but this is a corner case.

delvedor (approved):

PR #1194, opened 2021-12-23
Use nvmrc to specify the require node version

delvedor:
With this it will be easier to have everyone on the same nodejs version and handle upgrades.


swallez (approved):
Nice improvement!

PR #1212, opened 2021-12-24
Improve Search

philkra:
Request and response are not fully passing but, pushed the error count down.

Aggregations & Queries added:
* `bucket_ks`
* `bucket_correlations`
* `knn`

Investigated recordings for upstream test muting.


delvedor (approved):
LGTM

PR #1214, opened 2021-12-24
Fix Transform.* endpoints

philkra:
* `transform.preview_transform`
* `transform.put_transform`
* `transform.update_transform`
* `transform.get_transform_stats`


delvedor (approved):
LGTM

Issue #1011, opened 2021-11-08
Check the node/npm version in `make setup-env`


swallez:
In PR #990 we added the `package-lock.json` files to the git repo. Running `make setup-env` in a version of node/npm that is higher than `v14.x`/`6.x` will rewrite these files in a different format.

The `make setup-env` tasks should check that node/npm are in the expected version and fail otherwise.

PR #1219, opened 2021-12-27
Check nodejs version before running the compiler

delvedor:
With this change, the compiler will ensure that you are running it with the nodejs version configured in `.nvmrc`.
The `make setup` target will perform the same check as well.

Closes: #1011


philkra (approved):
LGTM

PR #1222, opened 2021-12-28
Upgrade node and deps

delvedor:
Once this pr is merged, we’ll require Node.js v16 to be used when working on this project.
This pr also updates the dependencies of our tooling.

Steps required once this pr is merged in your machine:
```sh
# pull latest changes in main
git pull

# install the nodejs version configured in .nvmrc
nvm install

# upgrade the project dependencies
make setup

# only for the very first validation, use the validate-no-cache target
make validate-no-cache api=… type=… stack-version=…
```

After running the commands above, I strongly suggest to run `nvm alias default 16`, in this way any new terminal instance will use Node.js v16 by default.


philkra (approved):
lgtm

PR #1225, opened 2021-12-28
Suggest the correct API name if the input is wrong

delvedor:
As titled.

Example:
```sh
▶ make validate stack-version=8.1.0-SNAPSHOT type=response api=indices.get_mappings
✖ The api 'indices.get_mappings' does not exists, did you mean 'indices.get_mapping'?
```

This check will be performed by both the compiler and validator.


philkra (approved):
LGTM 😄

PR #1240, opened 2022-01-06
Remove kibana private roles in IndexPrivilege

Anaethelion:


swallez (commented):
These `indices:...` privileges indeed don't seem to exist in the [ES index privileges enumeration](https://github.com/elastic/elasticsearch/blob/3d8107e5081667baa8c79075534fecd7c10bc936/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/IndexPrivilege.java#L121). However, the ES source code also mentions `none` that is missing from the enum.

PR #1242, opened 2022-01-10
Fix regression in aggregation changes

stevejgordon:
As discussed with @swallez last week. It appears that a mistake slipped into a prior PR which changed the `distribution` field in `StringStatsAggregate` to a `string` from a `Dictionary<string, double>`. This PR reverts this as .NET client tests are failing due to the currently generated type using a string for the property.

@swallez Not sure if we want this backported back to 7.x branches?


swallez (approved):
LGTM. And it should be backported to 7.17 even if it's a breaking change: this is an API spec bug that will to cause deserialization crashes in strongly typed clients like Java.

PR #1246, opened 2022-01-11
Add support for open/close enums

delvedor:
As titled.

**This should not be backported to v7**.

### DIFF
--- a/compiler/src/model/metamodel.ts
+++ b/compiler/src/model/metamodel.ts
@@ -325,6 +325,11 @@ export class EnumMember {
  */
 export class Enum extends BaseType {
   kind: 'enum'
+  /**
+   * If the enum is open, it mens that other than the specified values it can accept an arbitrary value.
+   * If this property is not present, it means that the enum is not open (in other words, is closed).
+   */
+  isOpen?: boolean
   members: EnumMember[]
 }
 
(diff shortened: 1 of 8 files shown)


typescript-generator/src/metamodel.ts
```diff
@@ -325,6 +325,11 @@ export class EnumMember {
  */
 export class Enum extends BaseType {
   kind: 'enum'
+  /**
+   * If the enum is open, it mens that other than the specified values it can accept an arbitrary value.
```
Anaethelion, line 329:
```suggestion
   * If the enum is open, it means that other than the specified values it can accept an arbitrary value.
```

compiler/src/model/metamodel.ts
```diff
@@ -325,6 +325,11 @@ export class EnumMember {
  */
 export class Enum extends BaseType {
   kind: 'enum'
+  /**
+   * If the enum is open, it mens that other than the specified values it can accept an arbitrary value.
```
Anaethelion, line 329:
```suggestion
   * If the enum is open, it means that other than the specified values it can accept an arbitrary value.
```

Anaethelion (approved):
Left some suggestions for typos in comments, other than that LGTM!

swallez (approved):
LGTM!

PR #1250, opened 2022-01-12
Change timestamps defined as integer to long

swallez:
Integers are not large enough to hold timestamps. This caused a failure in the Java client (see https://github.com/elastic/elasticsearch-java/issues/88).

This PR fixes fields declared as `*_timestamp: integer` (which includes the one that caused the original issue).


delvedor (approved):
LGTM

Issue #1253, opened 2022-01-12
Add health and status properties to the Indices Stats API spec


yuliacech:
## 🐛 Wrong type
The Indices Stats API has [been updated](https://github.com/elastic/elasticsearch/pull/81954) to include index `health` and `status`. It would be great to add those properties to the [IndicesStats spec](https://github.com/elastic/elasticsearch-specification/blob/main/specification/indices/stats/types.ts).

<img width="806" alt="Screenshot 2022-01-12 at 17 25 53" src="https://user-images.githubusercontent.com/6585477/149181107-3acb20a2-49ed-41d0-b4ee-bfa077255019.png">

## Definition

```diff
export class IndicesStats {
   primaries?: IndexStats
   shards?: Dictionary<string, ShardStats[]>
   total?: IndexStats
   uuid?: Uuid
+ health?: string // "green", "yellow, "red"
+ status?: string // "open", "close"
}
```

philkra:
Thank you @yuliacech - I opened https://github.com/elastic/elasticsearch-specification/pull/1254 to add the fields.

PR #1254, opened 2022-01-12
Add new 8.1.0 fields for indices.stats

philkra:
closes https://github.com/elastic/elasticsearch-specification/issues/1253

### DIFF
--- a/specification/indices/stats/types.ts
+++ b/specification/indices/stats/types.ts
@@ -94,7 +94,7 @@ export class IndicesStats {
   /** @since 8.1.0 */
   health?: HealthStatus
   /** @since 8.1.0 */
-  status?: string
+  status?: IndexMetadataState
 }
 
 export class ShardCommit {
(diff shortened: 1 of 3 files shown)


specification/indices/stats/types.ts
```diff
@@ -85,6 +91,10 @@ export class IndicesStats {
   shards?: Dictionary<string, ShardStats[]>
   total?: IndexStats
   uuid?: Uuid
+  /** @since 8.1.0 */
+  health?: HealthStatus
+  /** @since 8.1.0 */
+  status?: string
```
sethmlarson, line 97: I was surprised this wasn't an enum yet! Should we add a new enum `IndexState` with `open` and `closed` to mirror the enum in Elasticsearch?

sethmlarson (commented):
Additions look good, one thought:

specification/indices/stats/types.ts
philkra: Same here 😄  I was a bit cautious to add an enum directly since this is a new field/value - I'll check the server code and adjust to enum.

philkra: https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java#L125 

`IndexMetadata.State` is an enum `open` & `close`

sethmlarson (approved):

PR #1258, opened 2022-01-13
Address some of the Top 20 APIs

swallez:
Backport #1235 to 7.17.

These changes, even if they introduce breaking changes, have to be backported as they're fixing errors in the spec that will cause strongly types clients (like Java) to fail.


philkra (approved):
LGTM - thank you for backporting!

PR #1259, opened 2022-01-13
Refactor and fix task APIs

swallez:
This PR started because of a missing `codegen_names` for `TaskInfo[] | Dictionary<string, Info>` and after looking at the ES source code it ended up being a rather extensive refactoring/fix of the task list and cancel APIs .

It also renames the some very generic names like `Status` to `TaskStatus` to avoid name conflicts and moves `TaskFailure` to global types as it's used in a number of namespaces.

**This PR must not be packported to 7.x** because of the name changes, even if the 7.x spec is incomplete, as the renamings and refactorings would cause breaking changes.

Validation:

```
% make validate api=tasks.list stack-version=8.1.0-SNAPSHOT
- Validating endpoints
✔ tasks.list request has been successfully validated!

✔ 8 out of 8 test request cases are passing.
✔ 8 out of 8 test response cases are passing.
```

```
% make validate api=tasks.cancel stack-version=8.1.0-SNAPSHOT
- Validating endpoints
✔ tasks.cancel request has been successfully validated!

✔ 2 out of 2 test request cases are passing.
✔ 2 out of 2 test response cases are passing.
```


philkra (approved):
LGTM 🚀

Issue #339, opened 2021-04-14
Create Report of Import Graph for improving Organisation


philkra:
## 🚀 Feature Proposal

Track non Request & Response class usage for optimising organisation.

## Motivation

It's difficult to keep to identify common classes. Having usage graph would allow moving more classes into common and the other way around to push only one time used classes (& enums) back into the calling class.

## Example

```
{
"name": "ApiKey",
"namespace": "common.security",
"imported_by": [
  {
    "name": "GetApiKeyRequest",
    "namespace": "security"
  },
  {
    "name": "PutApiKeyRequest",
    "namespace": "security"
  }
]
}
```

philkra:
markdown is likely better

PR #1261, opened 2022-01-14
Create import graph of types and namespaces

delvedor:
This pr adds a new compiler step that generates two additional JSONs, one with the import graph of every type and another one with the import graph of every namespace.
From this data, we will generate a UI to visualize how different modules interacts with each other.

TODO:
- [x] generate a raw data JSON report
- [x] generate a graphical report to visualize dependencies
- [x] publish UI report to github pages

Closes: #339


compiler/src/steps/import-graph.ts
```diff
@@ -0,0 +1,141 @@
...
+        default:
+      }
+    }
+  }
+
+  const typeGraph: ImportTypesGraph[] = []
+  for (const [type, types] of importGraph.entries()) {
+    typeGraph.push({
+      type,
+      imported_by: types.map(type => type.name)
```
delvedor, line 97: TODO: this should be a unique array.

compiler/src/steps/import-graph.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+interface ImportNamespaceGraph {
+  namespace: string
+  imported_by: string[]
+}
+
+export default async function createImportGraph (model: model.Model, jsonSpec: Map<string, JsonSpec>): Promise<model.Model> {
+  const importGraph: Map<model.TypeName, model.TypeDefinition[]> = new Map()
+  for (const current of model.types) {
```
swallez, line 37: It's more complex than that, as we have to traverse every `ValueOf` (extends, generics, property types, etc) to build the graph from `TypeName` values.

output/schema/import-namespace-graph.json
```diff
@@ -0,0 +1,10954 @@
+[
+  {
+    "namespace": "_global.bulk",
```
philkra: https://github.com/elastic/elasticsearch-specification/pull/1261#discussion_r784709089 assume the same holds true for the namespace. Should be unique

philkra (changes requested):

delvedor:
The import map is not committed in the repo for now, it adds a lot of data that is only used for generating the graph (which is published on GH pages).
We'll generate some markdown representations of the json that will be published on GH pages as well.

swallez (approved):
LGTM!

philkra (approved):
LGTM

PR #1264, opened 2022-01-14
Fix some unlabelled unions

swallez:
Fixes some union types that had a missing `@codegen_names`, and also a property that only had a `null` type.

`AggregationProfileDebug` and `AggregationProfileDelegateDebug` have been merged into a single type. On the server side it's a `Map<String, Object>` and there's no strong indication that these are really two different types. Avoiding the union also simplifies the type definition.


philkra (approved):
LGTM

PR #1267, opened 2022-01-17
Remove old fix imports script

delvedor:
We can no longer use this script as it's not smart enough to actually fix the imports.
Better to remove it because if you run it accidentally it will screw up the specification.


philkra (approved):
LGTM - thank you!

PR #1272, opened 2022-01-18
Fix Action definitions for watcher.put_watcher API

sethmlarson:
Fixes all tests for `watcher.put_watch` except for search input test failures which look like bad NEST recordings, will make a separate PR for those.


delvedor (approved):

delvedor:
Should we backport this to 7.x?

sethmlarson:
It changes a ton of names so I don't think so?

PR #1276, opened 2022-01-19
Fix indices apis

delvedor:
This pr fixes many (but not all) errors we are getting in various indices apis.


stevejgordon (approved):
LGTM

PR #1278, opened 2022-01-19
Small api fixes

delvedor:
This pr fixes a bunch of apis that are currently failing (but it's not worth opening a separate pr for each one of them).

### DIFF
--- a/specification/_types/mapping/RuntimeFields.ts
+++ b/specification/_types/mapping/RuntimeFields.ts
@@ -21,7 +21,7 @@ import { Dictionary } from '@spec_utils/Dictionary'
 import { Field } from '@_types/common'
 import { Script } from '@_types/Scripting'
 
-export type RuntimeFields = Dictionary<Field, RuntimeField>
+export type RuntimeFields = Dictionary<Field, RuntimeField | RuntimeField[]>
 
 export class RuntimeField {
   format?: string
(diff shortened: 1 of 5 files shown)


specification/_types/mapping/RuntimeFields.ts
```diff
@@ -21,7 +21,7 @@ import { Dictionary } from '@spec_utils/Dictionary'
 import { Field } from '@_types/common'
 import { Script } from '@_types/Scripting'
 
-export type RuntimeFields = Dictionary<Field, RuntimeField>
+export type RuntimeFields = Dictionary<Field, RuntimeField | RuntimeField[]>
```
delvedor, line 24: @swallez @stevejgordon  is this a problem for you?

stevejgordon: @delvedor In regards to v8, this is fine for me. The generator simplifies such unions to prefer the array only. Today NEST has this as `Dictionary<Field, RuntimeField>` so it differs there, but if this is correct in terms of how these can be expressed, it makes sense. When I implemented runtime fields I didn't see a case where the array is used.

stevejgordon (approved):
LGTM

specification/_types/mapping/RuntimeFields.ts
stevejgordon: Looking at this again though, I'm trying to work out how it's valid to define more than runtime field for a particular field name? I don't see it documented either. Do we have an example of where this occurs as I probably warrants a test case in my client!

swallez: Java also simplifies `Foo | Foo[]` to just `Foo[]` because this type represents ES's leniency when it parses arrays (accepts a single object for a 1 element array).

As for the actual definition in ES, it's a `Map<String, Object>` and I wasn't able to go down to what is actually expected inside this `Object`.

sethmlarson (approved):

PR #1283, opened 2022-01-20
Align Datafeed+Job classes with read/write

sethmlarson:
```
$ make validate stack-version=8.1.0-SNAPSHOT api=ml.put_job
- Validating endpoints
⚠ It looks like ml.put_job request has some errors, take a look at the workbench.
/home/sethmlarson/clients-flight-recorder/scripts/types-validator/workbench

scripts/types-validator/workbench/f580409745707de8240570793edf99f1.test-d.ts:7:59
Argument of type '{ allow_lazy_open: false; analysis_config: { bucket_span: string; detectors: { by_field_name: string; custom_rules: { actions: "skip_result"[]; conditions: ({ applies_to: "actual"; operator: "lt"; value: number; } | { ...; })[]; scope: { ...; }; }[]; ... 4 more ...; partition_field_name: string; }[]; influencers: ne...' is not assignable to parameter of type 'ErrorResponseBase | MlPutJobResponse'.
  Property 'daily_model_snapshot_retention_after_days' is missing in type '{ allow_lazy_open: false; analysis_config: { bucket_span: string; detectors: { by_field_name: string; custom_rules: { actions: "skip_result"[]; conditions: ({ applies_to: "actual"; operator: "lt"; value: number; } | { ...; })[]; scope: { ...; }; }[]; ... 4 more ...; partition_field_name: string; }[]; influencers: ne...' but required in type 'MlPutJobResponse'.

✔ 226 out of 226 test request cases are passing.
✖ 225 out of 226 test response cases are passing.

$ make validate stack-version=8.1.0-SNAPSHOT api=ml.get_datafeeds
- Validating endpoints
✔ ml.get_datafeeds request has been successfully validated!

✔ 20 out of 20 test request cases are passing.
✔ 20 out of 20 test response cases are passing.

$ make validate stack-version=8.1.0-SNAPSHOT api=ml.put_datafeed
- Validating endpoints
✔ ml.put_datafeed request has been successfully validated!

✔ 71 out of 71 test request cases are passing.
✔ 71 out of 71 test response cases are passing.

make validate stack-version=8.1.0-SNAPSHOT api=ml.update_job
- Validating endpoints
✔ ml.update_job request has been successfully validated!

✔ 5 out of 5 test request cases are passing.
✔ 5 out of 5 test response cases are passing.

$ make validate stack-version=8.1.0-SNAPSHOT api=ml.get_jobs
- Validating endpoints
✔ ml.get_jobs request has been successfully validated!

✔ 31 out of 31 test request cases are passing.
✔ 31 out of 31 test response cases are passing.
```

### DIFF
--- a/specification/ml/_types/Datafeed.ts
+++ b/specification/ml/_types/Datafeed.ts
@@ -17,34 +17,16 @@
  * under the License.
  */
 
+import { OverloadOf } from '@spec_utils/behaviors'
 import { Dictionary } from '@spec_utils/Dictionary'
 import { AggregationContainer } from '@_types/aggregations/AggregationContainer'
-import { ExpandWildcards, Id, Indices, IndicesOptions } from '@_types/common'
+import { Id, IndexName, Indices, IndicesOptions } from '@_types/common'
 import { RuntimeFields } from '@_types/mapping/RuntimeFields'
 import { double, integer, long } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { ScriptField } from '@_types/Scripting'
 import { Time, Timestamp } from '@_types/Time'
 import { DiscoveryNode } from './DiscoveryNode'
-
-export class Datafeed {
-  /** @aliases aggs */
-  aggregations?: Dictionary<string, AggregationContainer>
-  chunking_config?: ChunkingConfig
-  datafeed_id: Id
-  frequency?: Timestamp
-  indices: string[]
-  indexes?: string[]
-  job_id: Id
-  max_empty_searches?: integer
-  query: QueryContainer
-  query_delay?: Timestamp
-  script_fields?: Dictionary<string, ScriptField>
-  scroll_size?: integer
-  delayed_data_check_config: DelayedDataCheckConfig
-  runtime_mappings?: RuntimeFields
-  indices_options?: IndicesOptions
-}
 
 export class DatafeedConfig {
   /**
@@ -68,15 +50,21 @@ export class DatafeedConfig {
    * The interval at which scheduled queries are made while the datafeed runs in real time. The default value is either the bucket span for short bucket spans, or, for longer bucket spans, a sensible fraction of the bucket span. For example: `150s`. When `frequency` is shorter than the bucket span, interim results for the last (partial) bucket are written then eventually overwritten by the full bucket results. If the datafeed uses aggregations, this value must be divisible by the interval of the date histogram aggregation.
    */
   frequency?: Timestamp
-  indexes?: string[]
+  /**
+   * Use `indices` instead.
+   */
+  indexes?: Indices
   /**
    * An array of index names. Wildcards are supported. If any indices are in remote clusters, the machine learning nodes must have the `remote_cluster_client` role.
    */
-  indices: string[]
+  indices?: Indices
   /**
    * Specifies index expansion options that are used during search.
    */
   indices_options?: IndicesOptions
+  /**
+   * The identifier for the anomaly detection job. This identifier can contain lowercase alphanumeric characters (a-z and 0-9), hyphens, and underscores. It must start and end with alphanumeric characters.
+   */
   job_id?: Id
   /**
    * If a real-time datafeed has never seen any data (including during any initial training period) then it will automatically stop itself and close its associated job after this many real-time searches that return no documents. In other words, it will stop after `frequency` times `max_empty_searches` of real-time operation. If not set then a datafeed with no end time that sees no data will remain started until it is explicitly stopped.
@@ -85,7 +73,7 @@ export class DatafeedConfig {
   /**
    * The Elasticsearch query domain-specific language (DSL). This value corresponds to the query object in an Elasticsearch search POST body. All the options that are supported by Elasticsearch can be used, as this object is passed verbatim to Elasticsearch.
    */
-  query: QueryContainer
+  query?: QueryContainer
   /**
    * The number of seconds behind real time that data is queried. For example, if data from 10:04 a.m. might not be searchable in Elasticsearch until 10:06 a.m., set this property to 120 seconds. The default value is randomly selected between `60s` and `120s`. This randomness improves the query performance when there are multiple jobs running on the same node.
    */
@@ -102,6 +90,25 @@ export class DatafeedConfig {
    * The size parameter that is used in Elasticsearch searches when the datafeed does not use aggregations. The maximum value is the value of `index.max_result_window`, which is 10,000 by default.
    * @server_default 1000
    */
+  scroll_size?: integer
+}
+
+export class DatafeedConfigRead implements OverloadOf<DatafeedConfig> {
+  /** @aliases aggs */
+  aggregations?: Dictionary<string, AggregationContainer>
+  chunking_config?: ChunkingConfig
+  datafeed_id: Id
+  delayed_data_check_config?: DelayedDataCheckConfig
+  frequency?: Timestamp
+  indexes?: Indices /** TODO: Remove but defined as alias above? */
+  indices: Indices /** TODO: Change to IndexName[] */
+  indices_options?: IndicesOptions
+  job_id: Id
+  max_empty_searches?: integer
+  query: QueryContainer
+  query_delay?: Timestamp
+  runtime_mappings?: RuntimeFields
+  script_fields?: Dictionary<string, ScriptField>
   scroll_size?: integer
 }
 
--- a/specification/ml/_types/Detector.ts
+++ b/specification/ml/_types/Detector.ts
@@ -34,11 +34,63 @@ export class Detector {
   /**
    * A description of the detector.
    */
+  description?: string
+  /**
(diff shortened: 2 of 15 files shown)


specification/ml/_types/Datafeed.ts
```diff
@@ -17,35 +17,17 @@
...
-import { ExpandWildcards, Id, Indices, IndicesOptions } from '@_types/common'
+import { Id, IndexName, Indices, IndicesOptions } from '@_types/common'
 import { RuntimeFields } from '@_types/mapping/RuntimeFields'
 import { double, integer, long } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { ScriptField } from '@_types/Scripting'
 import { Time, Timestamp } from '@_types/Time'
 import { DiscoveryNode } from './DiscoveryNode'
 
-export class Datafeed {
```
sethmlarson, line 30: Should we use `Datafeed` or `DatafeedConfig` for this type? Everywhere where users define a datafeed it's using `datafeed_config` but when we interact with the datafeed itself it's not called "config", just "datafeed". Maybe should change this to `Datafeed`?

specification/ml/_types/Datafeed.ts
```diff
@@ -68,15 +50,21 @@ export class DatafeedConfig {
    * The interval at which scheduled queries are made while the datafeed runs in real time. The default value is either the bucket span for short bucket spans, or, for longer bucket spans, a sensible fraction of the bucket span. For example: `150s`. When `frequency` is shorter than the bucket span, interim results for the last (partial) bucket are written then eventually overwritten by the full bucket results. If the datafeed uses aggregations, this value must be divisible by the interval of the date histogram aggregation.
    */
   frequency?: Timestamp
-  indexes?: string[]
+  /**
```
sethmlarson, line 53: The `indices`/`indexes` problem here is that on write we accept both but on read we always return `indices` and it's always an array of `IndexName`, not `Indices`.

specification/ml/_types/Detector.ts
```diff
@@ -30,6 +31,10 @@ export class Detector {
    * Custom rules enable you to customize the way detectors operate. For example, a rule may dictate conditions under which results should be skipped. Kibana refers to custom rules as job rules.
    */
   custom_rules?: DetectionRule[]
+  /**
+   * A description of the detector.
+   */
+  description?: string
```
sethmlarson, line 37: This situation is another strange one. On write we only accept `description`, but on read there's no `description` field but instead `datafeed_description`?

specification/ml/_types/Job.ts
```diff
@@ -44,50 +45,50 @@ export class JobStatistics {
...
   analysis_config: AnalysisConfig
   analysis_limits?: AnalysisLimits
   background_persist_interval?: Time
+  blocked?: JobBlocked
+  create_time?: integer
   custom_settings?: CustomSettings
   daily_model_snapshot_retention_after_days?: long
   data_description: DataDescription
-  datafeed_config?: DatafeedConfig
+  datafeed_config?: DatafeedConfig /** TODO: Make DatafeedConfigRead */
```
sethmlarson, line 76: I can't use `DatafeedConfigRead` here as must match `implemented OverloadOf<...>` type for the field. Can we loosen this requirement to allow for changing the type?

specification/ml/_types/Detector.ts
```diff
@@ -50,7 +54,7 @@ export class Detector {
   /**
    * The analysis function that is used. For example, `count`, `rare`, `mean`, `min`, `max`, or `sum`.
    */
-  function: string
+  function?: string
```
lcawl, line 57: >  function?: string

Is this actually optional?  I'm surprised if you can have a detector without a function. This should likely be reviewed by someone like @droberts195

specification/ml/_types/Detector.ts
```diff
@@ -67,15 +71,50 @@ export class Detector {
 }
 
 export class DetectorRead implements OverloadOf<Detector> {
-  by_field_name?: Field
-  custom_rules?: DetectionRule[]
-  detector_description?: string
-  detector_index?: integer
+  /**
+   * The field used to split the data. In particular, this property is used for analyzing the splits with respect to their own history. It is used for finding unusual values in the context of the split.
```
lcawl, line 75: I would expect most of these descriptions to match what's in the Detector type.  Is there a reason why we can't inherit the descriptions by default too?

specification/ml/put_job/MlPutJobRequest.ts
```diff
@@ -42,6 +42,10 @@ export interface Request extends RequestBase {
      */
     job_id: Id
   }
+  query_parameters: {
+    /** @deprecated 7.16.0 */
+    ignore_throttled?: boolean
```
lcawl, line 47: If it's true that this API supports ignore_throttled, it likely also supports the others mentioned here: https://www.elastic.co/guide/en/elasticsearch/reference/master/api-conventions.html#api-multi-index

specification/ml/put_job/MlPutJobRequest.ts
```diff
@@ -85,6 +89,10 @@ export interface Request extends RequestBase {
      * A list of job groups. A job can belong to no groups or many.
      */
     groups?: string[]
+    /**
+     * The identifier for the anomaly detection job. This identifier can contain lowercase alphanumeric characters (a-z and 0-9), hyphens, and underscores. It must start and end with alphanumeric characters.
+     */
+    job_id?: Id
```
lcawl, line 95: I'm not sure why we'd support putting a job_id in the body when it's required in the path. I think I vaguely remember knowing that if you *did* supply it in the body you got an error if it didn't match what was in the path, so I'm not sure why anyone would bother adding it. This is likely another question for someone like @droberts195

lcawl: I just did a quick test and this is the error I get if the job_id in the body differs from the job_id in the path:
> {
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "Inconsistent job_id; 'total-requests3' specified in the body differs from 'total-requests2' specified as a URL argument"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "Inconsistent job_id; 'total-requests3' specified in the body differs from 'total-requests2' specified as a URL argument"
  },
  "status" : 400
}

Therefore unless someone else knows why you'd ever want to use the value in the body, I'd recommend putting something like this in the description: "If specified, this value must match the path parameter; otherwise, an error occurs."

droberts195: Allowing the job ID in the body was just to make it easier for people who want to store configs as a single JSON document in some other location to then add into multiple clusters. That JSON document can then include the job ID. It has to be copied into the URL to add the config, but doesn’t also need to be removed from the JSON document.

specification/ml/put_job/MlPutJobRequest.ts
droberts195: I thought `ignore_throttled` would only be supported as part of `indices_options` in the body. I am on mobile so can’t easily confirm now. Can someone from @elastic/ml-core check the code and confirm whether indices options are only supported in the body?

specification/ml/put_job/MlPutJobRequest.ts
```diff
@@ -85,6 +89,10 @@ export interface Request extends RequestBase {
      * A list of job groups. A job can belong to no groups or many.
      */
     groups?: string[]
+    /**
+     * The identifier for the anomaly detection job. This identifier can contain lowercase alphanumeric characters (a-z and 0-9), hyphens, and underscores. It must start and end with alphanumeric characters.
```
lcawl, line 93: Based on Dave's feedback, I recommend setting the description to something like this:
```suggestion
     * If specified, this value must match the path parameter; otherwise, an error occurs.
```

specification/ml/put_job/MlPutJobRequest.ts
lcawl: Thanks for the info!  I've added a suggestion for the description.

specification/ml/put_job/MlPutJobRequest.ts
droberts195: Actually, this is on the job, not the datafeed so indices options don’t make sense at all. The indices options are in the body of the datafeed config.

specification/ml/_types/Detector.ts
lcawl: IMO it doesn't hurt to add this to the auto-generated docs and is probably harmless enough that it's not worth trying to introduce exceptional handling.

specification/ml/_types/Detector.ts
```diff
@@ -50,7 +54,7 @@ export class Detector {
   /**
    * The analysis function that is used. For example, `count`, `rare`, `mean`, `min`, `max`, or `sum`.
    */
```
lcawl, line 56:
```suggestion
   * @server_default metric
   */
```

sethmlarson:
Closing this as the changes are out of date.

PR #1285, opened 2022-01-20
Fix chained transform definition

swallez:
Follow-up to #1272

Chained transform is not a union but a simple array.

Backporting also to 7.17 even if #1272 was not as this field definition needs to be fixed.


philkra (approved):
LGTM

delvedor (approved):

PR #1295, opened 2022-01-20
Allow to override a property type

delvedor:
As titled.

Related: https://github.com/elastic/elasticsearch-specification/pull/1283


sethmlarson (approved):
I'm in favor but don't know if there are consequences for strongly typed langs. Deferring to @swallez here :)

sethmlarson (approved):
LGTM, thanks!

compiler/src/steps/read-definition-validation.ts
```diff
@@ -47,31 +47,77 @@ export default async function readDefinitionValidation (model: model.Model, json
           }
 
           const readProperty = type.properties.find(p => p.name === property.name) as model.Property
-          // in the future we might want to have a different type between read and write defintions
-          // but let's address it when it will happen
-          if (!deepEqual(readProperty.type, property.type)) {
+
+          if (property.type.kind === 'union_of' && property.type.items.some(contains(readProperty.type))) {
```
swallez, line 51: Will this allow arbitrary unions as long as one of its members is the read property, e.g. `ReadFoo | Bar`?

Should we rather test that all `unwrap`'ed union items are `readProperty.type`?

delvedor: This allows the read property to be a single element of the parent union.
Eg:
```ts
export class Foo {
  bar: string | string[]
}

export class FooRead implements OverloadOf<Foo> {
  bar: string[]
}
```

swallez: Ah right, I didn't read the relation check in the right direction.

swallez (approved):
LGTM

PR #1312, opened 2022-01-24
Add UpdateAction for bulk requests

swallez:
Adds a formal definition for the "document" part of the `update` bulk operation, similar to a regular document update request.

This also removes the generic parameter of the top-level `BulkRequest` (they're kept in invididual operations), considering that operations of a bulk request can have widely different unrelated types. The generic parameters are kept though in individual operation definitions.

Reported in https://github.com/elastic/elasticsearch-java/issues/81

### DIFF
--- a/specification/_global/bulk/BulkRequest.ts
+++ b/specification/_global/bulk/BulkRequest.ts
@@ -26,8 +26,9 @@ import {
   WaitForActiveShards
 } from '@_types/common'
 import { Time } from '@_types/Time'
-import { OperationContainer } from './types'
+import { OperationContainer, UpdateAction } from './types'
 import { SourceConfigParam } from '@global/search/_types/SourceFilter'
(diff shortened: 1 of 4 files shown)


specification/_global/bulk/BulkRequest.ts
```diff
@@ -36,7 +37,7 @@ import { SourceConfigParam } from '@global/search/_types/SourceFilter'
  * @doc_id docs-bulk
  *
  */
-export interface Request<TSource> extends RequestBase {
```
delvedor, line 39: I don't think we should remove the generic. It's true that bulk can send widely different documents, but it can also send the same "structure" over and over.
I would keep it and let language generators add a default value for the generic.

swallez: Good point, that makes sense. I added a commit that restores them (and adds the `TPartialDocument` for the new `UpdateAction`)

delvedor (approved):

PR #1313, opened 2022-01-24
Changes back indices.create:alias to a simple dictionary

swallez:
Follow-up to #1276 - reverts `indices.create:alias` to be a simple dictionary as it doesn't accept an array value.


delvedor (approved):
I'm not able to find the test that suggested we had to support the array value.

LGTM

PR #1320, opened 2022-01-25
Make MultiSearchItem.status optional

swallez:
The `MultiSearchItem.status` property is always returned in multisearch responses but is missing from multisearch-template responses.

This should be considered as an inconsistency/bug in the ES API, and this PR fixes it by making `status` optional rather than creating a separate type for multisearch-template responses.

Reported in https://github.com/elastic/elasticsearch-java/issues/110


philkra (approved):
LGTM, I closed https://github.com/elastic/elasticsearch-specification/pull/1178 as addresses the same missing property but differently

Anaethelion (approved):

swallez:
Thanks @philkra. Re-reading my comments in #1178 I realize that I only looked at the multisearch response and not the multisearch _template_ response, because I (wrongly) assumed that they were the same 🤷

swallez:
Opened an issue in Elasticsearch: https://github.com/elastic/elasticsearch/issues/83029

PR #1335, opened 2022-01-25
Add missing fields to SearchTemplateResponse

swallez:
Add missing fields to `SearchTemplateResponse` so that it has the same fields as `SearchResponse`.

Reported in https://github.com/elastic/elasticsearch-java/issues/89 and [the docs says](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#run-templated-search) _"The response uses the same properties as the search API's response"_.


sethmlarson (approved):
LGTM

PR #1323, opened 2022-01-25
Adds cat data frame analytics API spec

szabosteve:
## Overview

This PR adds missing fields to the [cat data frame analytics API](https://www.elastic.co/guide/en/elasticsearch/reference/master/cat-dfanalytics.html).

PR #1351, opened 2022-01-26
Fix cat api usage

delvedor:
Cat apis have a set of common query parameters that are defined [here](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_spec_utils/behaviors.ts#L86-L132). You can override them, but there is no need to redefine every property if the type does not change.

Related: https://github.com/elastic/elasticsearch-specification/pull/1304 https://github.com/elastic/elasticsearch-specification/pull/1323


szabosteve:
Thanks, @delvedor! I replaced #1323 with https://github.com/elastic/elasticsearch-specification/pull/1355. I think the CI will pass on my PR after yours is merged.

szabosteve (approved):

lcawl (approved):
Thanks, this is good to know!

swallez (approved):
LGTM

PR #1359, opened 2022-01-27
Fix security apis

delvedor:
As titled.

The APIs that are still failing should be fixed once https://github.com/elastic/clients-flight-recorder/pull/941 is merged.


philkra (approved):
LGTM

PR #1366, opened 2022-01-27
Run validate-pr action only if the pr is not from a fork

delvedor:
As titled. Otherwise the action will fail due to permissions.


philkra (approved):
LGTM

PR #1393, opened 2022-02-02
Missing migration types

philkra:
added and validated the endpoints:
* `migration.get_feature_upgrade_status`
* `migration.post_feature_upgrade`

### DIFF
--- a/specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
+++ b/specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
@@ -23,7 +23,7 @@ import { ErrorCause } from '@_types/Errors'
 export class Response {
   body: {
     features: MigrationFeature[]
-    migration_status: string
+    migration_status: MigrationStatus
   }
 }
 
@@ -37,7 +37,7 @@ export enum MigrationStatus {
 export class MigrationFeature {
   feature_name: string
   minimum_index_version: VersionString
-  migration_status: string
+  migration_status: MigrationStatus
   indices: MigrationFeatureIndexInfo[]
 }
 
--- a/specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
+++ b/specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
@@ -26,6 +26,13 @@ export class Response {
   }
 }
 
+export enum MigrationStatus {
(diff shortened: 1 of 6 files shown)


specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Indices, VersionString } from '@_types/common'
+
+export class Response {
+  body: {
+    features: MigrationFeature[]
+    migration_status: string
```
sethmlarson, line 25: Looks like this field is based on an enum, should we use one here too?

https://github.com/elastic/elasticsearch/blob/26c86871fc091900952e88e252c36fbfedf8d5fa/server/src/main/java/org/elasticsearch/action/admin/cluster/migration/GetFeatureUpgradeStatusResponse.java#L106

specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+export class MigrationFeature {
+  feature_name: string
+  minimum_index_version: VersionString
+  migration_status: string
+  indices: Indices
```
sethmlarson, line 33: From this XContent builder https://github.com/elastic/elasticsearch/blob/26c86871fc091900952e88e252c36fbfedf8d5fa/server/src/main/java/org/elasticsearch/action/admin/cluster/migration/GetFeatureUpgradeStatusResponse.java#L298 I think the structure for `indices` is:

```ts
export class MigrationFeatureIndexInfo {
  index: IndexName
  version: VersionString
  failure_cause?: ErrorCause
}
```

and then `MigrationFeature.indices` can be `MigrationFeatureIndexInfo[]`.

sethmlarson (changes requested):
Two comments:

specification/migration/get_feature_upgrade_status/GetFeatureUpgradeStatusResponse.ts
philkra: Thank you for elaborating, addressed the feedback in https://github.com/elastic/elasticsearch-specification/pull/1393/commits/794f0f4f2b0903dea03ebee6f794163aad6c46bf

sethmlarson (approved):

PR #1397, opened 2022-02-04
Remove generic on async search status response

stevejgordon:
As titled, this removes the generic type on the status response for async search which is not used in the body. 

Also, `completion_status` should be optional as it's only returned for non-running searches.

@swallez Should we backport to 7.17?

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -5727,7 +5727,7 @@ export interface AsyncSearchStatusRequest extends RequestBase {
   id: Id
 }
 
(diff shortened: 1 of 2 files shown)


swallez:
> @swallez Should we backport to 7.17?

No, as this changes the type signature, which impacts endpoint generation. If users raise issues about it, we'll have to tell them to use a dumb type (like `Void` in Java).

swallez (approved):
LGTM.

And as discussed, it should not be backported to 8.0 since the [last BC for 8.0.0](https://github.com/elastic/dev/issues/1923) has been built already.

PR #1418, opened 2022-02-09
Updates cat trained models API spec

szabosteve:
## Overview

This PR updates the cat trained model API spec.


delvedor (changes requested):
We are already tracking most cat query parameters here: https://github.com/elastic/elasticsearch-specification/blob/b3e872e566beba79bd34a8d8b54ef857e1278f1a/specification/_spec_utils/behaviors.ts#L91-L132

There is no need to type them again unless you are changing the type :)

szabosteve:
Thanks @delvedor for the info and for the review. I updated my PR accordingly.

specification/cat/ml_trained_models/CatTrainedModelsRequest.ts
```diff
@@ -40,7 +40,11 @@ export interface Request extends CatRequestBase {
   query_parameters: {
     allow_no_match?: boolean
     bytes?: Bytes
+    format?: string
```
delvedor, line 43:
```suggestion
```

You can remove this one, as it's already defined here: https://github.com/elastic/elasticsearch-specification/blob/b3e872e566beba79bd34a8d8b54ef857e1278f1a/specification/_spec_utils/behaviors.ts#L91-L132 and you aren't updating the type.

specification/cat/ml_trained_models/CatTrainedModelsRequest.ts
```diff
@@ -40,7 +40,11 @@ export interface Request extends CatRequestBase {
...
+    h?: CatTrainedModelsColumns
+    s?: CatTrainedModelsColumns
     from?: integer
     size?: integer
+    v?: boolean
```
delvedor, line 48:
```suggestion

```

Same here.

delvedor (approved):

PR #1419, opened 2022-02-10
Rename internal namespace

delvedor:
In light of the new Elasticsearch’s `internal` namespace, we must rename our `internal` namespace.
Given that our namespace was holding primitive values, I propose to rename it to `_primitive`.

Related: https://github.com/elastic/elasticsearch-specification/pull/1417

### DIFF
--- a/compiler/src/model/utils.ts
+++ b/compiler/src/model/utils.ts
@@ -82,7 +82,7 @@ export function modelType (node: Node): model.ValueOf {
         kind: 'instance_of',
         type: {
           name: 'boolean',
-          namespace: 'internal'
+          namespace: '_primitive'
         }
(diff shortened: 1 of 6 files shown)


swallez (commented):
Naming is hard. What about `_builtins` instead of `_primitive`? [Primitive types](https://en.wikipedia.org/wiki/Primitive_data_type) usually denote the very basic types from which all others derive, namely integers, floating point numbers and booleans. Our built in types also include strings and byte arrays and may be extended with even less primitive types in the future.

Looks good otherwise, and using an underscore-prefixed namespace will avoid any future conflict.

Anaethelion (approved):

PR #1431, opened 2022-02-10
Add missing properties to SQL query API

stevejgordon:
Adds several missing properties to the SQL query API. `wait_for_completion_timeout` is the most crucial as it's required to run asynchronous SQL queries. I added the others based on the documentation for completeness.

I've marked for backport to 8.0 as this is fairly crucial for a working API in the 8.0 clients and is all additive, so not technically breaking. :-)


swallez (approved):
LGTM. Addivite fixes like this one should be backported down to 7.17. There will be a number of 7.17 patch releases as there will be no new 7.x minor release.

PR #1436, opened 2022-02-11
Fix indices.field_usage_stats

delvedor:
The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/field-usage-stats.html) says it's optional, but the [rest-api-spec](https://github.com/elastic/elasticsearch/blob/e851efbaa1d0048909a96bdb942d6a2ebb462504/rest-api-spec/src/main/resources/rest-api-spec/api/indices.field_usage_stats.json#L15-L29) allows only url templates with the index defined.


Anaethelion (approved):
LGTM

PR #1443, opened 2022-02-11
Fix indices.simulate_template

philkra:
* removed three instances of `UserDefinedValue` in favour of once the correct [data_stream def](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-simulate-template.html#simulate-template-api-request-body) and twice for `IndexSettings`.
* moved shared type `IndexTemplate` into `indices._types`

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -9360,6 +9360,28 @@ export interface IndicesIndexState {
   data_stream?: DataStreamName
 }
 
+export interface IndicesIndexTemplate {
+  index_patterns: Names
+  composed_of: Name[]
+  template?: IndicesIndexTemplateSummary
+  version?: VersionNumber
+  priority?: long
+  _meta?: Metadata
+  allow_auto_create?: boolean
+  data_stream?: IndicesIndexTemplateDataStreamConfiguration
+}
+
+export interface IndicesIndexTemplateDataStreamConfiguration {
+  hidden?: boolean
+  allow_custom_routing?: boolean
+}
+
+export interface IndicesIndexTemplateSummary {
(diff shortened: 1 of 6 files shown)


philkra:
@swallez is this creating a breaking change in 7.x and 8.x in the Java client?

swallez:
This is a breaking change as the type of `data_streams` changes from a dict of raw json to a "real" data structure (and BTW `allow_custom_routing` was introduced in 8.0).

Now there are some breaking changes that IMHO are acceptable, as we cannot wait for the next major release to fix/improve the API spec:
- in patch releases (e.g. 8.0.1), we can accept breaking changes only to fix bugs that make the client unusable. This includes completly wrong types and required fields that are actually optional.
- in minor releases (e.g. 8.1),we should allow breaking changes that refine types and improve the spec and the developer experience. We should avoid though large refactorings that move things around.

In all cases, breaking changes should be documented so that users can be aware of what may break when they upgrade. I'm currently finishing this doc production for the Java client, and this could be also generated from `schema.json`.

Coming back to this PR, the fix of `overlapping` is fine to be backported down to 7.17 (this is a bug AFAIU) but the change to `data_stream?: IndexTemplateDataStreamConfiguration` should be backported only to 8.1 since it's the improvement of a definition that isn't nice but works, and thus doesn't justify a breaking change in patch releases.

philkra:
Thank you for sharing your thoughts, I agree with this statement:

> in minor releases (e.g. 8.1),we should allow breaking changes that refine types and improve the spec and the developer experience. We should avoid though large refactorings that move things around.

I will open a separate PR for 7.17

delvedor (approved):

Issue #928, opened 2021-10-21
How to represent API endpoints returning 404 (or other non-200) with a valid response?


swallez:
In https://github.com/elastic/elasticsearch-java/issues/36#issuecomment-948558659 I found that `clear_scroll` returns a `404` status code with a valid response when the `scroll_id` was not found.

Its successor `close_point_in_time` has the same behavior.

This is not documented in the docs ([scroll](https://www.elastic.co/guide/en/elasticsearch/reference/7.15/clear-scroll-api.html), [pit](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html#close-point-in-time-api)) nor in the json-spec. How can we represent this in the TS spec?

Are these snowflakes or do we know other APIs that behave similarly?

swallez:
Another nasty one, the [delete document API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html):
- it returns 404 with a valid response body if the index exists but doesn't contain a document with that id
- it returns 404 with an error response body if the index doesn't exist.

PR #1468, opened 2022-02-15
Track multiple responses per API

delvedor:
Proposal: https://gist.github.com/delvedor/eccf0936048dda2ecf0708156a8a726c

Fixes: https://github.com/elastic/elasticsearch-specification/issues/928


specification/cluster/remote_info/ClusterRemoteInfoResponse.ts
```diff
@@ -21,10 +21,11 @@ import { DictionaryResponseBase } from '@_types/Base'
 import { integer, long } from '@_types/Numeric'
 import { Time } from '@_types/Time'
 
-export class Response extends DictionaryResponseBase<
-  string,
-  ClusterRemoteInfo
-> {}
+export class Response {
+  '200': {
+    body: DictionaryResponseBase<string, ClusterRemoteInfo>
```
delvedor, line 26: Can we remove `DictionaryResponseBase` from the spec and always use `Dictionary`?

```suggestion
    body: Dictionary<string, ClusterRemoteInfo>
```

Anaethelion: If we are not extending anymore and we directly specify the body type in the property of the response we don't need the `DictionaryResponseBase` which is only there to be extended from.

In line with the rest of the changes I'd say yes, we can.

Anaethelion:
LGTM

specification/cluster/remote_info/ClusterRemoteInfoResponse.ts
stevejgordon: The .NET client will require such responses to derive from a `DictionaryResponseBase` base class which includes custom JSON formatting and a backing dictionary which derived type expose through a meaningful name. I don't think this change is too much a problem though as my generator could likely infer that for any responses where the body is a Dictionary type. I'm not currently generating these responses for the v8 alpha as they're on the long TODO list!

PR #1479, opened 2022-02-15
Fix ml.get_trained_models

philkra:
as titled.

### DIFF
--- a/specification/_types/aggregations/pipeline.ts
+++ b/specification/_types/aggregations/pipeline.ts
@@ -19,7 +19,7 @@
 
 import { Sort } from '@_types/sort'
 import { Dictionary } from '@spec_utils/Dictionary'
-import { Name, Field, EmptyObject } from '@_types/common'
+import { Name, Field, EmptyObject, IndexName } from '@_types/common'
 import { integer, double, float } from '@_types/Numeric'
 import { Script } from '@_types/Scripting'
 import { Aggregation } from './Aggregation'
@@ -162,6 +162,7 @@ export class InferenceConfigContainer {
   regression?: RegressionInferenceOptions
   /** Classification configuration for inference. */
   classification?: ClassificationInferenceOptions
+  ner?: NerInferenceOptions
 }
 
 export class RegressionInferenceOptions {
@@ -190,6 +191,25 @@ export class ClassificationInferenceOptions {
   results_field?: string
   /** Specifies the field to which the top classes are written. Defaults to top_classes. */
   top_classes_results_field?: string
+}
+
+export class NerInferenceOptions {
+  tokenization: NerInferenceTokenization
+  vocabulary: NerInferenceVocabulary
+}
+
+export class NerInferenceTokenization {
+  bert: NerInferenceTokenizationBert
+}
+export class NerInferenceTokenizationBert {
+  do_lower_case: boolean
+  max_sequence_length: integer
+  truncate: string
+  with_special_tokens: boolean
+}
+
+export class NerInferenceVocabulary {
+  index: IndexName
 }
 
 export class MaxBucketAggregation extends PipelineAggregationBase {}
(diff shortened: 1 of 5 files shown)


specification/_types/aggregations/pipeline.ts
```diff
@@ -162,6 +162,7 @@ export class InferenceConfigContainer {
   regression?: RegressionInferenceOptions
   /** Classification configuration for inference. */
   classification?: ClassificationInferenceOptions
+  ner?: NerInferenceOptions
```
lcawl, line 165: The inclusion of this ner option here is a bit confusing to me, since IMO it implies NER is supported in the inference bucket aggregation, which AFAIK is not true per the docs https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-pipeline-inference-bucket-aggregation.html.

lcawl (commented):
I added some comments, but think it should also be reviewed by @davidkyle

specification/_types/aggregations/pipeline.ts
lcawl: I think these NLP properties *are* supported in [inference processors](https://www.elastic.co/guide/en/elasticsearch/reference/master/inference-processor.html) in the create pipeline APi and in the [create trained model API](https://www.elastic.co/guide/en/elasticsearch/reference/master/put-trained-models.html), so iMO it makes more sense to add these in https://github.com/elastic/elasticsearch-specification/blob/main/specification/ingest/_types/Processors.ts or https://github.com/elastic/elasticsearch-specification/blob/main/specification/ml/_types/TrainedModel.ts

specification/ml/get_trained_models/MlGetTrainedModelRequest.ts
```diff
@@ -82,6 +82,10 @@ export interface Request extends RequestBase {
      * none. When supplied, only trained models that contain all the supplied
      * tags are returned.
      */
-    tags?: string
+    tags?: string | string[]
+    /**
+     * @since 7.17.0
```
lcawl, line 87: Looks like this is deprecated per https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ml.get_trained_models.json#L46
```suggestion
     * @since 7.17.0
     * @deprecated
```

davidkyle: The entire API and `include_model_definition` parameter were first added in 7.6: https://github.com/elastic/elasticsearch/pull/49052/

`include_model_definition` was later deprecated in 7.10 by https://github.com/elastic/elasticsearch/pull/62834

It looks like  theJSON spec was not correctly updated at the same time hence the confusion.

specification/ml/get_trained_models/MlGetTrainedModelRequest.ts
```diff
@@ -82,6 +82,11 @@ export interface Request extends RequestBase {
      * none. When supplied, only trained models that contain all the supplied
      * tags are returned.
      */
-    tags?: string
+    tags?: string | string[]
+    /**
+     * @since 7.13.0
```
davidkyle, line 87:
```suggestion
     * @since 7.6.0
```

specification/ml/get_trained_models/MlGetTrainedModelRequest.ts
```diff
@@ -82,6 +82,11 @@ export interface Request extends RequestBase {
...
+     * @deprecated 7.13.0
```
davidkyle, line 88:
```suggestion
     * @deprecated 7.10.0
```

benwtrent:
@philkra take a look at:  https://github.com/elastic/elasticsearch-specification/pull/1600

I think it addresses your issues here.

JoshMock:
This spec change hasn't been touched in over a year. Can we close?

davidkyle:
@JoshMock yes we can close this is hopelessly outdated. 

I will review which changes should be carried forward and create a new PR if necessary.

PR #1517, opened 2022-03-04
Fields in FieldCapabilities is required

swallez:
The `fields` property in `FieldCapabilitiesRequest` is required (see [API doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html#search-field-caps-api-query-params)).

Having it optional leads to all properties to be optional, causing the generation of a shortcut method in the Java client that always fails (reported in https://github.com/elastic/elasticsearch-java/issues/145).

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -346,7 +346,7 @@ export interface FieldCapsRequest extends RequestBase {
   index?: Indices
   allow_no_indices?: boolean
   expand_wildcards?: ExpandWildcards
-  fields?: Fields
+  fields: Fields
   ignore_unavailable?: boolean
   include_unmapped?: boolean
   filters?: string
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):

sethmlarson:
Are there tests that expect `fields` to be optional?

swallez:
Four recordings are missing `fields`, hence the validation errors above. Now when this property is missing, ES 7.17 returns an error:

```
{
  "error" : {
    "type" : "illegal_argument_exception",
    "reason" : "specified fields can't be null or empty"
  },
  "status" : 400
}
```

These come from Nest tests. Maybe Nest sets `fields` to `*` when unspecified? Should we add `fields` to these recordings?

sethmlarson:
@swallez Yeah let's update the NEST tests to `*`.

PR #1529, opened 2022-03-07
Schema version 2

delvedor:
Phase 2 of: #1468 

Proposal: [gist.github.com/delvedor/eccf0936048dda2ecf0708156a8a726c](https://gist.github.com/delvedor/eccf0936048dda2ecf0708156a8a726c)

This pr generates a new file: `schema-v2.json`. It has the new structure outlined in the proposal above. A new compiler step has been added to generate the schema v1 as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 output/dangling-types/dangling.csv linguist-generated=true
 output/schema/schema.json linguist-generated=true
+output/schema/schema-v2.json linguist-generated=true
 output/schema/import-type-graph.json linguist-generated=true
 output/schema/import-namespace-graph-compact.json linguist-generated=true
 output/schema/import-namespace-graph-expanded.json linguist-generated=true
--- a/compiler/src/compiler.ts
+++ b/compiler/src/compiler.ts
@@ -59,7 +59,7 @@ export default class Compiler {
     }
 
     await writeFile(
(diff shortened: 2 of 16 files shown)


swallez:
After some more thinking on the proposal, I see a potential issue by considering multiple response definitions as a flat list of equivalent definitions.

Client generators should be able to distinguish the nominal response from non-nominal error responses to represent API endpoints as function/methods that use the language's idiomatic representation of success and failure. That will be function result/exception in Java, a `Result` that can be either success or failure in Rust, the `result, error` result pair in Go, etc.

A way to address that is to add a flag on response definitions indicating if they're nominal or not, e.g. `nominal: true`. It doesn't have to be an enum as languages have a concept of success or failure, but nothing inbetween ("it kinda worked, but not completely").

Now there's also another issue with multiple responses in each category. For example, what should the response type look like for an endpoint that has several nominal responses? Does this introduce a union at a higher level, or should we merge all nominal property responses in a single type?

TL;DR: I'm wondering if we should limit ourselves to only one nominal response (using optional properties if needed) with multiple status codes and encode this constraint in the schema. Errors may not need that constraint as languages often have the ability to return a set of unrelated errors.

delvedor:
Superseded by https://github.com/elastic/elasticsearch-specification/pull/1536.

PR #1534, opened 2022-03-14
[ML] update put trained model vocab spec to allow merges

benwtrent:
With https://github.com/elastic/elasticsearch/pull/84777, the put trained model vocab can now accept an optional `merges` field that is a string array. This field is required for `roberta` tokenizers, but not for `mpnet` or `bert`.

### DIFF
--- a/specification/ml/put_trained_model_vocabulary/MlPutTrainedModelVocabularyRequest.ts
+++ b/specification/ml/put_trained_model_vocabulary/MlPutTrainedModelVocabularyRequest.ts
@@ -44,6 +44,7 @@ export interface Request extends RequestBase {
 
     /**
      * The optional model merges if required by the tokenizer.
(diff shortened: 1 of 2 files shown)


specification/ml/put_trained_model_vocabulary/MlPutTrainedModelVocabularyRequest.ts
```diff
@@ -41,5 +41,10 @@ export interface Request extends RequestBase {
      * The model vocabulary, which must not be empty.
      */
     vocabulary: string[]
+
+    /**
+     * The optional model merges if required by the tokenizer.
+     */
```
sethmlarson, line 47: Can we add `@since 8.2.0` since this is new in 8.2.0?

sethmlarson (approved):

PR #1536, opened 2022-03-16
Rollback #1468

delvedor:
Hey folks! After a long and productive discussion during the ES spec sync we decided to roll partially back the change that introduced the status codes and move to the following format:

```ts
export class Response {
  body: SuccessResponseDefinition
  failures: [{
    statusCode: 404
    body: NonStandardFailure
  }]
}
```

This will make the life of code generators easier as well as the docs team, where users will have the successful case, the standard error, and one or more optional non-standard failures. Furthermore, this change won't introduce a breaking change in the specification schema, as it's an additive change.

This pr only partially everts the spec to the previous format, a subsequent pr will introduce the `failures` support. I said partially because it keeps the improvements done meanwhile, such as the removal of `DictionaryResponseBase`.

Related: #928
Supersedes: #1529

### DIFF
--- a/compiler/src/model/build-model.ts
+++ b/compiler/src/model/build-model.ts
@@ -313,34 +313,28 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
         body: { kind: 'no_body' }
       }
 
-      for (const status of declaration.getMembers()) {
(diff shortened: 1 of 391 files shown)


swallez (approved):
LGTM

sethmlarson (approved):

PR #1541, opened 2022-03-16
Add new fields to ML GET Trained Model Stats response

davidkyle:
https://github.com/elastic/elasticsearch/pull/84628 added some new fields to the ML GET Trained model stats response. Here the types are updated with those new fields

### DIFF
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -55,40 +55,44 @@ export class TrainedModelStats {
 
 export class TrainedModelDeploymentStats {
   /** The detailed allocation status for the deployment. */
-  allocation_status: TrainedModelDeploymentAllocationStatus
+  allocation_status?: TrainedModelDeploymentAllocationStatus
   /** The sum of `error_count` for all nodes in the deployment. */
-  error_count: integer
+  error_count?: integer
   /** The sum of `inference_count` for all nodes in the deployment. */
-  inference_count: integer
+  inference_count?: long
   /** The number of threads used by the inference process. */
-  inference_threads: integer
+  inference_threads?: integer
   /** The unique identifier for the trained model. */
   model_id: Id
   /** The number of threads used when sending inference requests to the model. */
-  model_threads: integer
+  model_threads?: integer
   /** The deployent stats for each node that currently has the model allocated. */
   nodes: TrainedModelDeploymentNodesStats
-  /** The sum of `peak_throughput_per_minute` for all nodes in the deployment. @since 8.2.0 */
-  peak_throughput_per_minute: integer
+  /**
+   * The sum of `peak_throughput_per_minute` for all nodes in the deployment.
+   * @since 8.2.0
+   *
+   */
+  peak_throughput_per_minute: long
   /** The number of inference requests that can be queued before new requests are rejected. */
-  queue_capacity: integer
+  queue_capacity?: integer
   /**
    * The sum of `rejected_execution_count` for all nodes in the deployment.
    * Individual nodes reject an inference request if the inference queue is full.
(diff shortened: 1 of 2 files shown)


specification/ml/_types/TrainedModel.ts
```diff
@@ -121,6 +123,11 @@ export class TrainedModelSizeStats {
 export class TrainedModelDeploymentNodesStats {
   /** The average time for each inference call to complete on this node. */
   average_inference_time_ms: double
+  /**
+   * The average time for each inference call to complete on this node
+   * in the last minute. @Since 8.2.0
```
sethmlarson, line 128: Annotation should be `@since 8.2.0`. We typically put these on their own line too.

specification/ml/_types/TrainedModel.ts
```diff
@@ -121,6 +123,11 @@ export class TrainedModelSizeStats {
...
+   */
+  average_inference_time_ms_last_minute: double
```
sethmlarson, line 130: [From the PR](https://github.com/elastic/elasticsearch/pull/84628/files#diff-ef1e72f9ddf1f055b78b94a380131fe4d3c81a7e1926c9801616a25078a87011R272) this field looks optional?

specification/ml/_types/TrainedModel.ts
```diff
@@ -144,6 +151,8 @@ export class TrainedModelDeploymentNodesStats {
   node: DiscoveryNode
   /** The number of inference requests queued to be processed. */
   number_of_pending_requests: integer
+  /** The peak number of requests processed in a 1 minute period. @Since 8.2.0 */
```
sethmlarson, line 154: Same as above.

specification/ml/_types/TrainedModel.ts
```diff
@@ -152,6 +161,8 @@ export class TrainedModelDeploymentNodesStats {
   start_time: long
   /** The number of inference requests that timed out before being processed. */
   timeout_count: integer
+  /** The number of requests processed in the last 1 minute. @Since 8.2.0 */
+  throughput_last_minute: integer
```
sethmlarson, line 165: Same as above.

specification/ml/_types/TrainedModel.ts
```diff
@@ -68,6 +68,8 @@ export class TrainedModelDeploymentStats {
   model_threads: integer
   /** The deployent stats for each node that currently has the model allocated. */
   nodes: TrainedModelDeploymentNodesStats
+  /** The sum of `peak_throughput_per_minute` for all nodes in the deployment. @since 8.2.0 */
+  peak_throughput_per_minute: integer
```
sethmlarson, line 72: From the PR the `peak_throughput_per_minute` and `throughput_per_minute` appear to be longs instead of integers.

sethmlarson (changes requested):
Thanks for this, some comments:

davidkyle:
> this field looks optional?

This is actually true of many of the fields as some are not reported if the deployment is starting up or in an error state. I've updated the new fields added in this PR and the existing ones that should be optional

davidkyle:
The validation check is failing I'm not sure why

```
Run actions/checkout@v[2](https://github.com/elastic/elasticsearch-specification/runs/5584050421?check_suite_focus=true#step:4:2)
  with:
    repository: elastic/clients-flight-recorder
    path: ./clients-flight-recorder
    ref: dev
    ssh-strict: true
    persist-credentials: true
    clean: true
    fetch-depth: 1
    lfs: false
    submodules: false
Error: Input required and not supplied: token
```

davidkyle:
@sethmlarson I've updated the PR can you take another look please

davidkyle:
Closing as now outdated

PR #1546, opened 2022-03-17
Responses cannot inherit

delvedor:
As we have discussed in the past, response definitions should not be able to inherit from other classes, as responses are a custom type that we use to "connect the dots".
Using inheritance here could cause unwanted side effects. The specification was still using inheritance in a very few cases that have been fixed here, this should not break anyone, even tho it's actually a breaking change of the spec output format.


philkra (approved):
LGTM

PR #1556, opened 2022-03-22
Use correct type for Significant Terms Agg

### LINKED ISSUE #1553: significant terms aggregation missing regex include method
Found in https://github.com/elastic/elasticsearch-java/issues/85

The definition of `SignificantTermsAggregation.include: string|string[]` does not represent the usual array leniency. Both variants have a specific meaning, resp. regexp and term list.

It should be typed as `TermsInclude`.

philkra:
closes #1553


swallez (approved):
LGTM

PR #1564, opened 2022-03-22
Extract SearchTemplateRequestBody in watcher

Anaethelion:
Use the duplicate in watcher instead of referencing the request directly.


delvedor (changes requested):
LGTM

Can you run `make contrib` and push again?

Anaethelion:
> LGTM
> 
> Can you run `make contrib` and push again?

Done!

delvedor (approved):

delvedor:
Adding the `breaking` label. This pr fixes the definition, but it will "break" the generated types.

PR #1570, opened 2022-03-23
Use a version placeholder in doc_ids table

delvedor:
As titled.
This will solve backporting issues, but unfortunately, we can't automatically add the version in the output json, otherwise, we will have again the backporting issue.
The URL updating needs to be demanded to code generators.


lcawl:
Depending on where this is resolved, {version} might be a problematic choice, since we already use that as an attribute for the full version.release.modifier in our documentation (e.g. see https://github.com/elastic/docs/blob/master/shared/versions/stack/8.1.asciidoc).  What we currently use in the documentation builds for just version.release is {branch}.  Should we use the same placeholders here?

delvedor:
@lcawl done! :)

lcawl (approved):
Thanks! LGTM

PR #1586, opened 2022-03-28
Allow quoted strings in comma-separated jstags

swallez:
Allows quoted strings in `@aliases` and other comma-separated jstags.

This is needed for `StorageType` where the empty string is an alias for the `fs` enum member (a follow-up PR will add it).

Asking for index defaults with `include_default=true` will return `""` for the storage type. When parsing this setting, Elasticsearch interprets the empty string as `fs`.


delvedor (commented):
LGTM

Can you run the linter?
```
cd compiler
npm run lint:fix
```
Should do the trick.

swallez:
Lint done!

PR #1589, opened 2022-03-29
Add default value for internal tags

swallez:
Adds an optional `default` configuration to internally tagged union definitions, e.g. `@variants internal tag='type' default='object'`

This is needed for:
* property mappings where `object` is the default (reported in [elasticsearch-java#45](https://github.com/elastic/elasticsearch-java/issues/45#issuecomment-978033425))
* analyzers where `custom` is the default (reported in [elasticsearch-java#216](https://github.com/elastic/elasticsearch-java/issues/216))


delvedor (approved):
LGTM

PR #1590, opened 2022-03-29
Add the 'classification' inference config for pipelines

sethmlarson:
I don't think this is the complete set of options, however it's better than not being able to specify a `classification` inference config.


philkra (approved):
LGTM

PR #1595, opened 2022-03-31
Remove generic property

stevejgordon:
The `GenericProperty` is not a true property type in ES, and was ported from NEST in error originally. This PR removes it from the spec.


delvedor (approved):
LGTM

PR #1600, opened 2022-04-01
[ML] update inference configuration options for storage and inference

benwtrent:
This updates the inference configuration options for:

 - classification
 - regression
 - nlp tasks

All configurations have a stored type and an update type. It may or may not be the case that the update version contains a smaller number of allowable fields.


delvedor (commented):
The changes look good, but there are a few type renames, which might be a problem for some clients.
Please @elastic/es-clients take a look.

stevejgordon:
@delvedor For the .NET client, the renames are fine as we're in alpha and haven't yet generated the ML APIs.

specification/ml/_types/inference.ts
```diff
@@ -0,0 +1,364 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Field } from '@_types/common'
+import { integer } from '@_types/Numeric'
+
+/** Inference configuration provided when storing the model config*/
+export class InferenceConfigContainer {
```
sethmlarson, line 24: Let's add a `/** @variants container */` decoration to this class

specification/ml/_types/inference.ts
```diff
@@ -0,0 +1,364 @@
...
+  fill_mask?: FillMaskInferenceUpdateOptions
+  /** Named entity recognition configuration for inference. */
+  ner?: NerInferenceUpdateOptions
+  /** Pass through configuration for inference. */
+  pass_through?: PassThroughInferenceUpdateOptions
+  /** Text embedding configuration for inference. */
+  text_embedding?: TextEmbeddingInferenceUpdateOptions
+}
+
+export class PipelineAggInferenceConfigUpdateContainer {
```
sethmlarson, line 251: Let's add a `/** @variants container */` decorator here too

sethmlarson, line 251: Let's add a `/** @variants container */` decorator here too

specification/ml/_types/inference.ts
```diff
@@ -0,0 +1,364 @@
...
+  /** Regression configuration for inference. */
+  regression?: RegressionInferenceUpdateOptions
+  /** Classification configuration for inference. */
+  classification?: ClassificationInferenceUpdateOptions
+}
+
+export class RegressionInferenceUpdateOptions {
```
sethmlarson, line 258: The `<ModelType>UpdateOptions` classes look similar to the `<ModelType>Options` classes, is there a reason we need both?

specification/ml/_types/inference.ts
```diff
@@ -0,0 +1,364 @@
...
+  num_top_classes?: integer
+  /** The tokenization options to update when inferring */
+  tokenization?: TokenizationConfigContainer
+  /** The field that is added to incoming documents to contain the inference prediction. Defaults to predicted_value. */
+  results_field?: string
+}
+
+// Update containers that may only allow a subset of the options
+
+export class NlpInferenceConfigUpdateContainer {
```
sethmlarson, line 236: Same as above with `/** @variants container */`

sethmlarson, line 236: Same as above with `/** @variants container */`

sethmlarson (changes requested):
Thanks for this!

It looks like you've got `<ModelType>Options` classes and `<ModelType>UpdateOptions` classes which mostly match. Instead you should switch these to use the [`@overloadOf` behavior & convention](https://github.com/elastic/elasticsearch-specification/blob/2e3212fe71bce86b306609fe22c266b560b2ef47/docs/behaviors.md#overloadof) where the classes that are used for requests can be named `<ModelType>Options` and the classes for responses are named: `<ModelType>OptionsRead`. Does this construction work for the case we're covering here?

Also going to tag @swallez since this likely will mean breaks in Java due to renaming, wanted your thoughts as well.

specification/ml/_types/inference.ts
benwtrent: For NLP, it is definitely needed as the tokenization options are different (its a subset). 

For regression and classification, I can make them the same object

benwtrent:
@swallez @sethmlarson I can adjust the naming back, but its "non-optimal" IMO as its not descriptive of what the type means.

 IDK how ts file location adjusts the generated types, but it would be good to have all the inference objects in the same place (logically).

benwtrent:
>  Does this construction work for the case we're covering here?

From what I can tell, `Overload` requires the all the properties to be satisfied. For all the NLP update cases, its actually a subset of the fields. So, it doesn't really fit here.

specification/_types/aggregations/pipeline.ts
```diff
@@ -157,41 +161,14 @@ export class InferenceAggregation extends PipelineAggregationBase {
   inference_config?: InferenceConfigContainer
 }
 
-export class InferenceConfigContainer {
+/** @variants container */
+class InferenceConfigContainer {
```
sethmlarson, line 165: This name is unfortunate as it's in the global namespace, would've been better as `InferenceAggregationInferenceConfigContainer` which would be a breaking change but allows using `InferenceConfigContainer` within `ml/_types/inference`.

swallez (approved):
Thanks for this PR, it's very nice to see contributions to the spec from the server dev team!

I'm not overly concerned by the renaming even if it causes a breaking change as there are a lot of data structure refinements that would require some code changes anyway. Also I may be mistaken but I haven't heard yet of Java client users going deep in ML APIs, so the impact should be fairly limited.

So LGTM from me once the other issues reported have been resolved.

delvedor (approved):

sethmlarson (approved):
Thanks for this!

Issue #1644, opened 2022-04-14
`manage_enrich` is missing from the available cluster privileges


ywangd:
## 🐛 Missing type

The `manage_enrich` cluster privilege is missing from:
* Response of the [Get builtin privileges API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-builtin-privileges.html)
* Request of the [Create/Update role API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html) where this privilege should be a candidate for value of the `cluster` field.

## Definition

IIUC, `manage_enrich` is missing from this definition https://github.com/elastic/elasticsearch-specification/blob/e29671f624cf75c88b3c60fe347f6c4b44082d3b/specification/security/_types/Privileges.ts#L41

Relates: https://github.com/elastic/elasticsearch/pull/85877

PR #1645, opened 2022-04-14
Update cluster privileges list

delvedor:
Fixes: #1644

### DIFF
--- a/specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
+++ b/specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
@@ -23,5 +23,6 @@ import { RequestBase } from '@_types/Base'
  * @rest_spec_name security.get_builtin_privileges
  * @since 7.3.0
  * @stability stable
+ * @cluster_privileges manage_enrich
  */
 export interface Request extends RequestBase {}
--- a/specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
+++ b/specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
@@ -23,6 +23,6 @@ import { RequestBase } from '@_types/Base'
  * @rest_spec_name security.get_builtin_privileges
  * @since 7.3.0
  * @stability stable
- * @cluster_privileges manage_enrich
+ * @cluster_privileges manage_security
  */
 export interface Request extends RequestBase {}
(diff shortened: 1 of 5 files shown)


specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
```diff
@@ -23,5 +23,6 @@ import { RequestBase } from '@_types/Base'
  * @rest_spec_name security.get_builtin_privileges
  * @since 7.3.0
  * @stability stable
+ * @cluster_privileges manage_enrich
```
ywangd, line 26: Forgive my ignorance: Not sure why `manage_enrich` is specifically needed here. Is this comment kinda a change history?

specification/security/put_role/SecurityPutRoleRequest.ts
```diff
@@ -34,7 +34,7 @@ import { Metadata, Name, Refresh } from '@_types/common'
  * @rest_spec_name security.put_role
  * @since 0.0.0
  * @stability stable
- * @cluster_privileges manage_security
+ * @cluster_privileges manage_enrich, manage_security
```
ywangd, line 37: Same as above.

ywangd (approved):
LGTM

The following comments are rather questions due to my lack of knowledge of the project. I appreciate if you could help me understand the changes.

specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesRequest.ts
delvedor: Happy to answer every question! In the issue you wrote:

> The manage_enrich cluster privilege is missing from:
> Response of the [Get builtin privileges API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-builtin-privileges.html)
> Request of the [Create/Update role API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html) where this privilege should be a candidate for value of the cluster field.

This change adds the `manage_enrich` to the two APIs. If you look at the `schema.json`, now the new privileges are listed there as well.

sethmlarson: @delvedor From my reading the enum value was only missing in the response itself, not as a `@cluster_privileges` annotation for this API so adding the value to the enum is sufficient. From the docs it seems that the privilege needed for this API is actually `manage_security`.

ywangd: Sorry for any confusion that may have been caused by my original issue. @sethmlarson is right. The cluster privilege requried for calling this API is `manage_security`. The response of this API should return a list of cluster privileges which should include the missing `manage_enrich` privilege. 

Btw: I am assuming the `@cluster_privileges` directive is to specify what cluster privilege is required for calling the API.

specification/security/put_role/SecurityPutRoleRequest.ts
ywangd: Based on the above, we should remove `manage_enrich` from here as well.

Issue #1643, opened 2022-04-14
Update User Profile APIs with changes to properties


sethmlarson:
- https://github.com/elastic/elasticsearch/pull/85797
- https://github.com/elastic/elasticsearch/pull/85856

PR #1649, opened 2022-04-15
Apply changes to UserProfile model

sethmlarson:
Closes https://github.com/elastic/elasticsearch-specification/issues/1643

### DIFF
--- a/specification/security/_types/UserProfile.ts
+++ b/specification/security/_types/UserProfile.ts
@@ -19,7 +19,8 @@
 
 import { Dictionary } from '@spec_utils/Dictionary'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
-import { User } from './User'
+import { Username, Name } from '@_types/common'
+import { Metadata } from '@_types/common'
 import { long } from '@_types/Numeric'
 import { SequenceNumber } from '@_types/common'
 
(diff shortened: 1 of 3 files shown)


specification/security/_types/UserProfile.ts
```diff
@@ -28,11 +29,19 @@ export class UserProfileHitMetadata {
   _seq_no: SequenceNumber
 }
 
+export class UserProfileUser {
```
delvedor, line 32: How about renaming this type to `UserData`?
```suggestion
export class UserData {
```

delvedor: Ignore the above, you can import `User` from `./types'`:
https://github.com/elastic/elasticsearch-specification/blob/151d0b70a4c514a7986bf8543d13998837b418f4/specification/security/_types/User.ts#L22-L29

sethmlarson: Unfortunately we can't use `User` here, there is no `enabled` property in this nested object. I can switch the name to `UserData`.

sethmlarson: Hmmm, using the name here removes our ability to use `UserData` in all of `security`, I think sticking with `UserProfileUser` is best here for that reason?

PR #1662, opened 2022-04-26
Add support for non exhaustive variants

swallez:
Elasticsearch plugins can add new implementations for a significant number of components, such as for example queries, aggregations and property types.

This PR adds an additional `@non_exhaustive` annotation (term borrowed from [Rust's equivalent feature](https://doc.rust-lang.org/reference/attributes/type_system.html)) that can be added to `@variants` types and enums (see below) to capture the fact that variants or enum members not present in the spec can exist, and clients should allow that.

This PR also adds the new `@non_exhaustive` annotation to the relevant types, found from the [plugin interface definitions](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/plugins/) in the ES code base.

Example: 

```ts
/**
 * @variants container
 * @non_exhaustive
 */
export class QueryContainer {
  ...
}
```

The following pluggable components were left out from this PR as they require a bit more investigation in their specification: highlighters, score functions, significance heuristics.

**enums:** for consistency this PR also changes `@open_enum` for open enumerations to `@non_exhaustive`. The enum metamodel isn't changed, and this annotation still sets the `isOpen` property.

### DIFF
--- a/compiler/src/model/metamodel.ts
+++ b/compiler/src/model/metamodel.ts
@@ -183,7 +183,7 @@ export class VariantBase {
    * be extended with plugins. If true, target clients should allow for additional variants
    * with a variant tag outside the ones defined in the spec and arbitrary data as the value.
    */
-  isOpen?: boolean
+  nonExhaustive?: boolean
 }
 
 export class ExternalTag extends VariantBase {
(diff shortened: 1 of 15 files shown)


sethmlarson (approved):
This is super great, love what's been done with `@open_enum` to make everything even more consistent!

PR #1673, opened 2022-05-03
Removed redundant date types

delvedor:
As titled.

We will verify if we have used the wrong date type in the various properties in a separate PR.


swallez (changes requested):
`Time` represents durations (see the number of `timeout`, `offset` and `interval` fields where it's used) and so cannot replace `EpochMillis`. That being said, there are also occurrences of `Time` that seem wrong, e.g. `BucketSummary.timestamp`.

PR #1674, opened 2022-05-03
Add the "es_quirk" annotation to capture snowflakes in ES behavior

swallez:
Adds a new `@es_quirk` jsdoc tag to capture weird snowflakes in the specification. This tag's value is a description of the quirk.

It is exported in `schema.json` as an `esQuirk?: string` property on type definition and properties. This allows code generators to track the quirks they have effectively implemented and fail if a new quirk appears in the schema that needs to be handled in the code generator.


swallez:
Added some docs.

sethmlarson (approved):

PR #1683, opened 2022-05-06
[ML] add new infer endpoint that replaces deployment/_infer

benwtrent:
In https://github.com/elastic/elasticsearch/pull/86361 the API was renamed, must be renamed here too.


specification/ml/infer_trained_model/MlInferTrainedModelRequest.ts
```diff
@@ -45,13 +45,14 @@ export interface Request extends RequestBase {
...
   body: {
     /**
-     * An array of objects to pass to the model for inference. The objects should contain a field matching your
-     * configured trained model input. Typically, the field name is `text_field`. Currently, only a single value is allowed.
+     * An array of objects to pass to the model for inference. The objects should contain a fields matching your
+     * configured trained model input. Typically, for NLP models, the field name is `text_field`.
+     * Currently, for NLP models, only a single value is allowed.
      */
-    docs: Dictionary<string, string>[]
+    docs: Dictionary<string, any>[]
```
sethmlarson, line 52: Instead of `any` we should use `UserDefinedValue` for our "any" type as it has special meaning in implementing languages.

specification/ml/_types/inference.ts
```diff
@@ -328,3 +348,79 @@ export class FillMaskInferenceUpdateOptions {
...
+
+export class TrainedModelInferenceFeatureImportance {
+  feature_name: string
+  importance?: double
+  classes?: TrainedModelInferenceClassImportance[]
+}
+
+export type PredictedValue = string | double | boolean | integer
+
+/** @variants container */
```
sethmlarson, line 378: I don't this class is intended to be a container, the "container" variant means properties are mutually exclusive: https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#container

Since the `Container` suffix is typically used for these cases, maybe we should use a different name here? `TrainedModelInferenceResult`?

sethmlarson (commented):
Thanks for this, two comments:

specification/ml/infer_trained_model/MlInferTrainedModelRequest.ts
benwtrent: ok, as long as it can capture pretty much any value :)

sethmlarson (approved):
LGTM, thanks!

PR #1692, opened 2022-05-11
Fix security/has_privileges

Anaethelion:
Path parameter cannot be null


delvedor (approved):
I'm wondering what prompted this change. Anyhow, the validator is happy, so I am as well!

PR #1713, opened 2022-05-17
Fix time-related types and properties

swallez:
Supersedes #1673 

This PR does a big overhaul of time handling in the API spec: it removes redundant time-related types that existed, and refines the available types to capture the precise representation sent or expected by Elasticsearch.

This is a breaking change, in two main areas:
- most of the time-related types have been renamed or removed, and new ones added
- a lot of time-related fields have been updated to reflect their actual semantics

Full validation fixes a number of errors and does not add new ones.


specification/_spec_utils/Stringified.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/**
+ * Some APIs will return values such as numbers also as a string (notably epoch timestamps). This behavior
+ * is used to capture this behavior while keeping the semantics of the field type.
+ *
+ * Depending on the target language, code generators can keep the union or remove it and leniently parse
+ * strings to the target type.
+ */
+export type Stringified<T> = T | string
```
delvedor, line 27: Why add this here and not in [`_spec_utils/behaviors.ts`](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_spec_utils/behaviors.ts)?
Also, if this is a behavior, you should add the `@behavior` tag.

delvedor (changes requested):
Amazing work!
A couple of notes:

If `Stringified` is a behavior, you should also update:
https://github.com/elastic/elasticsearch-specification/blob/80618757a29d8017316d4d40c21f77eee6c0d6e6/compiler/src/model/utils.ts#L55-L61

The typescript generator must be updated as well.

specification/_spec_utils/Stringified.ts
swallez: I initially considered it as a behavior, but it's not. Behaviors are interfaces whose TypeScript definition is essentially empty and are meant to capture things we cannot easily represent in TypeScript.

In this case, it's just a regular type, that code generators may or may not consider in a special way.

swallez:
Additional thoughts (cc @technige). We may consider a more generic way of defining values with units, as follows:

```ts
export type UnitSeconds = long
export type UnitMillis = long
export type UnitNanos = long
export type UnitFloatMillis = double

/** A timestamp in time units since the Epoch */
export type EpochValue<TimeUnit> = TimeUnit

/** A timestamp in millis */
export type EpochMillis = EpochValue<UnitMillis>

/** A timestamp in seconds */
export type EpochSeconds = EpochValue<UnitSeconds>

export type TimeSpanValue<Unit> = Unit
export type TimeSpanSeconds = TimeSpanValue<UnitSeconds>
export type TimeSpanMillis = TimeSpanValue<UnitMillis>
export type TimeSpanNanos = TimeSpanValue<UnitNanos>
export type TimeSpanFloatMillis = TimeSpanValue<UnitFloatMillis>
```

This allows capturing the essential nature of a type (e.g. it's an epoch-based timestamp) separately from its unit. Same would apply to byte sizes.

For strongly typed languages this "essential nature" may map to a single type (or even a built-in type such as `java.util.Time`) whereas dynamically typed languages may just follow the alias.

stevejgordon:
@swallez This is great and I appreciate the effort you've gone to here. Makes so much sense to standardise the types and their usage. I'm all for these changes. I much prefer the way `EpochValue<T>` way to simplify the modelling and make intent clearer. Dates are handled so differently across languages, it makes sense for each generator to transform to the most appropriate built-in or custom implementations. As this is a breaking change, what version do we expect this to land in? For .NET since we're not GA'd yet, I'd happily take these sooner rather than later but for other clients that could be more challenging.

swallez:
@stevejgordon thanks for the feedback! I'll go ahead and add the `EpochValue<T>`.

I'm wondering if we should go even further and unify the number and text representations of date/time. The generic parameter would then not be a unit but a format (that also defines its unit).

```ts
export type DateTime<Format> = Format

export type DateText = DateTime<FmtIsoOrMillis>
export type EpochMillis = DateTime<FmtMillis>
export type EpochSeconds = DateTime<FmtSeconds>


export type TimeSpan<Format> = Format

export type TimeSpanSeconds = TimeSpan<FmtSeconds>
export type TimeSpanMillis = TimeSpan<FmtMillis>
export type TimeSpanNanos = TimeSpan<FmtNanos>
export type TimeSpanFloatMillis = TimeSpan<FmtFloatMillis>
```

```ts
type DateTime = string | long
```

```ts
export 
export type DateTime = string | long
```

specification/_global/delete_by_query/DeleteByQueryRequest.ts
```diff
@@ -56,18 +56,18 @@ export interface Request extends RequestBase {
     preference?: string
     refresh?: boolean
     request_cache?: boolean
-    requests_per_second?: long
+    requests_per_second?: float
```
sethmlarson, line 59: Should `requests_per_second` be `long|float` instead of only `float`?

sethmlarson:
General naming question: `TimeSpan` could elicit thoughts of a representation of "X-Y" instead of a duration. ISO 8601 uses the word "Period" to define this concept, should we use that instead?

specification/_global/update_by_query/UpdateByQueryResponse.ts
```diff
@@ -36,7 +36,7 @@ export class Response {
     total?: long
     updated?: long
     version_conflicts?: long
-    throttled_millis?: ulong
-    throttled_until_millis?: ulong
+    throttled_millis?: long
```
sethmlarson, line 39: Should these be:

```ts
throttled_millis?: TimeSpanMillis
throttled_until_millis?: EpochMillis
```

specification/cluster/health/ClusterHealthResponse.ts
```diff
@@ -31,7 +31,7 @@ export class Response {
     /** The number of active primary shards. */
     active_primary_shards: integer
     /** The total number of active primary and replica shards. */
-    active_shards: integer
+    active_shards: double
```
sethmlarson, line 34: Is this actually a `double`? It's documented as an integer.

specification/eql/_types/EqlSearchResponseBase.ts
```diff
@@ -37,7 +37,7 @@ export class EqlSearchResponseBase<TEvent> {
   /**
    * Milliseconds it took Elasticsearch to execute the request.
    */
-  took?: integer
+  took?: long
```
sethmlarson, line 40: Should we use a time time for milliseconds here?

specification/ilm/_types/Phase.ts
```diff
@@ -17,14 +17,17 @@
...
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 import { RolloverConditions } from '@indices/rollover/types'
-import { integer } from '@_types/Numeric'
+import { integer, long } from '@_types/Numeric'
 
 export class Phase {
   actions?: Actions
-  min_age?: Time
+  /**
+   * @es_quirk output as a millis number in XPack usage stats, which cannot roundtrip with a TimeSpan as it requires a unit.
```
sethmlarson, line 28: Is this an actual quirk with Elasticsearch's API or is this due to the way we're choosing to implement this type?

specification/indices/_types/IndexSettings.ts
```diff
@@ -311,8 +312,8 @@ export class IndexSettingsAnalysis {
 }
 
 export class IndexSettingsTimeSeries {
-  end_time?: DateOrEpochMillis
-  start_time?: DateOrEpochMillis
+  end_time?: DateTime
```
sethmlarson, line 315: This property type should still be a union, maybe `DateTime | EpochMillis`?

specification/indices/recovery/types.ts
```diff
@@ -96,15 +98,15 @@ export class TranslogStatus {
...
   total_time_in_millis: EpochMillis
 }
 
 export class VerifyIndex {
-  check_index_time?: Time
-  check_index_time_in_millis: EpochMillis
-  total_time?: Time
-  total_time_in_millis: EpochMillis
+  check_index_time?: TimeSpan
+  check_index_time_in_millis: long
```
sethmlarson, line 107: Should all the `long` types being changed still be `EpochMillis`?

specification/ml/_types/Datafeed.ts
```diff
@@ -24,21 +24,26 @@ import { RuntimeFields } from '@_types/mapping/RuntimeFields'
...
+} from '@_types/Time'
 import { DiscoveryNode } from './DiscoveryNode'
 
 export class Datafeed {
   /** @aliases aggs */
   aggregations?: Dictionary<string, AggregationContainer>
   chunking_config?: ChunkingConfig
   datafeed_id: Id
-  frequency?: Timestamp
+  frequency?: DateTime
```
sethmlarson, line 40: Should all instances of `frequency` and `query_delay` below be `TimeSpan`?

specification/ml/flush_job/MlFlushJobRequest.ts
```diff
@@ -48,7 +48,8 @@ export interface Request extends RequestBase {
      * Specifies to advance to a particular time value. Results are generated
      * and the model is updated for data from the specified time interval.
      */
-    advance_time?: DateString
+    // Also accepts `now` as a value, epoch seconds (< 10 digits) and epoch milliseconds
+    advance_time?: DateTime
```
sethmlarson, line 52: Should we capture the `DateTime | EpochMillis` aspect of this type? The `EpochSeconds` seems like it'd be hacky to annotate.

specification/nodes/info/types.ts
```diff
@@ -371,7 +372,7 @@ export class NodeOperatingSystemInfo {
   name: Name
   pretty_name: Name
   /** Refresh interval for the OS statistics */
-  refresh_interval_in_millis: integer
+  refresh_interval_in_millis: long
```
sethmlarson, line 375: `TimeSpanMillis`?

specification/shutdown/get_node/ShutdownGetNodeResponse.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { NodeId } from '@_types/common'
-import { EpochMillis, Timestamp } from '@_types/Time'
+import { EpochMillis, DateTime } from '@_types/Time'
```
sethmlarson, line 21: Is this import unused?

specification/watcher/_types/Watch.ts
```diff
@@ -40,16 +40,16 @@ export class Watch {
   input: InputContainer
   metadata?: Metadata
   status?: WatchStatus
-  throttle_period?: string
+  throttle_period?: TimeSpan
+  throttle_period_in_millis?: long
```
sethmlarson, line 44: TimeSpanMillis

sethmlarson (commented):
All the work done in this PR is Herculean, excellent work @swallez! :muscle: I've read every change top to bottom and left a bunch of small comments and a few questions too:

specification/_global/delete_by_query/DeleteByQueryRequest.ts
swallez: ES really expects a float here, which from a JSON perspective also includes integer values (i.e. no decimal point).

[The docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html) are a bit misleading through, saying "you can set requests_per_second to any positive decimal number" and later "requests_per_second (Optional, integer)"...

sethmlarson:
Sounds like `Duration` is a common one for languages, should we go with that then?

swallez:
> Sounds like Duration is a common one for languages, should we go with that then?

Ok for `Duration` (I just added Rust that also has `Duration`).

specification/ilm/_types/Phase.ts
swallez: It's an actual quirk. ES will not accept as input of `PUT /_ilm/policy/{name}` the ILM policy output by `GET _xpack/usage`. Not sure many users will try to do this instead of using the output of `GET _ilm/policy` that is correctly formatted, but if they do this will not work.  I opened https://github.com/elastic/elasticsearch/issues/87327 for this.

specification/indices/_types/IndexSettings.ts
swallez: `DateTime` is defined as `string | long` (actually will be `string | EpochTime<UnitMillis>` in the next commit) as ES accepts both a formatted date (generally ISO format) and an epoch-millis as input for `DateTime` even if it always outputs it as an ISO-formatted date, except, sometimes, in the settings.

The reason why in this specific case it can be a number is because settings are stored verbatim as expressed in the JSON that defines them. So if this setting was defined as an epoch millis number, it will be sent back as that same number.

swallez:
@sethmlarson I took all of your judicious comments into account (thanks for the thorough review) and found even more wrongly typed fields ;-)

I also added the "units as generic parameters" approach we discussed and we now have:
```ts
export type UnitSeconds = long
export type UnitMillis = long
... other units

export type EpochTime<Unit> = Unit
export type DateTime = string | EpochTime<UnitMillis>

export type DurationValue<Unit> = Unit
export type Duration = string | -1 | 0
```

Are we ok with this naming, in particular `Duration` and `DurationValue`?

sethmlarson (approved):
Re-reviewed all the changes and they look good! I'm good with `Duration` and `DurationValue`.

PR #1716, opened 2022-05-19
Add request & response types for cat.component_templates

Anaethelion:
Adds `cat.component_templates` request & response objects


delvedor (approved):
LGTM with a nit.

PR #1723, opened 2022-05-23
Fix spec issues reported in the Java client

### LINKED ISSUE #1555: CreateApiKeyRequest.RoleDescriptor is missing some fields
Found in https://github.com/elastic/elasticsearch-java/issues/106

`security/create_api_key/types.ts/RoleDescriptor` is missing some fields and should actually be replaced with `security/get_service_accounts/types.ts/RoleDescritor` (the docs -- and the code -- says they are the same structure).

Which probably means that RoleDescriptor should be moved to `security/_types`.

Also, the `indices` field needs an `@aliases index` (ES accepts both).

This is a breaking change as we move types around. It has to be backported to 7.17 (long term 7.x support), but probably not to 8.1.

### LINKED ISSUE #1558: Missing support for {dynamic_type} property mapping
Found in https://github.com/elastic/elasticsearch-java/issues/142

Index mapping [dynamic templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html) provide more control on dynamic field mappings. Using them allows using the [`{dynamic_type}`](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html#template-variables) property type that is missing in the API spec.

This requires a new `DynamicTypeProperty` class with a `type: '{dynamic_type}'` variant tag.

Dynamic properties can be of type `object`, `string`, `long`, `double`, `boolean`, `date` and `binary` (see [ES code](https://github.com/elastic/elasticsearch/blob/add9b6a5ebe68243e9b1d6a00f9c0ae36c948889/server/src/main/java/org/elasticsearch/index/mapper/DynamicTemplate.java#L64)). So `DynamicTypeProperty` must aggregate the properties of all these types.

swallez:
This PR fixes a number of API spec issues that were reported in the Java client. We're experimenting a batch processing of these issue, as they are most of the time very tiny problems such as missing fields or required fields that should be optional.

This PR has 2 larger changes: the new `dynamic_property` property type, and a refactoring of `Higlight` and `HighlightField` with a common base class, as they have most attribute in common (same in the ES code base).

* ILM Policy has an invalid "name" field.  
  Found in https://github.com/elastic/elasticsearch-java/issues/160

* CreateApiKeyRequest.RoleDescriptor is missing some fields. Fixes #1555.  
  Found in https://github.com/elastic/elasticsearch-java/issues/106

* Missing support for {dynamic_type} property mapping. Fixes #1558.  
  Found in https://github.com/elastic/elasticsearch-java/issues/142

* Missing support for auto slicing.  
  Found in https://github.com/elastic/elasticsearch-java/issues/147

* RemoteSource username/password should be optional.  
  Found in https://github.com/elastic/elasticsearch-java/issues/152

* SuggestFuzziness requires optional attributes to be set.  
  Found in https://github.com/elastic/elasticsearch-java/issues/162

* Missing "attributes" in explain token with custom analyzer.  
  Found in https://github.com/elastic/elasticsearch-java/issues/154

* Missing fields in MultisearchBody  
  Found in https://github.com/elastic/elasticsearch-java/issues/161 and https://github.com/elastic/elasticsearch-java/issues/170

* TermVectors.terms.tokens should be optional  
  Found in https://github.com/elastic/elasticsearch-java/issues/184

* Missing fields in Highlight (also refactored Highlight & HighlightField with a common base class)  
  Found in https://github.com/elastic/elasticsearch-java/issues/185

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -315,3 +315,15 @@ export class IndicesOptions {
    */
   ignore_throttled?: boolean
 }
+
+/**
+ * Uses sliced scroll to parallelize process. Using `auto` chooses a reasonable number for most data streams
+ * and indices.
+ *
+ * @codegen_names value, auto
+ */
+export type Slices = integer | SlicesEnum
+
+export enum SlicesEnum {
+  auto
+}
--- a/compiler/src/model/utils.ts
+++ b/compiler/src/model/utils.ts
@@ -301,6 +301,7 @@ export function modelType (node: Node): model.ValueOf {
           const generics = node.getTypeArguments().map(node => modelType(node))
           const identifier = node.getTypeName()
           assert(node, Node.isIdentifier(identifier), 'Not an identifier')
+          assert(node, identifier.getDefinitions().length > 0, 'Unknown definition (missing import?)')
 
           const declaration = identifier.getDefinitions()[0].getDeclarationNode()
           // We are looking at a generic parameter
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1620,7 +1620,8 @@
         "Request: query parameter 'allow_no_indices' does not exist in the json spec",
         "Request: query parameter 'expand_wildcards' does not exist in the json spec",
         "Request: query parameter 'ignore_throttled' does not exist in the json spec",
-        "Request: query parameter 'ignore_unavailable' does not exist in the json spec"
+        "Request: query parameter 'ignore_unavailable' does not exist in the json spec",
+        "Request: query parameter 'routing' does not exist in the json spec"
       ],
       "response": []
     },
@@ -1834,6 +1835,12 @@
       ],
       "response": []
     },
+    "security.create_api_key": {
+      "request": [
+        "interface definition security._types:RoleTemplateQueryContainer - Property template is a single-variant and must be required"
+      ],
+      "response": []
+    },
     "security.create_service_token": {
       "request": [
         "Request: missing json spec query parameter 'refresh'",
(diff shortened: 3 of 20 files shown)


specification/_types/common.ts
```diff
@@ -315,3 +315,15 @@ export class IndicesOptions {
    */
   ignore_throttled?: boolean
 }
+
+/**
+ * Uses sliced scroll to parallelize process. Using `auto` chooses a reasonable number for most data streams
+ * and indices.
+ *
+ * @codegen_names value, auto
```
sethmlarson, line 323: Does the name `auto` here still make sense if another value is added to the enum?

sethmlarson (approved):
Changes look good to me, had one question about the `Slices` type:

specification/_types/common.ts
Anaethelion: It does, `auto` is a special value for this field which is translated to `0` on ES side. Providing `0` in the payload doesn't apply the same behavior.

Anaethelion (approved):

specification/_types/common.ts
swallez: After a discussion with @technige we ended up considering that `all` is the name of a function used to compute the number of slices. There's only one such function as of now, but there could be others. So this becomes: 

```ts
/**
 * Slices configuration used to parallelize a process.
 *
 * @codegen_names value, computed
 */
export type Slices = integer | SlicesCalculation

/**
 * Named computations to calculate the number of slices
 */
export enum SlicesCalculation {
  /**
   * Let Elasticsearch choose a reasonable number for most data streams and indices.
   */
  auto
}
```

PR #1732, opened 2022-06-01
Add ml.get_model_snapshot_upgrade_stats

Anaethelion:
Adds request & response for `ml.get_model_snapshot_upgrade_stats`.

### DIFF
--- a/specification/ml/_types/Model.ts
+++ b/specification/ml/_types/Model.ts
@@ -20,6 +20,7 @@
 import { ByteSize, Id, VersionString } from '@_types/common'
 import { integer, long } from '@_types/Numeric'
 import { Time } from '@_types/Time'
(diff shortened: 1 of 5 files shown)


specification/ml/_types/Model.ts
```diff
@@ -44,6 +45,14 @@ export class ModelSnapshot {
   timestamp: long
 }
 
+export class ModelSnapshotUpgrade {
+  job_id: Id
+  snapshot_id: Id
+  state: string
```
swallez, line 51: The state is actually [defined as an enum](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/job/snapshot/upgrade/SnapshotUpgradeState.java). Should we capture this in the spec?

PR #1735, opened 2022-06-02
Extract security.privileges.query in a typealias

Anaethelion:


swallez (changes requested):
From the ES source code, the type for `IndicesPrivilegesQuery` is `string | QueryContainer | RoleTemplateQueryContainer | null` (not a string array).

I think we can ignore `null` as it has the same result as not including the `query` field.

The `string` is interpreted as `QueryContainer | RoleTemplateQueryContainer` in "json as text" (see also [the doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html#security-api-put-role-example)). ES uses this string format internally, and will always serialize this field as a json-as-text (i.e. the `string` variant).

Additionally, `RoleTemplateQueryContainer` is not a container, it's a simple structure that has a single `template: Script` field (we don't need `RoleTemplateQueryContainer`).

And finally, we need a `@codegen_name` on the union type to identify the variants.

So that should be:

```ts
export class IndicesPrivileges {
  ...
  query?: IndicesPrivilegesQuery
}

/**
 * ... explain that text and JSON are accepted in requests, but that responses
 * will always contain the text version.
 * 
 * @codegen_name text, query, template
 */
export type IndicesPrivilegesQuery = string | QueryContainer | RoleTemplateQuery

export class RoleTemplateQueryContainer {
  ...
  template?: Script
}
```

specification/security/_types/Privileges.ts
```diff
@@ -94,14 +94,27 @@ export class IndicesPrivileges {
...
    * @server_default false
    */
   allow_restricted_indices?: boolean
 }
 
+/**
+ * While creating or updating a role you can provide either a JSON structure or a string to the API.
+ * However, the response provided by Elasticsearch will only be string with a json-as-text content.
+ *
+ * Since this is embedded in IndicesPrivileges, the same structure is used for clarity in both contexts.
```
swallez, line 109: Add Markdown backquotes around type names for a nicer formatting? (also applies to `indices` in the response)

specification/security/_types/Privileges.ts
```diff
@@ -94,14 +94,27 @@ export class IndicesPrivileges {
...
+ *
+ * @codegen_names text, query, template
```
swallez, line 111: Wondering if we should use `json_text` instead of just `text` so that the expected format of this string is more explicit.

swallez (approved):
LGTM, with minor comments

specification/security/_types/Privileges.ts
Anaethelion: Done

PR #1761, opened 2022-06-14
Flatten type mapping property tree

Anaethelion:
This PR removes the non-existent types between the type mapping and the properties.
It simplifies the generated types and provide a better understanding of all the possibilities a user has for property mapping.


swallez (approved):
LGTM, once you've updated the output.

PR #1763, opened 2022-06-15
Add a Make target to expand generics in schema.json

swallez:
This PR adds a new `make transform-expand-generics` target.

This target transforms `schema.json` to expand all generic types to concrete types and remove all generic parameters. These concrete types can have two forms:
- for "internal generic types" that are instanciated by other specification types (e.g. `MultiBucketAggregateBase<TBucket>`), a new type is generated for every instance of this type, with the value of generic parameters replaced with the concrete value defined by the instanciating type.
- for "external generic types", i.e. requests and response with a user-provided generic parameter (e.g. `SearchRequest<TDocument>`), the generic parameters are replaced by `UserDefinedData`.

Currently `make transform-expand-generics` creates a new `output/schema/schema-no-generics.json` file, but ultimately we should be able to specify the input and output paths as arguments.

This transformation is the first of its kind, and is meant to produce a schema that is easier to consume by some code generators without removing features from the specification that would be useful for some other languages. It's the result of a discussion with @Anaethelion about the Go client generator, where generics bring significant complexity, but should be kept in the spec for languages with generic support.

Other transformations may be of interest, such as removing inheritance by expanding fields from a super class into its child classes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ clean-dep:	## Clean npm dependencies
 	@rm -rf compiler/node_modules
 	@rm -rf typescript-generator/node_modules
 
+transform-expand-generics: ## Create a new schema with all generics expanded
+	@npm run transform-expand-generics --prefix compiler
+
 contrib: | generate license-check spec-format-fix 	## Pre contribution target
 
 help:  ## Display help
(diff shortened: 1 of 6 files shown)


sethmlarson (approved):
I'm happy with this idea of transforming the schema.json, leaving @Anaethelion to review whether the implementation is effective for the Go client.

Anaethelion (approved):
Works beautifully, the generated code seems to match the API as far as my tests went.
LGTM

Issue #1742, opened 2022-06-03
Add types for security.has_privileges_user_profile


sethmlarson:
API was added in these two PRs:

- https://github.com/elastic/elasticsearch/pull/85898
- https://github.com/elastic/elasticsearch/pull/87360

The API types look very similar to the [`security.has_privileges` API](https://github.com/elastic/elasticsearch-specification/tree/main/specification/security/has_privileges), probably can use this API as a good starting point.

Request looks like this:

```json
{
  "uids": ["u_LQPnxDxEjIH0GOUoFkZr5Y57YUwSkL9Joiq-g4OCbPc_0", "u_rzRnxDgEHIH0GOUoFkZr5Y27YUwSk19Joiq=g4OCxxB_1"],
  "cluster": [ "monitor", "create_snapshot", "manage_ml" ],
  "index" : [
    {
      "names": [ "suppliers", "products" ],
      "privileges": [ "create_doc"]
    },
    {
      "names": [ "inventory" ],
      "privileges" : [ "read", "write" ]
    }
  ],
  "application": [
    {
      "application": "inventory_manager",
      "privileges" : [ "read", "data:write/inventory" ],
      "resources" : [ "product/1852563" ]
    }
  ]
}
```

Response looks like this:

```json
{
  "has_privilege_uids": ["u_rzRnxDgEHIH0GOUoFkZr5Y27YUwSk19Joiq=g4OCxxB_1"]
}
```

PR #1764, opened 2022-06-20
Add 'manage_user_profile' privilege, security.has_privileges_user_profile API

sethmlarson:
Closes https://github.com/elastic/elasticsearch-specification/issues/1742

### DIFF
--- a/specification/security/suggest_user_profiles/types.ts
+++ b/specification/security/suggest_user_profiles/types.ts
@@ -1,0 +1,17 @@
+import { UserProfileId } from "@security/_types/UserProfile"
+import { SingleKeyDictionary } from "@spec_utils/Dictionary"
+
+export class Hint {
+  /**
+   * A list of Profile UIDs to match against. 
+   */
+  uids?: UserProfileId[]
+  /**
+   * A single key-value pair to match against the labels section
+   * of a profile. A profile is considered matching if it matches
+   * at least one of the strings. 
+   */
+  labels?: SingleKeyDictionary<string, HintLabelValue>
+}
+
+export type HintLabelValue = string | string[]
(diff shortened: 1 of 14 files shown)


specification/security/suggest_user_profiles/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * A list of Profile UIDs to match against.
+   */
+  uids?: UserProfileId[]
+  /**
+   * A single key-value pair to match against the labels section
+   * of a profile. A profile is considered matching if it matches
+   * at least one of the strings.
+   */
+  labels?: SingleKeyDictionary<string, HintLabelValue>
```
swallez, line 33: A single key dictionary sounds weird for a `labels` (plural) property, and I wouldn't be surprised that it's extended in the future to support multiple entries. So what about making a regular `Dictionary`?

specification/security/suggest_user_profiles/types.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export type HintLabelValue = string | string[]
```
swallez, line 36: Not sure we need dedicated types to express array leniency, even more for a one-time use.

Ultimately, we should introduce a type for lenient arrays similar to the recent `Stringified`, i.e. `type LenientArray<T> = T | T[]`. Or just bite the bullet and have array leniency be an implicit contract throughout the entire spec since every array will have to be a `LenientArray`.

specification/security/suggest_user_profiles/types.ts
sethmlarson: The use of `labels` is to mirror the User Profile data property `labels`.

I was thinking the same but checking the server code it does check that there's on element. I'll set this to `Dictionary` because like you said, I think there's a good chance we see this expand to multiple filter values.

specification/security/suggest_user_profiles/types.ts
sethmlarson: I'll inline this `string | string[]`. Wasn't sure if having this inline would be a problem for generators.

swallez (approved):
LGTM!

PR #1765, opened 2022-06-22
Update versions used for validate-pr workflow

sethmlarson:
Noticed that the `validate-pr` workflow hasn't been working recently, wondering if it's related to a change in how we store recordings. If this fixes the issue then I'd like to introduce an automatic bump script so these values don't get out of date.


sethmlarson:
Looks to be the case, checking the recordings we always store with major.minor so we can't use `master` here anymore. Using `latest` will always used the latest major.minor.

PR #1771, opened 2022-06-27
Fix snapshot repository definitions

swallez:
In the `snapshot` namespace we only have a (partial) definition of the S3 repository. This PR defines the settings for all repository implementations and groups them in a `@non_exhaustive` union (plugins can define new snapshot repositories).

Found in https://github.com/elastic/elasticsearch-java/issues/207

### DIFF
--- a/specification/snapshot/_types/SnapshotRepository.ts
+++ b/specification/snapshot/_types/SnapshotRepository.ts
@@ -17,22 +17,119 @@
  * under the License.
  */
 
-import { Uuid } from '@_types/common'
+import { ByteSize, Uuid } from '@_types/common'
 import { integer } from '@_types/Numeric'
+import { Base } from '@xpack/usage/types'
+import { AdditionalProperties } from '@spec_utils/behaviors'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+import { Stringified } from '@spec_utils/Stringified'
 
-export class Repository {
-  type: string
+export class BaseRepository<Settings> {
   uuid?: Uuid
-  settings: RepositorySettings
+  settings: Settings
 }
 
-export class RepositorySettings {
-  chunk_size?: string
-  compress?: string | boolean
-  concurrent_streams?: string | integer
+/**
+ * @variants internal tag='type'
+ * @non_exhaustive
+ */
+export type Repository =
+  | S3Repository
+  | GoogleCloudRepository
+  | AzureRepository
+  | SourceRepository
+  | UrlRepository
+  | FsRepository
+  | HdfsRepository
+
+export class S3Repository extends BaseRepository<S3RepositorySettings> {
+  type: 's3'
+}
+
+class BlobStoreSettings {
+  chunk_size?: ByteSize
+  concurrent_streams?: Stringified<integer>
+  compress?: Stringified<boolean>
+  max_number_of_snapshots?: integer
+  max_snapshot_bytes_per_sec?: ByteSize
+  max_restore_bytes_per_sec?: ByteSize
+  /** @aliases read_only */
+  readonly?: Stringified<boolean>
+  use_for_peer_recovery?: boolean
+}
+
+export class S3RepositorySettings {
+  base_path?: string
+  bucket?: string
+  buffer_size?: ByteSize
+  client?: string
   location: string
-  /**
-   * @aliases readonly
-   */
-  read_only?: string | boolean
+  server_size_encryption?: boolean
+  storage_class?: string
+}
+
+export class AzureRepository extends BaseRepository<AzureRepositorySettings> {
+  type: 'azure'
+}
+
+export class AzureRepositorySettings extends BlobStoreSettings {
+  /** aliases account */
+  client?: string
+  container?: string
+  location_mode: AzureRepositoryLocation
+  max_single_part_upload_size: ByteSize
+}
+
+export enum AzureRepositoryLocation {
+  primary_only,
+  secondary_only,
+  primary_then_secondary,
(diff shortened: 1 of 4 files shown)


specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,119 @@
...
+
+export class AzureRepositorySettings extends BlobStoreSettings {
+  /** aliases account */
+  client?: string
+  container?: string
+  location_mode: AzureRepositoryLocation
+  max_single_part_upload_size: ByteSize
+}
+
+export enum AzureRepositoryLocation {
```
sethmlarson, line 83: nit: Should this be `AzureRepositoryLocationMode` to match the [`LocationMode` enum](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.retrypolicies.locationmode?view=azure-dotnet) name on the Azure side?

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,119 @@
...
+  implements AdditionalProperties<string, UserDefinedValue>
+{
+  delegate_type: string
+}
+
+export class FsRepository extends BaseRepository<FsRepositorySettings> {
+  type: 'fs'
+}
+
+export class FsRepositorySettings extends BlobStoreSettings {
```
sethmlarson, line 123: From docs, this is missing:

- chunk_size
- compress
- max_number_of_snapshots
- max_restore_bytes_per_sec
- max_snapshot_bytes_per_sec
- readonly

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,119 @@
...
+import { ByteSize, Uuid } from '@_types/common'
 import { integer } from '@_types/Numeric'
+import { Base } from '@xpack/usage/types'
+import { AdditionalProperties } from '@spec_utils/behaviors'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+import { Stringified } from '@spec_utils/Stringified'
 
-export class Repository {
-  type: string
+export class BaseRepository<Settings> {
```
sethmlarson, line 27: In the interest of avoiding generics if possible, what do you think about taking a more "Query"-like approach to these structures?

```ts
export class BaseRepository {
  uuid?: Uuid
}

export class BaseRepositorySettings {
  chunk_size?: ByteSize
  max_number_of_snapshots?: integer
  max_snapshot_bytes_per_sec?: ByteSize
  max_restore_bytes_per_sec?: ByteSize
  use_for_peer_recovery?: boolean
}

export class S3RepositorySettings extends BaseRepositorySettings {
  ...
}

export class S3Repository extends BaseRepository {
  type: 's3'
  settings: S3RepositorySettings
}

/**
 * @variants internal tag='type'
 * @non_exhaustive
 */
export type Repository = S3Repository | ...
```

sethmlarson (changes requested):
Thanks for this! Here's some comments and questions:

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,119 @@
...
+  compress?: Stringified<boolean>
+  max_number_of_snapshots?: integer
+  max_snapshot_bytes_per_sec?: ByteSize
+  max_restore_bytes_per_sec?: ByteSize
+  /** @aliases read_only */
+  readonly?: Stringified<boolean>
+  use_for_peer_recovery?: boolean
+}
+
+export class S3RepositorySettings {
```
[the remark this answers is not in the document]
swallez: Good catch!

specification/snapshot/_types/SnapshotRepository.ts
swallez: They are inherited from `BlobStoreSettings`

specification/snapshot/_types/SnapshotRepository.ts
swallez: Not sure if avoiding generics should be a goal, but this approach also works. Done.

swallez:
I addressed your comments. Can you review the latest changes?

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,132 @@
...
+}
+
+class BlobStoreSettings {
+  chunk_size?: ByteSize
+  concurrent_streams?: Stringified<integer>
+  compress?: Stringified<boolean>
+  max_number_of_snapshots?: integer
+  max_snapshot_bytes_per_sec?: ByteSize
+  max_restore_bytes_per_sec?: ByteSize
+  /** @aliases read_only */
```
sethmlarson, line 57: Is there a reason we switched the `@aliases` and the non-alias property name from the previous types?

sethmlarson (approved):
This looks good, I only had one question:

specification/snapshot/_types/SnapshotRepository.ts
swallez: This is because some recordings use `read_only` but the setting name in the ES code is `readonly`. I couldn't find though in the ES code where that `read_only` would be accepted.

steffenvan:
Hi, I was wondering what the timeline is to merge this PR?

pquentin:
Should we close this since https://github.com/elastic/elasticsearch-specification/pull/2255 from @JoshMock was merged instead? Or is there still work to do?

JoshMock:
@pquentin Yes, I believe we can close this.

PR #1772, opened 2022-06-27
Adds make.sh, Dockerfile, bump task

picandocodigo:
This commit adds a `bump` task to the Makefile, as well as the make.sh script and Dockerfile to automate the bump task.

I tried to keep it as simple as possible, not bringing in all the code (particularly in make.sh) from the clients.


sethmlarson (approved):
LGTM, thanks Fernando!

PR #1786, opened 2022-07-14
Add model for _internal.health API

sethmlarson:
This API is an internal one so won't be generated by clients yet, however having a model for this API would be useful to teams building on the new Health API.


sethmlarson:
I believe this PR is now very out of date and it looks like the health API is being promoted to non-internal soon. Going to close this one in favor of modeling the public API.

PR #1789, opened 2022-07-19
Fix snapshot.get validation errors

sethmlarson:
- Adds the missing `index_names` parameter which is a part of the JSON spec
- Only remaining validation errors will be fixed by https://github.com/elastic/elasticsearch/issues/88620

### DIFF
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -2166,10 +2166,7 @@
     },
     "snapshot.get": {
       "request": [
-        "Request: query parameter 'human' does not exist in the json spec",
-        "Request: missing json spec query parameter 'index_names'",
-        "Request: should not have a body",
-        "request definition snapshot.get:Request / query - Property 'human' is already defined in an ancestor class"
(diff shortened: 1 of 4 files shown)


sethmlarson:
Since this includes a breaking change in making the responses' `indices` property optional I won't backport this one.

output/schema/schema.json
```diff
@@ -164972,8 +164972,11 @@
           }
         },
         {
-          "name": "human",
+          "description": "If true, returns the name of each index in each snapshot.",
+          "name": "index_names",
           "required": false,
+          "serverDefault": true,
+          "since": "7.13.0",
```
pquentin, line 164979: I believe this was only introduced in 8.3.0?

https://github.com/elastic/elasticsearch/pull/86269

sethmlarson: Thanks for correcting this! I'll get the right version there.

stevejgordon (approved):
LGTM

PR #1790, opened 2022-07-21
add new execution hint for cardinality

not-napoleon:
This adds a specification for the new Cardinality aggregation execution hint parameter, which is optional and accepts an enum value from a list.  This parameter will be available from 8.4 forward.  Generally speaking, we do not expect users to set this value, but we are providing it as a option for tuning performance in extreme cases.


sethmlarson:
Thanks for opening this PR @not-napoleon, this looks good to me. Could you run `make contrib` to recompile the specification output and push the result?

not-napoleon:
@sethmlarson took a little while to get `nvm` to play nicely, but I managed to get it working.  Is there anything else you need from me for this?

sethmlarson (approved):
That's all, thank you!

PR #1817, opened 2022-08-04
Adds cache_size, fixes for TrainedModelDeployment models

sethmlarson:
While adding the `cache_size` parameter to `ml.start_trained_model_deployment` API for 8.4 I noticed a renaming of "allocation" to "assignment" had [happened during 8.3](https://github.com/elastic/elasticsearch/pull/85503) that we missed unfortunately. Made all the fixes for that renaming in this PR as well.

### DIFF
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -144,7 +144,7 @@ export class TrainedModelDeploymentNodesStats {
   /** The number of inference requests that were not processed because the queue was full. */
   rejection_execution_count: integer
   /** The current routing state and reason for the current routing state for this allocation. */
-  routing_state: TrainedModelAllocationRoutingTable
+  routing_state: TrainedModelAssignmentRoutingTable
   /** The epoch timestamp when the allocation started. */
   start_time: EpochTime<UnitMillis>
   /** The number of threads used by each allocation during inference. */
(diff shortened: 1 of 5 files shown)


specification/ml/start_trained_model_deployment/MlStartTrainedModelDeploymentRequest.ts
```diff
@@ -72,6 +78,6 @@ export interface Request extends RequestBase {
      * Specifies the allocation status to wait for before returning.
      * @server_default started
      */
-    wait_for?: DeploymentAllocationState
+    wait_for?: DeploymentAssignmentState
```
benwtrent, line 81:
```suggestion
    wait_for?: DeploymentAllocationState
```

specification/ml/_types/TrainedModel.ts
```diff
@@ -333,31 +350,39 @@ export class TrainedModelAllocationRoutingTable {
...
 }
 
-export class TrainedModelAllocation {
+export class TrainedModelAssignment {
   /**
-   * The overall allocation state.
+   * The overall assignment state.
    */
-  allocation_state: DeploymentAllocationState
+  assignment_state: DeploymentAssignmentState
```
benwtrent, line 376: This should be a new object. Assignment state is different than the allocation state (which we are keeping)

This is assignment state, which is an enumeration of: "starting", "started", "stopping", "failed".

sethmlarson:
Thanks for the review comments @benwtrent! Could you give this another quick check that nothing got missed?

benwtrent (approved):

PR #1823, opened 2022-08-23
Fix optional/missing fields reported via Java client issues

swallez:
This PR fixes a number of fields that should be optional and also adds `SearchRequest.ext` that is undocumented but is present in HLRC and used by some plugins to add new sections to search requests.

These are bug fixes that should be backported down to 7.17.

- `IndexVersioning.created` is optional (not present on default index settings)  
  Found in https://github.com/elastic/elasticsearch-java/issues/286

- `GeoBoundsAggregate.bounds` should be optional.  
  Found in https://github.com/elastic/elasticsearch-java/issues/383

- `RuntimeFields` is a dictionary of single values, not an array  
  Found in https://github.com/elastic/elasticsearch-java/issues/298

- Search requests have an `ext` field for plugin extensions  
  Found in https://github.com/elastic/elasticsearch-java/issues/264

- All fields in `IcuCollationTokenFilter` are optional  
  Found in https://github.com/elastic/elasticsearch-java/issues/353

- `ComponentTemplateSummary.settings` is optional  
  Found in https://github.com/elastic/elasticsearch-java/issues/282

- `GetUserAccessTokenResponse.refresh_token` is optional  
  Found in https://github.com/elastic/elasticsearch-java/issues/332

- `ErrorCause.reason` is optional  
  Found in https://github.com/elastic/elasticsearch-java/issues/339

### DIFF
--- a/specification/_types/Errors.ts
+++ b/specification/_types/Errors.ts
@@ -36,7 +36,7 @@ export class ErrorCause
   /**
    * A human-readable explanation of the error, in english
    */
-  reason: string
+  reason?: string
   /**
    * The server stack trace. Present only if the `error_trace=true` parameter was sent with the request.
    */
--- a/specification/_types/mapping/RuntimeFields.ts
+++ b/specification/_types/mapping/RuntimeFields.ts
@@ -21,7 +21,7 @@ import { Dictionary } from '@spec_utils/Dictionary'
 import { Field } from '@_types/common'
 import { Script } from '@_types/Scripting'
 
-export type RuntimeFields = Dictionary<Field, RuntimeField | RuntimeField[]>
+export type RuntimeFields = Dictionary<Field, RuntimeField>
 
 export class RuntimeField {
   format?: string
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -567,6 +567,7 @@ export interface MsearchMultisearchBody {
   collapse?: SearchFieldCollapse
   query?: QueryDslQueryContainer
   explain?: boolean
+  ext?: Record<string, any>
   stored_fields?: Fields
   docvalue_fields?: (QueryDslFieldAndFormat | Field)[]
   knn?: KnnQuery
@@ -1026,6 +1027,7 @@ export interface SearchRequest extends RequestBase {
     aggs?: Record<string, AggregationsAggregationContainer>
     collapse?: SearchFieldCollapse
     explain?: boolean
+    ext?: Record<string, any>
     from?: integer
     highlight?: SearchHighlight
     track_total_hits?: SearchTrackHits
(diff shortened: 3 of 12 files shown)


specification/_types/Errors.ts
```diff
@@ -36,7 +36,7 @@ export class ErrorCause
   /**
    * A human-readable explanation of the error, in english
    */
-  reason: string
```
sethmlarson, line 39: Hmm! When can an error not have a `reason`?

specification/_types/mapping/RuntimeFields.ts
```diff
@@ -21,7 +21,7 @@ import { Dictionary } from '@spec_utils/Dictionary'
 import { Field } from '@_types/common'
 import { Script } from '@_types/Scripting'
 
-export type RuntimeFields = Dictionary<Field, RuntimeField | RuntimeField[]>
+export type RuntimeFields = Dictionary<Field, RuntimeField>
```
sethmlarson, line 24: Do we know why this originally was made an array?

sethmlarson (commented):
Few questions, the rest of the changes look good.

specification/_types/Errors.ts
swallez: I went digging to understand this, as it was surprising but happening in the wild ;-)

`ErrorCause` is the [JSON representation of a Java exception](https://github.com/elastic/elasticsearch/blob/e03c2b0161e31f72dd848000efacab97f8c2cdbd/server/src/main/java/org/elasticsearch/ElasticsearchException.java#L523) (a `Throwable`). The `reason` field is the result of Java's `Throwable.getMessage()`, which [according to the docs](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#getMessage--) can be `null`. In practice this never (I guess) happens for the top-level error, which is generally an `ElasticsearchException`. The error here happened in a deeper root cause, where apparently `getMessage()` returned `null` 🤷

specification/_types/mapping/RuntimeFields.ts
swallez: No idea. There's no occurrence of the array variant in our recordings and [looking at the ES source code](https://github.com/elastic/elasticsearch/blob/2a00f25f573e1c1078c1f864eea52397a081c5c8/server/src/main/java/org/elasticsearch/index/mapper/RuntimeField.java#L186) it has always expected the mapping to be an object and not an array.

sethmlarson (approved):
LGTM

specification/_types/Errors.ts
Enerccio: null can happen when it's same exception as before in case of `NullPointerExceptions` - jvm can optimize them out

PR #1824, opened 2022-08-24
Add sortable option to filters agg

nik9000:
This adds the `sortable` option to the `filters` agg which modifies the
rendered json for named filters from
```
"buckets": {
  "filtera": {},
  "filterb": {},
  "filterc": {}
}
```

to:
```
"buckets": [
  { "filtera": {} },
  { "filterb": {} },
  { "filterc": {} }
]
```

That way when you sort the buckets, you can see the ordering. The json
parser won't throw it away.

Also drops the `keyed` parameter from the spec for `filters` - it
doesn't exist.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -3000,7 +3000,7 @@ export interface AggregationsFiltersAggregation extends AggregationsBucketAggreg
   filters?: AggregationsBuckets<QueryDslQueryContainer>
   other_bucket?: boolean
   other_bucket_key?: string
-  keyed?: boolean
+  sortable?: boolean
 }
 
 export interface AggregationsFiltersBucketKeys extends AggregationsMultiBucketBase {
--- a/specification/_types/aggregations/bucket.ts
+++ b/specification/_types/aggregations/bucket.ts
@@ -167,7 +167,7 @@ export class FiltersAggregation extends BucketAggregationBase {
   filters?: Buckets<QueryContainer>
   other_bucket?: boolean
   other_bucket_key?: string
-  keyed?: boolean
+  sortable?: boolean
 }
 
 export class GeoDistanceAggregation extends BucketAggregationBase {


nik9000:
This links ot https://github.com/elastic/elasticsearch/pull/89256

swallez (changes requested):
There's a fundamentally flawed assumption in https://github.com/elastic/elasticsearch/pull/89256, which is that JSON objects properties are ordered (filters in this context). This isn't the case, and depending on how users build their query, order in the JSON may not be the insertion order since a JSON object is basically a map.

Furthermore, it's good to limit the number of "shapes" of a data structure, as these lead to union types that 1/ aren't natively supported by every language (e.g. Java or Go) and 2/ require the user to downcast the object to a specific variant with the possible runtime errors that can result from that.

So for use cases where filter order matters, this should be implemented in the application and not as an additional representation of the result driven by a request parameter.

nik9000:
> There's a fundamentally flawed assumption in [elastic/elasticsearch#89256](https://github.com/elastic/elasticsearch/pull/89256), which is that JSON objects properties are ordered (filters in this context). This isn't the case, and depending on how users build their query, order in the JSON may not be the insertion order since a JSON object is basically a map.

That PR is attempting to fix that assumption. It provides the option to return an array.

> So for use cases where filter order matters, this should be implemented in the application and not as an additional representation of the result driven by a request parameter.

We've had support for ordering on the ES side for the better part of forever now. They are trying to make it so you don't have to rely on the ordering in the json objects. Personally, I'd do the sorting client side too. But because we support it for other aggs and don't fail when you point it at this one, it makes some sense to support it. Thus we never closed this bug as WONTFIX. But if the overhead from the extra shapes is too high then we should reject the PR and close the bug as something we won't implement.

sethmlarson:
@nik9000 Thanks for opening this PR to discuss. I'm in agreement with Sylvain about adding more potential shapes and unions to already complicated structures would be a lot of overhead for us.

sethmlarson:
Closing this PR in favor of the suggestion here: https://github.com/elastic/elasticsearch/issues/83957#issuecomment-1304115505

PR #1873, opened 2022-09-13
Rename position_length to positionLength

peedeeboy:
As reported in the Java Client repo, elastic/elasticsearch-java#391, `positionLength` is not being de-serialised by the Java API client, as the specification appears to mistakenly have `position_length` as the name of the property.

The change renames `position_length` to `positionLength` in the Specification, so the issue will be fixed in the Java API client when the client's entity classes are next generated.


sethmlarson (approved):
LGTM, thank you!

PR #1888, opened 2022-09-22
Add model for DownsampleConfig

sethmlarson:
Now that `indices.downsample` will become a non-feature flag API in 8.5 we should start modeling the config object. Right now there is only one option, `fixed_interval`.


swallez (approved):
LGTM

PR #1889, opened 2022-09-22
[validation-errors] Fixes in watcher API

### LINKED ISSUE #1886: Java-client+recordings validation errors -  Tasks & Watcher
These errors were found when loading recordings in the Java client. Do not forget to **backport relevant changes to 7.17**, which is still maintained.

Field names reported can be either in their json form (`snake_case`) or in their Java form (`camelCase`) depending on the context. If the messages sound cryptic, ask @swallez for an explanation!

Errors coming from Nest tests should be taken with a grain of salt, since these recordings are quite old now and were captured with ES 7.x

```
tasks.list response: Error deserializing co.elastic.clients.elasticsearch.tasks.TaskInfo: Unknown field 'cancelled' (JSON path: nodes['wA_RKu0tTx6t0-wY0aaTWA'].tasks['wA_RKu0tTx6t0-wY0aaTWA:58015'].cancelled)  (×1)
    elasticsearch/tmp-free/tasks.list/69f200ecca0f89a637335b0092b7ec56.json

tasks.list response: Error deserializing co.elastic.clients.elasticsearch.tasks.TaskInfo: Unknown field 'cancelled' (JSON path: nodes['wA_RKu0tTx6t0-wY0aaTWA'].tasks['wA_RKu0tTx6t0-wY0aaTWA:9438'].cancelled)  (×1)
    elasticsearch/tmp-free/tasks.list/2a4dd23af7659a3477da14641fc609b1.json

tasks.list response: Error deserializing co.elastic.clients.elasticsearch.tasks.TaskInfo: Unknown field 'cancelled' (JSON path: tasks['wA_RKu0tTx6t0-wY0aaTWA:58016'].children[0].cancelled)  (×1)
    elasticsearch/tmp-free/tasks.list/450e300fe192f3a646f09e38c0a24999.json

watcher.put_watch request: Error deserializing co.elastic.clients.elasticsearch.watcher.ArrayCompareCondition: Unknown field 'ctx.payload.search.aggregations.top_project_tags.buckets' (JSON path: condition.array_compare['ctx.payload.search.aggregations.top_project_tags.buckets'])  (×1)
    nest-tests/watcher.put_watch/XPack/Watcher/PutWatch/PutWatchApiTests/unknown__ClientUsage_b__17_0_2433__request.json

watcher.put_watch request: Error deserializing co.elastic.clients.elasticsearch.watcher.SearchInput: Unknown field 'search' (JSON path: input.chain.inputs[2].search.search)  (×8)
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2431__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2422__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2425__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2436__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2446__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2450__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2444__request.json
    nest-tests/watcher.put_watch/XPack/Watcher/GetWatch/GetWatchApiTests/unknown_PutWatch_2440__request.json
```

Anaethelion:
* Add a cancelled property to `TaskInfo`.
* Add `UserDefinedValue` in `FieldValue` union.
* Fix watcher input and conditions.

Fixes #1886

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -24,9 +24,8 @@ import { AdditionalProperties } from '@spec_utils/behaviors'
 
 /**
  * A field value.
- * @codegen_names long, double, string, boolean, user_defined_value
+ * @codegen_names long, double, string, boolean, any
  */
-// FIXME: representation of geopoints and ip addresses?
 export type FieldValue = long | double | string | boolean | UserDefinedValue
 
 export class UrlParameter {}
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -27,7 +27,7 @@ import { AdditionalProperties } from '@spec_utils/behaviors'
  * @codegen_names long, double, string, boolean
  */
 // FIXME: representation of geopoints and ip addresses?
-export type FieldValue = long | double | string | boolean
+export type FieldValue = long | double | string | boolean | UserDefinedValue
 
 export class UrlParameter {}
 
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -24,7 +24,7 @@ import { AdditionalProperties } from '@spec_utils/behaviors'
 
 /**
  * A field value.
- * @codegen_names long, double, string, boolean
+ * @codegen_names long, double, string, boolean, user_defined_value
  */
 // FIXME: representation of geopoints and ip addresses?
 export type FieldValue = long | double | string | boolean | UserDefinedValue
--- a/specification/watcher/_types/Conditions.ts
+++ b/specification/watcher/_types/Conditions.ts
@@ -17,44 +17,38 @@
  * under the License.
  */
 
-import { Dictionary } from '@spec_utils/Dictionary'
+import {Dictionary, SingleKeyDictionary} from '@spec_utils/Dictionary'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+import {FieldValue} from "@_types/common";
+import {AdditionalProperty} from "@spec_utils/behaviors";
 
 export class AlwaysCondition {}
 
-export class ArrayCompareCondition {
-  array_path: string
-  comparison: string
-  path: string
+export class ArrayCompareOpParams {
   quantifier: Quantifier
-  value: UserDefinedValue
+  value: FieldValue
 }
 
-export class CompareCondition {
-  comparison?: string
-  path?: string
-  value?: UserDefinedValue
-  'ctx.payload.match'?: CompareContextPayloadCondition
-  'ctx.payload.value'?: CompareContextPayloadCondition
+export class ArrayCompareCondition implements AdditionalProperty<ConditionOp, ArrayCompareOpParams> {
+  path: string
 }
 
-export class CompareContextPayloadCondition {
-  eq?: UserDefinedValue
-  lt?: UserDefinedValue
-  gt?: UserDefinedValue
-  lte?: UserDefinedValue
-  gte?: UserDefinedValue
+export enum ConditionOp {
+  not_eq,
+  eq,
+  lt,
+  gt,
+  lte,
+  gte,
 }
-
-export class Condition {}
 
 /**
  * @variants container
  */
 export class ConditionContainer {
   always?: AlwaysCondition
-  array_compare?: ArrayCompareCondition
-  compare?: CompareCondition
+  array_compare?: SingleKeyDictionary<string, ArrayCompareCondition>
+  compare?: SingleKeyDictionary<string, SingleKeyDictionary<ConditionOp, FieldValue>>
   never?: NeverCondition
   script?: ScriptCondition
 }
--- a/specification/watcher/_types/Conditions.ts
+++ b/specification/watcher/_types/Conditions.ts
@@ -17,10 +17,10 @@
  * under the License.
  */
 
-import {Dictionary, SingleKeyDictionary} from '@spec_utils/Dictionary'
+import { Dictionary, SingleKeyDictionary } from '@spec_utils/Dictionary'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
-import {FieldValue} from "@_types/common";
-import {AdditionalProperty} from "@spec_utils/behaviors";
+import { FieldValue } from '@_types/common'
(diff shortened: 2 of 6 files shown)


specification/_types/common.ts
```diff
@@ -24,10 +24,10 @@ import { AdditionalProperties } from '@spec_utils/behaviors'
 
 /**
  * A field value.
- * @codegen_names long, double, string, boolean
+ * @codegen_names long, double, string, boolean, user_defined_value
```
sethmlarson, line 27: `user_defined_value` feels a bit leaky here since these names are user-facing, should we use something else like `raw` or `any`?

specification/watcher/_types/Conditions.ts
```diff
@@ -17,44 +17,43 @@
...
+  value: FieldValue
 }
 
-export class CompareCondition {
-  comparison?: string
-  path?: string
-  value?: UserDefinedValue
-  'ctx.payload.match'?: CompareContextPayloadCondition
-  'ctx.payload.value'?: CompareContextPayloadCondition
+export class ArrayCompareCondition
```
sethmlarson, line 32: Should this be a `@variants container` object instead of an `AdditionalProperty<enum, value>`? A container with a required property seems like it fits this case better?

specification/watcher/_types/Conditions.ts
```diff
@@ -17,44 +17,43 @@
...
  * @variants container
  */
 export class ConditionContainer {
   always?: AlwaysCondition
-  array_compare?: ArrayCompareCondition
-  compare?: CompareCondition
+  array_compare?: SingleKeyDictionary<string, ArrayCompareCondition>
+  compare?: SingleKeyDictionary<
+    string,
+    SingleKeyDictionary<ConditionOp, FieldValue>
```
sethmlarson, line 55: Should we lift the inner `SingleKeyDictionary` to be a `CompareCondition` alias? I also wonder if `@variants container` is better here instead of Dictionary<enum, value>

specification/watcher/put_watch/WatcherPutWatchRequest.ts
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { Dictionary } from '@spec_utils/Dictionary'
+import { Dictionary, SingleKeyDictionary } from '@spec_utils/Dictionary'
```
sethmlarson, line 20: You can remove this type

specification/watcher/_types/Input.ts
```diff
@@ -29,12 +29,13 @@ import {
   Username
 } from '@_types/common'
 import { Host } from '@_types/Networking'
-import { uint } from '@_types/Numeric'
+import { integer, uint } from '@_types/Numeric'
```
sethmlarson, line 32: Are the new `integer` and `AggregationContainer` imports needed here?

sethmlarson (commented):
Looking great, here's a handful of comments:

sethmlarson:
Pasting a mockup here of what the structures would look like as container variants instead of `SingleKeyDictionary`/`AdditionalProperty` with an enum value:

```ts
/**
 * @variants container
 */
export class CompareCondition {
  not_eq?: FieldValue
  eq?: FieldValue
  lt?: FieldValue
  gt?: FieldValue
  lte?: FieldValue
  gte?: FieldValue
}

/**
 * @variants container
 */
export class ArrayCompareCondition {
  /** @variants container */
  path: string
  not_eq?: FieldValue
  eq?: FieldValue
  lt?: FieldValue
  gt?: FieldValue
  lte?: FieldValue
  gte?: FieldValue
}

/**
 * @variants container
 */
export class ConditionContainer {
  always?: AlwaysCondition
  array_compare?: SingleKeyDictionary<string, ArrayCompareCondition>
  compare?: SingleKeyDictionary<string, CompareCondition>
  never?: NeverCondition
  script?: ScriptCondition
}
```

specification/_types/common.ts
swallez: +1. `any` sounds good here.

It is to be noted that we could just now use `UserDefinedValue` everywhere instead of `FieldValue` but it has a specific semantics, and scalar values are used a lot more often than raw json. So having a dedicated `FieldValue` abstraction allows code language clients to use a specific type for it with convenience functions to create a `FieldValue` from simple scalar values.

BTW, the above could be captured in a code comment (and not a jsdoc comment) to explain the rationale for `FieldValue` :-)

swallez: ... and we can also remove the `FIXME` below since adding user defined value now handles geopoints and ip addresses!

specification/watcher/_types/Conditions.ts
swallez: This is really a pair of (operator, parameters) where parameters do not depend on the operator kind. A container is more suited to variants where the structure members depend on the variant discriminant.

specification/watcher/_types/Conditions.ts
swallez: Same as above, it's really a pair of (operator, params).

specification/_types/common.ts
Anaethelion: Agreed on `any`, I'll make the change!

specification/watcher/_types/Conditions.ts
sethmlarson: Makes sense, thanks for the explanation!

sethmlarson:
I've resolved my comment on structure, thanks for the explanation on how this is different from a container variant. Will approve once other changes are in :shipit:

sethmlarson (approved):
LGTM

Issue #1881, opened 2022-09-20
Java-client+recordings validation errors - Indices


swallez:
These errors were found when loading recordings in the Java client. Do not forget to **backport relevant changes to 7.17**, which is still maintained.

Field names reported can be either in their json form (`snake_case`) or in their Java form (`camelCase`) depending on the context. If the messages sound cryptic, ask @swallez for an explanation!

Errors coming from Nest tests should be taken with a grain of salt, since these recordings are quite old now and were captured with ES 7.x

```
indices.create request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.AggregateMetricDoubleProperty: Unknown field 'time_series_metric' (JSON path: mappings.properties.k8s.properties.pod.properties.network.properties.rx.time_series_metric)  (×1)
    elasticsearch/tmp-platinum/indices.create/f477b11251dd0beaab84f1c4d7e6eca0.json

indices.create request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.ObjectProperty: Unknown field 'id' (JSON path: mappings.properties._rollup.id)  (×1)
    elasticsearch/tmp-platinum/indices.create/372e8f9aeec225dac71eaf5e07df3ed2.json

indices.create request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.ObjectProperty: Unknown field 'purple' (JSON path: mappings.properties.object2.purple)  (×1)
    elasticsearch/tmp-free/indices.create/8e155644313c68b61a28d372bacb59f3.json

indices.create request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.RuntimeField: Unknown field 'text' (JSON path: mappings.runtime.message.text)  (×1)
    elasticsearch/tmp-free/indices.create/aeeb2d0a0b48f5c3613973f78e366de1.json

indices.create request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.ShortNumberProperty: Unknown field 'time_series_dimension' (JSON path: mappings.properties.k8s.properties.pod.properties.availability_zone.time_series_dimension)  (×3)
    elasticsearch/tmp-free/indices.create/689d7afabfb7d740a3dbf908569f2a80.json
    elasticsearch/tmp-free/indices.create/6358fccf54520bf7c4fcd0fc6df21c7f.json
    elasticsearch/tmp-free/indices.create/9fde535ea8c3e151e2abfb7a4968e356.json

[...]

PR #1920, opened 2022-09-30
Add IPProperty and missing properties in mappings

Anaethelion:
This PR adds the missing `IPProperty`, revamps the `NumberPropertyBase` to include missing possible properties and adds `time_series_metric` to `AggregateMetricDoubleProperty`.

Relates to #1881

### DIFF
--- a/specification/_types/mapping/core.ts
+++ b/specification/_types/mapping/core.ts
@@ -99,10 +99,34 @@ export class KeywordProperty extends DocValuesPropertyBase {
   type: 'keyword'
 }
 
-export class NumberPropertyBase extends DocValuesPropertyBase {
+export class IPProperty extends DocValuesPropertyBase {
   index?: boolean
   ignore_malformed?: boolean
+  null_value?: string
+  on_script_error?: OnScriptError
+  script?: Script
+  /** [experimental] For internal use by Elastic only. Marks the field as a time series dimension. Defaults to false. */
+  time_series_dimension?: boolean
+  type: 'ip'
+}
+
+export class NumberPropertyBase extends DocValuesPropertyBase {
+  boost?: double
+  coerce?: boolean
+  ignore_malformed?: boolean
+  index?: boolean
+  /**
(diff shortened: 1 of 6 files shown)


specification/_types/mapping/core.ts
```diff
@@ -99,54 +99,72 @@ export class KeywordProperty extends DocValuesPropertyBase {
   type: 'keyword'
 }
 
-export class NumberPropertyBase extends DocValuesPropertyBase {
+export class IPProperty extends DocValuesPropertyBase {
```
sethmlarson, line 102: Should be `IpProperty`

specification/_types/mapping/core.ts
```diff
@@ -99,54 +99,72 @@ export class KeywordProperty extends DocValuesPropertyBase {
...
   index?: boolean
   ignore_malformed?: boolean
+  null_value?: string
+  on_script_error?: OnScriptError
+  script?: Script
+  /** [experimental] For internal use by Elastic only. Marks the field as a time series dimension. Defaults to false. */
```
sethmlarson, line 108: I think we can use the `@stability experimental` annotation on properties? We can add `@server_default false` too.

sethmlarson (approved):
Looks good, few comments before we merge:

sethmlarson:
Also, this PR can't be directly backported to 7.17 since there's no time series metrics there.

specification/_types/mapping/core.ts
Anaethelion: Habits of Go uppercasing acronyms.
I'll change that!

sethmlarson (approved):

Anaethelion:
> Also, this PR can't be directly backported to 7.17 since there's no time series metrics there.

They [are](https://github.com/elastic/elasticsearch/blob/7.17/server/src/main/java/org/elasticsearch/index/mapper/TimeSeriesParams.java), behind an experimental shield!

PR #1921, opened 2022-09-30
Add Mode to Mappings for _source

Anaethelion:
This allows to set the `_source` storage to `synthetic`. ([doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html#synthetic-source))

Relates to https://github.com/elastic/elasticsearch-specification/issues/1881

### DIFF
--- a/specification/_types/mapping/meta-fields.ts
+++ b/specification/_types/mapping/meta-fields.ts
@@ -61,4 +61,15 @@ export class SourceField {
   enabled?: boolean
   excludes?: string[]
   includes?: string[]
+  mode?: Mode
+}
+
+export enum Mode {
+  disabled = 0,
+  stored = 1,
+  /**
+   *  Instead of storing source documents on disk exactly as you send them,
+   *  Elasticsearch can reconstruct source content on the fly upon retrieval.
+   */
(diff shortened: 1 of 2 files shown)


specification/_types/mapping/meta-fields.ts
```diff
@@ -61,4 +61,15 @@ export class SourceField {
   enabled?: boolean
   excludes?: string[]
   includes?: string[]
+  mode?: Mode
+}
+
+export enum Mode {
```
sethmlarson, line 67: You can remove the `= N` on enums

sethmlarson, line 67: You can remove the `= N` on enums. Also since this is a top-level class in `specification/_types` we should give it a specific name, maybe `SourceFieldMode`?

sethmlarson (changes requested):

specification/_types/mapping/meta-fields.ts
Anaethelion: Done!

sethmlarson (approved):
LGTM

PR #1934, opened 2022-10-03
Add missing @variant on IpPrefixAggregate

swallez:
Follow-up to #1795 - add `@variant` annotation to `IpPrefixAggregate`


sethmlarson (approved):
Thank you!

PR #1937, opened 2022-10-04
Remove overriden properties

swallez:
This removes some property overrides that were introduced by the refactoring in #1920

Note to self: this should be caught by the model validation rules.


Anaethelion (approved):
LGTM!

PR #1939, opened 2022-10-04
Use TaskId for DeleteByQueryRethrottleRequest path parameter

stevejgordon:
When using the delete by query API asynchronously, it returns a `TaskId` for the ongoing task. The DeleteByQueryRethrottleRequest accepts a `task_id` as a required path parameter, but it's currently typed as `Id`. For the .NET client, at least, it presents some usability hurdles as you can't just pass the value of the `Task` property from the `DeleteByQueryResponse` into subsequent API calls. Ideally, I believe we should align the typing here with the usage scenario.

My only concern is this being a breaking change for other clients that probably function today using the parameter typed as `Id`.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -238,7 +238,7 @@ export interface DeleteByQueryResponse {
 }
 
 export interface DeleteByQueryRethrottleRequest extends RequestBase {
-  task_id: Id
+  task_id: TaskId
   requests_per_second?: float
 }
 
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
This seems like the right thing to me, this API is seldom used and the chain-ability of the return type of `delete_by_query` into this API is important. I'll ping @swallez on this to confirm.

swallez (approved):
LGTM.

No problem for the Java client. Java doesn't support type aliases, so all aliases are resolved to their final base type (i.e. String in this case).

Also, `Id` is actually used for a lot of unrelated things. This "id" word is not specific enough and we may want to progressively get rid of it in favor of more specific id types like `TaskId`.

Issue #1878, opened 2022-09-20
Java-client+recordings validation errors - Core & search


swallez:
These errors were found when loading recordings in the Java client. Do not forget to **backport relevant changes to 7.17**, which is still maintained.

Field names reported can be either in their json form (`snake_case`) or in their Java form (`camelCase`) depending on the context. If the messages sound cryptic, ask @swallez for an explanation!

Errors coming from Nest tests should be taken with a grain of salt, since these recordings are quite old now and were captured with ES 7.x

```
core.delete_by_query response: Missing required property 'DeleteByQueryResponse.throttledMillis' (×2)
    nest-tests/delete_by_query/Document/Multiple/DeleteByQueryRethrottle/DeleteByQueryRethrottleApiTests/unknown_OnBeforeCall_2542__response.json
    nest-tests/delete_by_query/Document/Multiple/DeleteByQuery/DeleteByQueryApiTests/DeleteByQueryApiTests__ClientUsage_b__23_3_1991__response.json

core.field_caps request: Error deserializing co.elastic.clients.elasticsearch._types.mapping.RuntimeField: Unknown field 'keyword' (JSON path: runtime_mappings.text.keyword)  (×1)
    elasticsearch/tmp-free/field_caps/c438a7d2bed6bafcfd8addbb50c1a0e1.json

core.field_caps request: Error deserializing co.elastic.clients.elasticsearch.core.FieldCapsRequest: Unknown field 'fields' (JSON path: fields)  (×1)
    elasticsearch/tmp-free/field_caps/84053f0e657e0741237babf45b657eea.json

core.get request: Unknown parameter: force_synthetic_source (×1)
    elasticsearch/tmp-free/get/cde3f31f35f7852ac062258a84701911.json

core.mget request: Unknown parameter: force_synthetic_source (×2)
    elasticsearch/tmp-free/mget/7e3bb54ce970c7c4c951133186e9c6b3.json
    elasticsearch/tmp-free/mget/3afaca6e9b4df15e0904b40de062ffdf.json

core.msearch response: Error deserializing co.elastic.clients.elasticsearch._types.aggregations.SignificantStringTermsAggregate: Unknown field 'bg_count' (JSON path: responses[0].aggregations['sigsterms#test_significant_terms'].bg_count)  (×1)
    elasticsearch/tmp-free/msearch/c80bd4869c612a5a236cd267a1333a7b.json

[...]

PR #1940, opened 2022-10-05
Fix issues in aggregations found with java-client+recordings

swallez:
Fixes aggregation issues found loading test recordings with the Java client.

Fixes #1879

Issue #1767, opened 2022-06-23
Modelling `wait_for_completion`


swallez:
## The problem

We have 16 endpoints that have a `wait_for_completion` parameter that instructs the ES cluster to run potentially lengthy operations in a blocking fashion or to run them asynchronously as a background task.

The response structure of these endpoints varies heavily depending on the value of this parameter. The default value for this parameter depends on the endpoint (most often `true`, but not always).

When `wait_for_completion=true`, the request blocks until the processing is finished, and the "normal" response body is then returned. When it's `false`, the request processing is started in the background on the ES cluster and some information about the background process is returned. The background response can vary, e.g. `{"acknowledged": true}` ([snapshot-restore](https://www.elastic.co/guide/en/elasticsearch/reference/current/restore-snapshot-api.html)) or a task id `{ "task": "oTUltX4"}` for [ml-delete-job](https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-delete-job.html)

The current state of the API specification is not optimal when it comes to supporting this feature, as having a single response structure requires to merge both normal and background properties in a single type, and consequently make them all optional.

Furthermore, as the ES documentation is generally sparse about the background response format, it is often forgotten in the spec and the focus is put on normal response properties, including making them required, which can cause issues (see https://github.com/elastic/elasticsearch-java/issues/201).

## Semantics of `wait_for_completion`

From a semantic perspective, endpoints with a `wait_for_completion` are endpoints that are "background-capable". We can consider them as the combination of a foreground endpoint that returns the normal (foreground) response, and a background endpoint that starts a task in the background and returns information about the background process.

Those two endpoints have the exact same request path, parameters and body, with the exception of `wait_for_completion`, and only differ by their response types.

Some client generators may decide to have a single endpoint whose response is the a union of both foreground and background response types, while some others may want to distinguish the two endpoints to avoid exposing a union whose active variant depends on the value set for `wait_for_completion` (that parameter would then be removed from the request structure and set internally by the endpoint functions).

For example in the case of the Java client, that second form would be used and the snapshot-restore endpoint would result in two functions:
* `restore(RestoreRequest): RestoreResponse` (foreground, internally sets `wait_for_completion=true`)
* `restoreAsync(RestoreRequest): AcknowledgedResponse` (background, internally sets `wait_for_completion=false`)

In TypeScript however, the first form with a single function would be used:
* `restore(RestoreRequest): RestoreResponse | AcknowledgedResponse`

[...]

PR #1947, opened 2022-10-06
Start of fixes for core/search

sethmlarson:
Part of https://github.com/elastic/elasticsearch-specification/issues/1878

### DIFF
--- a/specification/_global/search/_types/hits.ts
+++ b/specification/_global/search/_types/hits.ts
@@ -39,7 +39,11 @@ import { Sort, SortResults } from '@_types/sort'
 
 export class Hit<TDocument> {
   _index: IndexName
-  _id?: Id
+  /**
+   * @es_quirk '_id' is not available when using 'stored_fields: _none_'
+   * on a search request. Otherwise the field is always present on hits.
+   */
+  _id: Id
   _score?: double | null
   _explanation?: Explanation
   fields?: Dictionary<string, UserDefinedValue>
--- a/specification/_global/search/_types/hits.ts
+++ b/specification/_global/search/_types/hits.ts
@@ -39,7 +39,7 @@ import { Sort, SortResults } from '@_types/sort'
 
 export class Hit<TDocument> {
   _index: IndexName
-  _id: Id
+  _id?: Id
   _score?: double | null
   _explanation?: Explanation
   fields?: Dictionary<string, UserDefinedValue>
--- a/specification/_types/Errors.ts
+++ b/specification/_types/Errors.ts
@@ -49,7 +49,7 @@ export class ErrorCause
 
 export class ShardFailure {
   index?: IndexName
-  node?: string | null
+  node?: string
   reason: ErrorCause
   shard: integer
   status?: string
--- a/specification/_types/Errors.ts
+++ b/specification/_types/Errors.ts
@@ -49,7 +49,7 @@ export class ErrorCause
 
 export class ShardFailure {
   index?: IndexName
-  node?: string
+  node?: string | null
   reason: ErrorCause
   shard: integer
   status?: string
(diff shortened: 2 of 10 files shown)


specification/_types/aggregations/Aggregate.ts
```diff
@@ -571,7 +573,9 @@ export class SignificantLongTermsBucket extends SignificantTermsBucketBase {
 
 /** @variant name=sigsterms */
 // ES: SignificantStringTerms & InternalSignificantTerms
-export class SignificantStringTermsAggregate extends MultiBucketAggregateBase<SignificantStringTermsBucket> {}
+export class SignificantStringTermsAggregate extends MultiBucketAggregateBase<SignificantStringTermsBucket> {
+  bg_count: long
```
swallez, line 577: I already added that one in https://github.com/elastic/elasticsearch-specification/pull/1940

Since it also appears in other `SignificantXxxAggregate`, we may want to factorize it into a `SignificantTermsAggregateBase<T> extends MultiBucketAggregateBase<T>`.

specification/_global/search/_types/hits.ts
```diff
@@ -39,7 +39,7 @@ import { Sort, SortResults } from '@_types/sort'
 
 export class Hit<TDocument> {
   _index: IndexName
-  _id: Id
+  _id?: Id
```
sethmlarson, line 42: This one is super-specific, the only time from what I can see where `_id` is optional on a Hit is when `stored_fields: _none_` which explicitly asks for zero fields on hits.

specification/_types/sort.ts
```diff
@@ -96,7 +98,12 @@ export type SortCombinations = Field | SortOptions
 
 export type Sort = SortCombinations | SortCombinations[]
 
-export type SortResults = Array<long | double | string | null>
+/**
+ * Only used when sorting by "_tsid"
+ */
+export type SortTsid = Dictionary<Field, FieldValue>
+
+export type SortResults = Array<long | double | string | null | SortTsid>
```
swallez, line 106: We should change this to `Array<FieldValue>`, because this is what it actually is, and `any` was recently added to `FieldValue`.

specification/_spec_utils/WaitForCompletion.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/**
+ * Union of two responses depending on input parameters
+ * and whether or not the server-side action completes
+ * before sending a response. Without setting any parameters
+ * that start with 'wait_for_*' the 'Complete' response is
+ * always returned.
+ */
+export type WaitForCompletion<Complete, Incomplete> = Complete | Incomplete
```
swallez, line 27: Should it be `Delayed` or `Accepted` instead of `Incomplete`?

Actually I'd rather delay this change to a further release, since it needs some special handling in code generators to provide a good developer experience.

For `wait_for_completion` (boolean) I'd like to generate two different endpoints that each hard-code the value of the `wait_for_completion` parameter and return only one side of the union.

Also two endpoints for wait_for_completion_timeout: one with the parameter not set that would return the `Complete` part, and another one with the timeout parameter that would return the union.

I'll capture this in https://github.com/elastic/elasticsearch-specification/issues/1767

sethmlarson: Oh dang, I didn't mean to include this in this PR, I'll section this off into it's own PR.

swallez (commented):
Looks good so far!

specification/_types/Errors.ts
```diff
@@ -49,7 +49,7 @@ export class ErrorCause
 
 export class ShardFailure {
   index?: IndexName
-  node?: string
+  node?: string | null
```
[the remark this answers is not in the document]
sethmlarson: Yeah this was in one of the recordings not a fix for Java specifically.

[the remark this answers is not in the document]
sethmlarson: I went looking for the test this is from and couldn't find it so reverted this change for now.

swallez (approved):
LGTM

PR #1951, opened 2022-10-06
SLM config and retention are optional

### LINKED ISSUE #1917: SLM Java Client Missing Required Policy Config Element (indices)
It seems that SLM Java Client is missing Required Policy Config Element (indices).

Example: 
Use the GUI interface to set up the SLM policy and take a snapshot of everything. 

When a user tries to use it in an application to validate that the SLM is running properly, it's not able to get the lifecycle because they don't have an `indices` element on the SLM policy configuration.
The documentation seems to say that the indices element on the policy config is optional, **but the Java api requires it.**
Tried using the `ApiTypeHelper.DANGEROUS_disableRequiredPropertiesCheck(true);` to try and get around this, but then got a null pointer exception.

swallez:
Fixes #1917

The `config` and `retention` fields in the SLM policy are optional, and also `indices` in `config`.

### DIFF
--- a/specification/slm/_types/SnapshotLifecycle.ts
+++ b/specification/slm/_types/SnapshotLifecycle.ts
@@ -74,10 +74,10 @@ export class Statistics {
 }
 
 export class Policy {
-  config: Configuration
+  config?: Configuration
   name: Name
   repository: string
-  retention: Retention
+  retention?: Retention
   schedule: CronExpression
 }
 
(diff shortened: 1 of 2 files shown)


specification/slm/_types/SnapshotLifecycle.ts
```diff
@@ -74,10 +74,10 @@ export class Statistics {
 }
 
 export class Policy {
```
sethmlarson, line 76: Is this another case of `Policy`/`PolicyRead`? If so can we split into two and use `OverloadOf`?

swallez: The issue was reported for a read operation, and I checked the code: parsing and serialization are symmetrical and have the same tests for those optional properties.

sethmlarson (approved):
LGTM

PR #1970, opened 2022-10-19
Fix optional Processor fields and add missing fields

swallez:
Fixes the optionality of ingest processor fields (most of them are optional) and add a few missing ones.

Checked with [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/processors.html).


Anaethelion (approved):
LGTM

PR #1981, opened 2022-11-01
Enforce path parameter required/optional with JSON specs

sethmlarson:
Path parameters aren't checked whether they're required or optional compared to the paths provided in the JSON specs. Found this through the newly added API `ml.clear_model_deployment_cache` which mistakenly was added with an optional path parameter `model_id` but the parameter should be marked as required in the spec.

### DIFF
--- a/compiler/src/steps/validate-rest-spec.ts
+++ b/compiler/src/steps/validate-rest-spec.ts
@@ -69,7 +69,7 @@ export default async function validateRestSpec (model: model.Model, jsonSpec: Ma
       }
 
       // are all path parameters properly required or optional?
-      let urlPartsRequired = new Set(urlParts.slice())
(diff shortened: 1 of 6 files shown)


compiler/src/steps/validate-rest-spec.ts
```diff
@@ -68,6 +68,35 @@ export default async function validateRestSpec (model: model.Model, jsonSpec: Ma
         }
       }
 
+      // are all path parameters properly required or optional?
+      let urlPartsRequired = new Set(urlParts.slice())
```
swallez, line 72: nit: no need to `slice()` to copy the array, as `new Set()` iterates on it, but doesn't modify it.

swallez (approved):
LGTM

PR #1988, opened 2022-11-06
Add 'time' parameter to 'cat.indices' API

miriam-eid:
Adds the missing `time` parameter to the `cat.indices` API.


specification/cat/indices/CatIndicesRequest.ts
```diff
@@ -36,5 +37,6 @@ export interface Request extends CatRequestBase {
     health?: HealthStatus
     include_unloaded_segments?: boolean
     pri?: boolean
+    time?: Duration
```
swallez, line 40: This should be `TimeUnit`, as the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-indices.html) says "Unit used to display time values" and links to [time unit definitions](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#time-units).

swallez (changes requested):

swallez (approved):
LGTM!

PR #1992, opened 2022-11-15
Dump routes to Go radix tree

Anaethelion:
This PR introduces a cli target that outputs the Elasticsearch routes as a Go radix-tree.


compiler/src/dump/extract-routes.ts
```diff
@@ -0,0 +1,550 @@
...
+  const str = serializeForest(forest)
+
+  await writeFile(outPath, str)
+}
+
+extractRoutesFromFiles(outputPath)
+  .catch(reason => console.error(reason))
+  .finally(() => console.log('Routes extraction complete.', outputPath))
+
+function defaultRoutes (): Trees {
```
sethmlarson, line 310: This defaultRoutes routine doesn't need to be kept up-to-date? Is this only used when we don't know the version of the server or for something else?

sethmlarson (approved):
Looks excellent! :star_struck: Only had one question:

compiler/src/dump/extract-routes.ts
Anaethelion: Yep, that's the one! I've extracted that list from existing targets.

PR #1997, opened 2022-12-12
Change StorageType to an open enumeration

swallez:
The `StorageType` enumeration is non exhaustive: some plugins may add additional storage types, and snapshot/frozen repositories, although they cannot be created directly, are identified with a storage type `snapshot`.

Reported in https://github.com/elastic/elasticsearch-java/issues/458


Anaethelion (approved):
LGTM!

PR #2008, opened 2023-01-05
Remove non-existent JSON specs when updating from artifacts API

sethmlarson:
Currently if a JSON spec is deleted upstream (like was done to `semantic_search.json` for 8.6) then the `download-artifacts.js` script doesn't detect and delete the file. This amends that missing behavior.


Anaethelion (approved):
LGTM

PR #2019, opened 2023-02-02
Update API spec from results of the Java flight records loader

swallez:
A collection or API spec fixes using flight-recorder recordings from a recent 8.7.0 snapshot and running them through the Java client's flight recordings validator. This validator is more strict than our JS/TS validator, and also reports unknown fields.

I've added the `backport 7.17` label, even if only a limited number of fixes actually apply to that branch. I'll cherry-pick those that are relevant.

### DIFF
--- a/specification/_global/search/_types/profile.ts
+++ b/specification/_global/search/_types/profile.ts
@@ -19,7 +19,21 @@
 
 import { integer, long } from '@_types/Numeric'
 import { DurationValue, UnitNanos } from '@_types/Time'
-import { Dictionary } from '@spec_utils/Dictionary'
+
+export class AggregationBreakdown {
+  build_aggregation: long
+  build_aggregation_count: long
+  build_leaf_collector: long
+  build_leaf_collector_count: long
+  collect: long
+  collect_count: long
+  initialize: long
+  initialize_count: long
+  post_collection?: long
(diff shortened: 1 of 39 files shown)


specification/_global/search/_types/profile.ts
```diff
@@ -75,7 +61,7 @@ export class AggregationProfileDelegateDebugFilter {
 }
 
 export class AggregationProfile {
-  breakdown: AggregationBreakdown
+  breakdown: Dictionary<string, long>
```
swallez, line 64: I made it a generic dictionary since this is what it is on the ES server, and the field names evolve quite often.

Mpdreamz: It'd be nice if we could ship the spec with an `AggregrationBreakdownKeys` that encodes known key constants into these dictionaries.

sethmlarson (commented):
I think I prefer the breakdowns as objects here over dictionaries since my understanding is the values aren't variable? Otherwise LGTM

swallez:
I've reverted the changes to breakdowns, so that we can move forward with the other changes and take the time to think carefully about them. I've opened https://github.com/elastic/elasticsearch-specification/issues/2021 for this.

sethmlarson (approved):
LGTM

PR #2026, opened 2023-03-07
Fix CalendarInterval year alias

esdrasbrz:
This PR fixes the `AggregationsCalendarInterval` type with `1y` instead of `1Y`. ElasticSearch 8.6 returns the following error when `1Y` is used:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "x_content_parse_exception",
        "reason": "[41:30] [date_histogram] failed to parse field [calendar_interval]"
      }
    ],
    "type": "x_content_parse_exception",
    "reason": "[41:30] [date_histogram] failed to parse field [calendar_interval]",
    "caused_by": {
      "reason": "The supplied interval [1Y] could not be parsed as a calendar interval."
    }
  },
  "status": 400
}
```

Also, from [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#calendar_intervals), the correct interval is `1y`

Closes #2627


esdrasbrz:
I'm not sure about the backports. It seems it needs backports as the API is only partially usable, but I can't change the PR labels.
LMK if there's something else I need to do.

pquentin (approved):
Great catch, thank you! And sorry for being so terribly late.

PR #2028, opened 2023-03-08
[KNN] Add query vector builder

Anaethelion:
This adds the `query_vector_builder` field to the KnnQuery for semantic search.
https://www.elastic.co/guide/en/elasticsearch/reference/master/knn-search.html#semantic-search

### DIFF
--- a/specification/_types/Knn.ts
+++ b/specification/_types/Knn.ts
@@ -21,11 +21,13 @@ import { Field } from '@_types/common'
 import { long, double, float } from '@_types/Numeric'
 import { QueryContainer } from './query_dsl/abstractions'
 
+export type QueryVector = float[]
+
 export interface KnnQuery {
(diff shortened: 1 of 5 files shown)


specification/_types/Knn.ts
```diff
@@ -35,3 +37,12 @@ export interface KnnQuery {
   /** Filters for the kNN search query */
   filter?: QueryContainer | QueryContainer[]
 }
+
+export interface QueryVectorBuilder {
```
swallez, line 41: On the server this is actually a variant. Although `text_embedding` is currently the only implementation, we may want to add a `@variants container` (and make `text_embedding` optional) to allow for future expansion.

swallez (changes requested):

swallez (approved):

PR #2033, opened 2023-03-13
Remove union for dynamic_templates

stevejgordon:
Creating an index with a mapping that includes `dynamic_templates` fails when sending the dictionary without wrapping it in an array: 

e.g.

```
PUT my-index
{
  "mappings": {
    "dynamic_templates": {
      "testTemplateName": {
        "mapping": {
          "type": "keyword"
        },
        "path_match": "testPathMatch"
      }
    }
  }
}
```

Error:

```
{
  "error": {
    "root_cause": [
      {
        "type": "mapper_parsing_exception",
        "reason": "Dynamic template syntax error. An array of named objects is expected."
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "Failed to parse mapping: Dynamic template syntax error. An array of named objects is expected.",
    "caused_by": {
      "type": "mapper_parsing_exception",
      "reason": "Dynamic template syntax error. An array of named objects is expected."
    }
  },
  "status": 400
}
```

I've made a check to the spec but can't validate locally, so pushing this WIP PR, for now, to see how the tests run in CI.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -5049,7 +5049,7 @@ export interface MappingTypeMapping {
   date_detection?: boolean
   dynamic?: MappingDynamicMapping
   dynamic_date_formats?: string[]
-  dynamic_templates?: Record<string, MappingDynamicTemplate> | Record<string, MappingDynamicTemplate>[]
+  dynamic_templates?: Record<string, MappingDynamicTemplate>[]
   _field_names?: MappingFieldNamesField
   index_field?: MappingIndexField
   _meta?: Metadata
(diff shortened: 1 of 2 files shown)


stevejgordon:
Relates to https://github.com/elastic/elasticsearch-net/issues/7234

stevejgordon:
I've been able to run the validation locally for all requests/responses that use this type, and there are no breaks, so this looks valid based on the tests.

swallez (approved):
LGTM. This is one of the few arrays that doesn't allow a single object for 1-sized arrays

PR #2044, opened 2023-03-23
Fix completion suggester

stevejgordon:
Fixes issues raised in https://github.com/elastic/elasticsearch-net/issues/7454.

The completion suggester doesn't support a field named "prefix" which is a FieldSuggester container property.

After reviewing the server code, I noticed that `regex` is not a string but an instance of `RegexOptions` used to control the regex provided via the container property.

### DIFF
--- a/specification/_global/search/_types/suggester.ts
+++ b/specification/_global/search/_types/suggester.ts
@@ -135,7 +135,7 @@ export class CompletionSuggester extends SuggesterBase {
 }
 
 export class RegexOptions {
-  flags?: string
+  flags?: integer | string
   max_determinized_states?: integer
 }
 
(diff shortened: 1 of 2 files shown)


specification/_global/search/_types/suggester.ts
```diff
@@ -130,11 +130,15 @@ export class SuggesterBase {
...
   contexts?: Dictionary<Field, CompletionContext | CompletionContext[]>
   fuzzy?: SuggestFuzziness
-  prefix?: string
-  regex?: string
+  regex?: RegexOptions
   skip_duplicates?: boolean
 }
 
+export class RegexOptions {
+  flags?: string
```
Anaethelion, line 138: Server code seems to suggest this could also be an integer.

PR #2047, opened 2023-03-29
Generate OpenAPI schema for Elasticsearch Serverless

swallez:
This PR adds a converter from `schema.json` to OpenAPI for Elasticsearch serverless.

`make contrib` now produces the `output/openapi/elasticsearch-serverless-openapi.json` file.


openapi-converter/openapi_to_clients_schema/src/openapi.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    ///
+    /// Is this type nullable?
+    ///
+    pub fn is_nullable(&self, id: &str) -> bool {
+        self.components
+            .as_ref()
+            .and_then(|c| c.schemas.get(id))
+            .map(|s| match s {
+                ReferenceOr::Reference { reference: _r } => false,
```
sethmlarson, line 74: Should this reference call recurse into the referenced type to determine if it's nullable?

openapi-converter/openapi_to_clients_schema/src/openapi.rs
```diff
@@ -0,0 +1,408 @@
...
+            all_of: merge_vec(acc.all_of, schema.all_of),
+            not: merge_option(acc.not, schema.not)?,
+
+            // Structure properties
+            additional_properties: merge_option(acc.additional_properties, schema.additional_properties)?,
+            items: merge_option(acc.items, schema.items)?,
+
+            unique_items: merge_option(acc.unique_items, schema.unique_items)?,
+            properties: Default::default(),
+            max_properties: merge_option(acc.max_properties, schema.max_properties)?,
```
sethmlarson, line 139: Idle thought: `max_properties` and `min_properties` might have special merging behavior even when the specified values aren't identical, maybe some others too? Probably don't have to handle this now until it comes up in a real spec.

openapi-converter/openapi_to_clients_schema/src/types.rs
```diff
@@ -0,0 +1,542 @@
...
+
+        // FIXME: need to set mappings, by setting the discriminator property in every variant to either
+        // the type name or the key of in discriminator.mapping
+    }
+
+    let type_alias = TypeAlias {
+        base,
+        typ: ValueOf::UnionOf(union_of),
+        generics: Vec::default(),
+        variants: variants, // May be set below
```
sethmlarson, line 336: nit: Is this comment out of date?

openapi-converter/openapi_to_clients_schema/src/types.rs
```diff
@@ -0,0 +1,542 @@
...
+    id: &str,
+    base: BaseType,
+    required: &Vec<String>,
+    properties: &IndexMap<String, ReferenceOr<Box<Schema>>>,
+    additional_properties: &Option<AdditionalProperties>,
+    types: &mut Types
+) -> anyhow::Result<()> {
+    // Regular type
+
+    // FIXME: max_properties -- if == 1 indicates a container
```
sethmlarson, line 384: Not necessarily with container properties?

sethmlarson (commented):
This is excellent, nice work @swallez! Left a few comments inline, some may not make sense due to unfamiliarity with Rust and this code:

jamietanna:
Great stuff @swallez :clap: I notice that there are a few errors (denoting actual issues with the produced formatting, rather than "nice to have"s) when linting the generated OpenAPI with https://github.com/stoplightio/spectral - just wondering if you're aware of or are validating the schema with any other tools?

swallez:
@jamietanna thanks for the pointer, Spectral is a great linter, picky as it should.

The conversion code has been updated and the linter is happy. The OpenAPI output now comes in two flavors, stack (full API) and serverless.

I still have to implement a few `todo` for some corner cases before merging this work, but it's now safe to start using the results.

jamietanna:
Awesome stuff @swallez :clap: Appreciate the tweaks.

A couple of plumbing points to note:

I believe the circular references may still be an issue (in `elasticsearch-openapi.json`):
- Code generation with https://github.com/deepmap/oapi-codegen/ v1.15.0 fails without configuring a high circular reference count
- Linting with https://github.com/daveshanley/vacuum fails (it crashes :sweat_smile:)

```
: kin-openapi bug found: circular schema reference not handled - #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:SpanFieldMaskingQuery -> #/components/schemas/_types.query_dsl:SpanQuery -> #/components/schemas/_types.query_dsl:SpanNotQuery -> #/components/schemas/_types.query_dsl:SpanQuery -> #/components/schemas/_types.query_dsl:SpanWithinQuery -> #/components/schemas/_types.query_dsl:SpanQuery -> #/components/schemas/_types.query_dsl:SpanContainingQuery -> #/components/schemas/_types.query_dsl:SpanQuery -> #/components/schemas/_types.query_dsl:SpanFirstQuery -> #/components/schemas/_types.query_dsl:SpanQuery -> #/components/schemas/_types.query_dsl:SpanMultiTermQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:FunctionScoreQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:PinnedQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:ConstantScoreQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:ScriptScoreQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:RuleQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:HasChildQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:BoostingQuery -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:HasParentQuery -> #/components/schemas/_global.search._types:InnerHits -> #/components/schemas/_types:Sort -> #/components/schemas/_types:SortCombinations -> #/components/schemas/_types:SortOptions -> #/components/schemas/_types:ScriptSort -> #/components/schemas/_types:NestedSortValue -> #/components/schemas/_types:NestedSortValue -> #/components/schemas/_types.query_dsl:QueryContainer -> #/components/schemas/_types.query_dsl:DisMaxQuery -> #/components/schemas/_types.query_dsl:QueryContainer
```

Some other validators (like `spectral`) appear to work OK.

There appears to be a clash in one of the related `snapshot.create_repository` types, but that may be a bug in the code generator.

Key errors I can see in Spectral are:

```
/home/jamie/workspaces/elasticsearch-specification/output/openapi/elasticsearch-openapi.json
 20265:24  error  path-params  Paths "/_nodes/{node_id}" and "/_nodes/{metric}" must not be equivalent.  paths./_nodes/{metric}
/home/jamie/workspaces/elasticsearch-specification/output/openapi/elasticsearch-serverless-openapi.json
 15946:24  error  path-params  Paths "/_nodes/{node_id}" and "/_nodes/{metric}" must not be equivalent.  paths./_nodes/{metric}
```

It loosk like `elasticsearch-serverless-openapi.json` can be used with `oapi-codegen`, but it generates with some compliation errors, so I'll look at raising a bug for that.

I also note that we've got generated code like:

```json
    "/_async_search/{id}": {
      "get": {
        "description": "Retrieves the results of a previously submitted async search request given its ID.",
        "externalDocs": {
          "url": "https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html"
        },
        "operationId": "async_search.get#0",
```

and

```json
    "schemas": {
      "_types:Id": {
        "type": "string"
      },
```
Is there anything we can do to maybe massage these into more human-readable names? This may make the generated code that's used with these a little better.

swallez:
@jamietanna thanks for the feedback. 

Regarding the failure of linters and code generators, this has to be put of into perspective with the context and and purpose of this conversion:

* there's no way we can remove circular dependencies as they're heavily used in the Elasticsearch query DSL. A boolean query contains nested queries, and aggregations can contain nested aggregations. Circular dependencies are indeed an issue to generate code in languages like Go/Rust/C++ that support both structure embedding and structure linking with a reference. This is one of the complex parts of the generator for [the Go client](https://github.com/elastic/go-elasticsearch). So if a linter or code generator fail on circular types used to represent hierarchical recursive data structures, I consider this to be a bug in these tools, and not a but in the OpenAPI schema 😉 

* this OpenAPI specification is not meant to generate code: OpenAPI cannot capture a number of the idiosyncrasies of the ES API or language features like generic type parameters that allow generating usable and idiomatic code in the 8 languages we support. This is why the clients team came up with this unconventional approach of using TypeScript to capture the Elasticsearch API request and response bodies. The conversion of this bespoke schema to OpenAPI is meant primarily for generating documentation and other materials for humans (e.g. Kibana auto-completion, CLIs, etc) where these quirks and fine-grained details are not of primary importance and existing OpenAPI tooling can be leveraged.

Regarding the other points you mention:

* The `"/_nodes/{node_id}" and "/_nodes/{metric}" must not be equivalent` error is caused by how the [Node info API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html) works: the placeholder value can be either a node identifier or a metric name. The server will [check if it's a metric name and consider it as a node id otherwise](https://github.com/elastic/elasticsearch/blob/16a4e542f09acef5e69daf41f5431bf2c4096dc2/server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestNodesInfoAction.java#L68). It's important to keep both variants when generating client libraries, even these paths are ambiguous from the server point of view.

* `"/_async_search/{id}"` and `"_types:Id"`: I'm not sure what you mean here. Is this about removing the `_types:Id` type and just use `string`? There are [a number of type aliases](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/common.ts) in the ES API specification meant to provide higher-level semantics to primitive types. Not sure if we should lose this information in the generated OpenAPI as it can be useful in some use cases. So maybe an option in the converter to resolve these type aliases into their primitive target?

Anaethelion (approved):
I'd change the name of the main directory from `openapi-converter` to something meaningful like `compiler-rs`. There's much more than OpenApi in that folder.

Except for a nit left in a comment, looks great!

openapi-converter/clients_schema_to_openapi/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Output file. Defaults to stdout.
+    output: Option<PathBuf>,
+
+    /// Elasticsearch flavor to produce
+    #[arg(short, long)]
+    flavor: Option<SchemaFlavor>
+}
+
+#[derive(Debug, Clone, PartialEq, ValueEnum)]
+pub enum SchemaFlavor {
```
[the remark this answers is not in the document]
swallez: Because there's we're likely to have a `Common` code generation target in the future (currently commented-out) that will be used to retain what's in both Stack and Serverless. So `client_schema::Flavor` represents the `@availability` annotation whereas here this represents the generation flavor that can be different from a simple XOR between Stack and Serverless.

PR #2070, opened 2023-04-12
Add composite fields to search API

### LINKED ISSUE #2059: Interface for RuntimeField missing string option
## 🐛 Wrong type

Interface for RuntimeFieldType missing string option "composite"

## Definition

Since Elasticsearch 7.15 the[ Runtime fields support "composite"](https://www.elastic.co/guide/en/elasticsearch/reference/7.15/runtime-mapping-fields.html#runtime-mapping-fields) fields: 

```
The runtime section can be any of these data types:

    boolean
    composite
    date
    double
    geo_point
    ip
    keyword
    long
```
And the RuntimeField / RuntimeFieldBuilder is missing the related "fields" field / method.

Has this been forgotten to update or have "composite" Runtime fields been replaced by something else?

See also:
https://www.elastic.co/guide/en/elasticsearch/reference/8.7/search-search.html
"runtime_mappings"

runtime_mappings

```
    (Optional, object of objects) Defines one or more [runtime fields](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/runtime-search-request.html) in the search request. These fields take precedence over mapped fields with the same name.
    Properties of runtime_mappings objects

    <field-name>

        (Required, object) Configuration for the runtime field. The key is the field name.
        Properties of <field-name>

        type

            (Required, string) [Field type](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/mapping-types.html), which can be any of the following:

                boolean
                composite
                date
                double
                geo_point
                ip
                keyword
                long
                [lookup](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/runtime-retrieving-fields.html#lookup-runtime-fields)

        script

            (Optional, string) [Painless script](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/modules-scripting-using.html) executed at query time. The script has access to the entire context of a document, including the original _source and any mapped fields plus their values.

  
```

MichaelOpitz2:
Closes #2059, closes https://github.com/elastic/elasticsearch-specification/pull/2812

Re-add composite runtime fields according to spec: [Runtime fields](https://www.elastic.co/guide/en/elasticsearch/reference/7.15/runtime-mapping-fields.html#runtime-mapping-fields)

The spec supports composite runtime fields sinde 7.15 but it seems they never made it into the new clients.
I am not sure, if these changes need to be backported, at least, composite runtime fields do not work in every version of the generated clients. I suppose number three applies: API is _partially_ usable and the fix unlocks a missing feature that has no workaround


pquentin:
Sorry for the delay! There were two changes in this pull request:

 1. Adding `composite` to `RuntimeFieldType`, which was done in https://github.com/elastic/elasticsearch-specification/pull/2485 and was applied to all 8.13 clients
 2. Adding `fields?: Dictionary<string, Dictionary<string, RuntimeFieldType>>` to `RuntimeField`

I merged main into this branch, so now there's only the second change left, and we need to review it.

pquentin:
So I studied composite runtime fields, introduced in https://github.com/elastic/elasticsearch/pull/75108. (Note that they're totally different from composite aggregations, despite sharing the same name.)

The point of a composite runtime field is to have one script which emits multiple values, and place those values in different fields: https://www.elastic.co/guide/en/elasticsearch/reference/8.15/runtime-examples.html#runtime-examples-grok-composite. It looks like this:

```json
PUT my-index-000001/_mappings
{
  "runtime": {
    "http": {
      "type": "composite",
      "script": "emit(grok(\"%{COMMONAPACHELOG}\").extract(doc[\"message\"].value))",
      "fields": {
        "clientip": {
          "type": "ip"
        },
        "verb": {
          "type": "keyword"
        },
        "response": {
          "type": "long"
        }
      }
    }
  }
}
```

The possible types, as correctly identified by @MichaelOpitz2, are the composite field types. However, one question remained: is it possible to use something else than "type"? The code is hard to read because the code to read the runtime fields is reused for composite fields, but [without allowing fields deeper than one level](https://github.com/elastic/elasticsearch/pull/75108#issuecomment-908845532). Also, whenever I tried using something else than type, I got errors. Two examples:

 * `Found sub-fields with name not belonging to the parent field they are part of [response]`
 * `Failed to parse mapping: Cannot use [script] parameter on sub-field [verb] of composite field [http]`

And finally, YAML tests about composite runtime fields only use type: see [110_composite.yml](https://github.com/elastic/elasticsearch/blob/main/modules/runtime-fields-common/src/yamlRestTest/resources/rest-api-spec/test/runtime_fields/110_composite.yml) and [114_composite_errors.yml](https://github.com/elastic/elasticsearch/blob/main/modules/runtime-fields-common/src/yamlRestTest/resources/rest-api-spec/test/runtime_fields/114_composite_errors.yml). (We don't currently use these files for validation.)

My conclusion is that only "type" is allowed, so I [pushed a commit](https://github.com/elastic/elasticsearch-specification/pull/2070/commits/8d1477d96b5c65c72bad1275aeab5b2736209951#diff-e7f4c495919c4b562400cbee4de3f60fbf6676f64c175ac92636ebebb4c909ee) to clarify that.

l-trotta (approved):
LGTM!

PR #2073, opened 2023-04-14
Fix IndexPrivileges.field_security

sethmlarson:
This field doesn't accept an array of `FieldSecurity` instances, only a single instance. This causes an issue in strongly-typed languages that collapse `X | X[]` type into `X[]` because users can't set a single object on a request.


swallez (approved):
LGTM

PR #2082, opened 2023-04-19
Add Search Applications API

carlosdelest:
Add Search Applications API, including both CRUD and _search request.

Reopening this from https://github.com/elastic/elasticsearch-specification/pull/2080, as Github actions need the PR to be done from the same repo


specification/search_application/get/SearchApplicationsGetRequest.ts
```diff
@@ -0,0 +1,35 @@
...
+ * under the License.
+ */
+import { RequestBase } from '@_types/Base'
+import { Name } from '@_types/common'
+
+/**
+ * Returns the details about a search application
+ * @rest_spec_name search_application.get
+ * @since 8.8.0
+ * @stability beta
```
kderusso, line 26: I think we need to update these `@stability` tags too

kderusso (commented):
Looks great so far, a couple comments! 💯

sethmlarson (approved):
This looks good to me!

@kderusso We'll eventually need to update the stability tags, but I was thinking we could get this PR merged and then I'll manually resync the `rest-api-spec` stubs before fixing the stabilities. Otherwise we need to resync, rebase, and all that, might be easier this way?

carlosdelest:
@kderusso , as @sethmlarson said, this can be done in a separate PR with the syncs already done. I'm merging this PR and we'll take care of that in a separate one 👍

PR #2085, opened 2023-04-20
Stringify relevant properties

Anaethelion:
While hardening the JSON decode tests for the Go client I've stumbled upon several instance of integer and one boolean that can also be a string in some instances.


swallez (approved):
LGTM

PR #2088, opened 2023-04-21
Move generic array to root of response body

Anaethelion:
Following tests and [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-preview-datafeed.html), ml.preview_data_feed should have the array of generics at the root of its response.


swallez (approved):
LGTM

PR #2089, opened 2023-04-21
Add 408 as exception status code for normal behaviour on cluster_health

Anaethelion:
`cluster_health` could return a valid payload with a `408` status code if one of the nodes from the cluster failed to respond before the timeout.

### DIFF
--- a/specification/cluster/health/ClusterHealthResponse.ts
+++ b/specification/cluster/health/ClusterHealthResponse.ts
@@ -27,16 +27,16 @@ import { IndexHealthStats } from './types'
  * @doc_id cluster-health
  */
 export class Response {
-  body: ResponseBody
+  body: HealthResponseBody
   exceptions: [
     {
       statusCodes: [408]
-      body: ResponseBody
+      body: HealthResponseBody
     }
   ]
 }
 
-export class ResponseBody {
+export class HealthResponseBody {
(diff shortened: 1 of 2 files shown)


specification/cluster/health/ClusterHealthResponse.ts
```diff
@@ -27,38 +27,46 @@ import { IndexHealthStats } from './types'
...
-    /** The time since the earliest initiated task is waiting for being performed. */
-    task_max_waiting_in_queue?: Duration
-    /** The time expressed in milliseconds since the earliest initiated task is waiting for being performed. */
-    task_max_waiting_in_queue_millis: DurationValue<UnitMillis>
-    /** If false the response returned within the period of time that is specified by the timeout parameter (30s by default) */
-    timed_out: boolean
-    /** The number of shards that are not allocated. */
-    unassigned_shards: integer
-  }
+  body: ResponseBody
```
swallez, line 30: Why moving it to a different type? If there's a good reason for this, we should at least name it with a more meaningful `HealthResponseBody`.

swallez: Doh, I did not see it's also used for the 408 response. Naming still applies ;-)

swallez (approved):
LGTM!

PR #2110, opened 2023-05-10
Added Behavioral Analytics APIs

carlosdelest:
Add Behavioral Analytics APIs.

POST is missing - I understood from @afoucret and @joemcelroy that this API should not yet be available for clients. It can be contributed as a separate PR however.

### DIFF
--- a/specification/search_application/get_behavioral_analytics/BehavioralAnalyticsGetRequest.ts
+++ b/specification/search_application/get_behavioral_analytics/BehavioralAnalyticsGetRequest.ts
@@ -1,0 +1,35 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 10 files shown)


specification/search_application/get_behavioral_analytics/BehavioralAnalyticsGetRequest.ts
carlosdelest, line 1: Overall question for @sethmlarson : Why are these APIs not grouped at the REST Spec level on their own namespace, like `behavioral_analytics.get`?

In my mind, they make more sense grouped separately from search applications. WDYT?

sethmlarson: If they aren't a submodel of a Search Application they can have their own top-level namespace. In that case I'd recommend having a different root path segment like `/_analytics` or `/_behavioral_analytics`.

sethmlarson (approved):
These changes as-is LGTM, will need to be updated if the API gets its namespace changed.

specification/search_application/get_behavioral_analytics/BehavioralAnalyticsGetRequest.ts
carlosdelest: Right now we have:
`/_application/search_applications` for search apps, and 
`/_application/analytics` for analytics.

I'd suggest we group them on a different namespace, or that both search apps and analytics have a common `application` namespace.

WDYT?

carlosdelest:
Hey @sethmlarson , I added a PR for changing the namespace [here](https://github.com/elastic/elasticsearch/pull/96225). Can you please double check it?

Once the above is merged, I'll update this PR with the changes.

carlosdelest:
I'll merge this PR as-is, and we can deal with namespace changes later

PR #2111, opened 2023-05-11
Make index settings blocks stringified

stevejgordon:
Raised in https://github.com/elastic/elasticsearch-net/issues/7714, it appears that the possible block values should all be stringified as they are each produced as an `APIBlock` ([see server code](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java#L264)).

This PR updates each possible block option to support the stringified value when getting index settings.


Anaethelion (approved):
LGTM

Issue #2107, opened 2023-05-04
Support excluding or including API and type definitions based on pre-defined annotations


sethmlarson:
## Feature Proposal

Add support for new JSDoc annotations on API endpoints and individual properties based on which API surface includes/excludes the feature being annotated. The proposed annotations would be `@include_if` and `@exclude_if` which each take a list of annotation names (ie `serverless`, `serverless-public`, `stack`)

Then the compiler can be called with an option `--annotations` with the same annotation names (ie `compiler --annotations serverless serverless-public` would generate the `schema.json` for public Serverless Elasticsearch API).

The motivation for doing this is to continue to have a strong cohesion between Stack and Serverless API surfaces while allowing for small altercations depending on the service.

The new JSDocs tags will be applicable to API endpoints and properties, implemented in that order, in order for other teams to start adding these annotations to APIs they own.

These annotations will not manifest in the `schema.json` output as new fields, instead the types and APIs will be modified so that client generators don't need to be aware of these differences.

## Example

Example that is only available internally for Serverless and publicly for Stack Elasticsearch.

```ts
/**
 * @rest_spec_name example
 * @since 0.0.0
 * @stability stable
 * @exclude_if serverless-public
 */
export interface Request {}
```

Example of an API that is only available on Stack Elasticsearch.

```ts
/**
 * @rest_spec_name example
 * @since 0.0.0
 * @stability stable
 * @exclude_if serverless
 */
export interface Request {}
```

Example that is only available on Serverless Elasticsearch

```ts
/**
 * @rest_spec_name example
 * @since 0.0.0
 * @stability stable
 * @include_if serverless
 */
export interface Request {}
```

Example of a property that is only available on Stack Elasticsearch

```ts
export interface Request {
  query_parameters: {
    /**
     * @exclude_if serverless
     */
    preference?: string
  }
}
```

tvernum:
A few thoughts:

`include_if` feels weird to me because the spec should describe the API rather than the code generation process. That is, I feel like we want to say _this request is `available_in serverless` rather than `include_if serverless`.

If also find it weird that we'll have a `@visibility` annotation that only means something on stack (because serverless visibility is handled by these new tags instead). 
Likewise `@since` will be weird on serverless - if there's a serverless only API what does `@since` mean?

I wonder whether we'll ever have a need for other annotations to have different values on serverless. For example I suspect we could need to have different `@stability` (an API that is stable on serverles but not yet GA in stack).

[...]

PR #2122, opened 2023-05-23
Add support for the `@availability` annotations

sethmlarson:
Closes https://github.com/elastic/elasticsearch-specification/issues/2107

- Removed the individual `@since`, `@stability`, and `@visibility` annotations in favor of `@availability` with those annotations now being set via key-value pairs.
- Always set `availability.stack.*` to the values from the rest-api-spec stubs for now, in the future once the Elasticsearch specification becomes the source of truth we can remove this.
- Updates all existing request definitions to use the new annotation.

### DIFF
--- a/compiler/src/model/metamodel.ts
+++ b/compiler/src/model/metamodel.ts
@@ -369,7 +369,7 @@ export enum Stability {
 }
 export enum Visibility {
   public = 'public',
-  featureFlag = 'feature_flag',
+  feature_flag = 'feature_flag',
   private = 'private'
 }
 
@@ -378,12 +378,30 @@ export class Deprecation {
   description: string
 }
 
+export class Availability {
+  stack?: StackAvailability
+  serverless?: ServerlessAvailability
+}
+
+export class StackAvailability {
+  since?: string
+  featureFlag?: string
+  stability?: Stability
+  visibility?: Visibility
+}
+
+export class ServerlessAvailability {
+  stability?: Stability
+  visibility?: Visibility
+}
+
 export class Endpoint {
   name: string
   description: string
   docUrl: string
   docId?: string
   deprecation?: Deprecation
+  availability: Availability
 
   /**
    * If the request value is `null` it means that there is not yet a
--- a/compiler/src/model/metamodel.ts
+++ b/compiler/src/model/metamodel.ts
@@ -378,19 +378,14 @@ export class Deprecation {
   description: string
 }
 
-export class Availability {
-  stack?: StackAvailability
-  serverless?: ServerlessAvailability
+export class Availabilities {
+  stack?: Availability
+  serverless?: Availability
 }
 
-export class StackAvailability {
+export class Availability {
   since?: string
   featureFlag?: string
-  stability?: Stability
-  visibility?: Visibility
-}
-
-export class ServerlessAvailability {
   stability?: Stability
   visibility?: Visibility
 }
@@ -401,7 +396,7 @@ export class Endpoint {
   docUrl: string
   docId?: string
   deprecation?: Deprecation
-  availability: Availability
+  availability: Availabilities
 
   /**
    * If the request value is `null` it means that there is not yet a
--- a/compiler/src/model/utils.ts
+++ b/compiler/src/model/utils.ts
@@ -1069,21 +1069,12 @@ export function parseAvailabilityTags (node: Node | Node[], values: string[]): m
     // Since we're using reduce() we need to check that there's no duplicates.
     assert(node, !(availabilityName in result), `Duplicate @availability tag: '${availabilityName}'`)
 
-    // Based on which availability definition we're parsing we have different valid tags.
-    let validKeys: string[] = []
-    switch (availabilityName) {
-      case 'stack':
-        validKeys = ['stability', 'visibility', 'since', 'feature_flag']
-        break
-      case 'serverless':
-        validKeys = ['stability', 'visibility']
-        break
-      default:
-        assert(node, false, 'The @availablility <name> value must either be stack or serverless')
-    }
+    // Enforce only known availability names.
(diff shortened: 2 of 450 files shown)


compiler/src/model/metamodel.ts
```diff
@@ -378,12 +378,30 @@ export class Deprecation {
   description: string
 }
 
+export class Availability {
```
swallez, line 381: Rename to `Availabilities` (plural) since it's a collection of named individual `Availability`?

compiler/src/model/metamodel.ts
```diff
@@ -378,12 +378,30 @@ export class Deprecation {
...
+}
+
+export class StackAvailability {
+  since?: string
+  featureFlag?: string
+  stability?: Stability
+  visibility?: Visibility
+}
+
+export class ServerlessAvailability {
```
swallez, line 393: I would make a single `Availability` class that also has `since` and `featureFlag`. In the context of serverless:
* `since` may be used for documentation purposes to indicate when a feature was added
* I'm pretty sure there will be some `featureFlag`s at some point. This is very common in SaaS products to provide early access to upcoming features to selected customers. In terms of code generation, this can materialize as custom builds.

Note BTW that we don't have any `feature_flag` visibility defined currently, so we may also remove it altogether (including in `Visibility`) and consider that APIs guarded by a feature-flag on the server are just `experimental` in the API specification.

compiler/src/model/utils.ts
```diff
@@ -1023,6 +1057,71 @@ export function parseVariantNameTag (jsDoc: JSDoc[]): string | undefined {
...
+    // Ensure that there is actually a name for this availability definition.
+    assert(node, value.split(' ').length >= 1, 'The @availability tag must include a name (either stack or serverless)')
+    const [availabilityName, ...values] = value.split(' ')
+
+    // Since we're using reduce() we need to check that there's no duplicates.
+    assert(node, !(availabilityName in result), `Duplicate @availability tag: '${availabilityName}'`)
+
+    // Based on which availability definition we're parsing we have different valid tags.
+    let validKeys: string[] = []
+    switch (availabilityName) {
```
swallez, line 1074: Can be simplified if we have a single `Availability`

swallez (commented):
Looking good overall! I left some comments about simplifying the definition of availability by using a single type, and also removing feature flag annotations

sethmlarson:
Thanks @swallez, I've applied your feedback!

compiler/src/model/metamodel.ts
tvernum: I'm not sure about the name `featureFlag` for serverless, but I agree with the sentiment that we will eventually have some concept of "invitation only" / "opt-in" / "pre-release" features. But since I don't know which of those concepts we will use (possibly more than 1), my inclination is that we shouldn't try to plan for it now.

tvernum (approved):
LGTM.

It looks like this covers us for Requests (in effect APIs).
Do we have any thoughts about how we'll handle stack/serverless differences for parameters and fields?

swallez (approved):
LGTM, although I think (and seems to be confirmed by the discussion) that we can remove `feature_flag`.

compiler/src/model/utils.ts
```diff
@@ -648,6 +635,34 @@ export function hoistRequestAnnotations (
...
+    } else if (tag === 'availability') {
+      // The @availability jsTag is different than most because it allows
+      // multiple values within the same docstring, hence needing to parse
+      // the values again in order to preserve multiple values.
+      const jsDocsMulti = parseJsDocTagsAllowDuplicates(jsDocs)
+      const availabilities = parseAvailabilityTags(jsDocs, jsDocsMulti.availability)
+
+      // Apply the availabilities to the Endpoint.
+      for (const availabilityName in availabilities) {
+        const availabilityValue = availabilities[availabilityName]
```
JoshMock, line 647:
```suggestion
      for (const [availabilityName, availabilityValue] of Object.entries(availabilities)) {
```
Nitpick, just drops the object iteration logic down to one line.

compiler/src/model/utils.ts
```diff
@@ -648,6 +635,34 @@ export function hoistRequestAnnotations (
...
+          }
+          if (availabilityValue.stability !== undefined) {
+            endpoint.stability = availabilityValue.stability
+          }
+          if (availabilityValue.visibility !== undefined) {
+            endpoint.visibility = availabilityValue.visibility
+          }
+          if (availabilityValue.featureFlag !== undefined) {
+            endpoint.featureFlag = availabilityValue.featureFlag
+          }
```
JoshMock, line 663:
```suggestion
          endpoint.since = availabilityValue?.since
          endpoint.stability = availabilityValue?.stability
          endpoint.visibility = availabilityValue?.visibility
          endpoint.featureFlag = availabilityValue?.featureFlag
```
optional chaining ftw :smile:

JoshMock (approved):
A couple nits, but LGTM overall from a TypeScript point of view. :trophy:

compiler/src/model/utils.ts
sethmlarson: This actually isn't the same operation, your suggestion unconditionally sets the value to `undefined` where there might already be a value defined on `endpoint.property`.

sethmlarson:
Thanks all for the reviews!

> It looks like this covers us for Requests (in effect APIs). Do we have any thoughts about how we'll handle stack/serverless differences for parameters and fields?

@tvernum I'm planning on using a similar approach with `@availability` to document parameters which are available on the Stack but not Serverless. I'll have to document in the modeling guide that if no `@availability` annotation is given then the property/request/value is available in all flavors of ES.

compiler/src/model/utils.ts
JoshMock: ooh good catch. :star:

PR #2134, opened 2023-06-02
Add `ignore_unavailable` to data stream api

sorenlouv:
The `ignore_unavailable` is valid but not allowed in the `getDataStream` api:

```ts
await esClient.indices.getDataStream({ name: 'foo-*', ignore_unavailable: true });
```


sethmlarson (changes requested):
The change looks good, but can you run `make contrib` locally and the push the resulting changes?

pquentin:
Hey Søren, thanks for your contribution. And sorry for taking more than one year (!) to get back to you. I checked with the team that maintains this endpoint, and while `ignore_unavailable` is accidentally accepted, it does not have any effect. Which is why it's not documented and why we're not going to add it to clients.

PR #2140, opened 2023-06-07
Add types for new cluster info api (`_info/<target>`)

HiDAl:
Closes elastic/elasticsearch#96477

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -339,3 +339,13 @@ export enum SlicesCalculation {
    */
   auto
 }
+
+
+export enum ClusterInfoTarget {
+  http,
+  ingest,
+  thread_pool,
+  script
+}
+
+export type ClusterInfoTargets = "_all" | ClusterInfoTarget | ClusterInfoTarget[]
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -340,13 +340,7 @@ export enum SlicesCalculation {
   auto
 }
 
-export enum ClusterInfoTarget {
-  http,
-  ingest,
-  thread_pool,
-  script
-}
-
+export type ClusterInfoTarget = 'http' | 'ingest' | 'thread_pool' | 'script'
 export type ClusterInfoTargets =
   | '_all'
   | ClusterInfoTarget
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -341,13 +341,11 @@ export enum SlicesCalculation {
 }
 
 export enum ClusterInfoTarget {
-  http = 'http',
-  ingest = 'ingest',
-  thread_pool = 'thread_pool',
-  script = 'script'
+  _all,
+  http,
+  ingest,
+  thread_pool,
+  script
 }
 
-export type ClusterInfoTargets =
-  | '_all'
-  | ClusterInfoTarget
-  | ClusterInfoTarget[]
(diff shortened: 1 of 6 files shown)


HiDAl:
Hey @sethmlarson :) when you have some spare time, can you review this PR? thanks!

sethmlarson:
Thanks for submitting this! I'm going to have someone from the team review it for correctness :)

specification/_types/common.ts
```diff
@@ -339,3 +339,9 @@ export enum SlicesCalculation {
    */
   auto
 }
+
+export type ClusterInfoTarget = 'http' | 'ingest' | 'thread_pool' | 'script'
```
Anaethelion, line 343: This seems like a good candidate for an [enum](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#enum).
It could encapsulate all the possibilities and `ClusterInfoTargets` would only be the union of the enum and an array of the same type.

Anaethelion (changes requested):

specification/_types/common.ts
HiDAl: Hi @Anaethelion! In [this commit](https://github.com/elastic/elasticsearch-specification/pull/2140/commits/72fa7e6ca8e37bc66c335b4025e1a6da2b910af0) I changed to a `type` because the value could be `_all OR <one value of ClusterInfoTarget> OR <two or more values of ClusterTargetInfo>`. Defining it as an `enum` made it fail with the following [error](https://github.com/elastic/elasticsearch-specification/actions/runs/5199181765/jobs/9376248950)

HiDAl: Look at this:

Screenshot 2023-06-08 at 10 31 21

it does work, but the code looks _odd_ to me. You can use a `string` or a value coming from an `enum`  mmh... this is clearly personal, but it does not look _pretty_ to me.

On the other hand, using the type union of strings, the API would look like:
Screenshot 2023-06-08 at 10 34 07

what is prettier –again, according to me.

I just want to show the potential options, I've no idea which one is better here.

HiDAl: nevermind all that I've said. The error was because I used double-quotes 🤦🏽 
pushed the changes. Thanks!

specification/cluster/info/ClusterInfoRequest.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @rest_spec_name cluster.info
+ * @availability stack since=8.9.0 stability=stable
+ * @availability serverless stability=stable visibility=public
+ * @doc_id cluster-info
+ */
+export interface Request extends RequestBase {
+  path_parts: {
+    /** Limits the information returned to the specific target. Supports a comma-separated list, such as http,ingest. */
+    target: ClusterInfoTargets
```
Anaethelion, line 32:
```suggestion
    target?: ClusterInfoTargets
```

Just noticed in the doc this field is marked as optional, it should be as well in the specification.

HiDAl: My bad here. it's not optional. Just sent a PR fixing it https://github.com/elastic/elasticsearch/pull/96707/files

specification/_types/common.ts
HiDAl: understood. changed it

Anaethelion (approved):
LGTM!

PR #2146, opened 2023-06-15
Add shards_capacity to health_report

JoshMock:
Adds new [`shards_capacity`](https://www.elastic.co/guide/en/elasticsearch/reference/8.8/health-api.html#health-api-response-details-shards-capacity) to the health API.

My first time making a spec change so let me know if I missed anything important here. :pray:

Done per @rudolf's needs mentioned in https://github.com/elastic/kibana/pull/159587#discussion_r1230997505.

### DIFF
--- a/specification/_global/health_report/types.ts
+++ b/specification/_global/health_report/types.ts
@@ -36,6 +36,7 @@ export class Indicators {
   repository_integrity?: RepositoryIntegrityIndicator
   ilm?: IlmIndicator
   slm?: SlmIndicator
+  shards_capacity?: ShardsCapacityIndicator
 }
 
 export class BaseIndicator {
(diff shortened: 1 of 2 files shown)


specification/_global/health_report/types.ts
```diff
@@ -166,3 +167,19 @@ export class SlmIndicatorUnhealthyPolicies {
...
+/** SHARDS_CAPACITY */
+
+export class ShardsCapacityIndicator extends BaseIndicator {
+  details?: ShardsCapacityIndicatorDetails
+}
+
+export class ShardsCapacityIndicatorDetails {
+  data: ShardsCapacityIndicatorTierDetail
+  frozen: ShardsCapacityIndicatorTierDetail
+}
```
swallez, line 180: This pattern where all fields are of the same type could also be a `Dictionary<string, SomeType>` to allow for future expansion, or also to account for values we haven't seen in the docs or tests.

In this particular case, looking at the ES code shows that these are hardcoded constants that are unlikely to have new members. So this approach works.

swallez (approved):
LGTM!

PR #2147, opened 2023-06-20
Throttling info in DeleteByQueryResponse is optional

swallez:
As titled. Reported in https://github.com/elastic/elasticsearch-java/issues/413


Anaethelion (approved):
LGTM

PR #2152, opened 2023-06-23
Add description to Authenticate API

abdonpijpelink:
Adds descriptions derived from https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api-authenticate.html

It's a trivial one, so it's a good one to do first :)


lcawl (approved):
LGTM!

PR #2156, opened 2023-06-27
Add exception to core.get to handle index_not_found & document not found

Anaethelion:
This adds an exception to the `core.get` response to handle `index_not_found_exception` and document not found.

### DIFF
--- a/specification/_global/get/GetResponse.ts
+++ b/specification/_global/get/GetResponse.ts
@@ -18,7 +18,14 @@
  */
 
 import { GetResult } from '@global/get/types'
+import { ErrorResponseBase } from '@_types/Base'
 
 export class Response<TDocument> {
   body: GetResult<TDocument>
+  exceptions: [
+    {
+      statusCodes: [404]
+      body: GetResult<TDocument> | ErrorResponseBase
+    }
+  ]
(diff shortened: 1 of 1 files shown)


specification/_global/get/GetResponse.ts
```diff
@@ -18,7 +18,14 @@
...
 
 import { GetResult } from '@global/get/types'
+import { ErrorResponseBase } from '@_types/Base'
 
 export class Response<TDocument> {
   body: GetResult<TDocument>
+  exceptions: [
+    {
+      statusCodes: [404]
+      body: GetResult<TDocument> | ErrorResponseBase
```
swallez, line 28: Can you add a comment explaining that `ErrorResponseBase` will be returned if the index doesn't exist, and `GetResult` if the id doesn't exist, and in that case it will only contain "found": false, _id and _index? 

And for the sake of nitpicking flip the two members of the union: start with the result for the unknown index and finish with the result for the unknown id 😉

Anaethelion: Documentation added!

The order is consistent with the existing exceptions in the rest of the specification; I suggest we keep it that way.

Issue #2049, opened 2023-03-31
Represent "meaningful `null`"


swallez:
_[edit 2023-04-03 - rewritten to explain "semantic null" and provide several proposals]_

Some fields in the settings API handle `null` values differently from missing values: sending `null` resets the setting to their default value, whereas omitting the field just keeps the current value.

We already use `SomeType | null` in the spec for fields where ES can return `null` for missing or unknown values. Semantically these are optional values where the ES server doesn't have a `if (value == null)` test before serializing the field. We added these `| null` mainly to allow validation tests to pass, even if semantically speaking they should not be there.

We need to distinguish "semantic null" from "null-as-optional".

There are different ways to represent this in the spec:

* Follow the example of [`Stringified`](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_spec_utils/Stringified.ts#L28-L27) to represent non-string values that ES can sometimes serialize as strings, and introduce `type OptionalNull<T> = T | null`:

  * semantic nulls: use `SomeType | null`.
  * null as optional: replace existing `SomeType | null` with `OptionalNull<SomeType>`.

* Consider that semantic null is important enough to have its own type defined as `type NullValue = null`:
  * semantic nulls: use `SomeType | NullValue`
  * null as optional: leave existing `SomeType | null` as is.

* Consider that the `null` value is an encoding concern and that it's really about default values:
  * semantic nulls: use `SomeType | Default` where `type Default = null`
  * null as optional: leave existing `SomeType | null` as is.

  Note: we can also consider `Reset` instead of `Default` but this name implies some action rather than a value or state. It makes sense for a request value but not so much for a response value.

* Something else?

Mpdreamz:
This has come up in the .NET world often too and while we offer workarounds the Developer Experience has never been great. +1 modelling this explicitly. 

In .NET I would still model this explicitly as `T | Reset | null` so folks can explicitly set the property to `Reset` while still serializing null. While `null` would retain its `unset` properties and not serialize the whole property.

This is also an issue with the Update API where people may use `null` to reset fields.

Since the update doc is userdefined there is nothing we can spec for it but something to consider in this context too.

Anaethelion:
I think its important we keep the distinction between `null` as a possible type and `null` as a possible value, not necessarily default, from an API perspective.

`Stringified` is not entirely the same since its purpose was the augmentation of a type to a union of said type and a string. Here we are trying to convey not-only a type but probably an API change.

[...]

PR #2160, opened 2023-06-27
Allow nullable `lifecycle` and `data_retention` template values

dakrone:
These keys allow using an explicit `null` as the value to "unset" the configuration when merging multiple component templates. As related to the merging tables seen in https://github.com/elastic/elasticsearch/pull/95979#issuecomment-1549717558

Relates also to the discussion in #2049

### DIFF
--- a/specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
+++ b/specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
@@ -66,5 +66,5 @@ export class IndexTemplateMapping {
    * @availability stack since=8.8.0 stability=experimental
    * @availability serverless stability=experimental
    */
-  lifecycle?: DataLifecycle
+  lifecycle?: DataLifecycle | null
 }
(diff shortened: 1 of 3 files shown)


dakrone:
I'm not sure how to fix the CI failure, it says: `Error: Input required and not supplied: token`

flobernd:
I think the CI workflow is faulty. This expression:

https://github.com/elastic/elasticsearch-specification/blob/9eff69a2a7b3f764a1e39ac50418ca184a68612f/.github/workflows/validate-pr.yml#L18

will always be `true` as `github.repository_owner` refers to the "target" repository for `pull_request` triggers.

It should instead check `github.event.pull_request.head.repo.organization` (if I remember correctly).

specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
```diff
@@ -66,5 +66,5 @@ export class IndexTemplateMapping {
    * @availability stack since=8.8.0 stability=experimental
    * @availability serverless stability=experimental
    */
-  lifecycle?: DataLifecycle
```
andreidan, line 69: @dakrone this shouldn't be nullable anymore as we now have the `enabled` field.

PR #2161, opened 2023-06-27
Fix incorrect "fork check" condition in `validate_pr` workflow

flobernd:
This expression:

https://github.com/elastic/elasticsearch-specification/blob/9eff69a2a7b3f764a1e39ac50418ca184a68612f/.github/workflows/validate-pr.yml#L18

always evaluates to `true` as `github.repository_owner` refers to the "target" repository for `pull_request` triggers.

It should instead check `github.event.pull_request.head.repo.owner.login`.

Related: https://github.com/elastic/elasticsearch-specification/pull/2160

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   validate-pr:
     name: build
-    if: github.repository_owner == 'elastic' # this action will fail if executed from a fork
+    if: github.event.pull_request.head.repo.owner.login == 'elastic' # this action will fail if executed from a fork
     runs-on: ubuntu-latest
     env:
       STACK_VERSION: 8.9-SNAPSHOT


.github/workflows/validate-pr.yml
```diff
@@ -15,7 +15,7 @@ on:
 jobs:
   validate-pr:
     name: build
-    if: github.repository_owner == 'elastic' # this action will fail if executed from a fork
+    if: github.event.pull_request.head.repo.owner.login == 'elastic' # this action will fail if executed from a fork
```
swallez, line 18: I'm not familiar with the `github` object model. This test is because the action checks out `elastic/clients-flight-recorder` which is a private repo. So the condition should test that the user running the action is a member of the `elastic` org. I don't think this is what this new expression will test?

flobernd: Correct. If the workflow is triggered from a fork, the `secrets` context is not available and in consequence no `token/PAT` is supplied to the `checkout` action which ultimately causes the workflow to fail.

The condition seems to be intended to work around that problem by disabling the workflow if it was triggered from a fork.

The expression used in the original condition however does not work for that purpose, because `github.repository_owner` always evaluates to the BASE/target repository (`elastic/elasticsearch-specification` in this case). This for example happened in #2160. 

To check the HEAD/source branch instead, we have to use `github.event.pull_request.head.repo.owner.login`. This expression would evaluate to e.g. `flobernd` if I trigger this workflow from a fork (e.g. `flobernd/elasticsearch-specification`).

flobernd: Probably this condition is still not the perfect one to use, because even a fork from inside the `elastic` organization (e.g. `elastic/elasticsearch-specification-clone`) might not have access to the `secrets` context. However there might be a special rule for "inside organization" forks that I don't remember right now.

PR #2166, opened 2023-06-30
add geo_grid query type

nreese:
Add types for [Geo-grid query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-grid-query.html)

### DIFF
--- a/specification/_types/query_dsl/geo.ts
+++ b/specification/_types/query_dsl/geo.ts
@@ -56,6 +56,15 @@ export class GeoDistanceQuery
   validation_method?: GeoValidationMethod
 }
 
+export class GeoGridQuery
+  extends QueryBase
+  implements AdditionalProperty<Field, unknown>
+{
+  geohash?: string;
+  geohex?: string;
+  geotile?: string;
+}
+
 export class GeoPolygonPoints {
   points: GeoLocation[]
 }


swallez (commented):
This should be represented in a different way. `AdditionalProperty` is used to represent an additional property with an unknown name can is added on top of those already defined in the class.

Here the field name is in the object that wraps the geo information. So this is similar to most queries, that are represented as `"foo"?: SingleKeyDictionary<Field, FooQuery>` in the `QueryContainer` class. Existing geo queries are kind of snowflakes in this regard and use a different nesting structure to define the target field.

So this should be something like:

```ts
/** @variants container */
class GeoGridQuery {
    geogrid?: GeoTile,
    geohash?: GeoHash,
    geohex?: GeoHexCell,
}

```
The `@variants` annotation indicates that only one of the fields can exist at a time.

And then we need to add this to `Query`: 
```ts
geo_grid?: SingleKeyDictionary<Field, GeoGridQuery>
```

l-trotta:
@nreese any update on this? should someone else take over? I can rewrite this

nreese:
> any update on this? should someone else take over? I can rewrite this

Please feel free to take over with this effort.

l-trotta:
closing in favor of https://github.com/elastic/elasticsearch-specification/pull/3860

PR #2167, opened 2023-07-05
Add descriptions for API key API

abdonpijpelink:
Adds descriptions for these API key APIs:

- Create API key
- Grant API key
- Update API key
- Get API key information
- Invalidate API key
- Query API key information
- Clear API key cache

Source for the descriptions: https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api.html#security-api-keys

I also had to add the `read_security` cluster privilege. It was missing from `utils.ts`, but it is mentioned in several docs. I don't know if that was the right thing to do?

### DIFF
--- a/specification/security/_types/RoleDescriptor.ts
+++ b/specification/security/_types/RoleDescriptor.ts
@@ -25,23 +25,61 @@ import { Metadata } from '@_types/common'
 import { OverloadOf } from '@spec_utils/behaviors'
 
 export class RoleDescriptor {
+  /**  
+   * A list of cluster privileges. These privileges define the cluster level actions that API keys are able to execute. 
+   */
   cluster?: string[]
-  /** @aliases index */
+  /** 
+   * A list of indices permissions entries.
+   * @aliases index 
+   */
   indices?: IndicesPrivileges[]
+  /**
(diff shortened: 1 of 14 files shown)


specification/security/_types/RoleDescriptor.ts
```diff
@@ -25,23 +25,61 @@ import { Metadata } from '@_types/common'
...
+   */
   cluster?: string[]
-  /** @aliases index */
+  /**
+   * A list of indices permissions entries.
+   * @aliases index
+   */
   indices?: IndicesPrivileges[]
+  /**
+   * An object defining global privileges. A global privilege is a form of cluster privilege that is request-aware. Support for global privileges is currently limited to the management of application privileges. This field is optional.
```
lcawl, line 38: > ... This field is optional.

Not a big deal but IMO this text is unnecessary, since the "?" already indicates it's optional.

lcawl (approved):
LGTM

specification/security/_types/RoleDescriptor.ts
abdonpijpelink: Thanks! I missed those. I've removed them.

PR #2098, opened 2023-04-27
Add text expansion config for the NLP APIs

davidkyle:
Updates the ML NLP APIs for new features added in 8.8. Namely the `text_expansion` config  and adding the new `deployment_id` field to stats results.

PR #2169, opened 2023-07-05
Fix put trained model spec

droberts195:
Built in models like ELSER do not require an inference_config be
specified. #2098 adjusted the spec to reflect this in one place,
but not in the put trained model API spec. This PR completes the
change.


davidkyle (approved):
LGTM

PR #2171, opened 2023-07-07
Add descriptions for Search API

abdonpijpelink:
Adds descriptions for the Search API. Source for the descriptions: https://www.elastic.co/guide/en/elasticsearch/reference/master/search-search.html (and pages linked from it)

**Does not include aggs and suggesters.**

### DIFF
--- a/specification/_global/search/SearchRequest.ts
+++ b/specification/_global/search/SearchRequest.ts
@@ -243,7 +243,6 @@ export interface Request extends RequestBase {
     /**
      * Specifies the period of time to wait for a response from each shard.
      * If no response is received before the timeout expires, the request fails and returns an error.
-     * @server_default no timeout
      */
     timeout?: Duration
     /**
--- a/specification/_global/search/SearchRequest.ts
+++ b/specification/_global/search/SearchRequest.ts
@@ -51,69 +51,290 @@ import { RankContainer } from '@_types/Rank'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 
 /**
+ * Returns search hits that match the query defined in the request.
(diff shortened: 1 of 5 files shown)


abdonpijpelink:
I'm not sure what to do about the validation error (if anything). Running `make validate api=search stack-version=8.8.0` locally just returns:

```
✖ 
    at run (/Users/abdon/projects/github/elasticsearch-specification/compiler/run-validations.js:175:12)
    exit code: 1
make: *** [validate] Error 1
```

specification/_global/search/SearchRequest.ts
```diff
@@ -51,69 +51,290 @@ import { RankContainer } from '@_types/Rank'
...
+     * When possible, let Elasticsearch perform early termination automatically.
+     * Avoid specifying this parameter for requests that target data streams with backing indices across multiple data tiers.
+     * If set to `0` (default), the query does not terminate early.
+     * @server_default 0
+     */
     terminate_after?: long
+    /**
+     * Specifies the period of time to wait for a response from each shard.
+     * If no response is received before the timeout expires, the request fails and returns an error.
+     * @server_default no timeout
```
lcawl, line 246: Is "no timeout" actually a valid value for this property? I'd have assumed it was a numeric value:
```suggestion
     * @server_default 0
```
If we're not sure, I think it's better to omit it.

lcawl:
> I'm not sure what to do about the validation error (if anything). 

I think the validation messages can be ignored in the context of adding descriptions. My interpretation of them is that they're indicating there are some request and response properties missing from the spec, but that's not what we're addressing with these PRs, unless we happen to notice an obvious omission.

lcawl (approved):
One suggestion, otherwise LGTM

PR #2177, opened 2023-07-12
[Transform] Filter availability

Anaethelion:
This PR adds a subscript to the transform list to filter the specification and types based on the `@availability` tags.

The script defaults to the standard `schema.json` and output a `schema-filtered.json` in the same directory based on the provided flags. (e.g. `--stack` or `--serverless`)
Input and output path the different source and target can be provided.

### DIFF
--- a/compiler/src/transform/filter-by-availability.ts
+++ b/compiler/src/transform/filter-by-availability.ts
@@ -166,6 +166,11 @@ function filterModel (inputModel: Model, stack, serverless: boolean): Model {
     }
   }
 
+  // Returns the fully-qualified name of a type name
+  function fqn (name: TypeName): string {
+    return `${name.namespace}:${name.name}`
+  }
+
   const seen = new Map<string, string[]>()
 
   const output: Model = {
@@ -174,18 +179,27 @@ function filterModel (inputModel: Model, stack, serverless: boolean): Model {
     endpoints: new Array<Endpoint>()
   }
 
+  const typeDefByName = new Map<string, TypeDefinition>()
+
+  for (const type of inputModel.types) {
+    typeDefByName.set(fqn(type.name), type)
+  }
+
   // we filter to include only the matching endpoints
   inputModel.endpoints.forEach((endpoint) => {
     if (include(endpoint.availability, stack, serverless)) {
       // add the current endpoint
       output.endpoints.push(endpoint)
 
-      // add the request and response type for current endpoint
-      inputModel.types.forEach((typeDef) => {
-        if (((endpoint.request !== null) && cmpTypeNames(endpoint.request, typeDef.name)) || ((endpoint.response !== null) && cmpTypeNames(endpoint.response, typeDef.name))) {
-          output.types.push(typeDef)
-        }
-      })
+      if (endpoint.request) {
+        let requestType = typeDefByName.get(fqn(endpoint.request));
+        requestType ? output.types.push(requestType) : void 0;
+      }
+
+      if (endpoint.response) {
+        let responseType = typeDefByName.get(fqn(endpoint.response));
+        responseType ? output.types.push(responseType) : void 0;
+      }
     }
   })
 
--- a/compiler/src/transform/filter-by-availability.ts
+++ b/compiler/src/transform/filter-by-availability.ts
@@ -236,6 +236,10 @@ function filterEndpoints(inputModel: Model, stack, serverless: boolean): Model {
 }
 
 async function filterSchema(inPath: string, outPath: string, stack: boolean, serverless: boolean): Promise<void> {
+  if (!stack && !serverless) {
+    throw new Error("Expected one of --stack or --serverless to be specified.");
+  }
+
   const inputText = await readFile(
     inPath,
     {encoding: 'utf8'}
@@ -243,9 +247,6 @@ async function filterSchema(inPath: string, outPath: string, stack: boolean, ser
 
   const inputModel = JSON.parse(inputText)
   const outputModel = filterEndpoints(inputModel, stack, serverless)
-
-  console.log("Endpoints: " + outputModel.endpoints.length)
-  console.log("Types: " + outputModel.types.length)
 
   await writeFile(
     outPath,
@@ -265,5 +266,7 @@ const serverless = !!argv.serverless
 // input, if not provided default to versioned schema.json
 // output, if not provided default to schema-filtered.json
 filterSchema(inputPath, outputPath, stack, serverless)
-  .catch(reason => console.error(reason))
+  .catch(reason => {
+    console.error(reason.message? reason.message:reason)
+  })
   .finally(() => console.log('Done, filtered schema is at', outputPath))
--- a/compiler/src/transform/filter-by-availability.ts
+++ b/compiler/src/transform/filter-by-availability.ts
@@ -189,11 +189,11 @@ function filterModel (inputModel: Model, stack, serverless: boolean): Model {
     }
   })
 
-  // from what was gathered we filter out anything that doesn't match the requested tags
+  // filter type items (properties / enum members)
   inputModel.types.forEach((typeDef) => {
     switch (typeDef.kind) {
       case 'interface':
-        typeDef.properties.filter(filterItem())
+        typeDef.properties = typeDef.properties.filter(filterItem())
         break
       case 'enum':
         typeDef.members = typeDef.members.filter(filterItem())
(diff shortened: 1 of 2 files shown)


compiler/src/transform/filter-by-availability.ts
```diff
@@ -66,9 +66,9 @@ function filterModel(inputModel: Model, stack, serverless: boolean): Model {
         output.types.push(typeDef)
 
         // store the type infos to prevent doubles & recursive calls
-        seen.get(typeName.namespace) ?
-          seen.get(typeName.namespace)?.push(typeName.name) :
-          seen.set(typeName.namespace, [typeName.name])
+        ;(seen.get(typeName.namespace) != null)
+          ? seen.get(typeName.namespace)?.push(typeName.name)
+          : seen.set(typeName.namespace, [typeName.name])
```
JoshMock, line 71: semi-nit: if you have to use a leading `;` to make a ternary readable, maybe switch to an `if`.

Also, change `!= null` to `!== null`, assuming you want a strict null check. If you use `!=`, some values like `undefined` will also compare as equal. Isn't JavaScript loose equality wonderful? :upside_down_face: 

```suggestion
        if (seen.get(typeName.namespace) !== null) {
          seen.get(typeName.namespace)?.push(typeName.name)
        } else {
          seen.set(typeName.namespace, [typeName.name])
        }
```

compiler/src/transform/filter-by-availability.ts
```diff
@@ -118,68 +118,69 @@ function filterModel(inputModel: Model, stack, serverless: boolean): Model {
...
   inputModel.endpoints.forEach((endpoint) => {
     if (include(endpoint.availability, stack, serverless)) {
       // add the current endpoint
-      output.endpoints.push(endpoint);
+      output.endpoints.push(endpoint)
 
       // add the request and response type for current endpoint
       inputModel.types.forEach((typeDef) => {
-        if ((endpoint.request && cmpTypeNames(endpoint.request, typeDef.name)) || (endpoint.response && cmpTypeNames(endpoint.response, typeDef.name))) {
+        if (((endpoint.request != null) && cmpTypeNames(endpoint.request, typeDef.name)) || ((endpoint.response != null) && cmpTypeNames(endpoint.response, typeDef.name))) {
```
JoshMock, line 183:
```suggestion
        if (((endpoint.request !== null) && cmpTypeNames(endpoint.request, typeDef.name)) || ((endpoint.response !== null) && cmpTypeNames(endpoint.response, typeDef.name))) {
```

compiler/src/transform/filter-by-availability.ts
```diff
@@ -257,8 +258,8 @@ async function filterSchema(inPath: string, outPath: string, stack: boolean, ser
 
 const inputPath: string = argv.input ?? join(__dirname, '..', '..', '..', 'output', 'schema', 'schema.json')
 const outputPath = argv.output ?? join(__dirname, '..', '..', '..', 'output', 'schema', 'schema-filtered.json')
-const stack = !!argv.stack
-const serverless = !!argv.serverless
+const stack: boolean = (argv.stack !== undefined)
```
JoshMock, line 261: This makes it more strict, which I assume is what you want. `!!argv.stack` would return `false` if its value was `undefined`, `0`, `""`, `null`, `-0` (don't get me started :roll_eyes:), and [a few others](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). Your change will now return `true` for all but `undefined`.

compiler/src/transform/filter-by-availability.ts
```diff
@@ -267,6 +268,6 @@ const serverless = !!argv.serverless
 // output, if not provided default to schema-filtered.json
 filterSchema(inputPath, outputPath, stack, serverless)
   .catch(reason => {
-    console.error(reason.message? reason.message:reason)
+    console.error((reason.message !== null) ? reason.message : reason)
```
JoshMock, line 271: Same here with the stricter type checks.

JoshMock (commented):
Lots of nit-picky little JS type-checking things. Otherwise the change looks good to me. :+1:

compiler/src/transform/filter-by-availability.ts
JoshMock: Just realized I must've been commenting on a single commit where you changed your own work in progress. :facepalm:

compiler/src/transform/filter-by-availability.ts
```diff
@@ -0,0 +1,273 @@
...
+    if ((availabilities.stack != null) && stack && !serverless) {
+      return true
+    }
+    if ((availabilities.serverless != null) && serverless && !stack) {
+      return true
+    }
+    if (((availabilities.serverless != null) || (availabilities.stack != null)) && serverless && stack) {
+      return true
+    }
+    return false
```
JoshMock, line 39: I'm trying to line this logic up with the comment from @swallez on https://github.com/elastic/elasticsearch-specification/issues/2107:

> If there's no `@availability` annotation, the endpoint or property is available in all environments.

In my serverless generator code, the filter for inclusion into the serverless client is much simpler, and defaults to `true` to recognize Sylvain's recommendation above:

```typescript
function serverlessAvailability(item: M.Property | M.Endpoint): boolean {
  if (item.availability) {
    return item.availability?.serverless?.visibility === 'public'
  }
  return true
}
```

To extend that to also support stack availability like your function, I would do something like:

```typescript
function availability(item: M.Property | M.Endpoint, stack: boolean, serverless: boolean): boolean {
  let isServerless = true
  let isStack = true

  if (item.availability) {
    isServerless = (item.availability?.serverless?.visibility === 'public')
    isStack = (item.availability?.stack?.visibility === 'public')
  }

  if (serverless && stack) {
    return isStack || isServerless
  } else if (serverless) {
    return isServerless
  } else if (stack) {
    return isStack
  }
}
```

This is more readable, to me at least. It also checks the string value of `visibility`. Based on my understanding, following Sylvain's comment above, the only way to exclude an item from the serverless spec is to include a serverless availability annotation marking it as private, e.g. `@availability serverless stability=stable visibility=private`.

Please correct me if I'm misunderstanding! The structure and implied defaults of the `@availability` tag values leave some things up to interpretation. It might be less confusing if the generated schema.json included default values for `availability.{stack,serverless}.*` to make the defaults explicit.

compiler/src/transform/filter-by-availability.ts
Anaethelion: You did, that's fined, the code is the same in the end. :)

compiler/src/transform/filter-by-availability.ts
Anaethelion: Fixed as well in a later commit.

compiler/src/transform/filter-by-availability.ts
Anaethelion: Ah! 

I'd prefer to keep visibility out of this script, different languages can have different use of that notion we could want to use that independently.

For the availability tag my understanding is a bit more complex because no tags means available everywhere, but presence of the tag is a direct exclusion if both are not defined.
That allows us to embed everything by default that is common to both APIs and later differentiate what could be stack or serverless only.

For example in the future we could have a public API that is specific to serverless with `@availability serverless` or only available to stack with `@availability stack`; the same logic goes for properties.

swallez: You both have good points.

Visibility is indeed an additional dimension. Private APIs should not be included in serverless clients we provide to the general public. They either expose things users should not care about or that may be removed from serverless in the future. Private APIs can however be part of internal clients built for Elastic products (e.g. Kibana).

What these internal clients look like is something we haven't discussed much, and their shape may depend on the target language. However, I think that ATM we should not care much about it: the Stack client includes everything and products such as Kibana already use it. So let just them keep using it and work on this later.

So back to this transformation, I think we should include visibility as an additional filtering parameter (with no default value to avoid implicit assumptions).

* if there is *some* availablity defined, reject if the target flavor is not present (items with no availability defined are included everywhere)
* if the visibility is private for the target flavor (Stack or Serverless), and the configuration is "public", reject the item
* otherwise keep the item.

BTW currently availability is defined with two distinct fields in the metamodel. We may want to change this to a dictionary, since this is what it conceptually is. This allows "some" visibility to be just a length check.

Anaethelion: Added a two step filter for both availability and if present visibility.

Visibility is a comma separated list to allow for cases like `public,feature_flag` since the visibility is an enum.

I completely agree changing the metamodel would allow to simplify usages, I'll do that change in another PR once we're through with this one.

swallez (approved):
LGTM

PR #2178, opened 2023-07-12
[NLP] Support the different mask tokens used by NLP models for Fill Mask

maxhniebergall:
Adds the specification for the new mask token field added in https://github.com/elastic/elasticsearch/pull/97453.


droberts195 (approved):
LGTM

PR #2181, opened 2023-07-13
Automatically compile/generate output artifacts on push to `main`

flobernd:
This PR removes the need to locally compile/generate the output artifacts before pushing changes.

Instead, output artifacts are created on demand when code is pushed to `main`.


.github/workflows/generate.yml
```diff
@@ -0,0 +1,90 @@
...
+        cache-dependency-path: '**/package-lock.json'
+
+    - name: Setup Dependencies
+      run: |
+        make setup
+
+    - name: Generate Output
+      run: |
+        make compile
+        make generate
```
swallez, line 47: [The `make contrib` target](https://github.com/elastic/elasticsearch-specification/blob/d31921a854f57ad4b9e7e6723d67a60e9b3f53d7/Makefile#L53) does a few more things, such as checking licenses, running a linter, fixing the format, and also generating the typescript version.

Having an action change the source code is not common and we may consider that part (code formatting) as a check, but in this particular case I think it can actually help "casual" contributors to easily contribute to the spec. So we may even go further and replace e.g. the `licence-check` called by `make contrib` with `license-add`.

Finally, what about adding a new make target that does all this so that this action only calls that single target? This may ease maintenance.

I'll let @picandocodigo look a the GitHub machinery around this.

flobernd: > [The `make contrib` target](https://github.com/elastic/elasticsearch-specification/blob/d31921a854f57ad4b9e7e6723d67a60e9b3f53d7/Makefile#L53) does a few more things, such as checking licenses, running a linter, fixing the format, and also generating the typescript version.

This new workflow only runs on pushes to `main`. Direct pushes should never occur due to branch protection, which means that all these checks have already run as part of the PR checks and review process.

> I think it can actually help "casual" contributors to easily contribute to the spec.

👍 

> So we may even go further and replace e.g. the `licence-check` called by `make contrib` with `license-add`.

I'm not sure if we should automatically modify non-generated files in general, but it's surely something to think about. On the other hand, pasting a license text is probably pretty straightforward and does not require e.g. NodeJS on the dev machine like for the compilation/generation.

swallez: > This new workflow only runs on pushes to `main`.

Oh, I missed that part! So this means PRs still need to pass the linter and license checks, and indeed we just need to addresses the compilation part here.

Note: if we don't include this on version branches (currently `8.[0-9]+` and `7.17`), we should keep the current behavior (fail a PR's check if compilation changes the output): we occasionally have to backport manually, and the cause is most often a merge conflict on `schema.json`.

flobernd: Ouch, good point! I forgot about the backports in general. I think we just have to make sure this workflow runs for the version branches as well and we should be fine.

picandocodigo (approved):
LGTM!

PR #2175, opened 2023-07-10
Add missing shard_min_doc_count to TermsAggregation

rand0m86:

PR #2183, opened 2023-07-15
Add missing sum_other_doc_count to TermsAggregation

tyrantkhan:
Hello There!

I've been working on moving some code from the legacy client to the java api client and noticed these two fields were missing from the spec.

btw I'm not sure there is a workaround for this currently, so i think this should be backported, but please advise if this is not right (the api is usable but if you require these fields there is no alternative)

Closes #2636


pquentin:
Hey @tyrantkhan, thanks for the contribution, and sorry for the delay getting back to you (nearly one year, wow). I really appreciate that you took the time to open a pull request here.

What I had done until now is to "blindly" merge main into your pull request to get up-to-date CI and code. Indeed, I merged https://github.com/elastic/elasticsearch-specification/pull/2175 yesterday which includes the first part of your change (`shard_min_doc_count`). Now, I'm reviewing the other part: `sum_other_doc_count`. But looking more closely, this field is set on responses, not requests, and is already correctly defined here:

https://github.com/elastic/elasticsearch-specification/blob/ce9d54c65d2c588b2576ab24cfaaa74f7281b37d/specification/_types/aggregations/Aggregate.ts#L381

As a result, there's nothing more to do here, and I'm going to close this pull request. If you have other issues with any of the Elasticsearch clients, let us know, and we'll get back to you faster next time!

PR #2186, opened 2023-07-17
Add serverless annotation to search mvt

iverase:
The search mvt API is a public API in serverless, therefore adding corresponding annotation.

### DIFF
--- a/specification/_global/search_mvt/SearchMvtRequest.ts
+++ b/specification/_global/search_mvt/SearchMvtRequest.ts
@@ -32,7 +32,8 @@ import { TrackHits } from '@global/search/_types/hits'
 
 /**
  * @rest_spec_name search_mvt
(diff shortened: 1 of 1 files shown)


specification/_global/search_mvt/SearchMvtRequest.ts
```diff
@@ -32,7 +32,8 @@ import { TrackHits } from '@global/search/_types/hits'
 
 /**
  * @rest_spec_name search_mvt
- * @availability stack since=7.15.0 stability=experimental
+ * @availability stack since=7.15.0 stability=stable
```
iverase, line 35: THe API has been stable since 7.16

swallez (approved):
LGTM. The "Validate APIs" checks fails as it only work with forks, but the PR changes are correct.

PR #2205, opened 2023-07-24
Remove node info from ML team serverless responses

droberts195:
Serverless will present the impression of having no underlying servers, so API responses in serverless should not contain any node information.

This PR removes the node fields from the responses owned by the ML team.


specification/ml/_types/TrainedModel.ts
```diff
@@ -71,7 +71,10 @@ export class TrainedModelDeploymentStats {
   inference_count: integer
   /** The unique identifier for the trained model. */
   model_id: Id
-  /** The deployent stats for each node that currently has the model allocated. */
+  /**
+   * The deployment stats for each node that currently has the model allocated.
+   * @availability stack
```
davidkyle, line 76: There are useful stats in this object so of which could be aggregated and lifted up to the top level as we have done with inference_count below.

We will want to review what model stats are presented in serverless, I'll open an issue for that

davidkyle (approved):
LGTM

specification/ml/_types/TrainedModel.ts
droberts195: Thanks.

The ES team is doing something similar for enrich stats - see https://github.com/elastic/elasticsearch/pull/97472.

specification/ml/_types/Model.ts
```diff
@@ -49,6 +49,9 @@ export class ModelSnapshotUpgrade {
   job_id: Id
   snapshot_id: Id
   state: SnapshotUpgradeState
+  /**
+   * @availability stack
+   */
   node: DiscoveryNode
```
jgowdyelastic, line 55: I believe all types which have had this `@availability stack` comment added to should also be make optional to communicate that this item will sometimes not exist. e.g.
```suggestion
  node?: DiscoveryNode
```

droberts195: For `DiscoveryNode` it's just this one place. I wouldn't like to do this for one of the more common APIs like open job or start datafeed, as it could break client code. However, since upgrading model snapshots is so esoteric I doubt anyone will be affected so I'll make the change.

PR #2207, opened 2023-07-24
Add Query Rules API

kderusso:
Adds the query rules CRUD API and the new Rule Query query DSL


specification/_types/query_dsl/specialized.ts
```diff
@@ -201,3 +202,13 @@ export class ShapeFieldQuery {
...
   shape?: GeoShape
 }
+
+/**
+ * @variants container
+ */
+export class RuleQuery extends QueryBase {
+  /** @variant container_property */
+  organic: QueryContainer
+  ruleset_id?: QueryRulesetId
```
carlosdelest, line 212: Are `ruleset_id` and `match_criteria` optional?

specification/query_ruleset/_types/QueryRuleset.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Query Ruleset unique identifier
+   */
+  ruleset_id: QueryRulesetId
+  /**
+   * Rules associated with the query ruleset
+   */
+  rules: QueryRule[]
+}
+
+export type QueryRulesetId = string
```
carlosdelest, line 36: Maybe using `Id` from `_types/common` would make more sense than defining a new type (even though it still maps to `string`) 😉

specification/query_ruleset/_types/QueryRuleset.ts
```diff
@@ -0,0 +1,78 @@
...
+  suffix,
+  contains,
+  lt,
+  lte,
+  gt,
+  gte
+}
+
+export class QueryRuleActions {
+  ids?: string[]
```
carlosdelest, line 71:
```suggestion
  ids?: Id[]
```

specification/query_ruleset/_types/QueryRuleset.ts
```diff
@@ -0,0 +1,78 @@
...
+  docs?: PinnedDocument[]
+}
+
+export class PinnedDocument {
```
carlosdelest, line 75: There's a `PinnedDoc` in `query_dsl/specialized` that might make sense here

specification/_types/query_dsl/specialized.ts
kderusso: They shouldn't be, but making them required results in the following error from `make generate`: 

```
Error: All @variants container properties must be optional
```

specification/query_ruleset/_types/QueryRuleset.ts
kderusso: That's a good callout, thanks.

specification/_types/query_dsl/specialized.ts
carlosdelest: Got it, thanks! I think that `ruleset_id` can be tagged as `container_property` in that case according to [this example](https://github.com/elastic/elasticsearch-specification/blob/5ccd90cfd3030b32527928484bfc255746124a3f/docs/modeling-guide.md?plain=1#L399).

kderusso: Fixed, thanks!

carlosdelest (approved):
💯 🚀

specification/_types/query_dsl/specialized.ts
```diff
@@ -201,3 +201,15 @@ export class ShapeFieldQuery {
   relation?: GeoShapeRelation
   shape?: GeoShape
 }
+
+/**
+ * @variants container
+ */
+export class RuleQuery extends QueryBase {
```
Anaethelion, line 208: What is the purpose of this type ? It doesn't seem to be used anywhere.

specification/query_ruleset/list/QueryRulesetListResponse.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Id } from '@_types/common'
+
+export class Response {
+  body: {
+    count: long
+    results: QueryRulesetListItem[]
+  }
+}
+
+export class QueryRulesetListItem {
```
Anaethelion, line 30: I would advise to extract this type in a `types.ts` file.
You can see several examples of that in the `_global` namespace.

Anaethelion (changes requested):
Left a question and a nit to align with existing spec.
Otherwise LGTM

specification/_types/query_dsl/specialized.ts
kderusso: Thanks for the question. I'd added it because it's a [new part of the query DSL](https://github.com/elastic/elasticsearch/pull/97250). If it's unnecessary I can remove it. What are your thoughts?

specification/query_ruleset/list/QueryRulesetListResponse.ts
kderusso: Refactored, thank you for the suggestion!

specification/_types/query_dsl/specialized.ts
kderusso: Per discussion, I have removed the container from this object, and added the rule query to the container object. Thank you!

Anaethelion (approved):
LGTM!

PR #2208, opened 2023-07-25
Update search application APIs from experimental to beta

kderusso:
Updates search applications APIs to beta.


JoshMock (changes requested):
I believe the correct order of operations for making these spec changes would be to update the relevant files in the `/specification` directory, at which point CI will generate an updated `schema.json` file and commit it for you.

kderusso:
🤦‍♀️ Thanks @JoshMock, fixed!

JoshMock (approved):

PR #2210, opened 2023-07-26
Add Synonyms API spec

carlosdelest:
Adds Synonyms API spec and types

### DIFF
--- a/specification/_json_spec/synonym_rule.get.json
+++ b/specification/_json_spec/synonym_rule.get.json
@@ -6,7 +6,6 @@
     },
     "stability": "experimental",
     "visibility": "public",
-    "feature_flag": "es.synonyms_api_feature_flag_enabled",
     "headers": {
       "accept": ["application/json"],
       "content_type": ["application/json"]
--- a/specification/_json_spec/synonym_rule.get.json
+++ b/specification/_json_spec/synonym_rule.get.json
@@ -8,20 +8,14 @@
     "visibility": "public",
     "feature_flag": "es.synonyms_api_feature_flag_enabled",
     "headers": {
-      "accept": [
-        "application/json"
-      ],
-      "content_type": [
-        "application/json"
-      ]
+      "accept": ["application/json"],
+      "content_type": ["application/json"]
     },
     "url": {
       "paths": [
         {
           "path": "/_synonyms/{synonyms_set}/{synonym_rule}",
-          "methods": [
-            "GET"
-          ],
+          "methods": ["GET"],
           "parts": {
             "synonyms_set": {
               "type": "string",
--- a/specification/_json_spec/synonym_rule.get.json
+++ b/specification/_json_spec/synonym_rule.get.json
@@ -1,0 +1,39 @@
+{
+  "synonym_rule.get": {
+    "documentation": {
+      "url": "https://www.elastic.co/guide/en/elasticsearch/reference/master/get-synonym-rule.html",
+      "description": "Retrieves a synonym rule from a synonym set"
+    },
+    "stability": "experimental",
+    "visibility": "public",
+    "feature_flag": "es.synonyms_api_feature_flag_enabled",
+    "headers": {
+      "accept": [
+        "application/json"
+      ],
+      "content_type": [
+        "application/json"
+      ]
+    },
+    "url": {
+      "paths": [
+        {
+          "path": "/_synonyms/{synonyms_set}/{synonym_rule}",
+          "methods": [
+            "GET"
+          ],
+          "parts": {
+            "synonyms_set": {
+              "type": "string",
+              "description": "The id of the synonym set to retrieve the synonym rule from"
+            },
+            "synonym_rule": {
+              "type": "string",
+              "description": "The id of the synonym rule to retrieve"
+            }
+          }
+        }
+      ]
+    }
+  }
+}
--- a/specification/_json_spec/synonyms.put.json
+++ b/specification/_json_spec/synonyms.put.json
@@ -1,7 +1,7 @@
 {
   "synonyms.put": {
     "documentation": {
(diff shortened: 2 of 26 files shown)


specification/_json_spec/synonyms.put.json
```diff
@@ -5,8 +5,7 @@
       "description": "Creates or updates a synonyms set"
     },
     "stability": "experimental",
-    "visibility": "feature_flag",
-    "feature_flag": "es.synonyms_feature_flag_enabled",
+    "visibility": "public",
```
carlosdelest, line 8: Feature flags were removed per https://github.com/elastic/elasticsearch/pull/97962

specification/synonyms/_types/SynonymRule.ts
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Id } from '@_types/common'
+
+export type SynonymString = string
+
+export class SynonymRuleOptionalId {
```
carlosdelest, line 24: There are operations (PUT) that don't need an ID to be included, so it will be generated by the API. I wanted to use a strict type for that, but it might be overkill...

specification/synonyms/_types/SynonymsSet.ts
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { SynonymRule, SynonymRuleOptionalId } from './SynonymRule'
+
+export class SynonymsSetUpdate {
```
carlosdelest, line 22: Same here, happy to hear thoughts on using specific types for including ids or not

specification/_json_spec/synonym_rule.get.json
```diff
@@ -0,0 +1,33 @@
+{
+  "synonym_rule.get": {
+    "documentation": {
+      "url": "https://www.elastic.co/guide/en/elasticsearch/reference/master/get-synonym-rule.html",
+      "description": "Retrieves a synonym rule from a synonym set"
+    },
+    "stability": "experimental",
+    "visibility": "public",
+    "feature_flag": "es.synonyms_api_feature_flag_enabled",
```
kderusso, line 9: I think we want to remove the FF from this file

specification/synonym_rule/put/SynonymRulePutResponse.ts
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { SynonymsUpdateResult } from '../../synonyms/_types/SynonymsUpdateResult'
+
+export class Response {
```
kderusso, line 22: There's some duplication between this response signature and others such as `SynonymRuleDeleteResponse`. Is this intentional or is there any way to consolidate?

specification/synonyms_sets/get/SynonymsSetsGetRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @availability serverless stability=experimental visibility=public
+ */
+interface Request extends RequestBase {
+  query_parameters: {
+    /**
+     * Starting offset (default: 0)
+     */
+    from?: integer
+    /**
+     * specifies a max number of results to get
```
kderusso, line 36: Should we include the default in the comment here too for consistency?

specification/synonyms/_types/SynonymRule.ts
kderusso: My two cents: I don't think it hurts, but I'm not sure what it adds. Happy to leave it in but I don't have strong feelings either way.

kderusso (approved):
LGTM with a couple of comments

specification/_json_spec/synonym_rule.get.json
carlosdelest: Yes, thanks for the catch! 👍

specification/synonym_rule/put/SynonymRulePutResponse.ts
carlosdelest: I tried to consolidate as much as possible by creating the `SynonymsUpdateResult` type, which is essentially the body for these responses.

I'd say we need the separate Request / Response classes for the codegen to work - hopefully the clients team can recommend a better pattern here.

specification/synonyms_sets/get/SynonymsSetsGetRequest.ts
carlosdelest: Sure! Thanks for the suggestion

specification/synonym_rule/put/SynonymRulePutResponse.ts
Anaethelion: > I'd say we need the separate Request / Response classes for the codegen to work - hopefully the clients team can recommend a better pattern here.

Yes we need those for codegen, each request/response couple will be generated into specific expanded types with their namespace.
In this case it does seem like there's duplication but I don't have a better pattern, you went as far as we can at the moment.

Kudos for a consistent API! :+1:

specification/synonyms/_types/SynonymRule.ts
carlosdelest: Thanks @Anaethelion - I added comments to clarify why they're used 👍

PR #2214, opened 2023-08-01
Introduce `PipeSeparatedFlags<T>` marker class

flobernd:
## What?

This PR introduces the `FlagsEnum<T>` marker class similar to the `Stringified<T>` concept.

Flags enums are defined as enums with a special case serialization. Values of this enum are represented as `string` (like for regular enums as well). In addition, individual enum member values can be combined using the pipe separator char (`|`) like e.g. described here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#supported-flags

## Why?

At the moment, this specific construct is defined in 2 parts:

1. The enum itself (`TEnum`) as a regular enum type
2. A type alias defined as `FlagsEnum = TEnum | string`

Using just a generic union makes it hard to generate correct (de-)serialization code and "hard" for the users to use properties of this type, as they have to manually craft strings when using multiple values (unintuitive, error prone, ...).

In case of C#, this marker would allow me to e.g. generate a native language construct:

```csharp
[Flags]
public enum MyEnum
{
    None = 0,
    A = 1 << 0,
    B = 1 << 1,
    ...
}
```

and allows the user to utilize the standard way of working with flags:

```csharp
var flags = MyEnum.A | MyEnum.B;
if (flags.HasFlag(MyEnum.B))
{
    ...
}
```

### DIFF
--- a/specification/_types/query_dsl/fulltext.ts
+++ b/specification/_types/query_dsl/fulltext.ts
@@ -29,7 +29,7 @@ import { Script } from '@_types/Scripting'
 import { QueryBase } from './abstractions'
 import { Operator } from './Operator'
 import { DateMath, TimeZone } from '@_types/Time'
-import { FlagsEnum } from '@spec_utils/FlagsEnum'
+import { FlagsEnum, PipeSeparatedFlags } from '@spec_utils/PipeSeparatedFlags'
 
 /**
  * @shortcut_property query
@@ -703,7 +703,7 @@ export class QueryStringQuery extends QueryBase {
  * Query flags can be either a single flag or a combination of flags, e.g. `OR|AND|PREFIX`
  * @doc_id supported-flags
  */
-export type SimpleQueryStringFlags = FlagsEnum<SimpleQueryStringFlag[]>
+export type SimpleQueryStringFlags = PipeSeparatedFlags<SimpleQueryStringFlag>
 
 export enum SimpleQueryStringFlag {
   /**
@@ -722,7 +722,6 @@ export enum SimpleQueryStringFlag {
    * Enables the `\|` OR operator.
    */
   OR = 1 << 2,
-  
   /**
    * Enables the `*` prefix operator.
    */
--- a/specification/_types/query_dsl/fulltext.ts
+++ b/specification/_types/query_dsl/fulltext.ts
@@ -29,6 +29,7 @@ import { Script } from '@_types/Scripting'
 import { QueryBase } from './abstractions'
 import { Operator } from './Operator'
 import { DateMath, TimeZone } from '@_types/Time'
+import { FlagsEnum } from '@spec_utils/FlagsEnum'
 
 /**
  * @shortcut_property query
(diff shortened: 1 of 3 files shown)


specification/_types/query_dsl/fulltext.ts
```diff
@@ -271,24 +272,23 @@ export class QueryStringQuery extends QueryBase {
 /**
  * Query flags can be either a single flag or a combination of flags, e.g. `OR|AND|PREFIX`
  * @doc_id supported-flags
- * @codegen_names single, multiple
  */
-export type SimpleQueryStringFlags = SimpleQueryStringFlag | string
+export type SimpleQueryStringFlags = FlagsEnum<SimpleQueryStringFlag[]>
 
 export enum SimpleQueryStringFlag {
```
flobernd, line 278: Does not make a real difference as these values are currently not part of the generated schema, but I adjusted them anyways based on their actual values:


https://lucene.apache.org/core/8_9_0/queryparser/constant-values.html#org.apache.lucene.queryparser.simple.SimpleQueryParser.AND_OPERATOR

specification/_types/query_dsl/fulltext.ts
```diff
@@ -271,24 +272,23 @@ export class QueryStringQuery extends QueryBase {
 /**
  * Query flags can be either a single flag or a combination of flags, e.g. `OR|AND|PREFIX`
  * @doc_id supported-flags
- * @codegen_names single, multiple
  */
-export type SimpleQueryStringFlags = SimpleQueryStringFlag | string
```
flobernd, line 276: @Anaethelion and me discussed about the semantics and we see multiple options here:

1. Always use explicit array brackets with this marker type in order to avoid changes to the spec compiler. This way even a completely generic generator would generate correct code that allows to pass `0..n` values of this enum.
2. Do not explicitly write out the array brackets, but implicitly assume them. This would make more sense from a semantic perspective as a flags enum value **per definition** has `0..n` bits set, but would either require changes to the spec compiler (to add the array brackets) or to each generator (to make them aware of the `FlagsEnum` concept).

For now, we do like the proposed option 1. better as it requires less code changes and as well does not force the generators to special case the new type alias.

flobernd: @Anaethelion As per discussion with @swallez , we agreed on a third option which looks like this:

```ts
export type SimpleQueryStringFlags = PipeSeparatedFlags<SimpleQueryStringFlag | string>
```

This allows the generator to unwrap the marker interface in order to preserve the current semantics, or to implement custom generation.

swallez: Actually that would be more

```ts
/**
 * A set of flags that can be represented as a single enum value or a set
 * of values that are encoded as a pipe-separated string
 */
export type PipeSeparatedFlags<T> = T | string

export type SimpleQueryStringFlags = PipeSeparatedFlags<SimpleQueryStringFlag>
```

The rule of thumb is that the TypeScript specification should accept Elasticsearch's JSON out of the box (except for behaviors), and code generators that want to provide more idiomatic types (e.g. an `EnumSet` in Java) can do so by adding specific code generation for the wrapper type.

The name `PipeSeparatedFlags` captures both the semantics (a set of flags) and the encoding, since code generators have to know about the encoding. If another encoding was used elsewhere for a set of flags, we would need another wrapper type.

specification/_types/query_dsl/fulltext.ts
swallez: Actually you can remove these constants altogether. They're some ancient leftovers from the very first iterations on the TypeScript spec where integer constants were added to the enums.

swallez (approved):
LGTM. As mentioned in the comment, we can remove the integer constants in the enum.

Issue #964, opened 2021-10-28
[DOCS] Add descriptions to API spec files


jrodewig:
Tracking issue for track doc updates to [specification files](https://github.com/elastic/elasticsearch-specification/blob/main/specification/). To prevent duplicated work, leave your name beside a namespace/API you start working on.

It is recommended to prioritize APIs that are listed as "public" instead of "blocked" in the serverless plans: https://docs.google.com/spreadsheets/d/1xFrpFP1dARey8A1QkAoC7JJCGhAZFexMLpRTeY9PAos/

API namespace inventory:
- [x] _global (we should probably subdivide these into distributed- and search-related APIs)
  - [x] bulk https://github.com/elastic/elasticsearch-specification/pull/2269
  - [x] close_point_in_time/open_point_in_time https://github.com/elastic/elasticsearch-specification/pull/2269
  - [x] create https://github.com/elastic/elasticsearch-specification/pull/2270
  - [x] delete https://github.com/elastic/elasticsearch-specification/pull/2270
  - [x] delete_by_query/delete_by_query_rethrottle/update_by_query/update_by_query_rethrottle https://github.com/elastic/elasticsearch-specification/pull/2270
  - [x] delete_script/get_script/put_script/scripts_painless_execute https://github.com/elastic/elasticsearch-specification/pull/2270
  - [x] exists/exists_source https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] explain https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] index https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] msearch_template https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] mtermvectors/termvectors
  - [x] reindex_rethrottle https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] render_search_template https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] search_shards https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] search_template https://github.com/elastic/elasticsearch-specification/pull/2271
  - [x] search https://github.com/elastic/elasticsearch-specification/pull/2171 
  - [x] queries https://github.com/elastic/elasticsearch-specification/pull/2204 https://github.com/elastic/elasticsearch-specification/pull/2213 https://github.com/elastic/elasticsearch-specification/pull/2217
[...]

PR #2216, opened 2023-08-02
[DOCS] Adds descriptions for EQL APIs

szabosteve:
## Overview

Related to https://github.com/elastic/elasticsearch-specification/issues/964.

Adds descriptions to EQL API properties. Source of info: https://www.elastic.co/guide/en/elasticsearch/reference/master/eql-apis.html

### DIFF
--- a/specification/eql/get/EqlGetRequest.ts
+++ b/specification/eql/get/EqlGetRequest.ts
@@ -22,8 +22,8 @@ import { Id } from '@_types/common'
 import { Duration } from '@_types/Time'
 
 /**
- * Returns the current status and available results for an async EQL search or a 
- * stored synchronous EQL search.
+ * Returns the current status and available results for an async EQL search or a stored synchronous EQL search.
+ * @doc_id eql-async-search-api
  * @rest_spec_name eql.get
  * @availability stack since=7.9.0 stability=stable
  */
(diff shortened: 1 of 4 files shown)


specification/eql/get/EqlGetRequest.ts
```diff
@@ -34,11 +36,13 @@ export interface Request extends RequestBase {
...
     /**
-     * Period for which the search and its results are stored on the cluster. Defaults to the keep_alive value set by the search’s EQL search API request.
+     * Period for which the search and its results are stored on the cluster.
+     * Defaults to the keep_alive value set by the search’s EQL search API request.
      */
     keep_alive?: Duration
     /**
-     * Timeout duration to wait for the request to finish. Defaults to no timeout, meaning the request waits for complete search results.
+     * Timeout duration to wait for the request to finish.
+     * @server_default no timeout
```
abdonpijpelink, line 45: Based on [an earlier review by @lcawl](https://github.com/elastic/elasticsearch-specification/pull/2171#discussion_r1255996084), I think we should only provide actual values as server defaults.

abdonpijpelink (approved):
LGTM. I left one commentt.

specification/eql/get/EqlGetRequest.ts
szabosteve: Thanks for the info, I have removed that line via [1f85bcb](https://github.com/elastic/elasticsearch-specification/pull/2216/commits/1f85bcb7b8794640324e4af17e36588f3bd9cd7d).

Issue #2149, opened 2023-06-20
Missing 'similarity' field in knn search request


adrian-arapiles:
From 8.8 version, elasticsearch has field 'similarity' for search in knn field.
Right now, the api doesn't support this field and fail.
## 🐛 Missing type

https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#_search_knn_with_expected_similarity
Unknown field 'similarity' (JSON path: similarity) (line no=2, column no=15, offset=16)
## Definition

If possible provide a basic definition that we can use as base.
You can see [here](https://github.com/elastic/elasticsearch-specification/blob/master/docs/add-new-api.md) how to create a basic definition for any API.

adrian-arapiles:
I tried to push a branch with the change, but I haven't permissions.

AlmostFamiliar:
https://github.com/elastic/elasticsearch-java/issues/411 Related. 

I also tried to open a PR with the fix, but same issue.

ezimuel:
@adrian-arapiles did you fork the project and send a PR from a branch in your fork? You cannot create a branch in the original repository. /cc @AlmostFamiliar

JoshMock:
#2261 is merged and backported to 8.8, 8.9 and 8.10, so this should be fixed in the next patch release for each of those, or 8.11.0 when it releases.

PR #2224, opened 2023-08-03
Add missing similarity field to KnnQuery and missing fields to KnnSea…

pstrsr:
Fix for: https://github.com/elastic/elasticsearch-specification/issues/2149
and: https://github.com/elastic/elasticsearch-java/issues/654
and: https://github.com/elastic/elasticsearch-java/issues/643


AlmostFamiliar:
@ezimuel I signed the Contributor Agreement on Tuesday, but the check still failed. Does it take a while for it to be available? Since then I have signed it 2 more times, just to be sure nothing went wrong. 

Could you help me getting the cla-check to pass?

flobernd:
@AlmostFamiliar You have to use the mail that is assigned to your commit. I can see you signed with 2 mails, but none of them equals the commit mail 🙂

output/schema/schema.json
```diff
@@ -25433,9 +25445,60 @@
...
+                }
+              }
+            ],
+            "kind": "union_of"
+          }
+        },
+        {
+          "description": "The required minimum similarity for a vector to be considered a match",
+          "name": "similarity",
+          "required": true,
```
mshameti, line 25491: Thanks for the contribution! 

Are we sure `similarity` is a required parameter? I believe it is a requirement for the similarity score when provided but not a required parameter for the knn query, per https://www.elastic.co/guide/en/elasticsearch/reference/8.9/knn-search.html#_search_knn_with_expected_similarity.

AlmostFamiliar: Yes that was a mistake on my part, thank you for catching that. 

I added a commit that fixes this.

AlmostFamiliar: Should I click on resolve conversation, or leave this as is for now?

output/schema/schema.json
```diff
@@ -25416,7 +25416,7 @@
         {
           "description": "The query vector",
           "name": "query_vector",
-          "required": true,
+          "required": false,
```
mshameti, line 25419: Is there a way to enforce that one of query_vector or query_vector_builder are defined through TS and/or the spec? The way the type would work if we were to merge this would inform the user that both fields are optional

AlmostFamiliar: Not that I know of. I just copied this over from the `KnnQuery` Interface, because I saw it was out of date and I thought I might as well fix that clearly out of date part of the code. 

I could just remove this part from the MR and this can be fixed another time. The Java Client only needs the `KnnQuery` interface part, as far as I can tell. That way the similarity field would work and people could use it. Should I do that?

bijela-gora: @mshameti 

> Is there a way to enforce that one of query_vector or query_vector_builder are defined through TS and/or the spec?

We can do this:
```ts
export type KnnQuery = {
  /** The name of the vector field to search against */
  field: Field
  /** The final number of nearest neighbors to return as top hits */
  k: integer
  /** The number of nearest neighbor candidates to consider per shard */
  num_candidates: integer
  /** The minimum similarity required for a document to be considered a match */
  similarity?: float
  /** Boost value to apply to kNN scores */
  boost?: float
} & (
  {
    /** The query vector */
    query_vector: QueryVector
  } & {
    /** The query vector builder. You must provide a query_vector_builder or query_vector, but not both. */
    query_vector_builder: QueryVectorBuilder
  }
)
```

But we will get this error:
```
Error: Unhandled node kind: IntersectionType
```

specification/_global/knn_search/_types/Knn.ts
```diff
@@ -18,16 +18,25 @@
...
   /** The final number of nearest neighbors to return as top hits */
   k: long
   /** The number of nearest neighbor candidates to consider per shard */
   num_candidates: long
+  /** Boost value to apply to kNN scores */
+  boost?: float
+  /** Filters for the kNN search query */
+  filter?: QueryContainer | QueryContainer[]
+  /** The minimum similarity for a vector to be considered a match */
+  similarity?: double
```
bijela-gora, line 41: @pstrsr hello, according to https://www.elastic.co/guide/en/elasticsearch/reference/8.10/knn-search-api.html the `_knn_search` API doesn't have `similarity` parameter.

I also want to see `similarity` parameter published: https://github.com/elastic/elasticsearch-specification/pull/2293 . Because of that I suggest to revert changes to this file completely.

pstrsr: Thanks, it seems that I confused the APIs. Reverted the changes.

pstrsr:
Closing because https://github.com/elastic/elasticsearch-specification/pull/2261 got merged

PR #2226, opened 2023-08-04
[DOCS] Adds descriptions to index APIs (A-F)

szabosteve:
## Overview

Related to https://github.com/elastic/elasticsearch-specification/issues/964.

Adds descriptions to index API properties (A-F). Source of info: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices.html

PR #2229, opened 2023-08-08
Fix broken doc url for indices-close

JoshMock:
CI found a broken doc link on my latest [build of the JS client](https://github.com/elastic/elasticsearch-js/pull/1946). Looks like a slash was missed in https://github.com/elastic/elasticsearch-specification/pull/2226.


abdonpijpelink (approved):
Thanks for fixing this!

PR #2234, opened 2023-08-11
Renamed synonyms API namespaces

carlosdelest:
Renaming synonyms API specs according to https://github.com/elastic/elasticsearch/issues/98364.

This is a follow-up PR for https://github.com/elastic/elasticsearch/pull/98383, and must be merged after that one.

Closes https://github.com/elastic/elasticsearch/issues/98364

PR #2235, opened 2023-08-11
Fix and tune synonyms API

swallez:
Follow-up to #2234

This PR fixes some modelling issues and refines the specification:
* Rename `SynonymRuleOptionalId` and `SynonymRule` to `SynonymRule` and `SynonymRuleRead`, respectively, and use the [`OverloadOf` feature](https://github.com/elastic/elasticsearch-specification/blob/main/docs/behaviors.md#overloadof).
* Removed inheritance between `SynonymRule` variations, since a subclass should not override a property defined in its parent class (need to add a validation rule for this)
* Use `Id` instead of `Name`
* Use `@server_default` to capture default values
* Fix response definitions where `_shards: ShardStatistics` was not a the correct nesting level


carlosdelest (approved):
LGTM, thank you for the improvements! 💯

PR #2242, opened 2023-08-17
TypeScript fixes

JoshMock:
Several new APIs were not exporting their `Request` interfaces. Also found some files importing modules they weren't using or that didn't exist, and noticed a few paths missing from the tsconfig.


JoshMock:
I haven't verified this, but I'm curious if some of these fixes will clear up some odd output I've seen from Python codegen this week.

kderusso (approved):
Thank you!

PR #2250, opened 2023-08-25
Change TaskStatus to UserDefinedValue

### LINKED ISSUE #2249: Revisit the TaskInfo API types
## 🐛 Wrong type

[`TaskStatus`](https://github.com/elastic/elasticsearch-specification/blob/7.17/specification/tasks/_types/TaskStatus.ts) has a number of fields that were gathered from flight recordings.

It turns out that this definition is wrong and this type should probably be a `UserDefinedValue`: in the [ES code base this is just an interface](https://github.com/elastic/elasticsearch/blob/47c1d99ae0eeb67806db35730f1510c70a9c0aef/server/src/main/java/org/elasticsearch/tasks/Task.java#L244) with this comment: "_These can vary wildly from task to task because each task is implemented differently but we should try to keep each task consistent from version to version where possible_".

swallez:
Replaces `TaskStatus` with a `UserDefinedValue`. Fixes #2249.

In the [ES code base this is just an interface](https://github.com/elastic/elasticsearch/blob/47c1d99ae0eeb67806db35730f1510c70a9c0aef/server/src/main/java/org/elasticsearch/tasks/Task.java#L244) with this comment: "_These can vary wildly from task to task because each task is implemented differently but we should try to keep each task consistent from version to version where possible_".

So we cannot make any assumption on the fields that will actually be there.


technige (approved):
Makes sense to me, given the reasoning you mentioned on Slack.

PR #2255, opened 2023-08-28
Expand snapshot repository types

JoshMock:
Use a different type definition for each repository type since differences are significant.

A community member [noted](https://github.com/elastic/elasticsearch-js/issues/1988) that some types were missing for S3 snapshot repositories, and inspecting the spec made it clear that it was an issue beyond just S3.


specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,126 @@
...
  */
 
-import { Uuid } from '@_types/common'
+import { ByteSize, Uuid } from '@_types/common'
 import { integer } from '@_types/Numeric'
+import { Duration } from '@_types/Time'
 
-export class Repository {
-  type: string
+export type Repository =
```
swallez, line 24: We need [a variant annotation indicating that `type` is the discriminator](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#variants), and also the indication that the enumeration isn't exhaustive as new implementations can (theoretically) be added using plugins.

```ts
/**
 * @variants internal tag='type'
 * @non_exhaustive
 */
```

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,126 @@
...
+  settings: ReadOnlyUrlRepositorySettings
+}
+
+export class SourceOnlyRepository {
+  type: 'source'
+  uuid?: Uuid
+  settings: SourceOnlyRepositorySettings
+}
+
+export type RepositorySettings =
```
swallez, line 68: Seems like we don't need this type.

specification/snapshot/_types/SnapshotRepository.ts
```diff
@@ -17,22 +17,126 @@
...
+  | AzureRepository
+  | GcsRepository
+  | S3Repository
+  | SharedFileSystemRepository
+  | ReadOnlyUrlRepository
+  | SourceOnlyRepository
+
+export class AzureRepository {
+  type: 'azure'
+  uuid?: Uuid
```
swallez, line 34: Move `uuid` to a common `RepositoryBase` so that we can access the id of a repository without having to downcast it to its actual type.

swallez (changes requested):

specification/snapshot/_types/SnapshotRepository.ts
JoshMock: It gets imported and used [here](https://github.com/elastic/elasticsearch-specification/blob/b3a1e6f30b6a0e19b644c406331e1d56bd2907a0/specification/snapshot/create_repository/SnapshotCreateRepositoryRequest.ts#L46). I'm not sure why the request `body` accepts both `repository` and `settings` keys when `repository.settings` also exists. :shrug:

specification/snapshot/_types/SnapshotRepository.ts
JoshMock: Fixed in ac703f9845694604552925c12332b73ceaf8cc85

specification/snapshot/_types/SnapshotRepository.ts
JoshMock: Fixed in f9550144a5cf45ba5389786ea73132c6859db39b.

DesAWSume:
For interface **SnapshotRepositorySettings** , it has **read_only** and **readonly**, any way we can make it clearer what each individual behave?

specification/snapshot/_types/SnapshotRepository.ts
swallez: Hmm... I'll look at the ES source code to understand if this is needed. Now `@variants internal tag='type'` won't work here since the union members don't have a `type` property.

JoshMock:
@DesAWSume `readonly` is an alias for `read_only` in the spec, I presume for some backward compatibility. I'm not sure if there's a clear way in TypeScript or TSDoc/JSDoc to annotate that fact, but I can take a look. I get that it can be confusing.

specification/snapshot/_types/SnapshotRepository.ts
JoshMock: Good catch on the `type` property. Fixed that.

> Hmm... I'll look at the ES source code to understand if this is needed.

I am still not great at navigating the Java codebase, but if [this is the request](https://github.com/elastic/elasticsearch/blob/1ca66bde9109bcc716a692be4a3531c0418299e8/server/src/main/java/org/elasticsearch/action/admin/cluster/repositories/put/PutRepositoryRequest.java#L34-L40), it certainly does not seem like it supports an optional `repository` property. I scrolled through a lot of the repository creation tests, too, and didn't see it. Feel free to correct my reading of it if I'm totally in the wrong place.

JoshMock:
@DesAWSume I did discover [JSDoc has an `@alias` tag](https://jsdoc.app/tags-alias.html), but its usefulness likely depends on how your IDE uses it. In mine (Neovim with language server support) it just adds an `@alias` to the type definition popup, but still allows you to use both.

2023-08-31T16:12:52,740847821-05:00

JoshMock:
@swallez could I get a final review on this? only outstanding thing is [this concern](https://github.com/elastic/elasticsearch-specification/pull/2255#discussion_r1312211500), but it was already in there like this so maybe we can address separately.

swallez (changes requested):

JoshMock:
@swallez One month and a parental leave later... I think I got the last of this sorted out. :laughing: Please take a look and let me know if I addressed what you were saying properly.

JoshMock:
thanks so much @swallez! :heart:

PR #2261, opened 2023-08-31
Added the similarity parameter to the KnnQuery type

pstrsr:
Fix for: https://github.com/elastic/elasticsearch-specification/issues/2149
and: https://github.com/elastic/elasticsearch-java/issues/643
and: https://github.com/elastic/elasticsearch-java/issues/654

Simplified version of: https://github.com/elastic/elasticsearch-specification/pull/2224, so that people can use vector search in their Java services.


mshameti:
It might make sense to just hold off or adjust based on the API changes that will stem from:
https://github.com/elastic/elasticsearch/pull/98916

AlmostFamiliar:
I guess that makes sense. 

It is always possible to not use the offical elastic clients in the meantime as a workaround, so it is not super critical I suppose.

mshameti:
In JS, you can still use the official elastic client of the =>8.8.0 version which is when `similarity` parameter was launched https://www.elastic.co/guide/en/elasticsearch/reference/8.8/knn-search.html.  

you'd just need to //@ts-ignore if you're using TS or amend the types with a .d.ts file.

AlmostFamiliar:
We are sadly using Java for our query services, so that is not possible. But until the feature is fully supported by Elastic, we can use the rest client.

Maybe a hint in the documentation that the vector search is not stable would be nice.

The python client also works at the moment, if that helps anyone. We have used it for some prototypes with semantic search.

specification/_types/Knn.ts
```diff
@@ -38,6 +38,8 @@ export interface KnnQuery {
   boost?: float
   /** Filters for the kNN search query */
   filter?: QueryContainer | QueryContainer[]
+  /** The minimum similarity for a vector to be considered a match */
+  similarity?: double
```
bijela-gora, line 42:
```suggestion
  similarity?: float
```

According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/search-search.html) it should be `float`.

bijela-gora:
@mshameti @pstrsr  hello

Can we also make `k` and `num_candidates` to have `integer` parameters in this MR?

bijela-gora:
@mshameti hello, you wrote
> It might make sense to just hold off

It may also make sense to align the capabilities of client libraries with those of the HTTP API. What do you think?

specification/_types/Knn.ts
pstrsr: thank you, changed it

bijela-gora:
@JoshMock hello

Just a kind reminder that we are one click away from improving the API.

JoshMock (approved):

JoshMock:
Thanks for the reminder @bijela-gora. I've merged and backported this to 8.8, 8.9 and 8.10 branches so we should expect this to be fixed in each client in the next patch release for any of those branches, or at the very least 8.11.0.

PR #2276, opened 2023-09-14
Update QueryRulesetPutRequest body to better align with REST API

kderusso:
Fix. See https://github.com/elastic/elasticsearch-py/pull/2299


carlosdelest (approved):

pquentin (approved):
Thanks! LGTM. I confirmed with @Anaethelion that the Python client should probably not be affected by the change here, but is in practice, so let's merge this to unblock you.

PR #2279, opened 2023-09-19
Fix include type in Significant Text agg

### LINKED ISSUE #2275: Significant Text aggregation missing regex include method
Found in https://github.com/elastic/elasticsearch-java/issues/544

The definition of SignificantTextAggregation.include: string|string[] does not represent the usual array leniency. Both variants have a specific meaning, resp. regexp and term list.

It should be typed as TermsInclude.

(Version 8.9.2)

pquentin:
Closes #2275 

I'm not sure where this should be backported exactly. This aggregation was introduced in Elasticsearch 6.0.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -3926,7 +3926,7 @@ export interface AggregationsSignificantTextAggregation extends AggregationsBuck
   field?: Field
   filter_duplicate_text?: boolean
   gnd?: AggregationsGoogleNormalizedDistanceHeuristic
-  include?: string | string[]
+  include?: AggregationsTermsInclude
   jlh?: EmptyObject
   min_doc_count?: long
   mutual_information?: AggregationsMutualInformationHeuristic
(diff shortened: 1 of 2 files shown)


JoshMock (approved):

JoshMock:
> I'm not sure where this should be backported exactly. This aggregation was introduced in Elasticsearch 6.0.

Since we're not likely to release any minor/patch versions older than 8.8, we could probably just backport to 8.8, 8.9 and 8.10 for now. For other versions we can probably address on a case-by-case basis so you don't have to open 20 backport PRs. :laughing:

pquentin:
Backports:

* https://github.com/elastic/elasticsearch-specification/pull/2292
* https://github.com/elastic/elasticsearch-specification/pull/2291
* https://github.com/elastic/elasticsearch-specification/pull/2290
* https://github.com/elastic/elasticsearch-specification/pull/2285

PR #2296, opened 2023-09-27
Fix typo in table.csv

lcawl:
There was an extra comma in this file, which was generating a warning in GitHub.


abdonpijpelink (approved):
Thanks for fixing this @lcawl !

PR #2301, opened 2023-10-02
Add additional counters in _clusters response

piergm:
Updates the spec for the response objects for `POST _async_search`, `GET _async_search/:id`, `GET _async_search/status/:id` and `GET _search`.

In `8.11` we are adding counters in the `_details` section of both async and synchronous search responses. The counters added are: `running`, `partial` and `failed`. https://github.com/elastic/elasticsearch/pull/99566


Anaethelion (approved):
LGTM

PR #2304, opened 2023-10-06
Add sparse_vector field type

carlosdelest:
Add sparse_vector field support (added to Elasticsearch https://github.com/elastic/elasticsearch/pull/98996) to mappings, and include it as usable in the docs for `text_expansion` query.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -5000,7 +5000,7 @@ export interface MappingFieldNamesField {
   enabled: boolean
 }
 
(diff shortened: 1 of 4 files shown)


pquentin (approved):
Thanks! LGTM.

By the way I think the documentation is missing for that field in the Elasticsearch docs. I can see rank_feature but not sparse_vector: https://www.elastic.co/guide/en/elasticsearch/reference/master/rank-feature.html

carlosdelest:
> I think the documentation is missing for that field in the Elasticsearch docs

Thanks for looking into that @pquentin ! It's just been merged in https://github.com/elastic/elasticsearch/pull/100348 👍

PR #2316, opened 2023-10-23
[WASM lib] Add compiler error with instructions to build the webassembly

Anaethelion:
This adds a simple compilation error in case someone tries to build the lib with a direct `cargo build`.


swallez (approved):
LGTM. Nice, as indeed this crate only makes sense in the context of a WASM compilation.

PR #2329, opened 2023-11-03
Add _inference API

davidkyle:
Add the _inference CRUD APIs.

Each model has a `service_settings` and `task_settings` configuration with is related to the service used to evaluate the model. These options are specific to the service,m for example if using the HuggingFace service the options are different to using the ELSER service.  The `ServiceSettings` and `TaskSettings` classes used to represent these options are currently empty as they are not explicitly modelled, I'm happy for these to be untyped at the moment and can be considered loose maps of <string, string>.

The APIs are not in serverless yet, the new APIs are tagged `visibility=private` for serverless availability.


specification/inference/_types/Results.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { double } from '@_types/Numeric'
+import { Dictionary } from '@spec_utils/Dictionary'
+
+/**
+ * Sparse Embedding tokens are represented as a dictionary
+ * of string to double.
+ */
+
+export type SparseVector = Dictionary<string, double>
```
jonathan-buttner, line 28: In the java code I think we store the results as floats right?

https://github.com/elastic/elasticsearch/blob/b055204b436f3b4d3729500b09bc932a21dd2e8c/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/results/TextExpansionResults.java#L27

specification/inference/inference/InferenceResponse.ts
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { InferenceResult } from '@inference/_types/Results'
+
+export class Response {
+  body: {
+    predicted_value: InferenceResult
```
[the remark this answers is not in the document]
jonathan-buttner: I wonder if we should include a field in the response that indicates what type it is. OpenAI does this using the `object` field. I wonder if it'd be easier for clients to key off of that field rather than remembering the task type in the request URL 🤷‍♂️

jonathan-buttner (commented):
Not sure if we need to include this now but will we support an array for the `input` request field?

If so, we might want to structure the responses something like this that way we don't have a breaking change in the response format:

```
{
  "<some field>": [
    {
     ...
    },
    {
     .... 
    }
  ]
}
```

For text embeddings, how do we feel about a structure like this?

```
{
  "<some field>": [
    {
      "embedding": [0.123, -0.456, ...]
    },
    {
      "embedding": [0.123, -0.456, ...]
    }
  ]
}
```

Having an array of objects in the response gives us more flexibility to add a field later if we need to. If they are just arrays it'll be harder to add a new field.

specification/inference/_types/Results.ts
davidkyle: ++ I've changed this and the dense vector to float

jonathan-buttner (approved):

davidkyle:
> I'm wondering if we actually need task_type in the URL path for all these APIs:

The response is different depending on task_type. If we know the task_type there is validation we can perform that the model is of the correct type and the input is properly constructed. Additionally, the input may be structured differently for certain task_types. 

I can see a cases where it would be preferable to elide task_type, one example is GET model by model Id without specifying the task type. That isn't implemented yet, when it is the spec will be updated. 

> does the task type define an isolated namespace for model_id, i.e. can we have both text_embedding/foobar and sparse_vector/foobar?

`model_id`s are globally unique you cannot have the same model_id for 2 models of different task types. 

> the task_type is also present in some of the response bodies (in ModelConfigContainer), and could equally be made part of ModelConfig, or maybe we don't even need it there since it is a characteristic of the chosen service?

`ModelConfig` is used at model creation where the task type is read from the URL path. `ModelConfigContainer` is used in the GET model responses and is the same as `ModelConfig` with the extra fields `model_id` and `task_type` both of which are read from the URL at model creation (PUT). Is there a preferred convention modelling this type of pattern?

specification/inference/inference/InferenceResponse.ts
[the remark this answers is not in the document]
jonathan-buttner: @davidkyle for `SparseVector` should we call it `SparseEmbedding` and have the field be `sparse_embedding`?

The `TaskType` in the java code is `SparseEmbedding`.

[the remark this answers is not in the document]
davidkyle: 👍 thanks for reviewing @jonathan-buttner. I've renamed the field `sparse_embedding` but kept the type as `SparseVector`. The reasoning being that the mapping types in Elasticsearch are `sparse_vector` and `dense_vector`. A `text_embedding` goes in a `dense_vector` and a `sparse_embedding` in a `sparse_vector`

specification/inference/_types/Services.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Task settings specific to the service and model
+   */
+  task_settings: TaskSettings
+}
+
+/**
+ * Represents a model as returned by the GET API
+ */
+export class ModelConfigContainer {
```
swallez, line 44: There's a lot of overlap between `ModelConfig` and `ModelConfigContainer`.

Do we need `task_type` and `model_id` here, considering that the user provided them in the request? If we remove them, then we can keep only `ModelConfig`.

If it's deemed necessary to repeat this information, then we should replace `service`, `service_settings` and `task_settings` with a single `config: ModelConfig` field.

swallez (changes requested):
Sorry for the delay. I left a comment about the redundancy between `ModelConfig` and `ModelConfigContainer` and how we can address it.

davidkyle:
Ping @swallez, could you take another look at this PR please

davidkyle:
> Ping @swallez, could you take another look at this PR please

Sorry @swallez I did not refresh the page and now I see your reply. Thanks I'll look at your comments now

specification/inference/_types/Services.ts
davidkyle: Thanks for the feedback @swallez, your comments have shown that the GET models API is not modelled correctly.

In future the GET inference model API will support wild card expansion. For example get all the text embedding  models:
```
GET _inference/text_embedding/*
```

To support this `GetModelResponse` should be an Array of `ModelConfigContainer` not a single value. Once we have an array of model configs it is clear why `model_id` and `task_type` are required in the response because `model_id` was not included in the request. The same applies to `task_type` if we were to ask for all models regardless of task the response should contain the task type.


> If it's deemed necessary to repeat this information, then we should replace service, service_settings and task_settings with a single config: ModelConfig field.

++ I like to reuse. How can I incorporate the fields from `ModelConfig` into `ModelConfigContainer` without having `ModelConfig` as a nested object? I want just the fields defined in `ModelConfig` to appear in `ModelConfigContainer`.

```
export class ModelConfigContainer {
  /**
   * The model Id
   */
  model_id: string
  /**
   * The model's task type
   */
  task_type: TaskType

  include all the fields defined in ModelConfig
}
```

swallez: Thanks for the explanation regarding wildcards.

Regarding reuse, we can have `ModelConfigContainer extends ModelConfig` with additional fields to achieve what you want.

Now there's nothing wrong with nesting (composition). Some of the target languages for client libraries (e.g. Go) don't support inheritance and will "flatten" inheritance hierarchies and produce unrelated types. So if we want to allow roundtripping (e.g. scenarios like get a ModelConfig, tweak it and Post it back), composition makes it easier.

davidkyle: 👍 I pushed the `ModelConfigContainer extends ModelConfig` change.

The inability to round trip config is a common problem in Elasticsearch and it's quite wrong headed that you can't GET then PUT. I will ask for some more opinions as I'd prefer not to change the format and the nested config object makes it harder to read

davidkyle: We looked at the round-tripping problem and realised there are other reason the prevent round-tripping anyway. The configuration contains secret settings such as an API key. The API key is included on PUT to create the model but because it is a secret it is never returned on GET. This means that you can't simply copy and paste the config as you need to insert the API key.

Our preference is to have the GET response contain flat fields as we think that is more readable than nesting the config part in another object.

davidkyle:
I've removed the backport 8.11 label as there have been many changes to the API since 8.11

swallez (approved):
LGTM. Thanks!

PR #2337, opened 2023-11-06
Index create request should support singular 'mapping'

JoshMock:
[The current indices create spec](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-settings-limit.html#mapping-settings-limit) expects the term `mapping` rather than `mappings`. Adding as an alias to avoid breaking compatibility.

See https://github.com/elastic/elasticsearch-js/issues/1753.

### DIFF
--- a/specification/indices/create/IndicesCreateRequest.ts
+++ b/specification/indices/create/IndicesCreateRequest.ts
@@ -71,6 +71,7 @@ export interface Request extends RequestBase {
      * - Field data types
      * - Mapping parameters
      * @doc_id mapping
+     * @aliases mapping
      */
     mappings?: TypeMapping
     /**


swallez (changes requested):
If we go back in time, [it was `mapping` for ages](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-settings-limit.html#mapping-settings-limit) (even forever ?). So we should just rename the property, since no user was ever able to use this property, either for configuring or reading the settings.

Note also that aliases are used for parsing and not serializing. As it is, in strongly typed clients the property will still be serialized as `mappings`.

JoshMock:
On closer inspection, the [indices-create docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-create-index.html) does use `mappings` at the root, and the [doc linked](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-settings-limit.html#mapping-settings-limit) in [the original issue](https://github.com/elastic/elasticsearch-js/issues/1753) is referring to `body.settings.mapping`.

I don't know why there are two ways to provide a mapping in `indices.create`, but the current spec does seem to reflect this properly.

PR #2317, opened 2023-10-26
Allow single fields in fields and _source parameters

pquentin:
This was initially reported in https://github.com/elastic/elasticsearch-py/pull/2347

The `_source` body field used to work but was broken in https://github.com/elastic/elasticsearch-specification/pull/1027/files#diff-4114e335296798822015a064d77c5c6d14c797a89c76a0d08b63dcf5d4028651.

Based on https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html, I fixed `_source`, `fields` and `docvalue_fields`.

PR #2341, opened 2023-11-07
Revert "Allow single fields in fields and _source parameters"

pquentin:
Reverts elastic/elasticsearch-specification#2317. The `FieldAndFormat` type has a `@shortcut_property` field which means that if it's a scalar value, it is equivalent to a `FieldAndFormat` with only field defined. The code generation strategy for Python should be to lift the shortcut property into a union, so that `FieldAndFormat` becomes `Field | FieldAndFormat`.


Anaethelion (approved):
LGTM

PR #2343, opened 2023-11-08
[JSON spec] sync spec files

Anaethelion:
This PR syncs the JSON spec files with the v8.11.0 Elasticsearch tag.


JoshMock (approved):
LGTM.

Do you just copy these over from the ES repo's `rest-api-spec` directory and run the compiler? Just curious if we can automate it.

Anaethelion:
Yep its a direct copy of the spec files from the Elasticsearch repo with the exception of `_common.json` which shouldn't be copied, `make contrib` takes care of reformatting the JSON files and regen the API.

PR #2345, opened 2023-11-14
Add sparse vector property

Anaethelion:
This add the SparseVectorProperty as part of the Property union.


swallez (approved):
LGTM.

Before merging, can you add `output/openapi/elasticsearch-serverless-openapi.json` to the `.gitattributes` file so that [its diff is not expanded by default](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)?

PR #2347, opened 2023-11-14
Enrich policy query should be a QueryContainer

JoshMock:
See https://github.com/elastic/elasticsearch-js/issues/2074


specification/enrich/_types/Policy.ts
```diff
@@ -34,7 +35,7 @@ export class Policy {
   enrich_fields: Fields
   indices: Indices
   match_field: Field
-  query?: string
+  query?: QueryContainer | QueryContainer[]
```
Anaethelion, line 38: Agreed this should be a `QueryContainer` but I'm failing to find an example for the array version.
Server code seem to only support one `query` per policy according to `EnrichPolicy.java`.

```suggestion
  query?: QueryContainer
```

Anaethelion (changes requested):

specification/enrich/_types/Policy.ts
JoshMock: @simianhacker since you opened the original issue on the JS client, do you have any insight into this?

simianhacker: Yeah... I think having it as `QueryContainer` instead of `QueryContainer | QueryContainer[]` is probably the correct type. I think I copied that type from an example but didn't think about how this is the top level `QueryContainer`. Good catch @Anaethelion!

Anaethelion (approved):
LGTM

PR #2349, opened 2023-11-15
Change replacement for PathHierarchyTokenizer to be optional

Anaethelion:
As per the docs for [Path hierarchy tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pathhierarchy-tokenizer.html#analysis-pathhierarchy-tokenizer) the replacement field should be optional.


JoshMock (approved):
looks good

PR #2354, opened 2023-11-22
Initial version of ES|QL query endpoint

Anaethelion:
Based on the following resources:

[ES|QL REST API](https://www.elastic.co/guide/en/elasticsearch/reference/current/esql-rest.html)
[EsqlQueryRequest.java](https://github.com/elastic/elasticsearch/blob/b08f1e4090f6dae3a582637caab9436fc471bcfb/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/EsqlQueryRequest.java)
[QueryPragmas.java](https://github.com/elastic/elasticsearch/blob/5e6bf3ceb1c6b1f9df7e169a57fb648d655aa320/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/plugin/QueryPragmas.java)

### DIFF
--- a/specification/esql/query/QueryRequest.ts
+++ b/specification/esql/query/QueryRequest.ts
@@ -1,0 +1,65 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { RequestBase } from '@_types/Base'
+import { QueryContainer } from '@_types/query_dsl/abstractions'
+import { TimeZone } from '@_types/Time'
+import { Pragmas } from 'esql/_types/Pragmas'
+
+/**
+ * Executes an ESQL request
+ * @rest_spec_name esql.query
+ * @availability stack since=8.11.0 stability=experimental
+ */
+export interface Request extends RequestBase {
+  query_parameters: {
+    /**
+     * A short version of the Accept header, e.g. json, yaml.
+     */
+    format?: string
+    /**
+     * The character to use between values within a CSV row. Only valid for the csv format.
+     */
+    delimiter?: string
+  }
+  /**
+   * Use the `query` element to start a query. Use `time_zone` to specify an execution time zone and `columnar` to format the answer.
+   */
+  body: {
+    /**
+     * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
+     */
+    query: string
+    /**
+     * By default, ES|QL returns results as rows. For example, FROM returns each individual document as one row. For the json, yaml, cbor and smile formats, ES|QL can return the results in a columnar fashion where one row represents all the values of a certain column in the results.
+     */
+    columnar?: boolean
+    time_zone?: TimeZone
+    /**
+     * Specify a Query DSL query in the filter parameter to filter the set of documents that an ES|QL query runs on.
+     */
+    filter?: QueryContainer
+    /**
+     * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
+     */
+    params?: Array<string>
+    pragmas?: Pragmas
+    locale?: string
+  }
+}
--- a/specification/esql/query/QueryResponse.ts
+++ b/specification/esql/query/QueryResponse.ts
@@ -1,0 +1,24 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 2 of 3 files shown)


specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Specify a Query DSL query in the filter parameter to filter the set of documents that an ES|QL query runs on.
+     */
+    filter?: QueryContainer
+    locale?: string
+    /**
+     * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
+     */
+    params?: Array<string>
+    pragmas?: Pragmas
```
bpintea, line 58: I don't think we need to expose these. They're mostly development aids and aren't documented.

specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @availability stack since=8.11.0 stability=experimental
+ */
+export interface Request extends RequestBase {
+  query_parameters: {
+    /**
+     * A short version of the Accept header, e.g. json, yaml.
+     */
+    format?: string
+    /**
+     * The character to use between values within a CSV row. Only valid for the csv format.
```
bpintea, line 37:
```suggestion
     * The character to use between values within a CSV row. Only valid for the CSV format.
```

specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+     */
+    delimiter?: string
+  }
+  /**
+   * Use the `query` element to start a query. Use `time_zone` to specify an execution time zone and `columnar` to format the answer.
+   */
+  body: {
+    /**
+     * By default, ES|QL returns results as rows. For example, FROM returns each individual document as one row. For the json, yaml, cbor and smile formats, ES|QL can return the results in a columnar fashion where one row represents all the values of a certain column in the results.
```
bpintea, line 46: Another comments pedantic nit, in case we want to capitalise the acronyms JSON, YAML, CBOR in this PR.

specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
+     */
+    params?: Array<string>
+    pragmas?: Pragmas
+    /**
+     * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
+     */
+    query: string
+    time_zone?: TimeZone
```
bpintea, line 63: This will be removed soon (and added back, but at a later time, since its support is currently patchy), so we can leave it out.

specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+ * under the License.
+ */
+
+import { RequestBase } from '@_types/Base'
+import { QueryContainer } from '@_types/query_dsl/abstractions'
+import { TimeZone } from '@_types/Time'
+import { Pragmas } from '@esql/_types/Pragmas'
+
+/**
+ * Executes an ESQL request
```
bpintea, line 26: A pedantic nit about comments' consistency: either `ESQL` or `ES|QL`.

specification/esql/query/QueryRequest.ts
```diff
@@ -0,0 +1,65 @@
...
+    columnar?: boolean
+    /**
+     * Specify a Query DSL query in the filter parameter to filter the set of documents that an ES|QL query runs on.
+     */
+    filter?: QueryContainer
+    locale?: string
+    /**
+     * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
+     */
+    params?: Array<string>
```
bpintea, line 57: These aren't string only, it can be a string, number, boolean or `null` (currently) and ESQL isn't tolerant and converting as needed.
I wonder if we should add links to the docs? Would it be useful in these comments? (Like https://www.elastic.co/guide/en/elasticsearch/reference/master/esql-rest.html#esql-rest-params in this case). I can provide them for the other params (but are mostly just from here: https://www.elastic.co/guide/en/elasticsearch/reference/master/esql-rest.html).

bpintea (commented):
Thanks for starting this, @Anaethelion!
I've left some comments.

specification/esql/query/QueryRequest.ts
Anaethelion: Granted, that line was taken verbatim from the `esql.query.json` spec file. I'll leave to you the choice to update that file as well, if you do I'll resync the one in the spec with the one in the server repo afterward.

specification/esql/query/QueryRequest.ts
Anaethelion: Also taken verbatime from the rest spec file.

specification/esql/query/QueryRequest.ts
Anaethelion: I'll remove that expecting it won't be present in `8.12`.

specification/esql/query/QueryRequest.ts
Anaethelion: The docs are usually inlined for each fields to be propagated to each language doc.
I'll add the second link you provided to the `table.csv` file so its found in the compiled and within the clients.

As for the type of `params`we can go for `FieldValue` which is `long | double | string | boolean | null | any`, does that make sense ?

specification/esql/query/QueryRequest.ts
bpintea: I see. I think that comes from the `format=` parameter value, which is "usually" not capitalised (`?format=csv`), though it can be. But yes, it should be at least quoted there. Thanks!

specification/esql/query/QueryRequest.ts
bpintea: I think Elasticsearch has a myriad of settings available, but not "advertised". This is an API and maybe different, but I doubt we want to provide things like backward compatibility for these parameters. 
I will check with the team, but would err on removing them.

specification/esql/query/QueryRequest.ts
Anaethelion: Thank you, I agree especially for arrays.

With @swallez we've come to the conclusion that `any` shouldn't be part of this union. To that effect I've added a new `ScalarValue` that maps to `long | double | string | boolean | null` which reflects better.

specification/esql/query/QueryResponse.ts
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+
+export class Response {
+  body: UserDefinedValue
```
swallez, line 23: `UserDefinedValue` assumes JSON. Maybe use the more general `ArrayBuffer` like for `SearchMvtResponse`?

specification/esql/query/QueryRequest.ts
Anaethelion: Thank you for diving on this! I'll remove `pragmas` from the PR.

PR #2356, opened 2023-11-28
Add transform-to-openapi to auto regen Github action

Anaethelion:
This add the `transform-to-openapi` target to the `generate` Github Action.


swallez (approved):
LGTM. Good catch!

PR #2362, opened 2023-12-08
Fix FieldRule

### LINKED ISSUE #2344: SecurityRealm type in SecurityPutRoleMappingRequest results in a parsing error.
## 🐛 Bug Report

The SecurityPutRoleMappingRequest type expects to specify a SecurityRealm in a SecurityFieldRule of a SecurityRule as:

```typescript
  {
  field: {
    realm: {
      name: "RealmName"
    }
  }
}
```

However doing this results in the following parsing error:

Error: parsing_exception
	Caused by:
		x_content_parse_exception: [1:77] [role-mapping] failed to parse field [rules]
	Root causes:
		parsing_exception: [1:77] [role-mapping] failed to parse field [rules]

I have played a bit around with it and got the desired behaviour by changing the input to:

```typescript
  {
  field: {
    "realm.name": "RealmName"
  }
}
```

It seems that the type SecurityRealm type should either be changed to use "realm.name" or handle the behaviour before forwarding the body to the api.

## To Reproduce

Attempt to call the elasticsearch client with .security.putRoleMapping() containing the input described above

## Expected behavior

I expected the elasticsearch client to map the specified type to an accepted api request before forwarding the apicall.

## Your Environment

- *node version*: ">=18.0.0"
- `@elastic/elasticsearch` *version*: "^8.7.0"
- *os*: Linux

pquentin:
Closes #2344

This fixes three issues with `FieldRule`:

* `username` can also be a list,
* `realm.name` is the field name, using an intermediate `realm` object leads to a parsing exception,
* `metadata` is not an object: values are of the form `metadata.key = value`.

Sources:
 * https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role-mapping.html for the general documentation of `POST /_security/role_mapping/<name>`, including complete examples
 * https://www.elastic.co/guide/en/elasticsearch/reference/current/jwt-auth-realm.html#jwt-authorization-role-mapping which has a an extensive list of possible rules

I only tested in Kibana, as the Python client only exposes `rules` and does not go deeper.

Regarding the backports, `realm.name` and `metadata` are unusable today, but `username` is. Should this be separated in two pull requests, so that the `realm.name` and `metadata` get backported to 8.12, 8.11 and 7.17?

### DIFF
--- a/specification/security/_types/RoleMappingRule.ts
+++ b/specification/security/_types/RoleMappingRule.ts
@@ -35,8 +35,7 @@ export class RoleMappingRule {
  * @variants container
  * @non_exhaustive
  */
-export class FieldRule
-{
+export class FieldRule {
   username?: Names
   dn?: Names
   groups?: Names
--- a/specification/security/_types/RoleMappingRule.ts
+++ b/specification/security/_types/RoleMappingRule.ts
@@ -33,9 +33,9 @@ export class RoleMappingRule {
 
 /**
  * @variants container
+ * @non_exhaustive
  */
 export class FieldRule
-  implements AdditionalProperties<string, UserDefinedValue>
 {
   username?: Names
   dn?: Names
--- a/specification/security/_types/RoleMappingRule.ts
+++ b/specification/security/_types/RoleMappingRule.ts
@@ -17,7 +17,6 @@
  * under the License.
  */
 
-import { AdditionalProperties } from '@spec_utils/behaviors'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 import { Name, Names } from '@_types/common'
 
--- a/specification/security/_types/RoleMappingRule.ts
+++ b/specification/security/_types/RoleMappingRule.ts
@@ -26,6 +26,9 @@ import { Name, Names } from '@_types/common'
 export class RoleMappingRule {
   any?: RoleMappingRule[]
   all?: RoleMappingRule[]
+  // `field` should have been defined as SingleKeyDictionary<String, ScalarValue|ScalarValue[]>
+  // However, this was initially defined as a container with a limited number of variants,
+  // and was later made non_exhaustive to limit breaking changes.
   field?: FieldRule
   except?: RoleMappingRule
 }
@@ -38,5 +41,4 @@ export class FieldRule {
   username?: Names
   dn?: Names
   groups?: Names
-  'realm.name'?: Name
 }
--- a/specification/security/_types/RoleMappingRule.ts
+++ b/specification/security/_types/RoleMappingRule.ts
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+import { AdditionalProperties } from '@spec_utils/behaviors'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 import { Name, Names } from '@_types/common'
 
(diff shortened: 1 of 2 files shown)


specification/security/_types/RoleMappingRule.ts
```diff
@@ -33,14 +34,11 @@ export class RoleMappingRule {
 /**
  * @variants container
```
Anaethelion, line 35:
```suggestion
 * @variants container
 * @non_exhaustive
```
Add a non exhaustive tag insteand of an additional properties implementation.

specification/security/_types/RoleMappingRule.ts
```diff
@@ -33,14 +34,11 @@ export class RoleMappingRule {
...
  */
-export class FieldRule {
-  username?: Name
+export class FieldRule
+  implements AdditionalProperties<string, UserDefinedValue>
```
Anaethelion, line 38:
```suggestion
```

specification/security/_types/RoleMappingRule.ts
```diff
@@ -33,14 +34,11 @@ export class RoleMappingRule {
...
+{
+  username?: Names
```
Anaethelion, line 40: According to the [source](https://github.com/elastic/elasticsearch/blob/fe1083f6c5f29e1fc28b82c1b71be9b7c1ec7a73/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/support/UserRoleMapper.java#L55) this should be a string, I'd go for `Name` instead.

Anaethelion (changes requested):
Left a few comments, unsure about the `non_exhaustive` tag but this seems the way to go. I'll defer to @swallez for confirmation.

specification/security/_types/RoleMappingRule.ts
pquentin: Multiple usernames work and are documented, see `"username": [ "es-admin", "es-system" ]` in https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role-mapping.html

Anaethelion: My bad, the API does allow for a list!

Anaethelion (approved):
LGTM

specification/security/_types/RoleMappingRule.ts
```diff
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+import { AdditionalProperties } from '@spec_utils/behaviors'
```
Anaethelion, line 20:
```suggestion
```

We should remove this import if its not used anymore.

JoshMock (approved):
Thanks for fixing this @pquentin. :pray:

swallez (commented):
Looking at the ES code, a field rule is a single key/value pair where the value can be a single scalar value or an array of scalar values. So in theory that should just be `type FieldRule = SingleKeyDictionary<String, ScalarValue|ScalarValue[]>`.

However, the approach used in this PR consisting in just adding `@non_exhaustive` is the one that limits the breaking changes in strongly typed clients. It will be a bit less easy to use and may allow users to provide arbitrary values and not only scalar values, but that's the tradeoff to limit breaking changes.

We should however add a code comment (and not jsdoc comment that ends up in API docs) explaining this decision. Something like:

```
// This should have been defined as SingleKeyDictionary<String, ScalarValue|ScalarValue[]>
// However, this was initially defined as a container with a limited number of variants,
// and was later made non_exhaustive to limit breaking changes.
```

Finally, we should remove `realm.name` that was added in this PR. It's very specific and covered by the `@non_exhaustive`.

pquentin:
Thanks for the review! Please take another look.

swallez (approved):
LGTM

PR #2363, opened 2023-12-08
Add prefix_string config option for ML models

davidkyle:
The `prefix_strings` option was added to support the E5 model in https://github.com/elastic/elasticsearch/pull/102089


Anaethelion (approved):
LGTM

PR #2364, opened 2023-12-08
Fix definition of `ManagedBy`

flobernd:
Changes the definition of `ManagedBy` to an `enum` to make consumption more idiomatic.

@andreidan 
We noticed that the string values have different casing (e.g. lowercase "stream" "lifecycle", but uppercase "Lifecycle" "Management"). Are these the exact strings returned by the ES server?


andreidan (approved):
Thanks for fixing this Florian. 

These are the correct returned values.

PR #2367, opened 2023-12-12
Add _connector API

jedrazb:
👋 Hey, I'm not sure if it's a best practice to add the client types after the FF. If I mislabelled it please let me know. The connectors API actions are introduced in `8.12`.

### Changes 

- Copy endpoints from es [rest-api-spec](https://github.com/elastic/elasticsearch/tree/8.12/rest-api-spec/src/main/resources/rest-api-spec/api) 
- Copy and adapt types from [Kibana ts definition](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/connectors.ts)
- Added types for
  - connector.put
  - connector.post
  - connector.delete
  - connector.get
  - connector.list
  - connector.check_in
  - connector.update_error
  - connector.update_scheduling
  - connector.update_filtering
  - connector.update_configuration
  - connector.update_pipeline
  - connector.last_sync <- it should be `update_last_sync`, we should followup to fix this in 8.13 on ES side


output/schema/validation-errors.json
```diff
@@ -482,6 +482,12 @@
         "response definition cluster.put_component_template:Response / body / instance_of - Non-leaf type cannot be used here: '_types:AcknowledgedResponseBase'"
       ]
     },
+    "connector.delete": {
+      "request": [],
+      "response": [
+        "response definition connector.delete:Response / body / instance_of - Non-leaf type cannot be used here: '_types:AcknowledgedResponseBase'"
```
jedrazb, line 488: This issue is common for other endpoints that deal with `delete` action - should I change something here?

Anaethelion: No, you can leave it as is.

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,241 @@
...
+  basic?: FeatureEnabled
+}
+
+export interface ConnectorFeatures {
+  [FeatureName.DOCUMENT_LEVEL_SECURITY]?: FeatureEnabled
+  [FeatureName.FILTERING_ADVANCED_CONFIG]?: boolean
+  [FeatureName.FILTERING_RULES]?: boolean
+  [FeatureName.INCREMENTAL_SYNC]?: FeatureEnabled
+  [FeatureName.SYNC_RULES]?: SyncRulesFeature
+}
```
Anaethelion, line 204:
```suggestion
export interface ConnectorFeatures {
  document_level_security?: FeatureEnabled
  filtering_advanced_config?: boolean
  filtering_rules?: boolean
  incremental_sync?: FeatureEnabled
  sync_rules?: SyncRulesFeature
}
```

That mechanism isn't implemented in the compiler, this results in an object with the properties called `[FeatureName.DOCUMENT_LEVEL_SECURITY]` which isn't what you wanted I presume.

These properties should be name directly from their counterparts in `FeatureName` and the enum removed.

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,241 @@
...
+type FilteringPolicy = 'exclude' | 'include'
+
+type FilteringRuleRule =
+  | 'contains'
+  | 'ends_with'
+  | 'equals'
+  | '>'
+  | '<'
+  | 'regex'
+  | 'starts_with'
```
Anaethelion, line 134: `FilteringPolicy` and `FilteringRuleRule` should be converted to enums.

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+enum ConnectorFieldType {
+  STRING = 'str',
+  INTEGER = 'int',
+  LIST = 'list',
+  BOOLEAN = 'bool'
+}
+
+interface ConnectorConfigCategoryProperties {
```
Anaethelion, line 48: This seems redundant with `ConnectorConfigProperties` and I can't find this in the server sources. If its not used, this should be removed and the `ConnectorConfiguration` simplified.

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,241 @@
...
+  draft: FilteringRules
+}
+
+enum FeatureName {
+  FILTERING_ADVANCED_CONFIG = 'filtering_advanced_config',
+  FILTERING_RULES = 'filtering_rules',
+  DOCUMENT_LEVEL_SECURITY = 'document_level_security',
+  INCREMENTAL_SYNC = 'incremental_sync',
+  SYNC_RULES = 'sync_rules'
+}
```
Anaethelion, line 187:
```suggestion
```

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,241 @@
...
+
+export interface ConnectorFeatures {
+  [FeatureName.DOCUMENT_LEVEL_SECURITY]?: FeatureEnabled
+  [FeatureName.FILTERING_ADVANCED_CONFIG]?: boolean
+  [FeatureName.FILTERING_RULES]?: boolean
+  [FeatureName.INCREMENTAL_SYNC]?: FeatureEnabled
+  [FeatureName.SYNC_RULES]?: SyncRulesFeature
+}
+
+export interface SchedulingConfiguraton {
```
Anaethelion, line 206:
```suggestion
export interface SchedulingConfiguration {
```

Nit: typo, should also be updated wherever referenced.

Anaethelion (changes requested):
Left a few comments.

I've also taken the liberty to add the links to the documentation within the endpoints, feel free to edit/update if I made any mistake.

specification/connector/_types/Connector.ts
jedrazb: Agreed, we can get rid of it (also checked in Kibana source).

jedrazb:
Hey @Anaethelion thank you for review and adding the documentation links 👍 . I addressed your feedback and also updated the json spec files with correct doc links. 

There are couple of more endpoints to be added (related to `_connector/_sync _job` API). Given the release timeline is it still ok to keep the `backport 8.12` label?

jedrazb:
Hey @swallez 👋  since Anaethelion is OOO this week, I was suggested to add you as a reviewer to this PR. Could you take a look?

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,231 @@
...
+  value: string
+}
+
+interface Dependency {
+  field: string
+  value: string | number | boolean | null
+}
+
+enum DisplayType {
+  TEXTBOX = 'textbox',
```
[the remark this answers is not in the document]
jedrazb: Thank you for pointing this out 👍 I'm guilty of just copy-pasting most of the types from defs in Kibana

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,231 @@
...
+  ConnectorConfigProperties | null
+>
+
+export interface ConnectorScheduling {
+  enabled: boolean
+  interval: string // interval has crontab syntax
+}
+
+interface CustomScheduling {
+  configuration_overrides: Dictionary<string, UserDefinedValue>
```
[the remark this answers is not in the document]
jedrazb: Yeah, my bad, it was defined like that [in Kibana](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/connectors.ts#L80), I fill fix this here and file a follow-up ticket to handle it in Kibana as well.

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,231 @@
...
+  tooltip: string | null
+  type: ConnectorFieldType
+  ui_restrictions: string[]
+  validations: Validation[]
+  value: string | number | boolean | null
+}
+
+export type ConnectorConfiguration = Dictionary<
+  string,
+  ConnectorConfigProperties | null
```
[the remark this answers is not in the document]
jedrazb: Again, I'm guilty of copy-pasting from (probably) slightly outdated types from [Kibana](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/connectors.ts#L65). According to my understanding `| null` here is indeed not needed

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+export interface ConnectorConfigProperties {
+  category?: string
+  default_value: string | number | boolean | null
+  depends_on: Dependency[]
+  display: DisplayType
+  label: string
+  options: SelectOption[]
+  order?: number | null
```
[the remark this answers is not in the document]
jedrazb: I revised the types and I think for majority of cases we can safely mark fields as optional (instead of making them nullable). The only exception are errors that are (string | null), I can think of some situations where we want to mark error as null (when connector gets back to a healthy state)

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,231 @@
...
+
+enum ConnectorFieldType {
+  STRING = 'str',
+  INTEGER = 'int',
+  LIST = 'list',
+  BOOLEAN = 'bool'
+}
+
+interface Validation {
+  constraint: string | number
```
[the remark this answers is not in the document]
jedrazb: It's limited to strings and numbers. It's represented as an `Object` type in ES, though the [parser](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/configuration/ConfigurationValidation.java#L65) validates that it's either of numeric or string type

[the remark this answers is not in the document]
swallez: Oh, as I understand, `type` is an [enumeration of the possible constraints](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/configuration/ConfigurationValidationType.java) that precisely defines what should be in the `constraint` field.

So this can be represented in a fully-typed way as follows using an union defining "[internally tagged variants](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#internal)":

```ts
/** @variants internal tag='type' */
export type Validation = LessThanValidation | GreaterThanValidation | ListTypeValidation | IncludedInValidation | RegexValidation | UnsetValidation

export interface LessThanValidation {
  type: 'less_than'
  constraint: double // we want to avoid 'number'
}

export interface GreaterThanValidation {
  type : 'greater_than'
  constraint: double
}

export interface ListTypeValidation {
  type: 'list_type'
  constraint: string
}

export interface IncludedInValidation {
  type: 'included_in'
  constraint: string // is it really a string?
}

export interface RegexValidation {
  type: 'regex'
  constraint: string
}

export interface UnsetValidation {
  type: 'unset'
  // no constraint ?
}
```

specification/connector/_types/Connector.ts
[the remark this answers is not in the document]
swallez: Ok. So we have to implement https://github.com/elastic/elasticsearch-specification/issues/2049 that clearly distinguishes "null as optional" and "null as an actual value" in the spec. Adding that on my immediate todo list.

specification/connector/_types/Connector.ts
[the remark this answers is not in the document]
jedrazb: Ack, thank you for explanation! Will include this into the type definitions

specification/connector/_types/Connector.ts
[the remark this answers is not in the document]
jedrazb: Does it imply that this PR needs to wait until #2049 is implemented? We are not in a rush, just want to clarify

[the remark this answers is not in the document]
swallez: Yes, because currently `null` is considered as the equivalent of "missing" in languages where objects are represented as data structures and not dictionaries (i.e. all strictly typed languages). #2049 is meant to capture "null is an actual value" in the API spec so that we can generate code that correctly handles this.

jedrazb:
CI steps are breaking in the `setup job` phase with 
```
Error: Internal server error occurred while resolving "actions/checkout@v[2](https://github.com/elastic/elasticsearch-specification/actions/runs/7461636399/job/20302186145?pr=2367#step:1:2)"
```

specification/connector/list/ConnectorListRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+     */
+    from?: integer
+    /**
+     * Specifies a max number of results to get
+     */
+    size?: integer
+    /**
+     * A comma-separated list of connector index names to fetch connector documents for
+     */
+    index_name?: Fields
```
swallez, line 43: `Fields` is meant to hold a list of index field names. If these are ES indices, then `Indices` should be used.

specification/connector/list/ConnectorListRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * A comma-separated list of connector names to fetch connector documents for
+     */
+    connector_name?: Fields
```
swallez, line 47: Use `Names` instead

specification/connector/list/ConnectorListRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * A comma-separated list of connector service types to fetch connector documents for
+     */
+    service_type?: Fields
```
swallez, line 51: Are service type identifiers? Then use `Ids`, or otherwise `Names`.

swallez (changes requested):
Looking good! There are a few minor naming issues left, and we'll be good to go!

specification/connector/_types/Connector.ts
```diff
@@ -0,0 +1,267 @@
...
+  full?: ConnectorScheduling
+  incremental?: ConnectorScheduling
+}
+
+export interface Connector {
+  api_key_id?: string
+  configuration: ConnectorConfiguration
+  custom_scheduling: ConnectorCustomScheduling
+  description?: string
+  error: WithNullValue<string>
```
l-trotta, line 243: error should be nullable, [server code](https://github.com/elastic/elasticsearch/blob/92f7e078071fba021f9784c98a7cea6e76c5710d/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/Connector.java#L82)

pquentin:
"Validate APIs / build" was failing because of me, I fixed it and retried the build.

specification/connector/list/ConnectorListRequest.ts
jedrazb: They are arbitrary strings, so it is `Names` for our case

specification/connector/_types/Connector.ts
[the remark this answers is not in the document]
jedrazb: Yes, I do agree with you here. Since the `Connector` denotes a response type. In a response, a missing (optional) field does indicate that the field is in fact null. 

I will adapt this, but I the meaningful is needed for requests definitions to set certain fields to null.

swallez (approved):
LGTM! 🎉

Anaethelion (approved):

specification/connector/update_index_name/ConnectorUpdateIndexNameRequest.ts
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * The unique identifier of the connector to be updated
+     */
+    connector_id: Id
+  }
+  /**
+   * The connector index name
+   */
+  body: {
+    index_name: WithNullValue<string>
```
flobernd, line 41: @swallez @Anaethelion I'm late to the party, but shouldn't we use the `IndexName` alias instead of `string` at these places?

Like:

```ts
index_name: WithNullValue<IndexName>
```

PR #2369, opened 2023-12-12
added note for known typo

l-trotta:
The `ManagedBy` enum has a field, `Data stream lifecycle`, which should have been written like `Data Stream Lifecycle`.
This is not correct, but it's already part of the API and there are special cases for it in tests, so this PR adds a comment to notify future contributors of this anomaly.


specification/indices/_types/DataStream.ts
```diff
@@ -32,7 +32,9 @@ import { DataStreamLifecycleWithRollover } from '@indices/_types/DataStreamLifec
 enum ManagedBy {
   /** @codegen_name ilm */
   'Index Lifecycle Management',
-  /** @codegen_name datastream */
+  /** This should have been written with capital letters, it's a known typo and should not be corrected
+   * since the test cases have already been written
+   * @codegen_name datastream */
```
swallez, line 37: Nitpicking: can you move `*/` to the next line?

specification/indices/_types/DataStream.ts
```diff
@@ -32,7 +32,11 @@ import { DataStreamLifecycleWithRollover } from '@indices/_types/DataStreamLifec
 enum ManagedBy {
   /** @codegen_name ilm */
   'Index Lifecycle Management',
-  /** @codegen_name datastream */
+  /**
```
swallez, line 35: Thinking further, this note should be a regular code comment aimed at API spec maintainers and not a JSDoc comment aimed at users (JSDoc comments end up in client libraries (e.g. Javadoc) and API documentation).

Something like:
```ts
/** codegen_name datastream */
// This should have been...
```

swallez (approved):
LGTM. Thanks!

PR #2371, opened 2023-12-22
Add `active_only` flag definition to Get API key spec

n1v0lg:


specification/security/get_api_key/SecurityGetApiKeyRequest.ts
```diff
@@ -68,5 +68,12 @@ export interface Request extends RequestBase {
      * @availability serverless
      */
     with_limited_by?: boolean
+    /**
+     * A boolean flag that can be used to query API keys that are currently active. An API key is considered active if it is neither invalidated, nor expired at query time. You can specify this together with other parameters such as owner or name. If `active_only` is false, the response will include both active and inactive (expired or invalidated) keys.
+     * @server_default false
+     * @availability stack since=8.12.1
```
n1v0lg, line 74: Technically it's available since 8.10 but based on the Slack convo we had, we don't backport so I'm assuming I should use the latest version we backported the fix to.

Anaethelion: Availability is mainly for documentation purpose, it should reflect the actual availability of the arg.
```suggestion
     * @availability stack since=8.10.0
```

Anaethelion (approved):
LGTM

PR #2376, opened 2024-01-03
Remove unknown setting index.shard

l-trotta:
The Java client has a `.shard()` method in the Index API  that returns `illegal_argument_exception` when called, with reason `unknown setting [index.shards]`.

For example:

```
esClient.indices().create(c -> c
        .index("test")
       .settings(st -> st
       .shards(1)
```

Trying the same call in Kibana yields the same result:

```
PUT test/
{
  "settings": {
      "shards": 1
  }
}
```

Given that `number_of_shards` exists, I suspect `shards` shouldn't be part of the IndexSettings type.


swallez (approved):
LGTM!

PR #2379, opened 2024-01-05
Cleanup enum values, allow using identifier+value as an alternative to @codegen_name

swallez:
This PR allows using the standard TypeScript enum syntax to define custom identifiers for enum members, in addition to the existing `@codegen_name` jsdoc tag.

Also cleans up the numeric values that existed on some enums for historical reasons (they were required in the early days of the API spec compiler) and refine the docs related to defining enums in the API spec.

Concretely, both definitions below are now equivalent (previously the `z-score` value was discarded, leading to a wrong declaration):

```ts
enum MyEnum {
    percent_of_sum,
    mean,
    softmax,
    /** @codegen_name z_score */
    'z-score'
}
```
 and

```ts
enum MyEnum {
    percent_of_sum,
    mean,
    softmax,
    z_score = 'z-score'
}
```

**Validation**: compiling the API spec after these changes only produces `specLocation` changes in `output/schema.json` becaused of the removed jsdoc tags.


ezimuel (approved):
LGTM

flobernd (approved):

davidkyle (approved):
ML changes LGTM

Thanks for updating

PR #2381, opened 2024-01-11
Update spec for `update_api_key` to include `expiration`

jfreden:


Anaethelion (approved):
LGTM

Endpoint validation ran locally is OK, Github automation cannot access the secrets from a fork.

PR #2382, opened 2024-01-11
Cleanup/improve compiler rs

swallez:
This PR improves the code in `compiler-rs`:
* the first commit introduces `derive_more` to automatically derive `From` for enums member values
* 2nd commit significantly reduces (by 1MB) the size of the `.wasm` file by not using `regex`
* subsequent commits are mechanical with rustfmt and clippy


compiler-rs/compiler-wasm-lib/src/lib.rs
```diff
@@ -16,12 +16,11 @@
...
 use anyhow::bail;
-use wasm_bindgen::prelude::*;
 use clients_schema::{Availabilities, Visibility};
+use wasm_bindgen::prelude::*;
 
-
-#[cfg(not(target_arch="wasm32"))]
-compile_error!("To build this crate use `make compiler-wasm-lib`");
+// #[cfg(not(target_arch = "wasm32"))]
+// compile_error!("To build this crate use `make compiler-wasm-lib`");
```
Anaethelion, line 23:
```suggestion
#[cfg(all(not(feature = "cargo-clippy"), not(target_arch = "wasm32")))]
compile_error!("To build this crate use `make compiler-wasm-lib`");
```

This should make that compatible with clippy while retaining the warning against direct compilation.

Anaethelion (approved):
LGTM

PR #2394, opened 2024-01-17
Specification of test_grok_pattern endpoint

jan-elastic:
For more details, see: https://github.com/elastic/ml-team/issues/1108 (item 1)

### DIFF
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1460,6 +1460,12 @@
       ],
       "response": []
     },
+    "text_structure.test_grok_pattern": {
+      "request": [
+        "Request: should not have a body"
+      ],
+      "response": []
+    },
     "transform.delete_transform": {
       "request": [],
       "response": [
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1460,12 +1460,6 @@
       ],
       "response": []
     },
-    "text_structure.test_grok_pattern": {
-      "request": [
-        "Request: should not have a body"
-      ],
-      "response": []
-    },
     "transform.delete_transform": {
       "request": [],
       "response": [
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1454,18 +1454,6 @@
       ],
       "response": []
     },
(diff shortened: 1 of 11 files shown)


output/schema/validation-errors.json
```diff
@@ -1460,6 +1460,12 @@
       ],
       "response": []
     },
+    "text_structure.test_grok_pattern": {
+      "request": [
+        "Request: missing json spec query parameter 'ecs_compatibility'"
```
droberts195, line 1465: Please can you add this to `specification/text_structure/test_grok_pattern/TestGrokPatternRequest.ts`. Then this new section should disappear from the validation errors.

output/schema/schema.json
```diff
@@ -18307,6 +18307,49 @@
...
     },
+    {
+      "availability": {
+        "serverless": {
+          "stability": "experimental",
+          "visibility": "private"
+        },
+        "stack": {
+          "since": "8.13.0",
+          "stability": "experimental",
```
droberts195, line 18318: I think this means you need to re-run `make generate` and commit the result.

specification/text_structure/test_grok_pattern/TestGrokPatternRequest.ts
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { RequestBase } from '@_types/Base'
+
+/**
+ * @rest_spec_name text_structure.test_grok_pattern
+ * @availability stack since=8.13.0 stability=experimental visibility=public
+ * @availability serverless stability=experimental visibility=private
```
droberts195, line 25: This should match the visibility of the find structure endpoint, as it will be used by the same UI functionality:

```suggestion
 * @availability stack since=8.13.0 stability=stable
 * @availability serverless stability=stable visibility=private
```

jan-elastic: fixed

droberts195 (approved):
LGTM, but don't merge it until tomorrow because the server-side changes have only just gone in

PR #2400, opened 2024-01-29
Fix script definition in processorcontainer

l-trotta:
Fixes [Issue 500](https://github.com/elastic/elasticsearch-java/issues/500)
ProcessorContainer class was referring to Script instead of ScriptProcessor


swallez (approved):
LGTM, good catch!

PR #2401, opened 2024-01-29
Make `PathHierarchyTokenizer` properties optional

stevejgordon:
When attempting to configure the path hierarchy tokenizer from a .NET app, I noticed that many of the fields are serialized with their defaults when unset, and this is incorrect. They are all optional.


flobernd (approved):
LGTM!

PR #2404, opened 2024-01-30
[7.17] Fix requests that were used as field types

swallez:
Requests should not be used as field types, as the generated structures contain much more than their body's properties.

This was already fixed in 8.x in #1564 and #1671 and but not backported to 7.17.


pquentin (approved):
Thanks! LGTM.

pquentin:
While it's OK to add (breaking) specification bugfixes to the Elasticsearch specification in 7.17.x, the concerned APIs can still be used here without this change, so we decided to close this.

PR #2407, opened 2024-02-08
Fix synonyms/put_synonym_rule body

pquentin:
synonyms is not a list, but a string in Solr format.


pquentin:
> The backport to `8.11` failed:

Opened https://github.com/elastic/elasticsearch-specification/pull/2409, the conflict was `output/openapi/elasticsearch-serverless-openapi.json` which isn't in 8.11.

PR #2398, opened 2024-01-22
Adding signed byte support from cohere integration

jonathan-buttner:
This PR is to reflect the proposed changes included in this PR: https://github.com/elastic/elasticsearch/pull/104559 which adds support for the Cohere 3rd party service. By adding Cohere we also would like to support signed bytes as a response format for dense vector.

I'll leave this in draft until the elasticsearch PR is merged.

PR #2410, opened 2024-02-08
[ML] Adding text embedding byte representation

jonathan-buttner:
This PR adds the text embedding byte representation. This is to resolve the problems with the original PR: https://github.com/elastic/elasticsearch-specification/pull/2398 which caused the client generates to fail because it was unable to deduce the response format.

These changes coincide with the server changes here: https://github.com/elastic/elasticsearch/pull/105290


jonathan-buttner:
Closing to open against repository directly.

PR #2412, opened 2024-02-08
Batch from java issues

l-trotta:
Fixes provided in this PR:

- [727](https://github.com/elastic/elasticsearch-java/issues/727) added `remove_binary` to AttachmentProcessor
- [242](https://github.com/elastic/elasticsearch-java/issues/242) made NodeInfoPath and `seed_hosts` optional to make the method `.nodes().info()` work properly, also made `initial_master_nodes` a string array, since the documentation allows for both a single string and an array, also added additional properties for NodeInfoDiscover
- [315](https://github.com/elastic/elasticsearch-java/issues/315) fixed description for update by query
- [299](https://github.com/elastic/elasticsearch-java/issues/299) allowed custom options for IndexSettings Similarity (via custom plugin)

### DIFF
--- a/specification/nodes/info/types.ts
+++ b/specification/nodes/info/types.ts
@@ -68,7 +68,7 @@ export class NodeInfo {
 export class NodeInfoSettings {
   cluster: NodeInfoSettingsCluster
   node: NodeInfoSettingsNode
-  path: NodeInfoPath
+  path?: NodeInfoPath
   repositories?: NodeInfoRepositories
   discovery?: NodeInfoDiscover
   action?: NodeInfoAction
@@ -155,9 +155,9 @@ export class NodeInfoSettingsNode {
 }
 
 export class NodeInfoPath {
-  logs: string
-  home: string
-  repo: string[]
+  logs?: string
+  home?: string
+  repo?: string[]
   data?: string[]
 }
 
--- a/specification/nodes/info/types.ts
+++ b/specification/nodes/info/types.ts
@@ -132,7 +132,7 @@ export class NodeInfoSettingsCluster {
   name: Name
   routing?: IndexRouting
   election: NodeInfoSettingsClusterElection
-  initial_master_nodes?: string
+  initial_master_nodes?: string[]
   /**
    * @availability stack since=7.16.0
    * @availability serverless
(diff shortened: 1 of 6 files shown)


specification/nodes/info/types.ts
```diff
@@ -170,7 +170,7 @@ export class NodeInfoRepositoriesUrl {
 }
 
 export class NodeInfoDiscover {
-  seed_hosts: string
+  seed_hosts?: string
```
swallez, line 173: Looking at [the ES source code](https://github.com/elastic/elasticsearch/blob/af5c11702bb1450e5b78684e9852dd746a9c3140/server/src/main/java/org/elasticsearch/discovery/SettingsBasedSeedHostsProvider.java#L33), `seed_hosts` should be an array, but I don't see it being optional as it's the single property of a parent object is itself optional.

swallez (changes requested):

specification/nodes/info/types.ts
l-trotta: with seed_hosts as not optional I get this error when calling nodes().info() locally

swallez: Interesting. With a single-node Cloud cluster I get

```json
"discovery": {
  "seed_hosts": [],
  "seed_providers": "file"
},
```

With a single-node local cluster started with the `"discovery.type": "single-node"` setting I get:
```json
"discovery": {
  "type": "single-node"
},
```

So for the sake of completeness and correctness, let's declare this type as:

```ts
export class NodeInfoDiscover implements AdditionalProperties<String, UserDefinedValue> {
  seed_hosts?: string[]
  type?: string
  seed_providers?: string[]
}
```

Rationale:
* we have a single value in our examples for `seed_providers` but [it's technically a list](https://github.com/elastic/elasticsearch/blob/cd5a19ac3c96ec779e998a584ae17522369bc343/server/src/main/java/org/elasticsearch/discovery/DiscoveryModule.java#L82)
* there are many more additional properties, including child objects for the configuration of discovery providers, which are pluggable (search for `"discovery.` in the ES code base to find them)
* there are other top-level properties, but I don't think we should formally add them to the spec are they are rather advanced settings, and they won't be present in responses unless they've been set explicitly.

swallez (approved):
LGTM!

PR #2413, opened 2024-02-09
Add optional query param in _field_caps to return only fields with value in index

piergm:
Updates the spec for the request object for all `_field_caps` endpoints.
In `8.13` we are adding a new optional query parameter `include_empty_fields` that defaults to `true`. If `false` empty fields are not included in the response. This change will not alter the `_field_caps` response schema.
https://github.com/elastic/elasticsearch/pull/103651

### DIFF
--- a/specification/_global/field_caps/FieldCapabilitiesRequest.ts
+++ b/specification/_global/field_caps/FieldCapabilitiesRequest.ts
@@ -78,6 +78,7 @@ export interface Request extends RequestBase {
     /**
      * If false, empty fields are not included in the response.
      * @availability stack since=8.13.0
+     * @availability serverless
      * @server_default true
      */
     include_empty_fields?: boolean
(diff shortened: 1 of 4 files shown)


specification/_global/field_caps/FieldCapabilitiesRequest.ts
```diff
@@ -75,6 +75,12 @@ export interface Request extends RequestBase {
      * @availability serverless
      */
     types?: string[]
+    /**
+     * If false, empty fields are not included in the response.
+     * @availability stack since=8.13.0
```
Anaethelion, line 80: I haven't seen anything that restrict this to `stack` only.
Do we miss a `serverless` availability as well ?

Anaethelion (commented):
Left a question, also could you sync the spec file located in:
https://github.com/elastic/elasticsearch-specification/blob/main/specification/_json_spec/field_caps.json

specification/_global/field_caps/FieldCapabilitiesRequest.ts
piergm: You are right, adding it! Thanks

Anaethelion (approved):
LGTM!

PR #2416, opened 2024-02-19
Remove `IcuTokenizer` from `TokenFilterDefinition` variants union

flobernd:
`IcuTokenizer` was accidentally added to the `TokenFilterDefinition` variants union a while ago. This PR removes the variant.

This as well removes a special case where `IcuTokenizer` is a variant of both, `Tokenizer` and `TokenFilterDefinition` and might allow for simplification of the code generators. cc @elastic/clients-team


swallez (approved):
LGTM, good catch!

PR #2427, opened 2024-02-20
Add _resolve/cluster API

quux00:
A new endpoint `_resolve/cluster` was added to Elasticsearch 8.13.0.
End user docs can be found here: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-resolve-cluster-api.html

This endpoint is not supported in serverless, so the spec only marks it as available in "stack".


quux00:
I don't understand the validation failure error message [here](https://github.com/elastic/elasticsearch-specification/actions/runs/7979858766/job/21788248916?pr=2427).

I am unable to run `make validate` locally. First, I'm not sure what command to run. When I try 
```
make validate api=index type=request stack-version=8.13.0-SNAPSHOT
```

It says "Pulling latest flight recorder changes" and spins, never doing anything. I clone the flight recorder repo moments ago.

specification/indices/resolve_cluster/ResolveClusterResponse.ts
```diff
@@ -0,0 +1,30 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { ElasticsearchVersionMinInfo } from '@_types/Base'
+
+export class Response {
+  body: {
```
Anaethelion, line 23: According to the documentation's examples, this should be a a dictionnary with this layout as value.

Something like:
```suggestion
  body: Dictionary<string, ClusterInfo>
```
Even better if we can name the string to give meaning.

quux00: Good catch. I had the response all wrong. Sorry for that. I've fixed it in the latest push.

quux00:
> Following you can find the validation results for the APIs you have changed.
> 
> API	Status	Request	Response
> `indices.resolve_cluster`	⚪	Missing test	Missing test
> You can validate these APIs yourself by using the `make validate` target.

Is there something else I need to do here (like add this new api to some metadata or config file)?

I tried running this locally but it fails:

```
└─ $ ▶ make validate api=indices.resolve_cluster type=request stack-version=8.13.0-SNAPSHOT
✖
    at run (/home/mpeterson/elasticsearch-specification/compiler/run-validations.js:175:12)
    exit code: 1
make: *** [Makefile:4: validate] Error 1
```

Anaethelion:
> > Following you can find the validation results for the APIs you have changed.
> > API	Status	Request	Response
> > `indices.resolve_cluster`	⚪	Missing test	Missing test
> > You can validate these APIs yourself by using the `make validate` target.
> 
> Is there something else I need to do here (like add this new api to some metadata or config file)?

Just noticed the flight recorder that produces the json for validation didn't have an `8.13` branch run yet. I've triggered that and will relaunch the validation once its run has completed.

Anaethelion (approved):
LGTM

Issue #2434, opened 2024-03-05
`Pipeline._meta` should be optional


pgayvallet:
## 🐛 Wrong type

`Pipeline._meta` is defined as mandatory, when it should be optional:

https://github.com/elastic/elasticsearch-specification/blob/main/specification/ingest/_types/Pipeline.ts#L23-L45

The field was introduced by https://github.com/elastic/elasticsearch-specification/pull/2348, and given the PR's description, I assume it was a mistake to have it mandatory in the first place.

Related: https://github.com/elastic/kibana/pull/175954

## Definition

```diff
export class Pipeline {
       // [other properties]

-     _meta: Metadata;
+     _meta?: Metadata;
}
```

PR #2348, opened 2023-11-14
Add optional _meta object to ingest pipeline response

JoshMock:
Support for `_meta` is mentioned in [the `PUT /_ingest/pipeline` docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-pipeline-api.html), and says it will be returned in `GET`:

> To check the `_meta`, use the [get pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/current/get-pipeline-api.html) API.

See https://github.com/elastic/elasticsearch-js/issues/1724

PR #2436, opened 2024-03-05
Define `Pipeline._meta` as optional

pgayvallet:
Fix https://github.com/elastic/elasticsearch-specification/issues/2434
Follow-up of PR https://github.com/elastic/elasticsearch-specification/pull/2348

Define the `_meta` field of the `Pipeline` type as optional instead of required, to reflect the actual shape of the requests/responses using the type


pquentin (approved):
Thanks! LGTM. Looking at the title of the pull request that introduced it (https://github.com/elastic/elasticsearch-specification/pull/2348), this was always meant to be optional

PR #2454, opened 2024-03-14
Rewrite similarity definition

swallez:
Rewrites the index similarity definition: this is a dictionary and not a container, as users can define several similarities. The name is also user-defined, indicating a dictionary.

Also adds the boolean similarity that was missing and makes some of the configuration properties optional after inspection of [the ES code base](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/index/similarity/SimilarityService.java).

### DIFF
--- a/specification/indices/_types/IndexSettings.ts
+++ b/specification/indices/_types/IndexSettings.ts
@@ -30,7 +30,7 @@ import {
   Uuid,
   VersionString
 } from '@_types/common'
-import { double, integer, long } from '@_types/Numeric'
+import { double, float, integer, long } from '@_types/Numeric'
 import { DateTime, Duration, EpochTime, UnitMillis } from '@_types/Time'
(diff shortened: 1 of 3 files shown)


specification/indices/_types/IndexSettings.ts
```diff
@@ -30,7 +30,7 @@ import {
   Uuid,
   VersionString
 } from '@_types/common'
-import { double, integer, long } from '@_types/Numeric'
+import { double, float, integer, long } from '@_types/Numeric'
```
l-trotta, line 33: is float needed?

l-trotta (approved):

specification/indices/_types/IndexSettings.ts
swallez: Good catch. Fixed!

PR #2455, opened 2024-03-14
Support for aggregations in Security's Query API Key Information API

albertzaharovits:
In 8.13 ES's [Query API Key Information API](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/security-api-query-api-key.html#_response_2) added support for [aggregations](https://github.com/elastic/elasticsearch/issues/101691).
This should reflect that change in the request/response of the specification.

### DIFF
--- a/specification/security/query_api_keys/QueryApiKeysRequest.ts
+++ b/specification/security/query_api_keys/QueryApiKeysRequest.ts
@@ -51,7 +51,7 @@ export interface Request extends RequestBase {
      * Additionally, aggregations only run over the same subset of fields that query works with.
      * @aliases aggs */
     aggregations?: Dictionary<string, APIKeyAggregationContainer>
-		/**
+    /**
      * A query to filter which API keys to return.
      * If the query parameter is missing, it is equivalent to a `match_all` query.
      * The query supports a subset of query types, including `match_all`, `bool`, `term`, `terms`, `match`,
--- a/specification/security/query_api_keys/QueryApiKeysRequest.ts
+++ b/specification/security/query_api_keys/QueryApiKeysRequest.ts
@@ -18,10 +18,11 @@
  */
 
 import { Dictionary } from '@spec_utils/Dictionary'
-import { AggregationContainer } from '@_types/aggregations/AggregationContainer'
+import { APIKeyAggregationContainer } from 'APIKeyAggregationContainer'
 import { RequestBase } from '@_types/Base'
 import { integer } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
+import { Sort, SortResults } from '@_types/sort'
 import { Sort, SortResults } from '@_types/sort'
 
 /**
@@ -50,7 +51,7 @@ export interface Request extends RequestBase {
      * `cardinality`, `value_count`, `composite`, `filter`, and `filters`.
      * Additionally, aggregations only run over the same subset of fields that query works with.
      * @aliases aggs */
-    aggregations?: Dictionary<string, AggregationContainer>
+    aggregations?: Dictionary<string, APIKeyAggregationContainer>
 		/**
      * A query to filter which API keys to return.
      * If the query parameter is missing, it is equivalent to a `match_all` query.
--- a/specification/security/query_api_keys/QueryApiKeysRequest.ts
+++ b/specification/security/query_api_keys/QueryApiKeysRequest.ts
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+import { Dictionary } from '@spec_utils/Dictionary'
+import { AggregationContainer } from '@_types/aggregations/AggregationContainer'
 import { RequestBase } from '@_types/Base'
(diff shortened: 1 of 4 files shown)


specification/security/query_api_keys/QueryApiKeysRequest.ts
```diff
@@ -42,9 +44,20 @@ export interface Request extends RequestBase {
   }
   body: {
     /**
+     * Any aggregations to run over the corpus of returned API keys.
+     * Aggregations and queries work together. Aggregations are computed only on the API keys that match the query.
+     * This supports only a subset of aggregation types, namely: `terms`, `range`, `date_range`, `missing`,
+     * `cardinality`, `value_count`, `composite`, `filter`, and `filters`.
+     * Additionally, aggregations only run over the same subset of fields that query works with.
+     * @aliases aggs */
+    aggregations?: Dictionary<string, AggregationContainer>
```
Anaethelion, line 53: If this only support a subset of the aggregation container we need to build a new container, otherwise its going to confuse users when they try to use those.

specification/security/query_api_keys/QueryApiKeysRequest.ts
```diff
@@ -42,9 +44,20 @@ export interface Request extends RequestBase {
...
      * A query to filter which API keys to return.
-     * The query supports a subset of query types, including `match_all`, `bool`, `term`, `terms`, `ids`, `prefix`, `wildcard`, and `range`.
-     * You can query all public information associated with an API key.
+     * If the query parameter is missing, it is equivalent to a `match_all` query.
+     * The query supports a subset of query types, including `match_all`, `bool`, `term`, `terms`, `match`,
+     * `ids`, `prefix`, `wildcard`, `exists`, `range`, and `simple_query_string`.
+     * You can query the following public information associated with an API key: `id`, `type`, `name`,
+     * `creation`, `expiration`, `invalidated`, `invalidation`, `username`, `realm`, and `metadata`.
      */
     query?: QueryContainer
```
Anaethelion, line 62: Same as aggregations, we should have a specific container for this.

Anaethelion (changes requested):
Since containers are property based unions I'd recommend we create the relevant versions with only the supported subset to avoid misleading users.

specification/security/query_api_keys/QueryApiKeysRequest.ts
albertzaharovits: Hi @Anaethelion ,
I addressed your feedback here, can you please checkup the `APIKeyAggregationContainer` and `APIKeyAggregator`.

specification/security/query_api_keys/QueryApiKeysRequest.ts
albertzaharovits: I agree, but I propose we have a separate PR for the query types restrictions. I feel this is already complex as it is. WDYT?

Anaethelion: A following PR to address the query seems like a good idea since this one was focused on aggregation.

specification/security/query_api_keys/QueryApiKeysRequest.ts
```diff
@@ -17,10 +17,13 @@
  * under the License.
  */
 
+import { Dictionary } from '@spec_utils/Dictionary'
+import { APIKeyAggregationContainer } from 'APIKeyAggregationContainer'
 import { RequestBase } from '@_types/Base'
 import { integer } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { Sort, SortResults } from '@_types/sort'
+import { Sort, SortResults } from '@_types/sort'
```
Anaethelion, line 26:
```suggestion
```

Double import should be removed

specification/security/query_api_keys/QueryApiKeysRequest.ts
```diff
@@ -17,10 +17,13 @@
  * under the License.
  */
 
+import { Dictionary } from '@spec_utils/Dictionary'
+import { APIKeyAggregationContainer } from 'APIKeyAggregationContainer'
```
Anaethelion, line 21:
```suggestion
import { APIKeyAggregationContainer } from './types'
```

This file should be called `types`.

specification/security/query_api_keys/QueryApiKeysResponse.ts
```diff
@@ -34,5 +51,29 @@ export class Response {
...
+     */
+    aggregations?: Dictionary<AggregateName, APIKeyAggregate>
   }
 }
+
+/**
+ * @variants external
+ * @non_exhaustive
+ */
+export type APIKeyAggregate =
```
Anaethelion, line 65: Can you move that over in the `types.ts` file ?

Anaethelion (changes requested):
Looks good, a few changes are still needed.

Anaethelion (approved):

PR #2457, opened 2024-03-14
Add missing `codegen_name` annotations to all value bodied responses

flobernd:
In the .NET client, these value-type bodies will be wrapped in a common response type like e.g.:

```csharp
public partial class EsqlQueryResponse : 
    ElasticsearchResponse
{
    public byte[] Data { init; }
}
````

In order to properly auto-generate the required body property (e.g. `Data` in this case) with a meaningful name, all of these bodies must be annotated with `codegen_name` attributes.

It would be nice having a compiler check for this rule in the future.

### DIFF
--- a/specification/_global/get_source/SourceResponse.ts
+++ b/specification/_global/get_source/SourceResponse.ts
@@ -18,5 +18,6 @@
  */
 
 export class Response<TDocument> {
+  /** @codegen_name source */
   body: TDocument
 }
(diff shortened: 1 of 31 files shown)


swallez (approved):
LGTM, but I'd like @Anaethelion to also have a look from the Go client perspective.

Anaethelion (approved):
LGTM, no impact on my part.

jonathan-buttner (approved):

PR #2460, opened 2024-03-14
Restrict query type when querying API keys

albertzaharovits:
Only a subset of query types are supported when [querying API keys](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/security-api-query-api-key.html).

### DIFF
--- a/specification/security/query_api_keys/types.ts
+++ b/specification/security/query_api_keys/types.ts
@@ -145,6 +145,10 @@ export type APIKeyAggregate =
   | DateRangeAggregate
   | CompositeAggregate
 
+/**
+ * @variants container
+ * @non_exhaustive
+ */
 export class APIKeyQueryContainer {
   /**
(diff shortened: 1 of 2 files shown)


specification/security/query_api_keys/types.ts
```diff
@@ -125,3 +144,88 @@ export type APIKeyAggregate =
   | RangeAggregate
   | DateRangeAggregate
   | CompositeAggregate
+
+export class APIKeyQueryContainer {
```
Anaethelion, line 148:
```suggestion
/**
 * @variants container
 * @non_exhaustive
 */
export class APIKeyQueryContainer {
```

This needs to be decorated properly to be a container. :christmas_tree:

Anaethelion (changes requested):
I've only left a suggestion for the missing annotations.

Anaethelion (approved):
LGTM

PR #2462, opened 2024-03-21
Expand attached behaviors for request in no generics transform

Anaethelion:
This PR adds a new function to embed the specific behaviors properties `CommonQueryParameters` and `CommonCatQueryParameters` in the query parameters list whenever they are present in a request as `attachedBehaviors`.

This adds the following query parameters to the OpenAPI export:
* Adds `error_trace`, `filter_path`, `human`, `pretty` from `CommonQueryParameters`
* Adds `format`, `h`, `help`, `local`, `master_timeout`, `s` and `v` from `CommonCatQueryParameters`


swallez (changes requested):
Why is this in `expand_generics.rs`? This is needed for the Schema.json to OpenAPI conversion, so should be better located in the `clients_schema_to_openapi` crate. The best place for this is probably `schemas.rs` in the ["Prepare query parameters"](https://github.com/elastic/elasticsearch-specification/blob/cf1b47abaade847e047cee7a6bb81d566e24aa07/compiler-rs/clients_schema_to_openapi/src/paths.rs#L97) section.

Also we should match on `behaviors` and not `attached_behaviors` since the latter is set on the type that has the behavior, but also all of its subtypes. Using it duplicates the common parameters on every subtype.

(IMHO `attached_behaviors` should be removed but this is another issue)

Anaethelion:
After an internal discussion, we've decided it is best to refactor the `*Common*QueryParameters` pattern instead.

Closing this draft.

PR #2463, opened 2024-03-22
Compiler: fix required body value

Anaethelion:
Propagate question mark from typescript spec
Overrides value taken from json spec file
Add an endpoint validation to detect discrenpancy


flobernd (approved):
LGTM! Thank you.

flobernd:
@Anaethelion I merged the predecessor PR. I think this should be good to go now 🙂

PR #2464, opened 2024-03-22
Fix body nullability

flobernd:
Removes incorrect nullability marker from certain bodies.

Related to #2463


davidkyle (approved):
ML change LGTM thanks @flobernd

PR #2471, opened 2024-03-28
Add Completion result and task type to the Inference API spec

timgrein:
This PR adds the new completion task type and the result to the specification.

Completion result example:

```
{ 
  "completion": [ 
     { 
       "result": "..." 
      } 
   ] 
}


davidkyle (approved):
LGTM

PR #2472, opened 2024-03-28
Add force_synthethic_source parameter

pquentin:
I noticed that `force_synthetic_source` was missing from this report: https://github.com/elastic/elasticsearch-specification/blob/071b2f79d37b92150d8ce42e3bd77951aa917f4f/output/schema/validation-errors.json#L655. Do we typically add fields behind feature flags?

I checked the following things:

 * this parameter is not accepted as body parameter, either in get, mget or search
 * this parameter is not accepted in msearch at all

### DIFF
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -650,12 +650,6 @@
       ],
       "response": []
     },
-    "get": {
-      "request": [
-        "Request: missing json spec query parameter 'force_synthetic_source'"
-      ],
-      "response": []
-    },
     "get_source": {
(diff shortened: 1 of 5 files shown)


pquentin:
It is an improvement! Here is the current state:

| API | Status | Request | Response |
| --- | --- | --- | --- |
| `get` | :red_circle: | 166/182 | 179/179 |
| `mget` | :red_circle: | 50/52 | 51/51 |
| `search` | :red_circle: | 1821/1925 | 1860/1907  |

PR #2476, opened 2024-04-02
Adding the API design guidelines.

technige:
This PR introduces the API Design Guidelines into this repository, giving them greater visibility. The content of these has not been changed, beyond conversion from `.mdx` to `.md`, which allows them to be viewed inline within GitHub (which is useful given that GitHub is a major part of the workflow for users of these).

The top-level README has also been updated to include reference to this new subdirectory, and a table of repo structure has been included.

### DIFF
--- a/api-design-guidelines/data-modelling.md
+++ b/api-design-guidelines/data-modelling.md
@@ -1,12 +1,4 @@
----
-id: esTmApiGuidelinesDataModelling
-slug: /elasticsearch-team/api-guidelines/data-modelling
-title: Data modelling
-image: https://source.unsplash.com/400x175/?Nature
-description: API guidelines - Data modelling
-date: 2022-06-26
-tags: ['elasticsearch','internal docs']
----
+# API guidelines - Data modelling
 
 Request and response bodies are typically encoded in JSON. The full JSON specification is defined in [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259).
 
--- a/api-design-guidelines/data-modelling.md
+++ b/api-design-guidelines/data-modelling.md
@@ -1,0 +1,275 @@
+---
+id: esTmApiGuidelinesDataModelling
+slug: /elasticsearch-team/api-guidelines/data-modelling
+title: Data modelling
+image: https://source.unsplash.com/400x175/?Nature
+description: API guidelines - Data modelling
+date: 2022-06-26
+tags: ['elasticsearch','internal docs']
+---
+
+Request and response bodies are typically encoded in JSON. The full JSON specification is defined in [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259).
+
+## JSON objects are not ordered
+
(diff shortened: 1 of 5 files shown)


api-design-guidelines/data-modelling.md
```diff
@@ -0,0 +1,275 @@
+---
```
flobernd, line 1: This kind of metadata is only partially supported by GH (it is displayed as a regular table). We might want to remove it or at least get rid of irrelevant fields like "id", "image", etc.

api-design-guidelines/naming.md
```diff
@@ -0,0 +1,170 @@
+---
```
flobernd, line 1: IDEM

api-design-guidelines/data-modelling.md
```diff
@@ -0,0 +1,275 @@
...
+
+Request and response bodies are typically encoded in JSON. The full JSON specification is defined in [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259).
+
+## JSON objects are not ordered
+
+JSON objects must be conceptually treated as **unordered**. Maintaining order of deserialized JSON objects is not supported in general, and isn’t possible in some programming languages.
+
+If preservation of order is important, such ordering information should be represented separately. The examples below shows a plain (unordered) object alongside three alternative representations which also provide ordering detail.
+
+```json
```
flobernd, line 19: The comments inside the JSON code block are highlighted as "invalid syntax". Let's maybe use this little hack to make it a little bit prettier (`jsonc` or `json5` could as well be used, but that does not highlight the field names in the same way as `json`):

``````suggestion
```yaml
``````

File: api-design-guidelines/requests-responses.md [hunk 4]
```diff
@@ -0,0 +1,157 @@
...
+  "indices": "index-name", // Input is a string.
+  ...
+}
+
+GET /_ml/datafeeds/feed-id
+{
+  "indices": ["index-name"], // Always returns a list of strings.
+  ...
+}
+```
```
flobernd, line 126:
``````suggestion
```yaml
PUT /_ml/datafeeds/feed-id
{
  "indices": "index-name", // Input is a string.
  // ...
}

GET /_ml/datafeeds/feed-id
{
  "indices": ["index-name"], // Always returns a list of strings.
  // ...
}
```
``````

flobernd (approved):
LGTM! Just a few style comments.

api-design-guidelines/data-modelling.md
technige: Yes, that was accidentally included. I hadn't saved the updates, thanks for pointing out :)

api-design-guidelines/data-modelling.md
technige: I agree that this is "broken" but I have ported over the files as-is from their source, where this invalid JSON also exists. I'm wondering whether this should be fixed here or in a separate commit later.

PR #2477, opened 2024-04-03
Changed json blocks to use yaml tag

technige:
Updating JSON blocks to use `yaml` tag to allow for invalid syntax in comments, etc


flobernd (approved):
LGTM!

PR #2490, opened 2024-04-08
Make k and num_candidates optional in kNN queries

pquentin:
This reflects the changes made in 8.13: https://github.com/elastic/elasticsearch/pull/101209. The current situation forces setting values: https://discuss.elastic.co/t/knnquery-in-java-client/356645/1


pquentin:
The validation results appear to be worse than what we have on main currently, so I will investigate.

swallez (changes requested):
`KnnSearch` is good, but there are some discrepancies on `KnnQuery`:
* it must extend `QueryBase` to have the `_name` and `boost` fields (the latter should be removed from the type)
* `num_candidates` should be optional
* `k` and `inner_hits` should be removed.

Can you add these fixes?

pquentin:
Thanks, done! The CI failure is addressed in https://github.com/elastic/elasticsearch-specification/pull/2492.

l-trotta (approved):

PR #2493, opened 2024-04-09
Fix 404 links in "add new api "docs

jedrazb:
Changing `__global` to `_global` in add new api docs

https://github.com/elastic/elasticsearch-specification/blob/main/specification/__global doesn't exist


pquentin (approved):
Thanks! LGTM.

PR #2495, opened 2024-04-09
Expose NullValue with export statement

jedrazb:
Add `export` statement to expose types globally


jedrazb:
fixed in https://github.com/elastic/elasticsearch-specification/pull/2496

PR #2497, opened 2024-04-09
Define global WithNullValue<T> type

jedrazb:
This defines a generic `WithNullValue` type that allows for semantic null assignment. 

Connectors type declarations would use this type in several different places so I'm proposing to make it a global utils type.


swallez (approved):
LGTM, thanks!

PR #2485, opened 2024-04-04
Java client issues batch 3

l-trotta:
Fixes provided in this PR:

- [456](https://github.com/elastic/elasticsearch-java/issues/456): policy name in ILM should be optional. [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-settings.html#_index_level_settings)
- [457](https://github.com/elastic/elasticsearch-java/issues/457): VariableWidthHistogramAggregation missing the `script` field. [server code class](https://github.com/elastic/elasticsearch/blob/0cc19f32704cfd168d3d3fa3cb2200c27e3b4c6b/server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/VariableWidthHistogramAggregationBuilder.java#L36), which extends [this class](https://github.com/elastic/elasticsearch/blob/0cc19f32704cfd168d3d3fa3cb2200c27e3b4c6b/server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceAggregationBuilder.java#L300), which has script. 
- [460](https://github.com/elastic/elasticsearch-java/issues/460): applies fix from [this comment](https://github.com/elastic/elasticsearch-java/issues/460#issuecomment-1402057328). @Anaethelion @flobernd in the java client generator I had to replace {dynamic_property} with {dynamic_type}, so there's a possibility that it could affect other client generators. 
- [466](https://github.com/elastic/elasticsearch-java/issues/466),[766](https://github.com/elastic/elasticsearch-java/issues/766): bucket aggregation subclasses should not have `name` or `meta` fields. 
- [472](https://github.com/elastic/elasticsearch-java/issues/472): `primaryTerm` in IndexResponse should be optional. no hard evidence from server, but running an indexing request with a noop pipeline (check the [issue](https://github.com/elastic/elasticsearch-java/issues/472)) effectively returns a response with no `primaryTerm`.
[...]

PR #2498, opened 2024-04-09
Index settings lifecycle name should be optional, not ILM policy name

JoshMock:
JavaScript codegen [started failing](https://github.com/elastic/clients-team/actions/runs/8611069657/job/23597614947#step:3:205) for me after [this PR](https://github.com/elastic/elasticsearch-specification/pull/2485) merged the other day, specifically breaking on the "policy name in ILM should be optional" change. [The docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html#ilm-put-lifecycle-path-params) say `policy_id` is a required parameter, since it is part of the URL path. 

I'm fairly certain the optional name that [the original issue](https://github.com/elastic/elasticsearch-java/issues/456) was referring to was for `IndexSettingsLifecycle` rather than `PutLifecycleRequest`.

### DIFF
--- a/specification/ilm/put_lifecycle/PutLifecycleRequest.ts
+++ b/specification/ilm/put_lifecycle/PutLifecycleRequest.ts
@@ -35,7 +35,7 @@ export interface Request extends RequestBase {
      * Identifier for the policy.
      * @codegen_name name
      */
-    policy?: Name
+    policy: Name
   }
   query_parameters: {
     /**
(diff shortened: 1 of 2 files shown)


l-trotta (approved):

l-trotta:
this is the correct fix, sorry for the oversight

JoshMock:
No problem! It took me a while to figure out that there were two separate possible places to look for this. :smile:

PR #2503, opened 2024-04-11
Connector SyncJob API spec

jedrazb:
### Changes

Added the API spec for endpoints related to connector sync job managements.


specification/connector_sync_job/_types/SyncJob.ts
```diff
@@ -0,0 +1,70 @@
...
+  canceled_at?: string
+  completed_at?: string
+  connector: SyncJobConnectorReference
+  created_at: string
+  deleted_document_count: long
+  error?: string
+  id: Id
+  indexed_document_count: long
+  indexed_document_volume: long
+  job_type: SyncJobType
```
l-trotta, line 62: can `job_type` be a SyncJobType or SyncJobType[] enum consistently?

l-trotta: same for trigger_method

l-trotta:
I have found a type that was introduced in the previous PR that I'm only now noticing is not matching the server code. Since this new API is using it, could you fix it here? 
[FilteringConfig](https://github.com/elastic/elasticsearch-specification/blob/4718254e0ea2bab696610ef8e141c140d116fa6e/specification/connector/_types/Connector.ts#L208) -> `active` and `draft` should be optional, `domain` should be removed, it seems like an [internal parameter](https://github.com/elastic/elasticsearch/blob/119f81f04305b91b9bfe9866525888ef17d35423/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/ConnectorFiltering.java#L47)

specification/connector_sync_job/_types/SyncJob.ts
```diff
@@ -0,0 +1,70 @@
...
+  SyncStatus
+} from '../../connector/_types/Connector'
+
+interface SyncJobConnectorReference {
+  configuration: ConnectorConfiguration
+  filtering: FilteringConfig
+  id: Id
+  index_name: string
+  language?: string
+  pipeline: IngestPipelineParams
```
l-trotta, line 36: pipeline should be [optional](https://github.com/elastic/elasticsearch/blob/119f81f04305b91b9bfe9866525888ef17d35423/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/Connector.java#L100C43-L100C51)

Anaethelion (changes requested):
Left a few comments, found some fields that could be hinted as datetime to enhance deserialization.

specification/connector_sync_job/_types/SyncJob.ts
jedrazb: In a `ConnectorSyncJob` representation they are not an array, there is singe value associated with `trigger_method` and `job_type`

jedrazb:
> I have found a type that was introduced in the previous PR that I'm only now noticing is not matching the server code. Since this new API is using it, could you fix it here?
[FilteringConfig](https://github.com/elastic/elasticsearch-specification/blob/4718254e0ea2bab696610ef8e141c140d116fa6e/specification/connector/_types/Connector.ts#L208) -> active and draft should be optional, domain should be removed, it seems like an [internal parameter](https://github.com/elastic/elasticsearch/blob/119f81f04305b91b9bfe9866525888ef17d35423/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/ConnectorFiltering.java#L47)

Yup, there were couple of changed to connector (not sync job) endpoints in 8.14 that were merged after we merged #2367 . I will followup to adapt the types in a separate PR.

specification/connector_sync_job/_types/SyncJob.ts
```diff
@@ -0,0 +1,70 @@
...
+  error?: string
+  id: Id
+  indexed_document_count: long
+  indexed_document_volume: long
+  job_type: SyncJobType
+  last_seen?: string
+  metadata: Dictionary<string, UserDefinedValue>
+  started_at?: string
+  status: SyncStatus
+  total_document_count?: long
```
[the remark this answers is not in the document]
jedrazb: Thank you for catching this

specification/connector_sync_job/_types/SyncJob.ts
```diff
@@ -0,0 +1,70 @@
...
+  access_control
+}
+
+enum TriggerMethod {
+  on_demand,
+  scheduled
+}
+
+export interface ConnectorSyncJob {
+  cancelation_requested_at?: string
```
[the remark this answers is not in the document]
jedrazb: Yup it can be `DateTime`, I adapted Sync Job representation to use this type. 

Note, there are several properties in `Connector` representation that should use it as well. I will follow up in a separate PR since I need to reflect a few changes that were merged in to ES after #2367

specification/connector_sync_job/_types/SyncJob.ts
l-trotta: What I mean is, in this class `job_type` and `trigger_method` are enums, while in the classes SyncJobPostRequest.ts and SyncJobListRequest.ts they are `Name`/`Names`, which are just strings, can't they be the same enum everywhere?

jedrazb:
Sync job endpoints were moved under the `connector` namespace, I reflected changes in this PR after https://github.com/elastic/elasticsearch/pull/107476 was merged.

specification/connector_sync_job/_types/SyncJob.ts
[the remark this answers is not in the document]
flobernd: @jedrazb I did that already a few days ago 🙂 https://github.com/elastic/elasticsearch-specification/pull/2507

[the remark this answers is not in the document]
jedrazb: Ahh thank you! Approved again!

specification/connector/_types/SyncJob.ts
```diff
@@ -0,0 +1,71 @@
...
+} from './Connector'
+
+interface SyncJobConnectorReference {
+  configuration: ConnectorConfiguration
+  filtering: FilteringConfig
+  id: Id
+  index_name: string
+  language?: string
+  pipeline?: IngestPipelineParams
+  service_type: string
```
l-trotta, line 38: `service type` seems to be [optional](https://github.com/elastic/elasticsearch/blob/eb6af0e6b549d0810d8b89791556bc791b766652/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/Connector.java#L103) as well. other than this, LGTM :D

specification/connector_sync_job/_types/SyncJob.ts
jedrazb: Ahh I see what you mean! For `job_type` this is passed as a comma-separated list in the URL param of `SyncJobListRequest`. I was not sure if defining this as `SyncJobType[]` would work with URL param list, and I saw `Names` in other examples of requests that pass comma separated lists in url params. 

If the usage of `SyncJobType[]` and `SyncJobTriggerMethod` is correct, I can adapt this in the PR.

specification/connector/_types/SyncJob.ts
jedrazb: For connector yes it can be optional. But you do need to have the `service_type` set to a non-null value before starting a sync job for a given connector. So we can expect this value to always be present.

I do see that our [sync job parser for its connector property](https://github.com/elastic/elasticsearch/blob/91d3bb026fa8272bb847e84f85cc18f6f93ebc2a/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/syncjob/ConnectorSyncJob.java#L377) in ES is lacking with that respect  I'm creating an issue to fix that https://github.com/elastic/enterprise-search-team/issues/7344

specification/connector_sync_job/_types/SyncJob.ts
l-trotta: thanks :) I think it's more coherent this way from the clients perspective, to always provide enums where there's fixed values

jedrazb: I addressed that feedback in the updated PR :)

l-trotta (approved):

jedrazb:
Thank you all for reviewing! I need to include this in the `8.14` release. I still don't see the `backport 8.14` label, do I need to do sth specific to make sure it lands in the 8.14 BC?

Anaethelion (approved):
LGTM!

PR #2512, opened 2024-04-15
Add esql version

nik9000:
Adds a `version` parameter to the ESQL `_query` request endpoint. We'll do some work to give this a default value in the generated clients soon, but for 8.14+ this'll be a required parameter.

### DIFF
--- a/specification/esql/query/QueryRequest.ts
+++ b/specification/esql/query/QueryRequest.ts
@@ -62,7 +62,7 @@ export interface Request extends RequestBase {
      */
     query: string,
     /**
-     * The version of the ES|QL language in which the "query" field uses was written.
+     * The version of the ES|QL language in which the "query" field was written.
      */
     version: Version
   }
--- a/specification/esql/query/QueryRequest.ts
+++ b/specification/esql/query/QueryRequest.ts
@@ -60,7 +60,7 @@ export interface Request extends RequestBase {
     /**
      * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
      */
-    query: string,
+    query: string
     /**
      * The version of the ES|QL language in which the "query" field was written.
      */
--- a/specification/esql/query/QueryRequest.ts
+++ b/specification/esql/query/QueryRequest.ts
@@ -20,6 +20,7 @@
 import { RequestBase } from '@_types/Base'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { ScalarValue } from '@_types/common'
+import { Version } from '@esql/_types/Version'
 
 /**
  * Executes an ES|QL request
@@ -59,6 +60,10 @@ export interface Request extends RequestBase {
     /**
      * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
      */
-    query: string
+    query: string,
+    /**
+     * The version of the ES|QL language in which the "query" field uses was written.
+     */
+    version: Version
   }
 }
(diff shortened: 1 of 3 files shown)


nik9000:
> You can validate these APIs yourself by using the `make validate` target.

I'm running:
```
make validate api=esql.query type=request stack-version=8.14.0-SNAPSHOT
```
and it's been half an hour or so installing a flight recorder.

output/schema/schema.json
```diff
@@ -110992,6 +110992,24 @@
...
     },
+    {
+      "kind": "enum",
+      "members": [
+        {
+          "description": "Run against the first version of ESQL.",
+          "name": "2024.04.01"
+        },
+        {
+          "description": "Run against the unreleased version of the ESQL language. This will\ncontain any changes staged for release that have yet to be included\nin a version.",
```
alex-spies, line 111003: IMHO we should mention that this only works on SNAPSHOT builds. ES on non-SNAPSHOT builds will return a 400 if it sees `"version": "snapshot"`.

I wonder if we should even make `snapshot` a part of the official specs - it's not like users miss out if they cannot use language feature xyz, given that existing language versions are going to still be updated with new features.

Can we maybe add `snapshot` here later, if it turns out that there's an important enough use case for it?

output/schema/schema.json
```diff
@@ -111062,6 +111080,18 @@
                 "namespace": "_builtins"
               }
             }
+          },
+          {
+            "description": "The version of the ES|QL language in which the \"query\" field uses was written.",
```
alex-spies, line 111085: Typo
```suggestion
            "description": "The version of the ES|QL language in which the \"query\" field was written.",
```

output/schema/schema.json
```diff
@@ -110992,6 +110992,24 @@
       },
       "specLocation": "eql/search/types.ts#L20-L32"
     },
+    {
+      "kind": "enum",
+      "members": [
+        {
+          "description": "Run against the first version of ESQL.",
```
alex-spies, line 110999: super nit, but we could be more consistent with our use of ESQL vs ES|QL:
```suggestion
          "description": "Run against the first version of ES|QL.",
```
Also applies below and in `Version.ts`.

specification/esql/query/QueryRequest.ts
```diff
@@ -59,6 +60,10 @@ export interface Request extends RequestBase {
     /**
      * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
      */
-    query: string
+    query: string,
+    /**
+     * The version of the ES|QL language in which the "query" field uses was written.
```
alex-spies, line 65:
```suggestion
     * The version of the ES|QL language in which the "query" field was written.
```

alex-spies (commented):
Can't comment on how this affects clients etc., but I checked/proofread the contained docs.

Thanks a lot @nik9000 !

Only reservation I have is whether official specs should already document `"version": "snapshot"`, which is intended for testing purposes only.

output/schema/schema.json
nik9000: Yeah, I'll drop it.

alex-spies (approved):
I committed the updated jsons obtained from `make generate`.

Looks good to me!

flobernd:
fyi: I created [this workflow](https://github.com/elastic/elasticsearch-specification/blob/main/.github/workflows/generate.yml) a while ago which automatically takes care of generating- and pushing the output json, if it is not part of the PR. This makes including the output json optional 🙂 It was primarily done to allow users to contribute without setting up Node etc., but might as well save some time for everybody else.

PR #2513, opened 2024-04-16
Change `QueryCacheStats` field types from `int` -> `long`

flobernd:
Server class: https://github.com/elastic/elasticsearch/blob/08a2f7991e214f2ac8e8b274ac51a540c30b6f3c/server/src/main/java/org/elasticsearch/index/cache/query/QueryCacheStats.java#L23

Related to: https://github.com/elastic/elasticsearch-net/issues/8015


pquentin (approved):
Thanks! LGTM.

PR #2522, opened 2024-04-18
Add missing fields property on top_hits aggregation

JoshMock:
See https://github.com/elastic/elasticsearch-js/pull/2209.

### DIFF
--- a/specification/_types/aggregations/metric.ts
+++ b/specification/_types/aggregations/metric.ts
@@ -338,7 +338,7 @@ export class TopHitsAggregation extends MetricAggregationBase {
   /**
    * Fields for which to return doc values.
    */
-  docvalue_fields?: Fields
+  docvalue_fields?: FieldAndFormat[]
   /**
    * If `true`, returns detailed information about score computation as part of a hit.
    * @server_default false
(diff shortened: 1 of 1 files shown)


specification/_types/aggregations/metric.ts
```diff
@@ -344,6 +344,11 @@ export class TopHitsAggregation extends MetricAggregationBase {
    * @server_default false
    */
   explain?: boolean
+  /**
+   * Array of wildcard (*) patterns. The request returns values for field names
+   * matching these patterns in the hits.fields property of the response.
+   */
+  fields?: (FieldAndFormat | Field)[]
```
l-trotta, line 351: should it be `FieldAndFormat[] | Field[]`?

JoshMock: That's a good question. My assumption was that `fields` could take an array that consisted of either, or a combination of the two (e.g. `['field_one', { field: 'field_two', format: 'xxx' }]`). [The docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#search-fields-request) suggest that's the case as well.

I was looking at the `FieldFetcher` subphase in the Elasticsearch repo to verify in the source, but my Java is pretty rusty so it's hard to tell. What does it seem like to you?

l-trotta (approved):

PR #2528, opened 2024-04-25
Use endpoint instead of model in the inference API

davidkyle:
Update the inference APIs to refer to Inference endpoints rather than models. This involves renaming classes and updating comments but also includes the potentially breaking changes:

- `inference.delete_model` renamed to `inference.delete`
- `inference.get_model` renamed to `inference.get`
- `inference.put_model` renamed to `inference.put`

The format of the `get` response has also changed, the `models` field is renamed to `endpoints`:

```
#previously
{
    "models": [...]  
}

#with this change
{
    "endpoints": [...]  
}
```

In `GetResponse.ts` I've used an `@alias` tag to retain the `models` option

Corresponding Elasticsearch change https://github.com/elastic/elasticsearch/pull/107704


pquentin (commented):
Thanks! Since this is for 8.15, I suggest waiting for https://github.com/elastic/elasticsearch/pull/107704 so that we make sure 1/ the JSON spec is exactly the same and 2/ allow clients-flight-recorder to keep up with the change and validate this properly.

I left a few comments there too.

davidkyle:
> Thanks! Since this is for 8.15, I suggest waiting for https://github.com/elastic/elasticsearch/pull/107704

That PR is green now. I hesitated before pressing the merge button as after merging we must have a compatible client ready for the next serverless release and when it comes to the breaking change I am hoping there is someway to _soften the blow_. 

Taking Python as an example, existing code will use `client.inference.put_model(...)` and after this change the Python function is `client.inference.put(...)`. Is there anyway of having `inference.put_model` resolve to `inference.put` so that we don't break existing code samples? Perhaps by having spec for both `inference.put` and `inference.put_model` but the latter is deprecated and not documented?

pquentin:
I'm not sure we can deprecate an endpoint today if the endpoint itself is gone from Elasticsearch. (The only existing example is _knn_search.) Also, the clients are one thing, but on Serverless right now the usage of Kibana is quite high. And it's not like deprecating a function has any practical effect: things will break eventually. I suppose that's why tech preview is for.

specification/inference/get/GetResponse.ts
```diff
@@ -17,10 +17,11 @@
  * under the License.
  */
 
-import { ModelConfigContainer } from '@inference/_types/Services'
+import { InferenceEndpointContainer } from '@inference/_types/Services'
 
 export class Response {
   body: {
-    models: Array<ModelConfigContainer>
+    /** @aliases models */
```
pquentin, line 24: Since we decided to break compatibility, should we keep models as an alias? For what it's worth, the current Python client does the wrong thing and would send `models` as is to the server. That could be fixed in time for 8.15 though.

specification/inference/inference/InferenceRequest.ts
```diff
@@ -54,7 +54,7 @@ export interface Request extends RequestBase {
      */
     query?: string
     /**
-     * Text input to the model.
+     * Inference input
```
pquentin, line 57: Adding a full stop here will clarify the one-line documentation in Python code.

```suggestion
     * Inference input.
```

pquentin (changes requested):
Thanks! Requested a few cosmetic fixes and asked about the `models` alias.

specification/inference/get/GetResponse.ts
davidkyle: I agree and have removed the alias.

The change breaks compatibility, we can do that because the API is neither GA or heavily used therefore it should be a clean break.

specification/inference/_types/Services.ts
```diff
@@ -33,21 +33,21 @@ export class ModelConfig {
...
   /**
-   * Task settings specific to the service and model
+   * Task settings specific to the service and task type
    */
   task_settings: TaskSettings
 }
 
 /**
- * Represents a model as returned by the GET API
+ * Represents a infenence endpoint as returned by the GET API
```
miguelgrinberg, line 42:
```suggestion
 * Represents a inference endpoint as returned by the GET API
```

pquentin (approved):
Thanks! LGTM.

davidkyle:
Thanks for the reviews

PR #2529, opened 2024-04-25
Add `TermsRangeQuery`

flobernd:
Currently, we only support `DateRangeQuery` and `NumberRangeQuery`. According to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#ranges-on-text-and-keyword), range queries can as well be used with `text` and `keyword` fields.

Related to: https://github.com/elastic/elasticsearch-net/issues/8004

### DIFF
--- a/specification/_types/query_dsl/term.ts
+++ b/specification/_types/query_dsl/term.ts
@@ -163,9 +163,30 @@ export class NumberRangeQuery extends RangeQueryBase {
   to?: double | null
 }
 
-/** @codegen_names date, number */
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM.

I think it also works with other types, like booleans or IP addresses (covered by strings?) but I'm not sure and this addition is correct.

Also note that by default such range queries are not allowed: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html#query-dsl-allow-expensive-queries

PR #2533, opened 2024-04-26
Fix spec compilation on Windows

flobernd:
`ts-morph` provides us with a normalized path that always uses `/` as the path separator (even on Windows). In this PR I replace some occurences of `sep` with just `/` in that context.

Besides that, on Windows, multi-line strings are emitted with CRLF (`/r/n`) instead of just LF (`/n`). I use some ugly `replace` calls to fix this. A better solution would probably be the usage of `.gitattributes` to make sure `.ts` files are always committed and checked out in LF style. But I'm not sure if that would cause any side-effects.

TODO: Order of endpoints in the output json is not consistent between Windows and Linux. I did not dig into that for now as it does not change semantics...

### DIFF
--- a/compiler/src/model/build-model.ts
+++ b/compiler/src/model/build-model.ts
@@ -400,7 +400,7 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
               )
               const jsDocs = child.getJsDocs()
               if (jsDocs.length > 0) {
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
LGTM! Does this resolve that `undefined` mess you found the other day?

flobernd:
@JoshMock Yes, that fixes the undefined namespace issue 🙂

PR #2536, opened 2024-04-29
Move from STACK_VERSION to SNAPSHOT_BRANCH

pquentin:
Following the changes in elasticsearch-specification.

~~The sed change can be tested with `VERSION=8.13 .ci/bump.sh`. (I would appreciate someone testing with BSD sed, as I tried to make it compatible which can make local testing easier.)~~ Since we use the branch name now, we can get that from GitHub Actions and don't need to bump anymore.

I had to make a random change to trigger validation.


l-trotta (approved):
LGTM!

PR #2544, opened 2024-05-06
Changed inference name to avoid duplicate

l-trotta:
The java generator was complaining about types `InferenceEndpointContainer` and `InferenceEndpoint`, changed the first one's name since it's not technically a container class.


flobernd (approved):
LGTM! I had a name collision with this as well.

Issue #2545, opened 2024-05-07
Proposal: Change declaration of `RangeQuery`


flobernd:
## Problem Description

`RangeQuery` is currently declared as:

```ts
/** @codegen_names date, number, terms */
// Note: deserialization depends on value types
export type RangeQuery = DateRangeQuery | NumberRangeQuery | TermsRangeQuery
```

This declaration is highly problematic for statically typed clients as there is no way to correctly **de**serialize into the correct union variant, because for example the following variants share the same underlying JSON type:

```ts
export class TermsRangeQuery extends RangeQueryBase {
  gt?: string
```

```ts
export class DateRangeQuery extends RangeQueryBase {
  gt?: DateMath // string
```

## Proposal

It seems like on server side, there is only one `RangeQuery` with an `object/any` type and the actual semantics are applied on base of the type mappings.

It would make sense to reflect this in our specification while still providing _hints_ about potential specializations to the code generator.

The naiv approach would be to change the specification like this:

```ts
export class RangeQuery extends RangeQueryBase {
  gt?: any
  // ...
}
```

This would allow serialization and de-serialization of arbitrary value types. However, this would as well complicate building queries in statically typed languages.

A better way could be a construct like this:

```ts
export type RangeQuery = GenericRangeQuery<any>;

export class GenericRangeQuery<T> extends RangeQueryBase {
  gt?: T
   // ...
}
```

Potentially with new annotations:

```ts
@known_generic = [DateMath, "DateRangeQuery" ]
@known_generic = [double  , "NumberRangeQuery"]
@known_generic = [string  , "TermRangeQuery"  ]
export type RangeQuery = GenericRangeQuery<any>;
```

Based on this, generators could generate specialized code only for serialization (deserialization always uses `DateRange<any>`):

```csharp
// Untyped/any variant always used for deserialization; can optionally be used by the user when building queries
public class RangeQuery
{
	public object? Gt { get; set; }
}

public abstract class GenericRangeQuery<T> : RangeQuery
{
	public new T? Gt { get; set; }
}

// Specialized range query for numbers
public sealed class NumberRangeQuery : GenericRangeQuery<double>
{
}

// Specialized range query for dates
public sealed class DateRangeQuery : GenericRangeQuery<DateMath>
{
}
```

## Side Notes

There are more types for which the same issue applies, for example:

```ts
/** @codegen_names date, numeric, geo */
// Note: deserialization depends on value types
export type DecayFunction =
[...]

PR #2548, opened 2024-05-08
Improve declaration of untagged unions to allow for generator optimizations

flobernd:
## Problem

`RangeQuery` is currently declared as:

```ts
/** @codegen_names date, number, terms */
// Note: deserialization depends on value types
export type RangeQuery = DateRangeQuery | NumberRangeQuery | TermsRangeQuery
```

This declaration is highly problematic for statically typed clients as there is no way to correctly **de**serialize into the correct union variant, because for example the following variants share the same underlying JSON type:

```ts
export class TermsRangeQuery extends RangeQueryBase {
  gt?: string
```

```ts
export class DateRangeQuery extends RangeQueryBase {
  gt?: DateMath // string
```

There are more types for which the same issue applies, for example:

```ts
/** @codegen_names date, numeric, geo */
// Note: deserialization depends on value types
export type DecayFunction =
  | DateDecayFunction
  | NumericDecayFunction
  | GeoDecayFunction
```

```ts
/** @codegen_names geo, date */
// Note: deserialization depends on value types
export type DistanceFeatureQuery =
  | GeoDistanceFeatureQuery
  | DateDistanceFeatureQuery
```

## Proposal

Rearrange and mark these kind of untagged unions, that only differ by the type of some fields, in a specific way so that code generators can drive specialized generation to allow for asymmetric (de-)serialization:

```ts
export class MyTypeBase<T1, T2, ...> { ... }

export class MyTypeSpecialized1 extends MyTypeBase<int> {}
export class MyTypeSpecialized2 extends MyTypeBase<string> {}
export class MyTypeSpecialized3 extends MyTypeBase<bool> {}

/**
 * @codegen_names 1, 2, 3 
 * @variant untagged
 */
// Note: deserialization depends on value types
export type MyType = MyTypeSpecialized1 | MyTypeSpecialized2 | MyTypeSpecialized3 
```

In .NET world, the resulting class design could look like this:

```csharp
// Untyped/any variant always used for deserialization; can optionally be used by the user when building queries
public class RangeQuery
{
    public object? Gt { get; set; }
}

public abstract class GenericRangeQuery<T> : RangeQuery
{
    public new T? Gt { get; set; }
}

// Specialized range query for numbers
public sealed class NumberRangeQuery : GenericRangeQuery<double> { }

// Specialized range query for dates
public sealed class DateRangeQuery : GenericRangeQuery<DateMath> { }
```

---

Motivation and initial discussion: #2545

### DIFF
--- a/compiler-rs/clients_schema/src/lib.rs
+++ b/compiler-rs/clients_schema/src/lib.rs
@@ -380,6 +380,7 @@ pub enum ServerDefault {
 pub enum Variants {
     ExternalTag(ExternalTag),
     InternalTag(InternalTag),
+    Untagged(Untagged),
     Container(Container),
 }
 
@@ -402,6 +403,13 @@ pub struct InternalTag {
     /// Default value for the variant tag if it's missing
     #[serde(skip_serializing_if = "Option::is_none")]
     pub default_tag: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Untagged {
+    #[serde(default)]
+    pub non_exhaustive: bool,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -799,6 +807,7 @@ impl TypeAlias {
 pub enum TypeAliasVariants {
     ExternalTag(ExternalTag),
     InternalTag(InternalTag),
+    Untagged(Untagged),
 }
 
 //------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


swallez (changes requested):
Can you add a brief description of this new variant type [in the docs](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#variants)?

And mentioning explicitly that this is used for legacy types and should generally avoided as it results in less optimal code generation in client libraries.

flobernd:
Done.

l-trotta:
@swallez added a commit to fix openapi generation (also enhanced error logging for the deserialization of the schema.json, thanks @Anaethelion)

flobernd:
## Spec Meeting 2024/06/27

1. Explicitly name the "deserialization"/untyped variant in the decorator
2. Add a strong validation to disallow adding new "untagged" variants
3. Check flattening of generics

Anaethelion (approved):
LGTM!

Issue #2508, opened 2024-04-15
Add the retriever API in _search


ezimuel:
We need to add the Retriever API, see documentation [here](https://www.elastic.co/guide/en/elasticsearch/reference/master/retriever.html).

This API is designed as parameter of the `_search` API.

The retriever API is already in elasticsearch main, see https://github.com/elastic/elasticsearch/pull/105470 and it will be available in 8.14.0.

PR #2551, opened 2024-05-10
Adding retriever specification to _search API

pmpailis:
This PR adds the `retriever` specification to the `_search` API, as is also defined in the 8.14.0 documentation to be released [here](https://www.elastic.co/guide/en/elasticsearch/reference/master/retriever.html) .

Closes https://github.com/elastic/elasticsearch-specification/issues/2508


pmpailis:
Running `make validate` locally I get many errors that seem to be unrelated to my changes: 
```
✖ ../output/typescript/types.ts(2894,3): error TS2411: Property 'filters' of type 'Record<string, QueryDslQueryContainer>' is not assignable to 'string' index type 'never'.
```
```
export interface AggregationsAdjacencyMatrixAggregation extends AggregationsBucketAggregationBase {
  filters?: Record<string, QueryDslQueryContainer>
  separator?: string
}
```

carlosdelest:
> Running `make validate` locally I get many errors that seem to be unrelated to my changes:

I wouldn't worry about those, as you mention they are not related to your changes. Given that `make compile` works and `make validate` does not refer to errors added by your code, I think this is ready for review by the clients team when you feel ready 👍

pmpailis:
Thanks @carlosdelest ! AFAICT there are no **new** validation errors introduced with these changes, as I'll mark it for review and we can take it from there. :)

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+  similarity?: float
+}
+
+export class RRFRetriever extends RetrieverBase {
+  /** A list of child retrievers to specify which sets of returned top documents will have the RRF formula applied to them.  */
+  retrievers: RetrieverContainer[]
+  /** This value determines how much influence documents in individual result sets per query have over the final ranked result set. */
+  rank_constant?: long
+  /** This value determines the size of the individual result sets per query.  */
+  window_size?: long
```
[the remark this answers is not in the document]
Anaethelion: Same for this one, reading the parser, I think this is mandatory.

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** Query vector. Must have the same number of dimensions as the vector field you are searching against. You must provide a query_vector_builder or query_vector, but not both. */
+  query_vector?: QueryVector
+  /** Defines a model to build a query vector. */
+  query_vector_builder?: QueryVectorBuilder
+  /** Number of nearest neighbors to return as top hits. */
+  k: long
+  /** Number of nearest neighbor candidates to consider per shard. */
+  num_candidates: long
+  /** Query to filter the documents that can match. The kNN search will return the top k documents that also match this filter. */
+  filter?: QueryContainer | QueryContainer[]
```
Anaethelion, line 69: `filter` should be in `RetrieverBase` as its found in all 3 retrievers and part of `RetrieverBuilder` server-side.

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** The name of the vector field to search against. */
+  field: string
+  /** Query vector. Must have the same number of dimensions as the vector field you are searching against. You must provide a query_vector_builder or query_vector, but not both. */
+  query_vector?: QueryVector
+  /** Defines a model to build a query vector. */
+  query_vector_builder?: QueryVectorBuilder
+  /** Number of nearest neighbors to return as top hits. */
+  k: long
+  /** Number of nearest neighbor candidates to consider per shard. */
+  num_candidates: long
```
Anaethelion, line 67:
```suggestion
  num_candidates: integer
```

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export class KnnRetriever extends RetrieverBase {
+  /** The name of the vector field to search against. */
+  field: string
+  /** Query vector. Must have the same number of dimensions as the vector field you are searching against. You must provide a query_vector_builder or query_vector, but not both. */
+  query_vector?: QueryVector
+  /** Defines a model to build a query vector. */
+  query_vector_builder?: QueryVectorBuilder
+  /** Number of nearest neighbors to return as top hits. */
+  k: long
```
Anaethelion, line 65:
```suggestion
  k: integer
```

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+  rrf?: RRFRetriever
+}
+
+export class RetrieverBase {}
+
+export class StandardRetriever extends RetrieverBase {
+  /** Defines a query to retrieve a set of top documents. */
+  query?: QueryContainer
+  /** Applies a boolean query filter to this retriever where all documents must match this query but do not contribute to the score. */
+  filter?: QueryContainer | QueryContainer[]
```
Anaethelion, line 44:
```suggestion
```

Same as below, filter should be in the `RetrieverBase`

Anaethelion (changes requested):
Overall looks good, I've left a few suggestions to make sure we get the field type right.

Can you check about the RRF Retriever field being optional ? I hav en't found anything that goes in that direction but I may have missed something.

specification/_types/Retriever.ts
pmpailis: This is how I had defined it as well, but I changed it to `long` to be consistent with the other`KnnSearch` interfaces ([here](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/knn_search/_types/Knn.ts#L30) and [here](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/Knn.ts#L38)). Maybe we should update those as well? 🤔

specification/_types/Retriever.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** The minimum similarity required for a document to be considered a match.  */
+  similarity?: float
+}
+
+export class RRFRetriever extends RetrieverBase {
+  /** A list of child retrievers to specify which sets of returned top documents will have the RRF formula applied to them.  */
+  retrievers: RetrieverContainer[]
+  /** This value determines how much influence documents in individual result sets per query have over the final ranked result set. */
+  rank_constant?: long
```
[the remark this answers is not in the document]
pmpailis: There is a [default value](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/rank-rrf/src/main/java/org/elasticsearch/xpack/rank/rrf/RRFRetrieverBuilder.java#L57) passed to the parser so a request missing `rank_constant` and `rank_window_size` for the RRF retriever would still succeed.

specification/_types/Retriever.ts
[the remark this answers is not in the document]
pmpailis: Same as above, there is a default value passed to the parser, so maybe this should be ok?

specification/_types/Retriever.ts
pmpailis: Yeap, you're right! Added as separate entities to have better clarity on the effect that each filter might have, but it makes sense to add to the parent class. 
Moved :)

specification/_types/Retriever.ts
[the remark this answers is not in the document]
pmpailis: For 8.14, this is defined as `window_size`, whereas for 8.15 we have changed this to `rank_window_size`. Should we merge this now as `rank_window_size` and backport a renaming?

pmpailis:
> Can you check about the RRF Retriever field being optional ? 

Do you mean the `rrf` field in `RetrieverContainer`  ([link](https://github.com/elastic/elasticsearch-specification/pull/2551/files#diff-36609334c1841ab59d50f58344776bba368488b79408959f5156f30d0c2fdb0eR35))?  If so, it shouldn't be mandatory, we could still have a retriever-based request without rrf (e.g. using just kNN or the `standard` retriever).

pmpailis:
Thanks for taking a look @Anaethelion ! I've pushed some changes as per your suggestions :)

specification/_types/Retriever.ts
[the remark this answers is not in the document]
Anaethelion: Oh I see, let's merge this as is so we can backport and do another change only for 8.15.

specification/_types/Retriever.ts
Anaethelion: Ah I see, you are right, those two need to be updated.

pmpailis: Updated in 364b782e0362df299f6e009d56d8b30f53f41c68

specification/_types/Retriever.ts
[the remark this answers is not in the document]
pmpailis: Renamed back to `window_size` in 364b782e0362df299f6e009d56d8b30f53f41c68

Anaethelion (approved):
LGTM!

PR #2552, opened 2024-05-10
Java issues batch 5

l-trotta:
Fixes provided in this PR:

- [666](https://github.com/elastic/elasticsearch-java/issues/666): [AggregationRange](https://github.com/elastic/elasticsearch-specification/blob/02109140a64ce247fbb3fc0e6e4dd7b4b0d98da6/specification/_types/aggregations/bucket.ts#L674) accepting both double and string was probably established because at some point it was also used for RangeQuery, which accepts Objects for parameters `to` and `from`. But in the spec we just use it in [GeoDistanceAggregation](https://github.com/elastic/elasticsearch/blob/8864058f83bc485ce6c89a196e9d64b04e79b1cf/server/src/main/java/org/elasticsearch/search/aggregations/bucket/range/GeoDistanceAggregationBuilder.java#L98) and [RangeAggregation](https://github.com/elastic/elasticsearch/blob/8864058f83bc485ce6c89a196e9d64b04e79b1cf/server/src/main/java/org/elasticsearch/search/aggregations/bucket/range/GeoDistanceAggregationBuilder.java#L98), both accepting only doubles. Since the java client generalizes `string | double` as just string, this results in the server throwing exceptions when called. 
- [673](https://github.com/elastic/elasticsearch-java/issues/673): [FieldType](https://github.com/elastic/elasticsearch-specification/blob/53edb7836cd4d15fc15c1341a5eb7bf9dfb5d50a/specification/_types/mapping/Property.ts#L160) enum missing `version`. [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html)
- [690](https://github.com/elastic/elasticsearch-java/issues/690): [KeywordProperty](https://github.com/elastic/elasticsearch-specification/blob/53edb7836cd4d15fc15c1341a5eb7bf9dfb5d50a/specification/_types/mapping/core.ts#L89) missing `script` and `onScriptError` fields. [server code](https://github.com/elastic/elasticsearch/blob/e7350dce291921bded600cf306e4dca6138fdb25/server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java#L179)
- [696](https://github.com/elastic/elasticsearch-java/issues/696): [Normalizer](https://github.com/elastic/elasticsearch-specification/blob/53edb7836cd4d15fc15c1341a5eb7bf9dfb5d50a/specification/_types/analysis/normalizers.ts#L24) is missing a default type, which for the java client means deserialization error whenever any normalizer was not created through the client (because of the missing `type` field).
-  [648](https://github.com/elastic/elasticsearch-java/issues/648): [DynamicTemplate](https://github.com/elastic/elasticsearch-specification/blob/53edb7836cd4d15fc15c1341a5eb7bf9dfb5d50a/specification/_types/mapping/dynamic-template.ts#L22) remodeled to match the current [server code](https://github.com/elastic/elasticsearch/blob/7e38ee13d593984406c9192c5ee31b6b351e99e7/server/src/main/java/org/elasticsearch/index/mapper/DynamicTemplate.java#L256). `mapping` and `runtime` can be used alternatively (it's explained in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html)) so both fields are maps and only one must be set (therefore the container variant).

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -39537,18 +39537,7 @@
         "properties": {
           "from": {
             "description": "Start of the range (inclusive).",
-            "oneOf": [
-              {
-                "type": "number"
-              },
-              {
-                "type": "string"
-              },
-              {
-                "nullable": true,
-                "type": "string"
-              }
-            ]
+            "type": "number"
           },
           "key": {
             "description": "Custom key to return the range with.",
@@ -39556,18 +39545,7 @@
           },
           "to": {
             "description": "End of the range (exclusive).",
-            "oneOf": [
-              {
-                "type": "number"
-              },
-              {
-                "type": "string"
-              },
-              {
-                "nullable": true,
-                "type": "string"
-              }
-            ]
+            "type": "number"
           }
         }
       },
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -47267,30 +47267,107 @@
         ]
       },
       "_types.mapping:DynamicTemplate": {
-        "type": "object",
-        "properties": {
-          "mapping": {
-            "$ref": "#/components/schemas/_types.mapping:Property"
+        "allOf": [
+          {
+            "type": "object",
+            "properties": {
+              "match": {
+                "oneOf": [
+                  {
+                    "type": "string"
+                  },
+                  {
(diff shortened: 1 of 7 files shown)


flobernd (approved):
LGTM!

flobernd:
The `onScriptError` is interesting. I did not find any mention of it in the docs.

PR #2557, opened 2024-05-17
Upgrade to checkout@v4

pquentin:
Validation currently fails, see https://github.com/elastic/elasticsearch-specification/actions/runs/9093977704/job/25087280424


pquentin:
I added a dummy file to trigger CI, which happened here: https://github.com/elastic/elasticsearch-specification/actions/runs/9123956086/job/25087262219. And the checkout worked, even though the dummy file broke validation itself.

l-trotta (approved):

PR #2563, opened 2024-05-17
fix   field_masking_span  to span_field_masking

vdemay:
field_masking_span has been deprecated since ES8.0 in favor of span_field_masking. Ref: https://github.com/elastic/elasticsearch/pull/74718/files

This change is not reflected in APIs causing exception in ES >= 8.0:
deprecated fields not supported here but got [field_masking_span] which has been replaced with [span_field_masking]

The following PR fix this replacing `field_masking_span queries` by `span_field_masking`


l-trotta (approved):
LGTM!

PR #2564, opened 2024-05-21
Add ml.update_trained_model_deployment API

pquentin:
I'm not sure if it should be added or not, as this API is currently in beta.

### DIFF
--- a/output/schema/validation-errors.json
+++ b/output/schema/validation-errors.json
@@ -1078,12 +1078,6 @@
       ],
       "response": []
     },
-    "ml.update_trained_model_deployment": {
-      "request": [
-        "Missing request & response"
-      ],
-      "response": []
-    },
     "ml.validate": {
       "request": [],
       "response": [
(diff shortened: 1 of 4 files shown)


davidkyle:
Elasticsearch is updated to make this API GA in https://github.com/elastic/elasticsearch/pull/108868

The same PR also adds `number_of_allocations` as a query param

pquentin:
@davidkyle What are your thoughts on merging this now, and then updating when the query parameters are available? For clients this makes no difference, the usage is the same.

davidkyle:
The PR that adds the query param (https://github.com/elastic/elasticsearch/pull/108868) is now merged, can the query param be added now?

davidkyle:
The API spec still says `beta` and needs updating from the ES repo. 

https://github.com/elastic/elasticsearch-specification/blob/e54499c8be3645d78f18c0692da4759ed2cd2984/specification/_json_spec/ml.update_trained_model_deployment.json#L7

pquentin:
I was thinking we could reflect the current reality in order to ship that today in 8.13, and then when 8.14.1 goes out we can update with the beta label and query parameters.

davidkyle (approved):
LGTM

davidkyle:
> I was thinking we could reflect the current reality in order to ship that today in 8.13, 

👍 I added the backport 8.13 label

PR #2565, opened 2024-05-22
OpenAPI output: add aliases and support for non exhaustive enums

swallez:
Improve the OpenAPI output for enumerations:
* add aliases to the list of allowed strings
* implement non-exhaustive enumerations as an any(<enum members>, <any string>). Keeping the enum values in the schema should allow tools to provide them in autocompletion while allowing any value.

Also updates Cargo dependencies (in particular wasm-bindgen) to fix https://github.com/rustwasm/wasm-bindgen/pull/3808


Anaethelion (approved):
LGTM!

PR #2577, opened 2024-05-24
Fix deserialization for PutSynonymResponse

GeorgeCiachir:
The ReloadResult should be a member of the PutResponseSynonym, instead of the ReloadDetails, so that the PutResponseSynonym is properly serialized.
Current behavior is that it throws a [MissingRequiredPropertyException](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/8.13/missing-required-property.html)

Closes https://github.com/elastic/elasticsearch-java/issues/784
Closes #2578

PR #2555, opened 2024-05-13
Add ES|QL endpoint for Serverless

luigidellaquila:
Fixes https://github.com/elastic/elasticsearch-specification/issues/2554

Adding ES|QL endpoint to Serverless specification

PR #2580, opened 2024-05-27
Fail explicitly with pull requests opened from forks

pquentin:
This is a common problem, see #2555 and #2577 for recent examples. But not *that* common, and I tend to forget about it after a few months. Indeed, the current condition does not actually skip validation, and we don't want that anyway: we want to fail when we should have run validation but could not do it. And explain *why* we failed.

If the pull request was from an Elastic employee (95% of cases), then they can reopen it from a branch. If it was community-contributed, we need to do it for them (without closing the original pull request, thus preserving attribution when merging).

From a branch, the build was correctly triggered: https://github.com/elastic/elasticsearch-specification/actions/runs/9250931808/job/25445405796?pr=2580 (and failed as expected due to the extra file I have now removed).

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Check pull request was opened from branch
-        if: ${{ github.event.pull_request.head.repo.full_name != 'elastic/elasticsearch-specification' }}
+        if: github.event.pull_request.head.repo.full_name != 'elastic/elasticsearch-specification'
         run: echo "Validation is not supported from forks"; exit 1
         
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM!

In the future, we could use the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) trigger instead.

This requires a little bit of extra work as we would have to check-out the user branch, copy only the specification (`*.ts`) files to the base branch and run the checks from here.

PRs that change the compiler itself can not be checked this way for security reasons, but it should be really really rare for an outside collaborator to change the compiler code.

pquentin:
I confirmed in #2580 that we now get a failure with the right message. And that from branches, validations does run. Merging.

image

PR #2599, opened 2024-06-05
Add `IcuCollationProperty`

flobernd:
Related to: https://github.com/elastic/elasticsearch-net/issues/8212

Server code:
https://github.com/elastic/elasticsearch/blob/c5892a7044b51012e5680346c3bf7e03a70ce70d/plugins/analysis-icu/src/main/java/org/elasticsearch/plugin/analysis/icu/ICUCollationKeywordFieldMapper.java#L221

### DIFF
--- a/specification/_types/analysis/icu-plugin.ts
+++ b/specification/_types/analysis/icu-plugin.ts
@@ -51,16 +51,16 @@ export class IcuFoldingTokenFilter extends TokenFilterBase {
 export class IcuCollationTokenFilter extends TokenFilterBase {
   type: 'icu_collation'
   alternate?: IcuCollationAlternate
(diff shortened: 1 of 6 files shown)


specification/_types/analysis/icu-plugin.ts
```diff
@@ -51,16 +51,16 @@ export class IcuFoldingTokenFilter extends TokenFilterBase {
 export class IcuCollationTokenFilter extends TokenFilterBase {
```
flobernd, line 51: Just noticed that the `IcuCollationTokenFilter` field names do not seem to be correct. Let`s fix them as well while we are on it...

l-trotta (approved):
LGTM!

PR #2600, opened 2024-06-05
Improve `.gitattributes`

flobernd:
Especially improves cross-platform compatibility for Windows by defining a sane LF vs CRLF handling.


miguelgrinberg (approved):

.gitattributes
```diff
@@ -1,3 +1,239 @@
+### Header #########################################################################################
+
+# Author: Florian Bernd
```
pquentin, line 3: Is this needed?

pquentin (approved):
Thanks! Does this come from an existing template?

.gitattributes
flobernd: Ah yes, I copied it from one of my personal templates. We can of course remove the header 😋

flobernd, line 3:
```suggestion
```

PR #2602, opened 2024-06-05
[ML] Update Delete Inference Endpoint for dry_run and force options

maxhniebergall:
Following on  https://github.com/elastic/elasticsearch/pull/109123 and https://github.com/elastic/elasticsearch/pull/109384, this change updates the elasticsearch specification to include the dry_run and force options for the Delete Inference Endpoint API.

### DIFF
--- a/specification/inference/_types/Results.ts
+++ b/specification/inference/_types/Results.ts
@@ -19,7 +19,7 @@
 
 import { float, byte, integer } from '@_types/Numeric'
 import { Dictionary } from '@spec_utils/Dictionary'
-import {AcknowledgedResponseBase} from "@_types/Base";
+import { AcknowledgedResponseBase } from '@_types/Base'
 
 /**
  * Sparse Embedding tokens are represented as a dictionary
@@ -92,5 +92,5 @@ export class InferenceResult {
  * Acknowledged response. For dry_run, contains the list of ingest processors which reference the inference endpoint
  */
 export class DeleteInferenceEndpointResult extends AcknowledgedResponseBase {
-    ingest_processors: Array<string>
+  ingest_processors: Array<string>
 }
--- a/specification/inference/_types/Results.ts
+++ b/specification/inference/_types/Results.ts
@@ -19,6 +19,7 @@
 
 import { float, byte, integer } from '@_types/Numeric'
 import { Dictionary } from '@spec_utils/Dictionary'
+import {AcknowledgedResponseBase} from "@_types/Base";
 
 /**
  * Sparse Embedding tokens are represented as a dictionary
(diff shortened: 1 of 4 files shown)


specification/inference/delete/DeleteRequest.ts
```diff
@@ -33,9 +33,22 @@ export interface Request extends RequestBase {
      * The task type
      */
     task_type?: TaskType
+
     /**
      * The inference Id
      */
     inference_id: Id
+
```
davidkyle, line 41:
```suggestion
query_parameters: {
```

These should be in query_parameters rather than path parts

specification/inference/delete/DeleteResponse.ts
```diff
@@ -17,8 +17,10 @@
  * under the License.
  */
 
-import { AcknowledgedResponseBase } from '@_types/Base'
+import { DeleteInferenceEndpointResult } from '@inference/_types/Results'
 
 export class Response {
-  body: AcknowledgedResponseBase
+  body: {
+    dryRunResult: DeleteInferenceEndpointResult
```
davidkyle, line 24: What does `dryRunResult` mean here? Should it be `body: DeleteInferenceEndpointResult` ?

specification/inference/_types/Results.ts
```diff
@@ -85,3 +87,10 @@ export class InferenceResult {
   completion?: Array<CompletionResult>
   rerank?: Array<RankedDocument>
 }
+
+/**
+ * Acknowledged response. For dry_run, contains the list of ingest processors which reference the inference endpoint
+ */
+export class DeleteInferenceEndpointResult extends AcknowledgedResponseBase {
+  ingest_processors: Array<string>
```
jonathan-buttner, line 95: I think the field is `pipelines` right? That's what I'm seeing in the response:

```
{
    "acknowledged": false,
    "pipelines": [
        "elser"
    ]
}
```

specification/inference/delete/DeleteResponse.ts
```diff
@@ -17,8 +17,10 @@
  * under the License.
  */
 
-import { AcknowledgedResponseBase } from '@_types/Base'
+import { DeleteInferenceEndpointResult } from '@inference/_types/Results'
 
 export class Response {
-  body: AcknowledgedResponseBase
+  body: {
```
jonathan-buttner, line 23: Hmm I think this should be:

```
export class Response {
  body: DeleteInferenceEndpointResult
}
```

I don't believe there's a `dryRunResult` field in the response.

jonathan-buttner (changes requested):

maxhniebergall:
Clearly I don't fully understand what the correspondence between the java code and the specification is. I'm going to do some reading and come back to this.

PR #2608, opened 2024-06-06
Splitting summary and description in OpenAPI output

l-trotta:
Requirements: `summary` should contain the first sentence of `description`, while `description` should contain the remaining phrases.


compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +313,25 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
     result
 }
 
+fn split_summary_desc(desc: String) -> SplitDesc{
```
swallez, line 316: Nit: this function doesn't need ownership of its parameter as it will create new strings. Changing the parameter to `&str` will avoid the caller to `clone()` the parameter.

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +313,25 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
...
+    let mut parts = desc.split(['.','\n',':']);
```
swallez, line 317: I don't think we should split on `:` as the description can become a bit weird.

We only have one occurrence of this: "Returns information about indices: number of primaries and replicas, document counts, disk size, ...". The entire sentence is ok IMHO to use as the summary.

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +313,25 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
...
+    let first_line = parts.next().unwrap_or_else(|| "");
```
swallez, line 318: You can use `desc.split_once()` that will return `Option<&str, &str>` which can greatly simplify the logic when there's only one sentence.

Also nit (running `cargo clippy` will show it): in that case you can use `.unwrap_or("")`: you don't need the closure since it returns a constant value. `unwrap_or_else()` is only useful when the "else" is an expression that we want to evaluate only when needed.

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +313,25 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
...
+    
+    let new_desc = desc.replace(first_line,"");
+    let trim = new_desc.trim();
+    let remove_period = trim.strip_prefix('.').unwrap_or_else(|| trim);
+    let remove_column = remove_period.strip_prefix(':').unwrap_or_else(|| remove_period);
```
swallez, line 323: Nit 1: you can simplify things by using the same pattern as in `split()` above: `trim.strip_prefix(['.', ':']`).

Nit 2: making the variable mutable avoids declaring lots of additional variables:
```rust
let mut new_desc = ... // second part of `split_once` if it succeeded
let new_desc = new_desc.trim();
let new_desc = new_desc.strip_prefix('.').unwrap_or(new_desc);
let new_desc = new_desc.strip_prefix(':').unwrap_or(new_desc);

```

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +313,25 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
...
+    SplitDesc {
+        summary: String::from(first_line.trim()),
+        description: String::from(remove_column.trim())
+    }
+}
+
+struct SplitDesc {
+    summary: String,
+    description: String
```
swallez, line 332: These should be `Option<String>`. There are lots of single-sentence descriptions in `schema.json` that result in a description that is `Some("")` while it should be `None` and not output in the resulting OpenAPI schema.

swallez (changes requested):
I left lots of nitpicking comments to improve your Rust-fu 😉

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +314,33 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
     result
 }
 
+fn split_summary_desc(desc: &str) -> SplitDesc{
```
Anaethelion, line 317: Can you add a quick comment to explain the logic ?

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -311,6 +314,33 @@ fn get_path_parameters(template: &str) -> Vec<&str> {
...
+        description: if !rest.is_empty() {Option::from(String::from(rest.trim()))} else {None}
+    }
+}
+
+struct SplitDesc {
+    summary: Option<String>,
+    description: Option<String>
+}
+
 #[cfg(test)]
```
Anaethelion, line 344: We should also add some tests to check the expected behavior.

Anaethelion (commented):
Left some comments, looks cleaner with the segmenter.
Some of @swallez comments still applies though.

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -325,4 +361,28 @@ mod tests {
...
+        assert_eq!(split_summary_desc("These are two totally. Separate sentences!"),
+                   SplitDesc{
+                       summary: Some(String::from("These are two totally")),
+                       description: Some(String::from("Separate sentences!"))
+                   });
+        assert_eq!(split_summary_desc(""),
+                   SplitDesc{
+                       summary: None,
+                       description: None
+                   });
```
swallez, line 386: There's a pathological case that isn't handled (I've seen a few of them IIRC):

```rust
assert_eq!(split_summary_desc("These are two totally\n\nSeparate sentences!"),
    SplitDesc {
        summary: Some(String::from("These are two totally")),
        description: Some(String::from("Separate sentences!"))
    }
);
```

This can be fixed by using `let desc_no_newlines = desc.replace("\n\n",".\n").replace('\n'," ");`

swallez (approved):
LGTM once we handle the additional pathological case.

PR #2610, opened 2024-06-10
[route_name] Add serverless

Anaethelion:
This adds a target to generate a serverless schema and the relevant output to dump the serverless routes to the Go map.


pquentin (approved):
Thanks! LGTM. I confirmed with `dump-routes --debug` that serverless has less routes than 8.

PR #2547, opened 2024-05-07
Serverless role and privilege model updates 

n1v0lg:
Brings back https://github.com/elastic/elasticsearch-specification/pull/2491 with a few corrected annotations. 

The changes included in this PR are ready to be published: the privileges pertain to the API key APIs which are already public in Serverless. However, the visibility of the Create or Update Roles API (as well as other Role-related APIs) remains private until the feature is ready. I will open a separate PR for those when the time has come.

PR #2611, opened 2024-06-10
Remove duplicate privilege

n1v0lg:
`read_slm` was defined twice in https://github.com/elastic/elasticsearch-specification/pull/2547. Removing the dup.


pquentin (approved):
Thanks! LGTM.

PR #2619, opened 2024-06-18
Attach generic type parameters to the declaring type

flobernd:
Attach generic type parameters to the declaring type instead of the parent namespace.

This more precisely reflects the actual declaration hierarchy and allows type parameters to be uniquely identified by their FQNs.

The following definition:

```ts
export class A<T> { }
```

caused a type parameter with the FQN `some.namespace.T` to get emitted.

The same type parameter is now emitted as `some.namespace.A.T` instead.

### DIFF
--- a/compiler/src/model/json-spec.ts
+++ b/compiler/src/model/json-spec.ts
@@ -67,12 +67,13 @@ export default function buildJsonSpec (): Map<string, JsonSpec> {
   const files = readdirSync(jsonSpecPath)
     .filter(file => file.endsWith('.json'))
 
-  const map = new Map()
+  const map: Map<string, JsonSpec> = new Map()
   for (const file of files) {
     const json = require(join(jsonSpecPath, file))
     const name = Object.keys(json)[0]
     map.set(name, json[name])
   }
 
-  return map
+  // Ensure deterministic ordering
+  return new Map([...map.entries()].sort((a, b) => a[0].localeCompare(b[0])))
 }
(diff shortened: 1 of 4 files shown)


compiler/src/model/utils.ts
```diff
@@ -343,6 +343,20 @@ export function modelType (node: Node): model.ValueOf {
...
           }
+
+          if (Node.isTypeParameterDeclaration(declaration)) {
+            const parent = declaration.getParent()
+            assert(
+              parent,
+              Node.isClassDeclaration(parent) ||
+              Node.isInterfaceDeclaration(parent) ||
+              Node.isTypeAliasDeclaration(parent),
+              'It should be a class, interface, enum, type alias, or type parameter declaration'
```
l-trotta, line 354: comment needs to be changed too

compiler/src/model/json-spec.ts
```diff
@@ -74,5 +74,6 @@ export default function buildJsonSpec (): Map<string, JsonSpec> {
     map.set(name, json[name])
   }
 
-  return map
+  // Ensure deterministic ordering
+  return new Map([...map.entries()].sort())
```
flobernd, line 78: Maps in JS are iterated in the order of insertion. This map is created by enumerating the filesystem which causes the entries to be inserted in a different order on Linux and Windows.

swallez (approved):
LGTM.

@Anaethelion @JoshMock @l-trotta make sure this doesn't break anything in your respective clients. It should not, as it only changes the namespace of generic parameters, not their names.

The rationale behind this change is that it ensures global uniqueness of generic parameter names, even if their scope is limited to the enclosing type definition, which can help when processing a global type graph.

l-trotta (approved):

Anaethelion (approved):

JoshMock (approved):

PR #2622, opened 2024-06-18
Add semantic_text and semantic query

carlosdelest:
Add support for the upcoming [semantic_text](https://www.elastic.co/guide/en/elasticsearch/reference/master/semantic-text.html) and [semantic query](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-semantic-query.html).

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -5338,7 +5338,6 @@ export interface MappingSemanticTextProperty {
   type: 'semantic_text'
   meta?: Record<string, string>
   inference_id: Id
-  copy_to?: Fields
 }
 
 export interface MappingShapeProperty extends MappingDocValuesPropertyBase {
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -5080,7 +5080,7 @@ export interface MappingFieldNamesField {
   enabled: boolean
 }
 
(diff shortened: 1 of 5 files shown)


l-trotta:
@carlosdelest from the [server code](https://github.com/elastic/elasticsearch/blob/d5488ad78dd615fb46f791ae09355c8ed768c022/x-pack/plugin/inference/src/main/java/org/elasticsearch/xpack/inference/mapper/SemanticTextFieldMapper.java#L311) it seems like there are some other parameters for the Semantic Text property. Are all of those internal? Like `model_settings` for example

carlosdelest:
hey @l-trotta ! Good catch - `model_settings` is used internally and should not be set by the user, but internally from the mapping itself when a document is added.

I believe these are the properties exposed by the [field type](https://www.elastic.co/guide/en/elasticsearch/reference/master/semantic-text.html).

l-trotta:
Thanks for the clarification, just making sure :D 
What about `copy_to`? From the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/master/semantic-text.html#copy-to-support) it looks like it's accepted, but then the [server code](https://github.com/elastic/elasticsearch/blob/227034e64d59c484418a4c9fb4cccb4a5a74fffc/x-pack/plugin/inference/src/main/java/org/elasticsearch/xpack/inference/mapper/SemanticTextFieldMapper.java#L149) says otherwise. Tested this in the devtools:
```
PUT my-index-000003
{
  "mappings": {
    "properties": {
      "inference_field": {
        "type": "semantic_text",
        "inference_id": "my-elser-endpoint",
        "copy_to": "field"
        }
    }
  }
}
``` 
returns:
```
{
  "error": {
    "root_cause": [
      {
        "type": "mapper_parsing_exception",
        "reason": "Failed to parse mapping: semantic_text field [inference_field] does not support [copy_to]"
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "Failed to parse mapping: semantic_text field [inference_field] does not support [copy_to]",
    "caused_by": {
      "reason": "semantic_text field [inference_field] does not support [copy_to]"
    }
  },
  "status": 400
}
```

carlosdelest:
My bad @l-trotta ! semantic_text can be the _target_ of copy_to, but not have copy_to fields itself. Thanks for the catch, fixed it!

l-trotta (approved):
LGTM! thank you

Issue #2621, opened 2024-06-18
Get ML request validation to 100%


pquentin:
For ML requests (not responses), 97.9% or the YAML tests pass type validation. Passing means that the specification is enough to type the request, which means that all clients using the validation will be able to send the request. If we get to 100%, then we can start treating any regression as errors, to hopefully stay at 100%.

Here are all the current issues, sorted by API. Sometimes the issue is in our tooling, so I'd appreciate if @elastic/ml-core and/or @elastic/ml-ui could verify what I'm saying here, and ideally send fixes. Thank you!

* get_trained_models
  * include_model_definition is missing
* post_data
  * body is specified as Array of data, but actual type is just a string that is somehow parsed afterwards?
* preview_datafeed
  * The datafeed config accepts "indices" according to docs but 7 YAML tests send "indexes". [Example](https://github.com/elastic/elasticsearch/blob/e66cb6c632dd164a3810154ce00c8b3a3c0e8b5e/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/ml/preview_datafeed.yml#L137)
* put_data_frame_analytics
  * `body._meta` is missing
* put_datafeed
  * The datafeed config accepts "aggregations" according to docs (and spec) but "Test put datafeed with aggregations" test uses "aggs"
* put_job
  * `body._job_id` is missing
  * `body.analysis_limits.model_memory_limit` can be sent as an integer
  * `body.ignore_throttled` is missing from docs and spec. In docs, it's supposed to only be possible inside `indices_options` 
* put_trained_model
  * `bert_ja` used by "Test put model config with Japanese tokenizer" missing from https://github.com/elastic/elasticsearch-specification/blob/33e044d31224bd96da660e99fdb0af668b6404f3/specification/ml/_types/inference.ts#L114-L129
* update_job
  * description is missing from https://github.com/elastic/elasticsearch-specification/blob/33e044d31224bd96da660e99fdb0af668b6404f3/specification/ml/_types/Detector.ts#L25-L67

Also, `put_trained_model_definition_part` and `put_trained_model_vocabulary` don't have any known YAML tests.

And finally, there are issues caused by our recording of tests that split all comma-separated into lists which allows validating things like `expand_wildcards` but breaks for `job_id` and `model_id`. I'll fix that.

PR #2623, opened 2024-06-19
Fix errors around arrays in ML requests

pquentin:
Relates #2621 

When query parameters accept comma-separated strings, we should type them as `string | string[]` (which is what `Ids` and `Indices` are), so that clients can offer nicer list-based APIs. In fact, it's currently mandated for validation to pass, and I did not realize it when listing issues in #2621, so I am fixing those unlisted issues.

This fixes `ml.put_calendar_job`, fixes `ml.preview_datafeed` and improves `ml.get_trained_models` (fixed 5 examples, one example left).


l-trotta (approved):
LGTM!

PR #2638, opened 2024-06-21
Maintain "behavior" meta-data in specification

### LINKED ISSUE #2637: Proposal: Maintain "behavior" meta-data in specification
## Problem

In the current state it's not possible to generate good code for the `AdditionalProperty` and `AdditionalProperties` behaviors without hardcording supplementary meta-data in the generators.

### AdditionalProperty

```ts
export class TermsQuery
  extends QueryBase
  implements AdditionalProperty<Field, TermsQueryField> {}
```

To generate named fields for the additional property "name" and "value" parts, code generators must map the generic type arguments to property names like e.g.:

```
{ "Field" , "field" }
{ "TermsQueryField" , "terms" }
```

### AdditionalProperties

```ts
export class ErrorCause
  implements AdditionalProperties<string, UserDefinedValue>
{
```

The mapping for `AdditionalProperties` is based on the type of the implementing class and (in most generators) includes a description string:

```
{ "ErrorCause" , ("metadata", "Additional details about the error") }
```

### Conclusion

Spreading the meta-data all over the place in different generators makes it very hard to keep it on sync and to maintain this meta-data in general.

### Specification

Similar to the existing `@codegen_name` and `@codegen_names` decorators, we should introduce a mechanic that allows us to annotate the `AdditionalProperty` and `AdditionalProperties` behaviors with the required codegen related meta-data.

A generic approach to provide meta-data for arbitrary behaviors could look like this:

```ts
/** @behavior_meta {name} {meta1}={value1} {meta2}={value2} ... {metaN}={valueN} */
```

For example:

```ts
/** @behavior_meta AdditionalProperty name=field value=terms */
export class TermsQuery
  extends QueryBase
  implements AdditionalProperty<Field, TermsQueryField> {}
```

```ts
/** @behavior_meta AdditionalProperty name=metadata description="Additional details about the error" */
export class ErrorCause
  implements AdditionalProperties<string, UserDefinedValue>
{
```

### Metamodel

In our metamodel, behaviors are mapped to `Inherits` objects:

```ts
export class Inherits {
  type: TypeName
  generics?: ValueOf[]
}
```

A new field must be added to hold the meta-data information:

```ts
export class Inherits {
  type: TypeName
  generics?: ValueOf[]
  meta: Map<string, string>
}
```

cc @elastic/devtools-team

flobernd:
Adds a new `behavior_meta` decorator that allows to specify supplementary meta-data per implemented behavior.

Closes #2637

### DIFF
--- a/compiler/src/model/build-model.ts
+++ b/compiler/src/model/build-model.ts
@@ -459,7 +459,8 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
       properties: new Array<model.Property>()
     }
 
-    hoistTypeAnnotations(type, declaration.getJsDocs())
+    const jsDocs = declaration.getJsDocs()
+    hoistTypeAnnotations(type, jsDocs)
 
     const variant = parseVariantNameTag(declaration.getJsDocs())
     if (typeof variant === 'string') {
@@ -532,7 +533,7 @@ function compileClassOrInterfaceDeclaration (declaration: ClassDeclaration | Int
     if (Node.isClassDeclaration(declaration)) {
       for (const implement of declaration.getImplements()) {
         if (isKnownBehavior(implement)) {
-          type.behaviors = (type.behaviors ?? []).concat(modelBehaviors(implement))
+          type.behaviors = (type.behaviors ?? []).concat(modelBehaviors(implement, jsDocs))
         } else {
           type.implements = (type.implements ?? []).concat(modelImplements(implement))
         }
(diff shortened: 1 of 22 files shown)


pquentin (approved):
While I don't think this affects any top-level types and thus does not affect the generator I work on, this makes sense to me! LGTM.

JoshMock (approved):
This shouldn't have any effect on the JS generator, but I do see some possible opportunities to use behavior metadata to simplify some code. :raised_hands:

PR #2639, opened 2024-06-21
[Query Rules] Update Query Rules API for 8.15

kderusso:
The following changes are going live in 8.15 for query rules: 

- The feature is moving from Technical Preview to GA 
- The `rule_query` has been renamed to `rule` and support for multiple rulesets was added in https://github.com/elastic/elasticsearch/pull/108831
- A CRUD API for individual rule updates was added in https://github.com/elastic/elasticsearch/pull/109042 and https://github.com/elastic/elasticsearch/pull/109554

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -24523,6 +24523,183 @@
         }
       }
     },
+    "/_query_rules/{ruleset_id}/_rule/{rule_id}": {
+      "get": {
+        "tags": [
+          "query_rule.get"
+        ],
+        "summary": "Returns the details about a query rule within a query ruleset",
+        "externalDocs": {
+          "url": "https://www.elastic.co/guide/en/elasticsearch/reference/master/get-query-rule.html"
+        },
+        "operationId": "query-rule-get",
+        "parameters": [
+          {
+            "in": "path",
+            "name": "ruleset_id",
+            "description": "The unique identifier of the query ruleset containing the rule to retrieve",
+            "required": true,
+            "deprecated": false,
+            "schema": {
+              "$ref": "#/components/schemas/_types:Id"
+            },
+            "style": "simple"
+          },
+          {
+            "in": "path",
+            "name": "rule_id",
+            "description": "The unique identifier of the query rule within the specified ruleset to retrieve",
+            "required": true,
+            "deprecated": false,
+            "schema": {
+              "$ref": "#/components/schemas/_types:Id"
+            },
+            "style": "simple"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "",
+            "content": {
+              "application/json": {
+                "schema": {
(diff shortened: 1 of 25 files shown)


l-trotta:
the spec is missing the [Always criteria type](https://github.com/elastic/elasticsearch/blob/28d7743a26497eda1361636dd545c292abb8af2c/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/rules/QueryRuleCriteriaType.java#L19), and if I got this correctly metadata [here](https://github.com/elastic/elasticsearch-specification/blob/ce9d54c65d2c588b2576ab24cfaaa74f7281b37d/specification/query_ruleset/_types/QueryRuleset.ts#L48) should be optional.

l-trotta:
QueryRulesetListItem probably has to be updated as well, [server](https://github.com/elastic/elasticsearch/blob/57f99e6cc3a3f9f2343f00841d5981f1ce850d16/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/rules/QueryRulesetListItem.java#L28), [current spec](https://github.com/elastic/elasticsearch-specification/blob/ce9d54c65d2c588b2576ab24cfaaa74f7281b37d/specification/query_ruleset/list/types.ts#L22)

kderusso:
@l-trotta thanks for the review! 

> the spec is missing the [Always criteria type](https://github.com/elastic/elasticsearch/blob/28d7743a26497eda1361636dd545c292abb8af2c/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/rules/QueryRuleCriteriaType.java#L19), and if I got this correctly metadata [here](https://github.com/elastic/elasticsearch-specification/blob/ce9d54c65d2c588b2576ab24cfaaa74f7281b37d/specification/query_ruleset/_types/QueryRuleset.ts#L48) should be optional.

Thanks for the catch, fixed!

> QueryRulesetListItem probably has to be updated as well, [server](https://github.com/elastic/elasticsearch/blob/57f99e6cc3a3f9f2343f00841d5981f1ce850d16/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/rules/QueryRulesetListItem.java#L28), [current spec](https://github.com/elastic/elasticsearch-specification/blob/ce9d54c65d2c588b2576ab24cfaaa74f7281b37d/specification/query_ruleset/list/types.ts#L22)

Updated!

l-trotta:
one last thing! in QueryRulesetListItem, `rules_count` should be `rule_total_count` and `rule_criteria_type_counts` should be `rule_criteria_types_counts`

pquentin:
(I sound like a broken record, but those would be easier to catch if we could collect the Query Rules YAML tests.)

kderusso:
> (I sound like a broken record, but those would be easier to catch if we could collect the Query Rules YAML tests.)

I know, I promise it's on my list to look into!

l-trotta (approved):
LGTM!

PR #2640, opened 2024-06-24
Populate endpoint deprecation from request definition

flobernd:
Populates the `deprecation` field for endpoints from the value of the corresponding request definition.

Currently, the `deprecation` info is only available on the request type, but not on the corresponding endpoint.


pquentin (approved):
This looks great, thanks.

PR #2642, opened 2024-06-24
Add specification for new sparse vector query; deprecate specification for text_expansion and weighted_tokens queries

kderusso:
Adds specification for the new `sparse_vector` query and deprecates the specifications for `text_expansion` and `weighted_tokens` queries.

### DIFF
--- a/compiler/src/index.ts
+++ b/compiler/src/index.ts
@@ -27,6 +27,7 @@ import addDescription from './steps/add-description'
 import validateModel from './steps/validate-model'
 import addContentType from './steps/add-content-type'
 import readDefinitionValidation from './steps/read-definition-validation'
+import addDeprecation from './steps/add-deprecation'
 
 const nvmrc = readFileSync(join(__dirname, '..', '..', '.nvmrc'), 'utf8')
 const nodejsMajor = process.version.split('.').shift()?.slice(1) ?? ''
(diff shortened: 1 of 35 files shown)


l-trotta:
if I got this right every sparse query should have either `query_vector` or `inference_id`, the way we model this to have it checked client side is to add `/** @variants container */` on top of the class and mark all properties that can always be added with `@variant container_property`, an example of this is [PinnedQuery](https://github.com/elastic/elasticsearch-specification/blob/4a3a7aee6783a8fe9e6e2cb1dc07da4aeb6773ea/specification/_types/query_dsl/specialized.ts#L235) where you must set either `ids` or `docs`

kderusso:
> if I got this right every sparse query should have either query_vector or inference_id, the way we model this to have it checked client side is to add /** @variants container */ on top of the class and mark all properties that can always be added with @variant container_property, an example of this is [PinnedQuery](https://github.com/elastic/elasticsearch-specification/blob/4a3a7aee6783a8fe9e6e2cb1dc07da4aeb6773ea/specification/_types/query_dsl/specialized.ts#L235) where you must set either ids or docs

Thanks @l-trotta! I think I got this right?

l-trotta:
almost! all properties that aren't `query_vector` or `inference_id` should be `@variant container_property`, so not just `field`. basically this is to assure that either `query_vector` or `inference_id` are set, but not both of them, while all the other properties are unaffected by this check.

l-trotta (approved):
LGTM!

PR #2646, opened 2024-06-25
Fix 'Non-leaf type' errors in xpack

pquentin:
And also remove unused x_pack special cases. The namespace is named xpack now, so they did not have any effect.

This fixes around 25 validation errors. More generally, I'm not sure how useful those "Non-leaf type cannot be used here" errors really are. It's pretty common to use a class in a property and then subclass it to add more things, and then use that too.


pquentin:
Closed in favor of #2648.

PR #2648, opened 2024-06-25
Remove 'Non leaf-type cannot be used here' rule

pquentin:
It has too many false positives to be useful in practice.


flobernd (approved):

swallez (approved):
LGTM!

PR #2656, opened 2024-06-26
Merge `query_rule` and `query_ruleset` namespaces into `query_rules`

kderusso:
Moves the `query_rule` and `query_ruleset` namespaces into a single top level `query_rules` namespace. 

Followup to https://github.com/elastic/elasticsearch-specification/pull/2639


pquentin (commented):
Thanks for this! I believe having a single namespace is much better! I'm only wondering about the endpoint names defined in the JSON spec, as they are used in client calls. With this PR, taking `get` as an example, here's how users will call those APIs:

| Client call | API call |
| - | - |
|  `client.query_rules.query_ruleset_get(...)` | `GET /query_rules/{ruleset_id}`
| `client.query_rules.query_rule_get(...)` | `GET /query_rules/{ruleset_id}/_rule/{rule_id}` |

The two issues I'm seeing are:

 1. Other APIs put the verb first, eg. `indices.put_mapping` or `inference.put_model`
 2. There's a repetition around `query_rule`

What are your thoughts around using those names instead?

* `client.query_rules.get_ruleset(...)`
* `client.query_rules.get_rule(...)`

kderusso:
> What are your thoughts around using those names instead?
>
> client.query_rules.get_ruleset(...)
> client.query_rules.get_rule(...)

😅 Yeah, those are cleaner, I'll update both PRs today.

pquentin (approved):
Thanks! LGTM. I had to run `git merge` locally as GitHub reported conflicts. (The actual `git merge` did not have any conflicts.) And I ran `make contrib` to fix the OpenAPI specs.

PR #2660, opened 2024-06-28
Set fail-fast: false on update-rest-api workflow

pquentin:
Otherwise, as soon as a job like 7.17 fails, everything else is canceled. But they are independent, and 1/ any success is welcome 2/ understanding fully all failures is welcome too.


flobernd (approved):
LGTM! Good change.

PR #2671, opened 2024-07-02
Fix most esql.query request types

pquentin:


specification/esql/query/QueryRequest.ts
```diff
@@ -51,12 +57,17 @@ export interface Request extends RequestBase {
...
+     * @doc_id esql-returning-localized-results
+     */
     locale?: string
     /**
      * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
      * @doc_id esql-query-params
      */
-    params?: Array<ScalarValue>
+    params?: Array<EsqlParam>
+    profile?: boolean
```
pquentin, line 70: This one is undocumented but used in the YAML tests, which is why I'm adding without details.

pquentin:
With `make validate`, I get `✔ 174 out of 174 test request cases are passing.`. I'll need to understand what is going on.

pquentin:
So I get `224/227` now (both locally and in CI thanks to #2690!), and all the remaining failures are related to `LOOKUP` and its new `tables` parameter that I'd like to tackle later.

specification/esql/query/QueryRequest.ts
```diff
@@ -51,12 +57,17 @@ export interface Request extends RequestBase {
...
+     * Returns results (especially dates) formatted per the conventions of the locale.
+     * @doc_id esql-returning-localized-results
+     */
     locale?: string
     /**
      * To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
      * @doc_id esql-query-params
      */
-    params?: Array<ScalarValue>
+    params?: Array<UserDefinedValue>
```
pquentin, line 69: I'm not sure about this. It allows values like `{"foo": "bar"}`, which sounds difficult to type?

Anaethelion: We should go for `Array<ScalarValue | UserDefinedValue>`, which can be simplified to `Array<FieldValue>`.

pquentin: Thanks, done!

l-trotta (approved):

PR #2675, opened 2024-07-04
[Connector APIs] Update existing endpoints

jedrazb:
Hey 👋 

As discussed, I reviewed the existing Connector API endpoints and added patches to update them to the `8.15` state.

### Changes
- `index_name` and `error` (including `last_sync_error` and `last_access_control_sync_error`) are the only properties that use semantic null. `index_name` can be set to null to detach the connector, error can be reset to `null` to indicate healthy state again. We are using optional for everything else.
- Add `sync_cursor` field in the Connector definition (we missed it before) and add this as an arg to `_last_sync` endpoint (new payload field added in 8.15)
- Updated the `ConnectorFeatures` representation as it included long-deprecated feature names (`filtering_advanced_config` and `filtering_rules`), add missing `native_connector_api_keys`

### Note
We have this open [issue](https://github.com/elastic/search-team/issues/7835) with `_last_sync` endpoint (confusion between semantic null and optional when updating `last_sync_*_error `). We are looking into fixing this. 

Also, this PR should address existing endpoints, there are couple of endpoints that are still missing. Let's work on this patch first. :)


jedrazb:
Hey @l-trotta 👋  I see you assigned yourself to the issue. I tried addressing the concerns we discussed in the last call. Happy to answer any questions and address any other concerns!

pquentin:
Hey @jedrazb, this is good timing because we just managed to add Connectors YAML tests to the flight-recorder, so I reran validation and you can see the results in the above comment.

I can report that compared to the current state, this pull requests:

 * fixes `last_sync`, `post` and `update_name` requests
 * fixes `sync_job_get` responses

The rest is unchanged. I can open an issue listing the failures to not block this pull requests which definitely improves things.

jedrazb:
Hey @pquentin ! The changes that fix remaining issues should be minimal, I'm having issues running `make validate` locally due to vault permission issues. I will ask more in Slack!

jedrazb:
Now it looks better! Two issues:
- The issues around `connector.get/list` endpoint responses: 
  - Upon connector creation some fields get preset to `null` (to match the status quo from before). When handling the response, response definition is unhappy about `null` fields which are actually optional. If this doesn't break client we can probably ignore this until we fix this in ES (?)
- The issue with `connector.update_configuration`
  - In ES implementation we have a yaml test case where a crawler-specifc setting is updated via this endpoint. We should not be doing this - since we are not planning to manage crawler via this API. 
  - Another test failure is with literal `null` tooltip being passed in the body (we could make it just optional in server implementation)

I think we can address those issues as a followup.

l-trotta:
does this field exist only in the connector response? https://github.com/elastic/elasticsearch/blob/e1a091f1a2a7444873a1887c686186af580649e3/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/Connector.java#L211

l-trotta:
I'm starting to consider mapping the connector request and response in two different classes, as you can see from the validator:
`connector.get` | :red_circle: | 59/59 | 1/59 |  
this is because in the response many types are not optional, for example 
`api_key_id?: string` in the request
is actually 
`api_key_id: string | null` in the response

jedrazb:
Hey @l-trotta 👋 As we discussed in the call, I'm updating the spec to address main issues:

> I'm starting to consider mapping the connector request and response in two different classes
++ Actually `Connector` type was only used to describe responses, so I updated the type accordingly. 

I had to put `tooltip` property as `string | null` for the automated tests to pass. We always return `tooltip` field from the ES (even if it's null). 

I marked `last_sync_error` and `last_access_control_sync_error` as just optional, until we address https://github.com/elastic/elasticsearch/pull/110543 in ES. 

There are just 2 failing test cases related to crawler state (in connector doc!). Due to historical reasons, some parts of the crawler state were, and still are, stored in the connector index. However, it was agreed that the connector API should not interact with the crawler, so those tests should be removed from ES. I don’t think these failures should block us in this PR, as this is a quirky edge case.

jedrazb:
> got it we'll wait for the yaml tests to be fixed!

Okok I see, let me get rid of those tests from the server first.

l-trotta (approved):

PR #2692, opened 2024-07-09
Remove `read_connector_secrets` cluster privilege

n1v0lg:
ES PR https://github.com/elastic/elasticsearch/pull/110414 removes `read_connector_secrets` -- this cluster privilege was introduced for an undocumented beta API. It was never meant to be user-facing, and instead meant to be used by an internal service account. Since there is no need for this privilege for end-users, and it grants access to an internal API only, we have decided to remove it altogether and rely on raw action names for the internal service account's role instead.

This PR removes the privilege from the spec.


l-trotta (approved):

l-trotta:
LGTM, just waiting for the validation to be available again

PR #2704, opened 2024-07-10
Java issues batch 7

l-trotta:
- [826](https://github.com/elastic/elasticsearch-java/issues/826): `language` in snowball token filter should be optional, server defaults to English. [server code](https://github.com/elastic/elasticsearch/blob/89113e9b3c2fc73b5d94213285197ecc39ae94b5/modules/analysis-common/src/main/java/org/elasticsearch/analysis/common/SnowballTokenFilterFactory.java#L28)
- [831](https://github.com/elastic/elasticsearch-java/issues/831): `nested` missing from geo distance sort. [server code](https://github.com/elastic/elasticsearch/blob/2a193b53d84b65395003e5be9cb5502c764be8e4/server/src/main/java/org/elasticsearch/search/sort/GeoDistanceSortBuilder.java#L90)
- [834](https://github.com/elastic/elasticsearch-java/issues/834): `unhealthy_policies` is optional in SlmIndicator, no hard proof found in the [server code](https://github.com/elastic/elasticsearch/blob/97650b02b9a90608f539fb19d8846e6dea5243ae/x-pack/plugin/slm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java#L203) except that it can be an empty collection, but it's easily verifiable by calling `GET _health_report` in a new cloud instance. 
- [835](https://github.com/elastic/elasticsearch-java/issues/835): AllocationRecord was missing some fields. [server code](https://github.com/elastic/elasticsearch/blob/97650b02b9a90608f539fb19d8846e6dea5243ae/server/src/main/java/org/elasticsearch/rest/action/cat/RestAllocationAction.java#L93), [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-allocation.html#cat-allocation-api-response-body). 
- [842](https://github.com/elastic/elasticsearch-java/issues/842): in TransformProgress, `total_docs` can be null, and if it is, `docs_remaining` and `percent_complete` are null too. [server code](https://github.com/elastic/elasticsearch/blob/4a77e062334b06b88ab437014a48b7e402ff4b2c/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/transform/transforms/TransformProgress.java#L152)

### DIFF
--- a/specification/_global/health_report/types.ts
+++ b/specification/_global/health_report/types.ts
@@ -160,7 +160,7 @@ export class SlmIndicator extends BaseIndicator {
 export class SlmIndicatorDetails {
   slm_status: LifecycleOperationMode
   policies: long
-  unhealthy_policies: SlmIndicatorUnhealthyPolicies
+  unhealthy_policies?: SlmIndicatorUnhealthyPolicies
 }
 
 export class SlmIndicatorUnhealthyPolicies {
--- a/specification/cat/allocation/types.ts
+++ b/specification/cat/allocation/types.ts
@@ -19,7 +19,7 @@
 
 import { ByteSize } from '@_types/common'
 import { Host, Ip } from '@_types/Networking'
-import { Percentage } from '@_types/Numeric'
+import { double, Percentage } from '@_types/Numeric'
 
 export class AllocationRecord {
   /**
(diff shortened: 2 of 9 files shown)


specification/cat/allocation/types.ts
```diff
@@ -19,14 +19,31 @@
...
   /**
    * Number of primary and replica shards assigned to the node.
    * @aliases s
    */
   shards?: string
+
+  /**
+   * Amount of shards that are scheduled to be moved elsewhere in the cluster or -1 other than desired balance allocator is used
+   */
+  'shards.undesired'?: string | null
```
flobernd, line 34: From the linked server-code it seems like these properties are always present. Should we remove the optional modifier?

specification/_global/health_report/types.ts
```diff
@@ -150,6 +150,7 @@ export class IlmIndicator extends BaseIndicator {
 export class IlmIndicatorDetails {
   ilm_status: LifecycleOperationMode
   policies: long
+  stagnating_indices: long
```
Anaethelion, line 153: nit: `stagnating_indices` is returned as a size so it should be an integer.

Anaethelion (commented):
Only left a nit!

l-trotta:
I think I'll try to turn those dots green in another PRs, there's too many types missing

PR #2706, opened 2024-07-10
ESQL: define the `tables` parameter

nik9000:
Adds support for the `tables` parameter and documents the `profile` option.


nik9000:
I'm not sure what's up with the failure here. It looks like it doesn't like:
```
    tables?: Dictionary<
      string,
      Dictionary<string, SingleKeyDictionary<TableValuesType, any>>
    >
```
The `SingleKeyDictionary<TableValuesType, any>` doesn't feel quite right, but I'm not sure how tightly you need it here. That could be: `{"long": [long|[long]]}`/`{"int": [int|[int]]}`/`{"keyword": [string|[string]]}`/`{"double": [double|[double]]}`. But I don't know how to write that.

flobernd:
I would model it like this:

```ts
export type TableValuesData = long | int | string | double
```

```ts
tables?: Dictionary<
     string,
     Dictionary<string, SingleKeyDictionary<TableValuesType, TableValuesData | TableValuesData[]>>
>
```

Btw: `any` is invalid in the specification (note the failed validation). Instead, the `UserDefinedValue` might be used.

nik9000:
> Btw: `any` is invalid in the specification (note the failed validation). Instead, the `UserDefinedValue` might be used.

Gotcha. I'll do it like you said without `any`. That's just better and not hard.

nik9000:
@flobernd have a look now!

nik9000:
> `esql.query`

Hey! I was testing that with `make validate api=esql.query type=request branch=main` locally. That does seem better. Though the response is defined as an `ArrayBuffer` so I think that should be skipped.

specification/esql/_types/TableValuesType.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+type TableValuesKeywordValue = string | string[]
+export interface TableValuesLong {
+  long: TableValuesLongValue[]
+}
+type TableValuesLongValue = long | long[]
+export interface TableValuesDouble {
+  double: TableValuesLongDouble[]
+}
+type TableValuesLongDouble = double | double[]
```
flobernd, line 37: Since these types are mutually exclusive, we should consider modelling them as a `container` like:

```ts
/** @variants container */
export interface TableValues
{
  integer: integer | integer[]
  keyword: string | string[]
  long: long | long[]
  double: double | double[]
}

File: specification/esql/query/QueryRequest.ts [hunk 2]
```diff
@@ -67,10 +74,29 @@ export interface Request extends RequestBase {
...
+     * and its format can change at any time but it can give some insight into the performance
+     * of each part of the query.
+     */
     profile?: boolean
     /**
      * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
      */
     query: string
+    /**
+     * Tables to use with the LOOKUP operation.
```
flobernd, line 89: We should probably document the dictionary keys a little bit more in detail. I assume the outer key is the table name and the inner one the column name?

specification/esql/_types/TableValuesType.ts
flobernd: Would that work with your code-generators @Anaethelion @l-trotta ?

Anaethelion: nit: for it to be a proper container all the fields should be optional.

To explain further the `@container` was designed primarily for `query` in search where the fields you set defines the type of query you use. Thus each [QueryContainer](https://github.com/elastic/elasticsearch-specification/blob/f737f34e18daa7626ac141171a005ca28372dada/specification/_types/query_dsl/abstractions.ts#L107) can only have one field populated. [link to modeling guide](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#container)

I think it fits the intent.

flobernd: @nik9000 

> I did it this way because you can't specify more than one key in the object. Is that what @container does?

Yes, the keys inside the container are mutually exclusive. I updated the above code with the missing optional `?` modifier.

flobernd: Oh sorry, I meant the code with the container example in the above comment.

specification/esql/query/QueryRequest.ts
nik9000: Done.

specification/esql/_types/TableValuesType.ts
nik9000: Cool. I've pushed some updates.

specification/esql/_types/TableValues.ts
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { double, integer, long } from '@_types/Numeric'
+
+/** @variants container */
+export interface TableValues {
```
flobernd, line 23:
```suggestion
export interface TableValueContainer {
```

Just a naming suggestion to be consistent with the existing specification, LGTM otherwise!

specification/esql/query/QueryRequest.ts
```diff
@@ -67,10 +69,21 @@ export interface Request extends RequestBase {
...
     profile?: boolean
     /**
      * The ES|QL query API accepts an ES|QL query string in the query parameter, runs it, and returns the results.
      */
     query: string
+    /**
+     * Tables to use with the LOOKUP operation. The top level key is the table
+     * name and the next level key is the column name.
+     */
+    tables?: Dictionary<string, Dictionary<string, TableValues>>
```
flobernd, line 87:
```suggestion
    tables?: Dictionary<string, Dictionary<string, TableValueContainer>>
```

specification/esql/_types/TableValues.ts
flobernd: (or `TableValuesContainer` - not sure if singular or plural makes more sense here)

nik9000: I think plural does. I'll make the change.

nik9000:
Renamed. I've marked this auto-merge because I think that's what we want. I *think* I'm waiting on a final review.

flobernd (approved):
LGTM! Thanks for the effort 🙂

nik9000:
Thanks for the feedback folks!

PR #2708, opened 2024-07-11
[Proposal] Transform script into a container

Anaethelion:
`@shortcut_property` on `InlineScript` is unreachable from the union.
From the server code, `Script` could benefit from being a container instead.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -58375,43 +58375,32 @@
         ]
       },
       "_types:Script": {
-        "allOf": [
-          {
+        "type": "object",
+        "properties": {
+          "source": {
+            "description": "The script source.",
+            "type": "string"
+          },
+          "id": {
+            "$ref": "#/components/schemas/_types:Id"
+          },
+          "params": {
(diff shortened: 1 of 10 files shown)


swallez (commented):
If we're going that route, I would go as far as not making it a container: `lang` and `options` only apply to inline script, so it's weird to allow `id` _xor_ `source` while having these container properties that will cause a server-side failure if set for the wrong variant.

Anaethelion:
I'm OK with merging everything in the same interface.
pinging @l-trotta @flobernd for visibility!

flobernd:
I like the container solution a little bit better over the single interface, but no strong preference here. Fine with a single interface as well. This is definitely better than what we currently have.

l-trotta:
RoleTemplateInlineScript has duplicate fields from ScriptBase, we should remove lang and options

l-trotta:
LGTM but it breaks deserialization for RoleTemplateScript the java client, if we could wait until I fix it before merging this

swallez (approved):
LGTM.

flobernd (approved):

l-trotta (approved):

Issue #2641, opened 2024-06-24
Regression on `AggregationRange` - incorrect types for `from` and `to` fields


pgayvallet:
## 🐛 Wrong type

`AggregationRange.from` and `AggregationRange.to` were changed from `string | number` to `number` in  https://github.com/elastic/elasticsearch-specification/pull/2552

From the PR:

> [AggregationRange](https://github.com/elastic/elasticsearch-specification/blob/02109140a64ce247fbb3fc0e6e4dd7b4b0d98da6/specification/_types/aggregations/bucket.ts#L674) accepting both double and string was probably established because at some point it was also used for RangeQuery, which accepts Objects for parameters to and from. But in the spec we just use it in [GeoDistanceAggregation](https://github.com/elastic/elasticsearch/blob/8864058f83bc485ce6c89a196e9d64b04e79b1cf/server/src/main/java/org/elasticsearch/search/aggregations/bucket/range/GeoDistanceAggregationBuilder.java#L98) and [RangeAggregation](https://github.com/elastic/elasticsearch/blob/8864058f83bc485ce6c89a196e9d64b04e79b1cf/server/src/main/java/org/elasticsearch/search/aggregations/bucket/range/GeoDistanceAggregationBuilder.java#L98), both accepting only doubles. Since the java client generalizes string | double as just string, this results in the server throwing exceptions when called.

This was supposed to be done to address https://github.com/elastic/elasticsearch-java/issues/666, but the assumption that was used for the PR is wrong, For `RangeAggrations` at least,  `from` and `to` do support `string` type value, for patterns such as `now` or `now+X`.

E.g from the Kibana repo:

https://github.com/elastic/kibana/blob/1c1e20afdb883fc279c5a06d4b71f8e577f8dad6/x-pack/plugins/task_manager/server/monitoring/workload_statistics.ts#L169-L180

flobernd:
I think this type follows the same semantics as `RangeQuery` which means that it should accept arbitrary typed values for the `from` and `to` fields.

The corresponding `RangeAggregate` looks suspicious to me as well. We should double check the server code here as well.

PR #2725, opened 2024-07-16
making range aggregation untagged union

l-trotta:
fixes https://github.com/elastic/elasticsearch-specification/issues/2641
~~draft for now because I still have to figure out if we're mapping the responses right~~

### DIFF
--- a/compiler/src/steps/validate-model.ts
+++ b/compiler/src/steps/validate-model.ts
@@ -597,9 +597,9 @@ export default async function validateModel (apiModel: model.Model, restSpec: Ma
 
       validateValueOf(valueOf, openGenerics)
     } else if (variants.kind === 'untagged') {
-      if (fqn(parentName) !== '_types.query_dsl:DecayFunction' &&
+      if (fqn(parentName) !== '_types.aggregations:AggregationRange' &&
+          fqn(parentName) !== '_types.query_dsl:DecayFunction' &&
           fqn(parentName) !== '_types.query_dsl:DistanceFeatureQuery' &&
-          fqn(parentName) !== '_types.aggregations:AggregationRange' &&
           fqn(parentName) !== '_types.query_dsl:RangeQuery') {
         throw new Error(`Please contact the devtools team before adding new untagged variant ${fqn(parentName)}`)
       }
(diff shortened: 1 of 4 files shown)


compiler/src/steps/validate-model.ts
```diff
@@ -610,6 +610,7 @@ export default async function validateModel (apiModel: model.Model, restSpec: Ma
     } else if (variants.kind === 'untagged') {
       if (fqn(parentName) !== '_types.query_dsl:DecayFunction' &&
           fqn(parentName) !== '_types.query_dsl:DistanceFeatureQuery' &&
+          fqn(parentName) !== '_types.aggregations:AggregationRange' &&
```
flobernd, line 613: Nit: Let's keep the names in alphabetical order please :-)

specification/_types/aggregations/bucket.ts
```diff
@@ -671,21 +671,36 @@ export class RangeAggregation extends BucketAggregationBase {
...
 
-export class AggregationRange {
+/**
+ * @codegen_names untyped, number, string
+ * @variants untagged untyped=_types.aggregations.UntypedAggregationRange
+ */
+export type AggregationRange =
+  | UntypedAggregationRange
+  | NumberAggregationRange
+  | StringAggregationRange
```
flobernd, line 681: Let's check, if we can have "term" range aggregations and if yes, rename this variant to TermAggregationRange.

We as well most likely want to have either a Date- or even a DateMath variant.

l-trotta: [DateRangeAggregation](https://github.com/elastic/elasticsearch-specification/blob/5615a46a870f15fa51edf6a3c4e8c2ed8cc3ebf7/specification/_types/aggregations/bucket.ts#L307) can be part of the union yes. so untyped | number | term | date,  like range query?

l-trotta: just noticed that DateRangeAggregation, like IpRangeAggregation, has its own field in the AggregationContainer (`date_range`). does it make sense to also add it here?

flobernd (approved):
LGTM!

Anaethelion (approved):

PR #2728, opened 2024-07-17
[OAS] Add OpenAPI overlays

lcawl:
This PR:

- Adds an overlay file to add the missing OpenAPI `info.description`.
- Updates the linting rules to check for operation identifiers, since poor or inconsistent URLs might be created in their absence
- Adds an overlay command to the makefile for testing purposes

Per https://docs.bump.sh/guides/openapi/augmenting-generated-openapi/#deploy-overlay-changes-to-bumpsh overlays could ultimately be applied during a GitHub action so we don't need to store the modified version of the file.

### DIFF
--- a/.spectral.yaml
+++ b/.spectral.yaml
@@ -14,9 +14,9 @@ rules:
   oas3-valid-schema-example: false
   oas2-valid-media-example: false
   # Operations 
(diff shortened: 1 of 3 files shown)


karenzone (approved):
LGTM

PR #2729, opened 2024-07-17
Added include_named_queries_score to search

### LINKED ISSUE #2405: Missing `include_named_queries_score` in search and boolean query
`include_named_queries_score` is missing in the API spec, although it's present in the rest-spec.json files.

The [comment in the rest-spec file](https://github.com/elastic/elasticsearch-specification/blob/298881468492420ba265e2526df25769c9cd8e83/specification/_json_spec/search.json#L224) doesn't match [the explanation in the docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/search-search.html). 

If this parameter really changes the shape of the result, we don't want to enable it for strongly-typed languages as it would lead to a more complex API (changing an array to a union of map and array) that would also be a breaking change.

Investigation needed as to what doc is right.

l-trotta:
Added in [8.8](https://github.com/elastic/elasticsearch/pull/94564), `include_named_queries_score` adds the score to the result of named query. 
In practice, it changes the type of [matched_queries](https://github.com/elastic/elasticsearch-specification/blob/5615a46a870f15fa51edf6a3c4e8c2ed8cc3ebf7/specification/_global/search/_types/hits.ts#L52) in Hit from `string[]` to `string[] | Dictionary<string,double>`, because the server will return a different data structure depending on the value of `include_named_queries_score` in the request. 

After discussing this, we decided that for now strongly typed clients (java, go, .net) won't support it, but we're still adding this to the spec for completeness and for other clients that might be able to support it. 

@Anaethelion, @flobernd before merging this we need to exclude `include_named_queries_score` from the available query parameters and ignore the secondary type of `matched_queries`.

@JoshMock does this cause any problem for javascript?

fixes #2405

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -23418,6 +23418,9 @@
             "$ref": "#/components/parameters/msearch#ignore_unavailable"
           },
           {
+            "$ref": "#/components/parameters/msearch#include_named_queries_score"
+          },
+          {
             "$ref": "#/components/parameters/msearch#max_concurrent_searches"
           },
           {
@@ -23473,6 +23476,9 @@
           },
           {
             "$ref": "#/components/parameters/msearch#ignore_unavailable"
+          },
+          {
+            "$ref": "#/components/parameters/msearch#include_named_queries_score"
           },
           {
             "$ref": "#/components/parameters/msearch#max_concurrent_searches"
@@ -23537,6 +23543,9 @@
             "$ref": "#/components/parameters/msearch#ignore_unavailable"
           },
           {
+            "$ref": "#/components/parameters/msearch#include_named_queries_score"
+          },
+          {
             "$ref": "#/components/parameters/msearch#max_concurrent_searches"
           },
           {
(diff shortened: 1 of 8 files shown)


JoshMock:
Nope, no problems from JS. :+1:

pquentin (changes requested):
Thank you! This looks great. I confirmed that this does not work as part of the request body.

However, this needs to be added to MultiSearchRequest.ts.

l-trotta:
@pquentin added it to multisearch, thank you!

pquentin (commented):
Should we also add the performance caveat from the docs?

> Typically, this adds a small overhead to a request. However, using computationally expensive named queries on a large number of hits may add significant overhead.

Also, the `server_default false` annotation is missing.

pquentin (approved):
Thanks! LGTM.

l-trotta:
@flobernd @Anaethelion let me know when the generators are ready so that we can merge this

flobernd:
@l-trotta My generator is adjusted.

Anaethelion:
@l-trotta all good for me.

PR #2730, opened 2024-07-17
Make sure that the type discriminator is the first property

flobernd:
This changes allows to speed up deserialization a lot (for statically typed languages) as the polymorhic variants can be determined by peeking a single JSON key-value pair instead of deserializing a complete object into an intermediate representation.

This as well allows to use built-in polymorphic JSON deserialization with several JSON frameworks (like e.g. `System.Text.Json` in .NET).

Note:
For other polymorhic objects like `Body`, `Value`, `Variant`, the discriminator is already emitted as the first key already.

Edit:
Needs further work to ensure rest of the properties stay sorted. In the best case, [safe-stable-stringify](https://github.com/BridgeAR/safe-stable-stringify/issues/48) would provide an option to customize the sort order.

### DIFF
--- a/compiler/package-lock.json
+++ b/compiler/package-lock.json
@@ -13,7 +13,7 @@
         "compiler-wasm-lib": "file:../compiler-rs/compiler-wasm-lib/pkg",
         "fastest-levenshtein": "^1.0.12",
         "ora": "^5.4.1",
-        "safe-stable-stringify": "^2.3.1",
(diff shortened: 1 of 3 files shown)


flobernd:
I have to wait for the author of `safe-stable-stringify` to release version 2.5.0 on NPM, before we can merge this. Using the GH repo works locally, but not in the CI.

PR #2719, opened 2024-07-15
Add more summaries

shainaraskas:
This PR adds operation summaries for the following areas, using text from the appropriate pages from the [elasticsearch API reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-apis.html).

- component template
- scripts
- tasks
- security

part of https://github.com/elastic/elasticsearch-specification/issues/2635

will regenerate the reference after https://github.com/elastic/elasticsearch-specification/pull/2718 is approved and merged and this PR is rebased

PR #2718, opened 2024-07-15
Update summaries for several endpoints

shainaraskas:
This PR adds operation summaries for the following areas, using text from the appropriate pages from the [elasticsearch API reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-apis.html).

- alias
- license
- lifecycle
- refresh
- reindex
- rollover

part of https://github.com/elastic/elasticsearch-specification/issues/2635

Issue #2635, opened 2024-06-20
[DOCS] Improve serverless operation summaries


lcawl:
Now that https://github.com/elastic/elasticsearch-specification/issues/2314 is fixed, we can have operation summaries that diverge from operation descriptions.  This meta-issue tracks the operation summary review and edits for the serverless APIs in each of these sections:

## Done

- [Alias APIs](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-indices-get-alias-1) https://github.com/elastic/elasticsearch-specification/pull/2718
- [Analyze](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-indices-analyze) https://github.com/elastic/elasticsearch-specification/pull/2924
- [Application](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-search-application-get) https://github.com/elastic/elasticsearch-specification/pull/2926
- [Async search](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-async-search-get) https://github.com/elastic/elasticsearch-specification/pull/2924
- [Bulk](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-bulk-1) https://github.com/elastic/elasticsearch-specification/pull/2719
- [Cat](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-cat-aliases) https://github.com/elastic/elasticsearch-specification/pull/2734
- [Component template](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-cluster-get-component-template-1) https://github.com/elastic/elasticsearch-specification/pull/2719
- [Connector](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/group/endpoint-connector) https://github.com/elastic/elasticsearch-specification/pull/3051
- [Data stream](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-indices-get-data-stream-1) https://github.com/elastic/elasticsearch-specification/pull/2711
- [Delete by query](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-delete-by-query) https://github.com/elastic/elasticsearch-specification/pull/2734
- [Doc](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-index-2) https://github.com/elastic/elasticsearch-specification/pull/2734
[...]

PR #2732, opened 2024-07-19
Add summaries for root operations

shainaraskas:
This PR adds operation summaries for operations at the root `/`, using text from the appropriate pages from the [elasticsearch API reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-apis.html).

part of https://github.com/elastic/elasticsearch-specification/issues/2635

will regenerate the reference after https://github.com/elastic/elasticsearch-specification/pull/2718 + https://github.com/elastic/elasticsearch-specification/pull/2719 is approved and merged and this PR is rebased

note for reviewers: most of these operations will probably eventually make it out of root and into their respective categories (mostly index/doc related)

### DIFF
--- a/specification/cluster/reroute/ClusterRerouteRequest.ts
+++ b/specification/cluster/reroute/ClusterRerouteRequest.ts
@@ -46,7 +46,7 @@ export interface Request extends RequestBase {
      */
     metric?: Metrics
     /**
-     * If true, then retries allocation of shards that are ed due to too many subsequent allocation failures.
+     * If true, then retries allocation of shards that are blocked due to too many subsequent allocation failures.
      * @server_default false
      */
     retry_failed?: boolean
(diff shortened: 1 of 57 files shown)


specification/cluster/reroute/ClusterRerouteRequest.ts
```diff
@@ -46,7 +46,7 @@ export interface Request extends RequestBase {
      */
     metric?: Metrics
     /**
-     * If true, then retries allocation of shards that are blocked due to too many subsequent allocation failures.
+     * If true, then retries allocation of shards that are ed due to too many subsequent allocation failures.
```
lcawl, line 49: I think this must be a typo
```suggestion
     * If true, then retries allocation of shards that are blocked due to too many subsequent allocation failures.
```

lcawl (approved):
One typo otherwise LGTM

PR #2736, opened 2024-07-23
put_user: Drop username from body

JoshMock:
Username is present in the path, and then again redundantly in the body. [The docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.14/security-api-put-user.html) indicate that it should only be in the path.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -17625,7 +17625,6 @@ export interface SecurityPutUserRequest extends RequestBase {
   username: Username
   refresh?: Refresh
   body?: {
-    username?: Username
     email?: string | null
     full_name?: string | null
     metadata?: Metadata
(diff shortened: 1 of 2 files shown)


l-trotta:
The docs say so... but the server says otherwise, and the validation reflects that. [username in the body](https://github.com/elastic/elasticsearch/blob/0c814b1c39f7f9c3e119ebb0fe9367dd5ba01b74/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/PutUserRequest.java#L31), [username in the path](https://github.com/elastic/elasticsearch/blob/0c814b1c39f7f9c3e119ebb0fe9367dd5ba01b74/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/user/RestPutUserAction.java#L49). I think we should keep this as is since it can accept both.

JoshMock:
Yeah, the flight recorder validation indicated the same. Never mind!

PR #2746, opened 2024-07-29
Clean up markdown in spec

shainaraskas:
Based on feedback, we realized that certain markdown was breaking other docs sets that rely on these specifications. Removing the problematic markdown while we decide on a strategy.

~holding this until I figure out whether additional markdown such as code markup and bulleted lists also break things~ <- will follow up on this in a later pr if needed

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -849,7 +849,7 @@
           "cat.aliases"
         ],
         "summary": "Get aliases",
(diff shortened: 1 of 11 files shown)


pquentin (commented):
Thanks! Great step in the right direction for now.

Can you please tell us which APIs have bullet point or code samples?

shainaraskas:
@pquentin 

An example of bullets: https://github.com/elastic/elasticsearch-specification/blob/main/specification/cat/indices/CatIndicesRequest.ts#L31-L36

An example of code formatting: https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/exists_source/SourceExistsRequest.ts#L33

kilfoyle (approved):
LGTM! 🦕

PR #2752, opened 2024-07-31
Removes the weighted_tokens query from the specification

kderusso:
The `weighted_tokens` query is removed in 8.16 in https://github.com/elastic/elasticsearch/pull/111492. This PR removes it from our specification.


kderusso:
Won't do this until 9.0

Issue #2482, opened 2024-04-04
[DOCS] Add examples to API endpoints


szabosteve:
## Description

Add examples to the API endpoints. 

* Use YAML format instead of JSON, since it supports comments. For example, conversion can be done via https://marketplace.visualstudio.com/items?itemName=tuxtina.json2yaml
* Put the examples in "request" and "response" subfolders alongside the appropriate specifications. If you know the response code, use a "200_response" subfolder, for example.
* To prevent duplicated work, leave your name beside a namespace/API you start working on.
* The list contains only the API namespaces (for example, `indices`), you can expand the list granularly by adding the endpoints related to the namespace you work on in case it helps track the effort (for example, if you don't cover the whole namespace in one PR).
* Add your PR URL next to the related namespace/API.
* Adjust the list if you find any missing namespaces/endpoints.
* Leave relevant notes either in the list or in a comment.

NOTE: Until there's an automated method for associating examples with endpoints, we're accomplishing it via [overlays](https://github.com/elastic/elasticsearch-specification/tree/main/docs/overlays). Reach out to @lcawl if assistance is required.

API namespace inventory:
- [x] _global
  - [x] bulk https://github.com/elastic/elasticsearch-specification/pull/3510
  - [x] clear scroll https://github.com/elastic/elasticsearch-specification/pull/3510
  - [x] close_point_in_time https://github.com/elastic/elasticsearch-specification/pull/3510
  - [x] open_point_in_time https://github.com/elastic/elasticsearch-specification/pull/3561
  - [x] [count](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-count.html) https://github.com/elastic/elasticsearch-specification/pull/3510
  - [x] create https://github.com/elastic/elasticsearch-specification/pull/3540
  - [x] delete https://github.com/elastic/elasticsearch-specification/pull/3540
  - [x] delete_by_query/delete_by_query_rethrottle https://github.com/elastic/elasticsearch-specification/pull/3574
  - [x] update_by_query/update_by_query_rethrottle https://github.com/elastic/elasticsearch-specification/pull/3597
  - [x] delete_script/get_script/put_script https://github.com/elastic/elasticsearch-specification/pull/3613
  - [x] scripts_painless_execute https://github.com/elastic/elasticsearch-specification/pull/3589
  - [x] exists/exists_source https://github.com/elastic/elasticsearch-specification/pull/3540
[...]

PR #2755, opened 2024-07-31
Add transform API examples via overlay

lcawl:

This PR demonstrates a way we could short-term use [overlays](https://docs.bump.sh/guides/openapi/augmenting-generated-openapi/) to insert examples into the openAPI document.

For now the updated OpenAPI document is generated by using the "make overlay-docs" command and it's stored in the output folder. Long-term we could potentially perform that processing as part of the OpenAPI generator or as part of the deployment process so that we didn't need to maintain otherwise identical files.


szabosteve (approved):
LGTM, thanks!

PR #2756, opened 2024-08-01
Update rest-api-spec only once a day

pquentin:
It currently runs 8 times per day, which is just too much. But maybe it should be once a week instead?


picandocodigo (approved):
I think once a day is good :+1:

PR #2760, opened 2024-08-01
Create user feedback issue template

KOTungseth:
User feedback issue template for Bump.sh API docs.

To do: Create `Feedback` label.


lcawl (approved):
LGTM!

pquentin:
@KOTungseth @lcawl  Any reason not to apply my fix above and merge?

.github/ISSUE_TEMPLATE/docs_feedback.yaml
```diff
@@ -0,0 +1,26 @@
...
+description: Tell us about your experience with Elastic API documentation.
+title: "[Feedback]: "
+labels: ["feedback"]
+body:
+  - type: dropdown
+    id: was-this-helpful
+    attributes:
+      label: Was this helpful?
+      options:
+        - Yes
```
lcawl, line 11:
```suggestion
        - "Yes"
```

.github/ISSUE_TEMPLATE/docs_feedback.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - No
```
lcawl, line 12:
```suggestion
        - "No"
```

lcawl (commented):
When I used this template in docs-content, it generated an error on the option values. It seems you need to put them in quotes per https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#bodyi-options-must-not-include-booleans-please-wrap-values-such-as-yes-and-true-in-quotes

pquentin:
Good catch! YAML is annoying like that: yes and no are treated as booleans, not strings: https://yaml.org/type/bool.html

Issue #2758, opened 2024-08-01
Allow relative links to other parts of the documentation


pquentin:
## 🚀 Feature Proposal

Relative links in the Markdown description should be interpreted as links to endpoint and/or namespaces:

 * #cat would link to the cat namespace, such as https://elasticsearch-py.readthedocs.io/en/v8.14.0/api/cat.html for Python.
 * #cat.aliases would link to the specific function, such as https://elasticsearch-py.readthedocs.io/en/v8.14.0/api/cat.html#elasticsearch.client.CatClient.aliases in the documentation.

There's a catch with global endpoints, such as search or bulk. Should they be defined as `_global.bulk` or just `bulk`? The Python link is https://elasticsearch-py.readthedocs.io/en/v8.14.0/api/elasticsearch.html#elasticsearch.Elasticsearch.bulk.

The OpenAPI export would use links that work for the Serverless OpenAPI docs. Currently, there's no grouping, so for like `cat.aliases` it would `#endpoint-alias` according to https://github.com/elastic/elasticsearch-specification/pull/2746/files#diff-4f456691f5983599680ab2e423b475e86dc8ea112ef4e7ed91e691cc6c14c567.

Using other forms of grouping than namespaces is out of scope.

## Motivation

Linking between endpoints is useful. The current use case is linking from the human-friendly cat APIs to the corresponding machine-friendly API.

## Example

Please provide an example for how this feature would be used.

```
/*
 * Get aliases.
 * 
 * Retrieves the cluster’s index aliases, including filter and routing information.
 * The API does not return data stream aliases.
 * > [!NOTE]
 * > CAT APIs are only intended for human consumption using the command line or the Kibana console. They are not intended for use by applications. For application consumption, use [the Aliases API](#alias).
 */
```

## Open questions

* Should we use a prefix to make it clear that the link is relative?
* (Does our API docs support GFM admonitions?)

shainaraskas:
>The OpenAPI export would use links that work for the Serverless OpenAPI docs. Currently, there's no grouping, so for like cat.aliases it would #endpoint-alias according to https://github.com/elastic/elasticsearch-specification/pull/2746/files#diff-4f456691f5983599680ab2e423b475e86dc8ea112ef4e7ed91e691cc6c14c567.

I don't think this is exactly correct. cat.aliases would be `#operation-cat-aliases` and the aliases grouping would be `#endpoint-alias`

[...]

PR #2761, opened 2024-08-01
Generalize references to APIs

shainaraskas:
To avoid confusing people reading these docs in the client references, I'm removing explicit references to endpoints and endpoint names, and going back to general API names. We might revisit this [as we improve our link tech](https://github.com/elastic/elasticsearch-specification/issues/2758).


lcawl (approved):
LGTM!

PR #2773, opened 2024-08-06
Update CatCountRequest.ts documentation

picandocodigo:
Small fix on the documentation. I found this while adding markdown support to the Ruby client code generator.


shainaraskas (approved):
thanks for catching that!

PR #2784, opened 2024-08-13
Fix indices.get_data_stream response type

pquentin:
We used to score 3/11 on responses, it's now 11/11.

| API | Status | Request | Response |
| --- | --- | --- | --- |
| `indices.get_data_stream` | :green_circle: | 11/11 | 11/11 |

### DIFF
--- a/specification/indices/_types/DataStream.ts
+++ b/specification/indices/_types/DataStream.ts
@@ -76,11 +76,11 @@ export class DataStream {
   /**
    * Name of the lifecycle system that'll manage the next generation of the data stream.
    */
-  next_generation_managed_by?: ManagedBy
+  next_generation_managed_by: ManagedBy
   /**
    * Indicates if ILM should take precedence over DSL in case both are configured to managed this data stream.
    */
-  prefer_ilm?: boolean
+  prefer_ilm: boolean
   /**
    * Array of objects containing information about the data stream’s backing indices.
    * The last item in this array contains information about the stream’s current write index.
@@ -103,7 +103,7 @@ export class DataStream {
   /**
    * If `true`, the next write to this data stream will trigger a rollover first and the document will be indexed in the new backing index. If the rollover fails the indexing request will fail too.
    */
-  rollover_on_write?: boolean
+  rollover_on_write: boolean
   /**
    * Health status of the data stream.
    * This health status is based on the state of the primary and replica shards of the stream’s backing indices.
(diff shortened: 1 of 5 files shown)


specification/indices/_types/DataStream.ts
```diff
@@ -47,6 +53,11 @@ export class DataStream {
    *  If `true`, the data stream allows custom routing on write request.
    */
   allow_custom_routing?: boolean
+  /**
+   * Information about failure store backing indices
+   *
+   */
+  failure_store?: FailureStore
```
pquentin, line 60: I've made it optional because it's behind a feature flag (and undocumented): https://github.com/elastic/elasticsearch/blob/03384b5f6ffdbafbb57c261c798911303801c4d1/server/src/main/java/org/elasticsearch/action/datastreams/GetDataStreamAction.java#L350-L360. There's a `feature_flag` annotation in the specification, but only for requests.

When the feature flag go away, this will no longer be optional.

specification/indices/_types/DataStream.ts
```diff
@@ -47,6 +53,11 @@ export class DataStream {
    *  If `true`, the data stream allows custom routing on write request.
    */
   allow_custom_routing?: boolean
+  /**
+   * Information about failure store backing indices
```
pquentin, line 57: I would appreciate a better description here - I usually use the docs for this but there aren't any for now.

specification/indices/_types/DataStream.ts
```diff
@@ -36,6 +36,12 @@ enum ManagedBy {
   unmanaged = 'Unmanaged'
 }
 
+export class FailureStore {
+  enabled: boolean
+  indices: DataStreamIndex[]
+  rollover_on_write: boolean
```
pquentin, line 42: Descriptions would be useful here.

specification/indices/_types/DataStream.ts
```diff
@@ -134,11 +149,11 @@ export class DataStreamIndex {
   /**
    * Name of the lifecycle system that's currently managing this backing index.
    */
-  managed_by: ManagedBy
+  managed_by?: ManagedBy
   /**
    * Indicates if ILM should take precedence over DSL in case both are configured to manage this index.
    */
-  prefer_ilm: boolean
+  prefer_ilm?: boolean
```
pquentin, line 156: Those properties are indeed optional: https://github.com/elastic/elasticsearch/blob/03384b5f6ffdbafbb57c261c798911303801c4d1/server/src/main/java/org/elasticsearch/action/datastreams/GetDataStreamAction.java#L372-L378

pquentin (commented):
&nbsp;

PR #2785, opened 2024-08-13
Mark role APIs serverless public

n1v0lg:
Marks relevant role APIs as public in Serverless.

Resolves: ES-7833


pquentin (approved):
Thanks! LGTM.

PR #2786, opened 2024-08-14
Fix infer_trained_model API

pquentin:
It can return list of lists of integers and floats.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -91193,10 +91193,22 @@
             "type": "number"
           },
           {
+            "type": "array",
+            "items": {
+              "type": "number"
+            }
+          },
+          {
             "type": "boolean"
           },
           {
             "type": "number"
+          },
+          {
+            "type": "array",
+            "items": {
+              "type": "number"
+            }
           }
         ]
       },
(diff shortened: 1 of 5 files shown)


pquentin:
@davidkyle Sorry that I missed your comment. It happens in four "3rd party deployment" ML YAML tests, for example https://github.com/elastic/elasticsearch/blob/0d38528e0e6f01010ba4a6d11d84ff7a5ddf32f4/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/ml/3rd_party_deployment.yml#L600-L622. In all cases, the following JSON is returned:

```json
{
  "inference_results": [
    {
      "predicted_value": [
        [
          1,
          1
        ]
      ]
    }
  ]
}
```

pquentin:
Actually yes it *is* a list of lists, so there's a missing list somewhere, `PredictedValue[]` isn't enough. Reopening.

davidkyle (approved):
LGTM

Thanks @pquentin, the yml test you linked to uses the special `pass_through` type that returns the raw response. I'd forgotten about that one sorry

pquentin:
Merging, assuming that only ints and floats can be lists of lists. Happy to open another PR if that's not the case.

davidkyle:
Yes only ints and floats

PR #2793, opened 2024-08-20
Fix hot threads to always return strings

### LINKED ISSUE #1206: Incorrect Response content-type for nodes.hot_threads
The response of `nodes.hot_threads`: 

```
make validate stack-version=8.0.0-SNAPSHOT type=response api=nodes.hot_threads
```

Text instead of JSON is returned. Can you confirm that the correct header is emitted?

pquentin:
Closes #1206


dadoonet (approved):
LGTM ;)

swallez (dismissed):
Please also remove the `HotThread` type.

l-trotta (approved):

l-trotta:
I think we missed the change because we thought the hidden file was an output file ^^" LGTM!

PR #2802, opened 2024-08-22
making predicted_value scalar

l-trotta:
simplifying the definition of predicted_value


l-trotta:
unfortunately this reverts some green dots to red compared to https://github.com/elastic/elasticsearch-specification/pull/2786

flobernd (approved):
LGTM!

pquentin:
I'm not following the logic here. Don't we want to allow lists of lists?

flobernd (approved):

PR #2803, opened 2024-08-22
Add exclude query rule type to specification

kderusso:
The `exclude` query rule type will go live with 8.16.0, this PR updates the specification accordingly.


Mikep86 (approved):

pquentin (approved):
Thanks! LGTM.

PR #2805, opened 2024-08-22
[DOCS] Add API docs feedback link

lcawl:
This PR: 

- adds a feedback link to the Elasticsearch openAPI documents via an overlay, as described in https://docs.bump.sh/help/publish-documentation/feedback/. The feedback link creates an issue in the docs_content repo for now.
- fixes a path error in the overlay, since some folders had changed in the paths to the examples
- removes the output files from the "make overlay-docs", since I don't think there's value to checking them in and adds them to the .gitignore file


shainaraskas (approved):
given that you've previewed this and it's working, good to go! 🚀

PR #2807, opened 2024-08-23
[DOCS] Add more transform example overlays

lcawl:
This PR fixes the existing "create transform" example overlay, since the file path had changed. It also adds the "update transform" and "preview transform" examples.

NOTE: You need to run "make overlay-docs" to see the updated output, so I'm no longer checking in that file.


szabosteve (approved):
Thank you, LGTM!

PR #2808, opened 2024-08-26
Fix Painless execute API

pquentin:
According to https://github.com/elastic/elasticsearch/blob/44758b3823a4982e9c62598f184cf90b67ab5d2a/modules/lang-painless/src/main/java/org/elasticsearch/painless/action/PainlessExecuteAction.java#L167-L223 all `context_setup` fields are optional, but [docs](https://www.elastic.co/guide/en/elasticsearch/painless/8.15/painless-execute-api.html) say that `index` and `document` are required, [all YAML tests](https://github.com/search?q=repo%3Aelastic%2Felasticsearch+scripts_painless_execute+language%3AYAML&type=code&l=YAML) include them and `context_setup` itself is already optional. As a result, I left them required.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -96876,8 +96876,7 @@
         },
         "required": [
           "document",
-          "index",
-          "query"
+          "index"
         ]
       },
       "_types:RankContainer": {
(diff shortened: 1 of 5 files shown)


l-trotta (approved):
LGTM!

PR #2815, opened 2024-08-27
Add Get geoip database configuration API

pquentin:
Part of #2810 

Thanks to https://github.com/elastic/elasticsearch/pull/112216 we have validation for this API, which helpfully fixed a few errors I made during development.

Issue #2810, opened 2024-08-26
Ingest GeoIP database APIs


yuliacech:
The types for the GeoIP APIs are currently not complete in the specification: 
- Get geoip database configuration API [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/get-geoip-database-api.html)
 The response of the API is not specified, it should be something like 
```ts
{
  databases: {
    id: string;
    version: number;
    modified_date_millis: number;
    database: {
      name: string;
      maxmind: {
        account_id: string;
      };
    };
  }[]
}
```
- Create or update geoip database configuration API [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-geoip-database-api.html)
  The specification doesn't have the correct parameters for request body, so it's not possible to pass the `body` object to the js client, should be something like 
```ts
{
    name: string,
    maxmind: {
      account_id: string,
    },
  }
```
- Delete geoip database configuration API [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/delete-geoip-database-api.html)
  The specification doesn't have the correct parameters for the request, should be 
```ts
{
  id: string
}
```

pquentin:
Done in #2815, #2823 and #2824. Backported to 8.15 in #2822, #2828 and #2827.

Thank you, closing!

PR #2824, opened 2024-08-28
Add Delete geoip database configuration API

pquentin:
Part of https://github.com/elastic/elasticsearch-specification/issues/2810, like https://github.com/elastic/elasticsearch-specification/pull/2815 and #2823. There's an output conflict with #2823 that I will fix when one of the two will have been merged.

### DIFF
--- a/specification/ingest/delete_geoip_database/DeleteGeoipDatabaseRequest.ts
+++ b/specification/ingest/delete_geoip_database/DeleteGeoipDatabaseRequest.ts
@@ -1,0 +1,48 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


specification/ingest/delete_geoip_database/DeleteGeoipDatabaseRequest.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @rest_spec_name ingest.delete_geoip_database
+ * @availability stack since=8.15.0 stability=stable
+ * @availability serverless visibility=private
+ */
+export interface Request extends RequestBase {
+  path_parts: {
+    /**
+     * A comma-separated list of geoip database configurations to delete
+     */
+    id?: Ids
```
flobernd, line 35: @pquentin Is this really an optional parameter? This triggered a validation rule in my code generator.

The rule requires a mandatory parameter to be used in all available urls for an endpoint and an optional parameter to **not** be used in all available urls for an endpoint.

flobernd: We currently have only one url for this endpoint:

```json
"urls": [
        {
          "methods": [
            "DELETE"
          ],
          "path": "/_ingest/geoip/database/{id}"
        }
]
```

In other words: If that endpoint does not support an additional url like `"/_ingest/geoip/database"` (without the `{id}`), we should make the path parameter mandatory.

PR #2825, opened 2024-08-28
Fix ingest.simulate API response

pquentin:
The errors on the request side are due to the missing `redact` processor.

### DIFF
--- a/specification/ingest/simulate/SimulatePipelineRequest.ts
+++ b/specification/ingest/simulate/SimulatePipelineRequest.ts
@@ -45,10 +45,6 @@ export interface Request extends RequestBase {
   }
   body: {
     /**
-     * JSON body for the document.
-     */
-    _source?: UserDefinedValue
-    /**
      * Sample documents to test in the pipeline.
      */
     docs: Document[]
--- a/specification/ingest/simulate/SimulatePipelineRequest.ts
+++ b/specification/ingest/simulate/SimulatePipelineRequest.ts
@@ -21,6 +21,7 @@ import { Pipeline } from '@ingest/_types/Pipeline'
 import { RequestBase } from '@_types/Base'
 import { Id } from '@_types/common'
 import { Document } from './types'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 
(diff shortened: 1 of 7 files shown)


l-trotta:
hmm I'm suspecting the whole definition of the response might be wrong, I'll need some time to check this thoroughly. Just to be sure, this is [Simulate Pipeline API](https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-pipeline-api.html) right?

pquentin:
> Just to be sure, this is [Simulate Pipeline API](https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-pipeline-api.html) right?

Correct!

pquentin:
Not to be confused with the "Simulate ingest API": https://www.elastic.co/guide/en/elasticsearch/reference/8.15/simulate-ingest-api.html

specification/ingest/simulate/SimulatePipelineRequest.ts
```diff
@@ -43,10 +44,14 @@ export interface Request extends RequestBase {
     verbose?: boolean
   }
   body: {
+    /**
+     * JSON body for the document.
+     */
+    _source?: UserDefinedValue
```
pquentin, line 50: The docs suggest that this should be in `Document`. But maybe the server code supports passing the source of a single document? But then how does that interact with `docs`?

pquentin (commented):
Thanks! LGTM, but I have one concern about `_source` in body.

specification/ingest/simulate/SimulatePipelineRequest.ts
l-trotta: not sure... I did some tests and the server just seems to ignore it. I'll ask around

l-trotta: so this is another case where the parser will accept anything, but the only fields that actually get processed are `docs` and `pipeline`, so I removed `_source`

pquentin (commented):
Thanks! LGTM.

l-trotta (approved):

PR #2829, opened 2024-08-29
Fix Open Point in time API

pquentin:
The remaining failures will be fixed when the point in time NEST tests will be removed.

### DIFF
--- a/specification/_global/open_point_in_time/OpenPointInTimeResponse.ts
+++ b/specification/_global/open_point_in_time/OpenPointInTimeResponse.ts
@@ -18,7 +18,12 @@
  */
 
 import { Id } from '@_types/common'
+import { ShardStatistics } from '@_types/Stats'
 
 export class Response {
-  body: { id: Id }
+  body: {
+    /** Shards used to create the PIT */
+    _shards: ShardStatistics
+    id: Id
+  }
 }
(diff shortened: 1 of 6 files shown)


specification/_global/open_point_in_time/OpenPointInTimeResponse.ts
```diff
@@ -18,7 +18,12 @@
  */
 
 import { Id } from '@_types/common'
+import { ShardStatistics } from '@_types/Stats'
 
 export class Response {
-  body: { id: Id }
+  body: {
+    /** Shards used to create the PIT */
+    _shards: ShardStatistics
```
l-trotta, line 26: _shards is [optional](https://github.com/elastic/elasticsearch/blob/5e455db10ecbb1a31cad58ecb1120a66fc50079f/server/src/main/java/org/elasticsearch/action/search/OpenPointInTimeResponse.java#L49)

l-trotta: we could merge this in main and wait for the backport?

pquentin: Right, we won't backport because 8.15 will never have `_shards`. (I could backport the `index_filter` part but I'm being lazy here.)

We agreed offline to wait for the `_shards` change to land in serverless before merging.

l-trotta (approved):

PR #2834, opened 2024-08-30
Fix search_shards API

pquentin:


specification/_types/Node.ts
```diff
@@ -55,6 +55,10 @@ export class NodeAttributes {
    * @availability serverless
    */
   external_id?: string
+  // version, min_index_version and max_index_version are set when using the search_shards API
+  version?: VersionString
+  min_index_version?: integer
+  max_index_version?: integer
```
pquentin, line 61: I hesitated a lot on this one. NodeAttributes is used by transform.get_transform_stats, cat.ml_data_frame_analytics, and cat.nodeattrs, but they're much ligther versions, which is why many fields are optional. At this point I think it would be better to have a specific `NodeAttributes` just for `search_shards`, which would remove a lot of optional keys. But I'm not sure about the impact on typed clients.

flobernd: Both solutions should be fine. It might as well be worth to consider which solution better reflects the server code.

pquentin: It's definitely the second one. I'll update my pull request then, thanks.

pquentin: Done in [`08babed` (#2834)](https://github.com/elastic/elasticsearch-specification/pull/2834/commits/08babede4976d1a5f1797c2059a7c635c3535aca). Our tooling forced me to use a different class name.

I hope this won't be breaking any clients.

l-trotta (approved):

Issue #2231, opened 2023-08-09
Multiple processor classes are missing


flobernd:
Multiple processor classes are missing from the current spec:
https://github.com/elastic/elasticsearch-specification/blob/main/specification/ingest/_types/Processors.ts

Missing processors:

- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/community-id-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/fingerprint-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-geo-grid-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/htmlstrip-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/network-direction-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/redact-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/registered-domain-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/urldecode-processor.html (class exists; missing in the container)
- [ ] https://www.elastic.co/guide/en/elasticsearch/reference/current/uri-parts-processor.html

Related to: https://github.com/elastic/elasticsearch-net/issues/7881

PR #2835, opened 2024-08-30
Adds `HtmlStripProcessor` and `UriPartsProcessor`

flobernd:
Adds `HtmlStripProcessor` and `UriPartsProcessor`.

Related to #2231
Related to https://github.com/elastic/elasticsearch-net/issues/7881


pquentin (approved):
Thanks! Processors are our biggest gap currently. I checked that those additions match with the Elasticsearch documentation.

Unfortunately, we don't have specific YAML tests for those processors, so we can't validate those changes more.

PR #2844, opened 2024-09-02
Add geo-grid ingest processor

pquentin:


specification/ingest/_types/Processors.ts
```diff
@@ -348,6 +354,60 @@ export class AttachmentProcessor extends ProcessorBase {
...
+
+export enum GeoGridTileType {
+  geotile,
+  geohex,
+  geohash
+}
+
+export enum GeoGridTargetFormat {
+  geojson,
+  wkt
```
pquentin, line 408: While the documentation says it should be `GeoJSON` and `WKT`, the YAML tests are using `geojson` and `wkt`. I talked to @craigtaverner who confirms that this processor was specifically designed to be case-insensitive so lowercase is OK.

flobernd (commented):
Nice, more processors 🙌

flobernd (approved):
LGTM! Thank you!

PR #2847, opened 2024-09-02
Fix cat allocation API response

pquentin:


specification/cat/allocation/types.ts
```diff
@@ -26,73 +26,73 @@ export class AllocationRecord {
...
    * Amount of shards that are scheduled to be moved elsewhere in the cluster or -1 other than desired balance allocator is used
    */
-  'shards.undesired': string | null
+  'shards.undesired'?: string | null
 
   /**
    * Sum of index write load forecasts
    * @aliases wlf,writeLoadForecast
    */
-  'write_load.forecast': double | null
```
flobernd, line 40: We might want to use `Stringified<double> | null` here - this way, statically typed clients will expose the field as `double` while transparently converting any `string` value to `double` during deserialization.

pquentin: Thanks, TIL! Done in [`d98209c` (#2847)](https://github.com/elastic/elasticsearch-specification/pull/2847/commits/d98209cafac8513193e84418659e347c3a3e3387).

flobernd (approved):
LGTM! Thank you!

PR #2851, opened 2024-09-03
Add unassigned primary shards to health reports

pquentin:
This was added in https://github.com/elastic/elasticsearch/pull/112024

Validation will fail because we're still using some old NEST client tests that don't have this field. I'll merge tomorrow when they will be gone.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -66514,6 +66514,10 @@
             "description": "number of initializing nodes",
             "type": "string"
           },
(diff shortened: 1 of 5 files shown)


lukewhiting (approved):
LGTM 👍🏻 Seems to represent the original change correctly. Thanks again for the PR!

pquentin:
I had missed that index/shard stats should also have that field. Thankfully the YAML tests caught it. Please take another look.

l-trotta (approved):

Issue #2109, opened 2022-08-10
`function_score` type in `QueryDslQueryContainer` interface seems incorrect


ohueter:
## :bug: Bug Report

The `function_score` type in the `QueryDslQueryContainer` interface seems to be incorrect.

I implemented the first example from the [Function score query docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.3/query-dsl-function-score-query.html) in TypeScript:
```ts
import type { QueryDslQueryContainer } from '@elastic/elasticsearch/api/types'

const query: QueryDslQueryContainer = {
  function_score: {
    query: { match_all: {} },
    boost: 5,
    random_score: {},
    ^^^^^^^^^^^^^^^^^
    boost_mode: 'multiply',
  },
}
```
which results in the following type error:
```
(property) random_score: {}
Type '{ query: { match_all: {}; }; boost: number; random_score: {}; boost_mode: "multiply"; }' is not assignable to type 'QueryDslFunctionScoreQuery'.
  Object literal may only specify known properties, and 'random_score' does not exist in type 'QueryDslFunctionScoreQuery'.
types.d.ts(4824, 3): The expected type comes from property 'function_score' which is declared here on type 'QueryDslQueryContainer'
```

## Expected behavior

No type error should occur. I'm not familiar with the code base, the correct fix seems to change the `function_score` type in the `QueryDslQueryContainer` interface from

```ts
interface QueryDslQueryContainer {
  function_score?: QueryDslFunctionScoreQuery
}
```

to 

```ts
interface QueryDslQueryContainer {
  function_score?: QueryDslFunctionScoreQuery | QueryDslFunctionScoreContainer
}
```

Hoping this is the correct solution, I went ahead and created a PR for this change: elastic/elasticsearch-js#1745

It's tested to be working with a running Elastic instance :-)

## Your Environment

- `@elastic/elasticsearch` *version*: >=7.16.0

JoshMock:
@ohueter Just moved this to the `elasticsearch-specification` repo, since this concern you have with function score queries would apply to all client libraries.

The fix that you made in https://github.com/elastic/elasticsearch-js/pull/1745 would be accomplished via the code generator by modifying the following line:

https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/query_dsl/abstractions.ts#L112

to be

```typescript
function_score?: FunctionScoreQuery | FunctionScoreContainer
```

[...]

PR #2852, opened 2024-09-03
Break function_score out into single- and multi-function types

JoshMock:
Fixes https://github.com/elastic/elasticsearch-specification/issues/2109.


l-trotta:
`FunctionScoreQuery` was recently updated with the `@shortcut` property which allowed us to accept every variant of the query, are we sure we need this?

JoshMock:
I'm not entirely sure `@shortcut_property` covers all of this. As @swallez [noted here](https://github.com/elastic/elasticsearch-specification/issues/2109#issuecomment-1569808519), the issue is that all the properties of `FunctionScoreContainer` can be used on the same level as `boost_mode`, `max_boost`, etc. OR a `functions` property can be provided that includes one function object or an array of function objects.

pquentin (approved):
Thanks! LGTM.

l-trotta:
we need to rename FunctionScoreContainer, the java generator (and maybe others?) removes the "Container" part of the name, and since FunctionScore already exists there's a duplicate error. Does FunctionScoreCommonContainer sounds okay?

l-trotta:
After discussing it with the other static clients maintainer we have decided that supporting this would require too much effort compared to the benefits (since it's not a unique feature, but just an alternative syntax). If we were to discover a case where this is absolutely necessary to implement in the dynamic clients then we can review this solution again as a starting point and the static client can ignore it from the generator.

JoshMock:
Thanks @l-trotta!

PR #2853, opened 2024-09-04
Organize imports with Prettier plugin

pquentin:
Prettier does not organize imports by default, but there's a popular plugin to do so: https://www.npmjs.com/package/prettier-plugin-organize-imports. My initial motivation was to avoid unused imports, but not having to worry about import order is nice too.


specification/indices/put_mapping/IndicesPutMappingRequest.ts
```diff
@@ -18,7 +18,6 @@
  */
 
 import { Dictionary } from '@spec_utils/Dictionary'
-import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
```
pquentin, line 21: Here's an example of unused import.

specification/cluster/_types/ComponentTemplate.ts
```diff
@@ -18,14 +18,11 @@
  */
 
 import { AliasDefinition } from '@indices/_types/AliasDefinition'
+import { DataStreamLifecycleWithRollover } from '@indices/_types/DataStreamLifecycle'
 import { IndexSettings } from '@indices/_types/IndexSettings'
 import { Dictionary } from '@spec_utils/Dictionary'
 import { IndexName, Metadata, Name, VersionNumber } from '@_types/common'
 import { TypeMapping } from '@_types/mapping/TypeMapping'
-import {
-  DataStreamLifecycle,
```
pquentin, line 26: And this is the import that actively misled me, as it shows up when I asked my editor to find references or `DataStreamLifecycle`.

pquentin (commented):
There are plenty of unused imports, but they can be hidden in the noise, so I've highlighted two.

flobernd (approved):
LGTM!

If it does not change the output, I assume it's fine.

pquentin:
Done:

 * https://github.com/elastic/elasticsearch-specification/pull/2868
 * https://github.com/elastic/elasticsearch-specification/pull/2867

PR #2854, opened 2024-09-04
Fix most pipeline API errors

pquentin:
This was done purely looking at the YAML test errors. The Put pipeline API has one error left: the Redact processor is still missing.

Note that even when we hit 100%, the work on ingest pipelines won't be over. Not only there are number of pipelines still missing (tracked in https://github.com/elastic/elasticsearch-specification/issues/2231), but we have not added all ingest modules to YAML tests.


l-trotta:
In `UserAgentProcessor` we should remove `options`, it's not a field it seems. [server code](https://github.com/elastic/elasticsearch/blob/c21fc968d4180b025e2be2eb1bd7a4b503aeffbd/modules/ingest-user-agent/src/main/java/org/elasticsearch/ingest/useragent/UserAgentProcessor.java#L38)

pquentin:
> In `UserAgentProcessor` we should remove `options`, it's not a field it seems. [server code](https://github.com/elastic/elasticsearch/blob/c21fc968d4180b025e2be2eb1bd7a4b503aeffbd/modules/ingest-user-agent/src/main/java/org/elasticsearch/ingest/useragent/UserAgentProcessor.java#L38). Other than that, LGTM!

Thanks Laura! Removing `options` made me realize that we had an `UserAgentProperty` enum, so I've repurposed it for properties. I also opened https://github.com/elastic/elasticsearch/pull/112518 to fix the relevant Elasticsearch docs.

Please take another look.

flobernd (approved):
LGTM!

PR #2856, opened 2024-09-04
Update resolve index missing type definition

mohamedhamed-ahmed:
This PR adds the missing type definition for `ignore_unavailable` and `allow_no_indices` query param` which are [accepted by ES](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-resolve-index-api.html#resolve-index-api-query-params) but the definition is missing.

### DIFF
--- a/specification/_json_spec/indices.resolve_index.json
+++ b/specification/_json_spec/indices.resolve_index.json
@@ -29,10 +29,6 @@
         "options": ["open", "closed", "hidden", "none", "all"],
         "default": "open",
         "description": "Whether wildcard expressions should get expanded to open or closed indices (default: open)"
-      },
-      "ignore_unavailable": {
-        "type": "boolean",
-        "description": "Ignore unavailable indexes (default: false)"
       }
     }
   }
--- a/specification/_json_spec/indices.resolve_index.json
+++ b/specification/_json_spec/indices.resolve_index.json
@@ -29,6 +29,10 @@
         "options": ["open", "closed", "hidden", "none", "all"],
         "default": "open",
         "description": "Whether wildcard expressions should get expanded to open or closed indices (default: open)"
+      },
+      "ignore_unavailable": {
+        "type": "boolean",
+        "description": "Ignore unavailable indexes (default: false)"
       }
     }
   }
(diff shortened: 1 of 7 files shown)


specification/_json_spec/indices.resolve_index.json
```diff
@@ -29,6 +29,10 @@
         "options": ["open", "closed", "hidden", "none", "all"],
         "default": "open",
         "description": "Whether wildcard expressions should get expanded to open or closed indices (default: open)"
+      },
+      "ignore_unavailable": {
+        "type": "boolean",
+        "description": "Ignore unavailable indexes (default: false)"
```
pquentin, line 35: The source of truth of the `_json_spec` folder is in Elasticsearch, see https://github.com/elastic/elasticsearch/blob/main/rest-api-spec/src/main/resources/rest-api-spec/api/indices.resolve_index.json. Can you please open an Elasticsearch pull request? Otherwise this will change will be overwritten in our next weekly sync.

pquentin (commented):
Thanks, this looks good! Can you please remove the `_json_spec` change for now?

Also, as a follow-up and only if you're interested, we could add `allow_no_indices` that is also missing.

specification/indices/resolve_index/ResolveIndexRequest.ts
```diff
@@ -44,5 +44,17 @@ export interface Request extends RequestBase {
...
+     * If `false`, the request returns an error if it targets a missing or closed index.
+     * @server_default false
+     */
+    ignore_unavailable?: boolean
+    /**
+     * If false, the request returns an error if any wildcard expression, index alias, or _all value targets only missing
+     * or closed indices. This behavior applies even if the request targets other open indices. For example, a request
+     * targeting foo*,bar* returns an error if an index starts with foo but no index starts with bar.
+     * @server_default true
+     */
```
pquentin, line 57: We need to use Markdown backticks to avoid `foo*,bar*` be interpreted as Markdown. And we tend to have one line per sentence.

```suggestion
    /**
     * If `false`, the request returns an error if any wildcard expression, index alias, or `_all` value targets only missing or closed indices.
     * This behavior applies even if the request targets other open indices.
     * For example, a request targeting `foo*,bar*` returns an error if an index starts with `foo` but no index starts with `bar`.
     * @server_default true
     */
```

pquentin (approved):

PR #2858, opened 2024-09-04
Add Redact processor

pquentin:
When coupled with #2854, this makes all ingest APIs pass validation (specifically, a redact processor is used in `ingest.put_pipeline` and `ingest.simulate`)


specification/_doc_ids/table.csv
```diff
@@ -405,6 +405,7 @@ query-dsl-wildcard-query,https://www.elastic.co/guide/en/elasticsearch/reference
 query-dsl-wrapper-query,https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/query-dsl-wrapper-query.html
 query-dsl,https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/query-dsl.html
 realtime,https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/docs-get.html#realtime
+redact-processor,https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/redact-processor.html
```
flobernd, line 408: Unrelated, but I think I forgot to add these urls in my previous PR where I added the HtmlStripProcessor 😅

flobernd (approved):
LGTM! Just wondering if we should introduce a GrokPattern alias to allow the clients to implement custom functionality for this type.

pquentin:
> Just wondering if we should introduce a GrokPattern alias to allow the clients to implement custom functionality for this type.

We could, but we would need to do that for the `GrokProcessor` too. Would it be breaking?

flobernd:
It would not be breaking for .NET and Java as we simplify to the aliased type by default. Not 100% sure if that's the case for Go as well. cc @Anaethelion

specification/ingest/_types/Processors.ts
```diff
@@ -1046,6 +1053,42 @@ export class PipelineProcessor extends ProcessorBase {
...
+  field: Field
+  /**
+   * A list of grok expressions to match and redact named captures with
+   */
+  patterns: GrokPattern[]
+  /*
+   * A map of pattern-name and pattern tuples defining custom patterns to be used by the processor.
+   * Patterns matching existing names will override the pre-existing definition
+   */
+  pattern_definitions?: Dictionary<string, string>
```
flobernd, line 1069: Should the value part be of type `GrokPattern` as well here?

pquentin: No, because they're not grok patterns, see this example: https://www.elastic.co/guide/en/elasticsearch/reference/current/grok-processor.html#custom-patterns

PR #2861, opened 2024-09-05
Fix code style

pquentin:
#2829 was merged after #2853, and introduced an import error.


pquentin:
I wanted to backport #2853, not this one. Indeed, this pull request is only needed because of #2829 which is 8.16 only.

I removed the backport labels.

PR #2870, opened 2024-09-06
Adds cluster API example requests and responses

szabosteve:
## Overview

This PR adds request and response examples to the cluster APIs.


lcawl (approved):
LGTM. Let me know if you want me to add the updates to the overlay files in this same PR

specification/nodes/info/nodesInfoResponseExample1.json
```diff
@@ -0,0 +1,7 @@
...
...ansport_version\": 100000298,\n      \"index_version\": 100000074,\n      \"component_versions\": {\n        \"ml_config_version\": 100000162,\n        \"transform_config_version\": 100000096\n      },\n      \"build_flavor\": \"default\",\n      \"build_type\": \"{build_type}\",\n      \"build_hash\": \"587409e\",\n      \"roles\": [\n        \"master\",\n        \"data\",\n        \"ingest\"\n      ],\n      \"attributes\": {},\n      \"plugins\": [\n        {\n          \"name\": \"analysis-icu\",\n          \"version\": \"{version}\",\n          \"description\": \"The ICU Analysis plugin integrates Lucene ICU module into elasticsearch, adding ICU relates analysis components.\",\n          \"classname\": \"org.elasticsearch.plugin.analysis.icu.AnalysisICUPlugin\",\n          \"has_native_controller\": false\n        }\n      ],\n      \"modules\": [\n        {\n          \"name\": \"lang-painless\",\n          \"version\": \"{version}\",\n          \"description\": \"An easy, safe and fast scripting language for Elasticsearch\",\n          \"classname\": \"org.elasticsearch.painless.PainlessPlugin\",\n          \"has_native_controller\": false\n        }\n      ]\n    }\n  }\n}"
```
lcawl, line 6: >  "value": "{\n  \"_nodes\": ...\n  \"cluster_name\": \"elasticsearch\",\n  \"nodes\": {\n    

I think this one will fail because it's not a complete example ("...").

szabosteve: I modified the key:value pair to form a valid JSON. Is it good enough do you think for this particular example?

lcawl: That seems to work, thanks (though I needed to add two more escape characters around the quotes for it to work).

lcawl:
I added a *.yaml file to this PR for your consideration.  Right now all the *.json examples are returning linting errors like this:

```
 105666:17      warning  oas3-schema                 Property "type" is not expected to be here.                                         components.examples.nodesInfoResponseExample1.type
 105667:26      warning  oas3-schema                 Property "response_code" is not expected to be here.                                components.examples.nodesInfoResponseExample1.response_code
```

I tried converting the example to a YAML file then commenting out the extra fields (since yaml allows comments whereas json doesn't).  This succeeds in solving the linting errors (since the comment fields are not carried forward to the output).  Is this acceptable to use YAML instead of JSON format and is it okay to comment out those extra fields?

PR #2874, opened 2024-09-06
Add roles to cluster.allocation_explain response

pquentin:
As done in https://github.com/elastic/elasticsearch/pull/98550. The remaining failure is a NEST test that will be removed.


l-trotta:
In the spec we have `NodeRoles` that is an alias for `NodeRole[]`, to be coherent with the other uses I'd keep using it

pquentin:
Thanks, done!

PR #2877, opened 2024-09-08
Add missing tokenizers

pquentin:
And also:

 * Sort tokenizers alphabetically
 * Move NoriTokenizer to its own file like other plugins

I can split this in multiple commits if that would help review.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -74766,9 +74766,6 @@
             "$ref": "#/components/schemas/_types.analysis:PathHierarchyTokenizer"
           },
           {
-            "$ref": "#/components/schemas/_types.analysis:PathHierarchyPascalCaseTokenizer"
-          },
-          {
             "$ref": "#/components/schemas/_types.analysis:PatternTokenizer"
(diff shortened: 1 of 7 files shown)


l-trotta:
Where did you find PathHierarchyPascalCaseTokenizer? asking because I didn't find it in the server code or the documentation so I'm wondering if I'm missing some references. Also taking a look at the server code and discovering how many tokenizers/analyzers/normalizers we're missing... :')

pquentin:
> Where did you find PathHierarchyPascalCaseTokenizer?

There are currently two ways to use a Path hierarchy tokenizer: with `path_hierarchy` (snake_case, the default) and `PathHierarchy` (PascalCase, presumably for legacy reasons). OK, definitely for legacy reasons: https://github.com/elastic/elasticsearch/pull/15785. (Looks like Elasticsearch developers call it camelCase instead, but the idea is the same.)

When I initially added the PascalCase version I was simply fixing tests and did not stop to understand why we had such a  test. But now I'll make sure to skip that test in the flights recorder.

l-trotta (approved):
LGTM!

Issue #2144, opened 2023-06-12
Support for request and response body examples


lcawl:
## 🚀 Feature Proposal

It would be useful to be able to attach examples for any request and response specifications.

## Motivation

OpenAPI specifications make use of [examples](https://spec.openapis.org/oas/latest#example-object) to demonstrate valid use cases.
They are also useful for validating the accuracy of the specification. 

## Example

There are some examples for Kibana alerting OAS in https://github.com/elastic/kibana/tree/main/x-pack/plugins/alerting/docs/openapi/components/examples

They can be validated against the specification by using commands like these:

```
 npx swagger-cli validate entrypoint.yaml
npx @redocly/cli lint bundled.json
```

... which generate errors and warnings if the examples don't align with the specifications.

Per discussions at recent conference, it would be very nice if these examples were somehow tested against the code not just validated against the specification. There was discussion that this could be accomplished by validating them as part of the flight recorder(?), but I have no further details about this proposal.

lcawl:
My hope is that this can be as simple as creating (one or many) JSON/YAML files alongside the specifications, which contain examples for that request or response body.

For example, in the https://github.com/elastic/elasticsearch-specification/tree/main/specification/ml/put_job folder, it would be nice to simply add a file that contains the examples from the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-put-job.html#ml-put-job-example), in a format something like this:

MlPutJobRequestExample1.json:
````
{
  "summary": "Create an anomaly detection job and datafeed.",
  "description": "An optional additional description that can contain commonmark syntax.",
  "value":
    {
      "analysis_config": {
        "bucket_span": "15m",
        "detectors": [
          {
            "detector_description": "Sum of bytes",
            "function": "sum",
            "field_name": "bytes"
          }
...
````

MlPutJobResponseExample1.json:
````
{
  "summary": "A successful response with details about the new anomaly detection job. ",
  "value":
    {
    "job_id" : "test-job1",
    "job_type" : "anomaly_detector",
    "job_version" : "8.4.0",
    "create_time" : 1656087283340,
    "datafeed_config" : {
      "datafeed_id" : "datafeed-test-job1",
      "job_id" : "test-job1",
      "authorization" : {
        "roles" : [
          "superuser"
        ]
      ...
}
````

Then this information would be output in the *openapi.json file something like this (as described in https://spec.openapis.org/oas/v3.0.3#example-object)

````
...
[...]

PR #2882, opened 2024-09-11
Add examples to both OpenAPI documents via overlays

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2144, https://github.com/elastic/elasticsearch-specification/issues/2482

Until we can automatically include examples in the OpenAPI documents, they are being applied by using https://github.com/elastic/elasticsearch-specification/tree/main/docs/overlays

In preparation for publishing https://github.com/elastic/elasticsearch-specification/blob/main/output/openapi/elasticsearch-openapi.json, this PR splits the existing overlay into (1) an overlay that is specific to the serverless openAPI document, (2) an overlay that is specific to the non-serverless openAPI document, and (3) an overlay that contains any common updates such as examples that are applicable to both.

The make overlay-docs command is updated to apply all those overlays. I've also created a make lint-docs-errs command to only return linting errors rather than all the warnings and information messages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,5 +66,7 @@ output/schema/schema
 compiler/test/**/output/
 
 # Temporary openAPI files
-output/openapi/elasticsearch-serverless-openapi.tmp.json
+output/openapi/elasticsearch-serverless-openapi.tmp*.json
 output/openapi/elasticsearch-serverless-openapi.examples.json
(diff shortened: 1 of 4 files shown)


szabosteve (approved):
Thank you for this, LGTM!

PR #2883, opened 2024-09-11
Fix ingest statistics in cluster.info

pquentin:
Splitting IngestTotal in two allowed me to remove all optional fields. I have no idea if this can break typed clients or not.


l-trotta (approved):
LGTM and the java client

PR #2886, opened 2024-09-12
Add HTTP routes to cluster/nodes stats

pquentin:
This was added in https://github.com/elastic/elasticsearch/pull/99852. The `toXContent` code is fairly easy to follow in [HttpRouteStats.java](https://github.com/elastic/elasticsearch/blob/v8.15.1/server/src/main/java/org/elasticsearch/http/HttpRouteStats.java).

This also happens to fix the cluster.info API.


l-trotta (approved):
LGTM!

PR #2888, opened 2024-09-12
[Serverless] Update role name description

slobodanadamovic:
I've update the role name parameter description to point out the restrictions we have. But since the specification is generated from `SecurityPutRoleRequest.ts` (and is shared), it makes this proposed change true for Serverless, but completely wrong for stateful.

Also this change is fragile, anytime someone runs `make contrib` it will override the specs for stateful as well.

It's not nice, but I could potentially word it in a way to mention that this restriction only applies in Serverless. LMKWYT?

Ideally, we should have this documented in the dedicated docs for Serverles.


n1v0lg (approved):
LGTM -- just in case, can you run it my @pquentin before merging? (Only when we're ready to merge ofc)

pquentin (approved):
Thanks! We don't have the option to have serverless-specific descriptions. Any reason not to add "On Elasticsearch Serverless, the role name must ..."?

slobodanadamovic:
> Any reason not to add "On Elasticsearch Serverless, the role name must ..."?

I wasn't sure if there is a better way to document this and avoid mentioning Serverless. But I'm fine with this proposal.

pquentin (approved):

Issue #2310, opened 2023-10-10
Add support for tag groupings


lcawl:
## 🚀 Feature Proposal

Add support for "tags" on each operation in the specification.

## Motivation

Per https://spec.openapis.org/oas/latest#operationObject, "Tags can be used for logical grouping of operations by resources or any other qualifier".

Since we have such a long list of Elasticsearch APIs, it would be helpful to group them. For example, we could use the groupings that already exist in the documentation (https://www.elastic.co/guide/en/elasticsearch/reference/master/rest-apis.html) such as machine learning APIs, enrich APIs, Fleet APIs, etc.

At the moment, the tags that are generated are too low-level for our doc purposes (e.g. "async_search.get").

NOTE: OpenAPI specifications also optionally have a global list of the tags with descriptions for each of them per https://spec.openapis.org/oas/v3.0.3#tag-object (and this list can be used to [organize the API documentation in Bump.sh](https://docs.bump.sh/help/customization-options/operations-navigation/#group-operations-by-tag)).  We'll ultimately want that but it could feasibly be added as a overlay via post-processing. IMO the more important action item is to curate the operation-level tags.

## Example

It would be nice to have something like `@tag` that we could add as a modifier, like what exists in typespec.io per https://github.com/elastic/elasticsearch-specification/issues/2310#issuecomment-2076076208

lcawl:
There are now tags in the specifications (e.g. "async_search.get" in https://github.com/elastic/elasticsearch-specification/blob/7e140693f2bbd392b95f54f2cd2b0c0895db2e5d/output/openapi/elasticsearch-serverless-openapi.json#L15), however I think from a docs point of view those values are too low-level. I suspect we'd want "asynch_search", "ml", "cat" tags instead.

lcawl:
Per https://typespec.io/docs/getting-started/typespec-for-openapi-dev#tags in TypeSpec "Tags can be specified using the @tag decorator on an operation".

flobernd:
As discussed today, it would probably be sufficient to use the existing namespace (e.g. async_search) as the tag. This could be done in the OAPI generator and would not require spec changes.

If required, the @tag functionality could be introduced to the spec at a later point to allow overriding the tag/group without changing the namespace.

lcawl:
> As discussed today, it would probably be sufficient to use the existing namespace (e.g. async_search) as the tag. This could be done in the OAPI generator and would not require spec changes.

That would be great!

[...]

PR #2889, opened 2024-09-12
Use a simplified namespace value for the OpenAPI tags

lcawl:
This is an attempt to accomplish simplified OpenAPI tags per https://github.com/elastic/elasticsearch-specification/issues/2310#issuecomment-2263773091

The intention is to have tags like "asynch_search" instead of "async_search.delete"


l-trotta (approved):

lcawl:
> Looks good! Just to be sure, the second part of the tag isn't needed at all?

That's right!

PR #2894, opened 2024-09-17
Fix get field mapping API response

pquentin:
The [default value](https://github.com/elastic/elasticsearch/blob/8e289f2b10f86959393b43bbf6c77abeb3a12622/server/src/main/java/org/elasticsearch/index/mapper/TextFieldMapper.java#L274-L277) of the indexPrefixes [Parameter](https://github.com/elastic/elasticsearch/blob/8e289f2b10f86959393b43bbf6c77abeb3a12622/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java#L764-L785) is null, which is what gets returned with the `include_defaults` parameter.

I'm not sure why `similarity` can be null (other than seeing it in a YAML test), but I did notice that https://github.com/elastic/elasticsearch/pull/58439 (shipped to 8.0) moved it so that it only applies to text, keyword and search as you type.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -60539,9 +60539,6 @@
               "copy_to": {
                 "$ref": "#/components/schemas/_types:Fields"
               },
-              "similarity": {
-                "type": "string"
-              },
               "store": {
                 "type": "boolean"
               }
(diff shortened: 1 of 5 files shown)


flobernd (approved):
LGTM!

PR #2896, opened 2024-09-17
Fix make validate generation

pquentin:
Without this fix, `make validate` does not take into account specification changes as it was generating them in the wrong directory. (You can check that this is the correct directory in the Makefile.) This is also why we had to add output/schema/schema to the .gitignore file.


flobernd (approved):
LGTM! Good catch.

PR #2900, opened 2024-09-17
Adds graph and ILM API example requests and responses

szabosteve:
## Overview

This PR adds request and response examples to the graph and ILM APIs.


lcawl (approved):
I renamed a few files to add ILM to the title so they're more unique; I also added the overlays for these examples. LGTM!

PR #2901, opened 2024-09-17
Add minimum_should_match to terms_set

mspielberg:
Elasticsearch has supported the `minimum_should_match` parameter to `terms_set` since 8.10.0 (see https://github.com/elastic/elasticsearch/pull/96082), but it is not reflected here, so the parameter is not exposed in the autogenerated clients.

Closes https://github.com/elastic/elasticsearch-specification/pull/2908

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -46869,6 +46869,9 @@
           {
             "type": "object",
             "properties": {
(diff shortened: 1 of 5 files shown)


mspielberg:
I have signed the CLA, but it does not appear to be (yet?) reflected by the automation.

pquentin (approved):
Thanks! LGTM.

PR #2905, opened 2024-09-18
[OpenAPI] Add list of tags in overlay

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2310

As a follow-on to https://github.com/elastic/elasticsearch-specification/pull/2961, this PR adds `@doc_tag` values where we want to override the default tags. There are a few remaining (e.g. `health_report`, `mget`, `mtermvectors`) that we can't override yet because the typescript files for those APIs don't exist yet.

This PR also adds the missing document-level list of tags via an overlay. Doing this in an overlay is not perfect since it means that new tags must be manually added. Any endpoints associated with tags that do not appear in this list will not be published on Bump.sh. I've therefore also changed the [operation-tag-defined linting rule](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#operation-tag-defined) to an error instead of a warning.

I've also changed the make commands to only lint the OpenAPI documents that have overlays applied, since the others will contain errors that are fixed by the overlays.

NOTE: Even though the list of tags used differs in the two OpenAPI documents (e.g. some endpoints do not exist in the serverless context), it seems okay to use a single tag list since I've tested in Bump.sh and any tags for which there aren't any endpoints are simply omitted.


docs/overlays/elasticsearch-shared-overlays.yaml
```diff
@@ -4,6 +4,274 @@ info:
...
+      # C 
+        - name: cat
+          x-displayName: Compact and aligned text (CAT)
+          description: >
+            The compact and aligned text (CAT) APIs aim are intended only for human consumption using the Kibana console or command line.
+            They are not intended for use by applications.
+            For application consumption, it's recommend to use a corresponding JSON API.
+            
+            All the cat commands accept a query string parameter `help` to see all the headers and info they provide, and the `/_cat` command alone lists all the available commands.
+        - name: cluster
```
shainaraskas, line 24: unsure why, but items at the `/_component_template` endpoint are displaying as ` cluster` items. I see that reflected [in the tag fix PR](https://github.com/elastic/elasticsearch-specification/pull/2889/files#diff-3a3f55821e7fc4208762783b2a0635ea25a647938275ac13931747a2e8b6498dR3227), but this one doesn't make as much logical sense to me. Could the tags be updated in cases like this?

docs/overlays/elasticsearch-shared-overlays.yaml
shainaraskas, line 1: missing categories / perhaps endpoints: 

[`/_application/analytics/...`](https://www.elastic.co/guide/en/elasticsearch/reference/current/behavioral-analytics-apis.html) (Behavioral analytics)

[`/_data_stream/...`](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-stream-apis.html) (Data streams)

[`/`](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-api-root.html) hilariously, root - after being the biggest section forever


ML miiiiight be too big? Wonder what it looks like to add some nuance to the tags for cases where we have like 100 endpoints

shainaraskas (approved):
Couple of comments on funny things, otherwise LGTM

also noting for posterity that trying to preview this with `make overlay-docs` didn't work for me locally and might need some love to get runnable

docs/overlays/elasticsearch-shared-overlays.yaml
lcawl: I've drafted https://github.com/elastic/elasticsearch-specification/pull/2961 to enable us to override the tag values (e.g. to put "data stream" instead of defaulting to "indices").  Seems like it works 🤞

docs/overlays/elasticsearch-shared-overlays.yaml
lcawl: That seems to be derived from the original spec in https://github.com/elastic/elasticsearch/blob/main/rest-api-spec/src/main/resources/rest-api-spec/api/cluster.get_component_template.json, though I'm not sure why. It's another scenario that we could override if https://github.com/elastic/elasticsearch-specification/pull/2961 works

lcawl: I've added @doc_tag for those component template APIs in 3cba8c9210e49d97a8566737fbdeccd6634dd878

PR #2906, opened 2024-09-18
Accept any FieldValue for terms_set

mspielberg:
Change `terms_set` definition to accept any FieldValue, as the `terms` query already does, and as the Elasticsearch server implementation already does.

This is not a breaking change at the Typescript level, as FieldValue includes `string` as one of its alternatives, but the materialization of the change may result in a breaking API change in specific language (e.g. Java) compiled clients.


l-trotta:
Looks correct, we probably got this from the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-set-query.html#terms-set-field-params), which says that `terms` is a string array. Going from String to FieldValue breaks the java constructor for this, could we not backport this and leave it for 9? @flobernd @pquentin

flobernd:
Fine with not backporting this for now. For C# it's not a breaking change as we have an implicit conversion operator from string -> FieldValue.

mspielberg:
As a side note, I already submitted https://github.com/elastic/elasticsearch/pull/113043 to correct the documentation as well.

l-trotta (approved):
LGTM! Thanks for the contribution @mspielberg

l-trotta:
correctly validated [here](https://github.com/elastic/elasticsearch-specification/pull/2929), merging

PR #2914, opened 2024-09-19
Add overlay for deep object in security APIs

lcawl:
This PR adds the `x-model` extension for the role template query schema in the role APIs, since it is too deep to display well at this time. It also adds a link to https://www.elastic.co/guide/en/elasticsearch/reference/master/field-and-document-access-control.html#templating-role-query


szabosteve (approved):
LGTM!

PR #2916, opened 2024-09-19
[ML] Add new calendar event fields

davidkyle:
ML Calendar Events were updated with new fields in https://github.com/elastic/elasticsearch/pull/112837, this is the corresponding PR to update the spec

### DIFF
--- a/specification/ml/_types/CalendarEvent.ts
+++ b/specification/ml/_types/CalendarEvent.ts
@@ -18,6 +18,7 @@
  */
 
 import { Id } from '@_types/common'
+import { integer } from '@_types/Numeric'
 import { DateTime } from '@_types/Time'
 
 export class CalendarEvent {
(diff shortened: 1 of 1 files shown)


specification/ml/_types/CalendarEvent.ts
```diff
@@ -30,4 +31,10 @@ export class CalendarEvent {
   end_time: DateTime
   /** The timestamp for the beginning of the scheduled event in milliseconds since the epoch or ISO 8601 format. */
   start_time: DateTime
+  /** When true the model will not create results for this calendar period. */
+  skip_result: boolean
+  /** When true the model will not be updated for this calendar period. */
+  skip_model_update: boolean
+  /** Shift time by this many hours. For example adjust time for daylight savings changes */
+  force_time_shift: integer
```
pquentin, line 39: Thanks! As you can see, the calendar events validation still fails, because those fields are currently marked as required.

While `force_time_shift` is the [only optional field](https://github.com/elastic/elasticsearch/blob/4ff4384550bf5e038447ed263e84a0d03babb775/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/calendars/ScheduledEvent.java#L222-L226) *in the response*, this class is used both for response and request, so they should all be optional. And it's good practice to specify the default values in those case:

```suggestion
  /** When true the model will not create results for this calendar period.
   * @server_default true
   */
  skip_result?: boolean
  /** When true the model will not be updated for this calendar period.
   * @server_default true
   */
  skip_model_update?: boolean
  /** Shift time by this many hours. For example adjust time for daylight savings changes */
  force_time_shift?: integer
```

pquentin (changes requested):

PR #2930, opened 2024-09-24
Approximate snapshot verify integrity mapping

l-trotta:
Mapping the new [repository integrity verification API](https://github.com/elastic/elasticsearch/pull/112348), allowing to merge the [latest spec update](https://github.com/elastic/elasticsearch-specification/pull/2890) without introducing new validation errors.

The query parameter `max_bytes_per_sec` is a string, but it has a specific format: `<<size-units, size units>>`. Considering it's the same in the DevTools I think we can keep it as a string for now.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -33298,6 +33298,125 @@
         "x-available-since": "0.0.0"
       }
     },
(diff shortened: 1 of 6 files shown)


flobernd (approved):
LGTM from my side as well.

PR #2949, opened 2024-09-27
Fetch JSON spec from Artifacts Snapshot API

pquentin:
We were still using the Artifacts API, which has no notion of "8.x", which is why builds were still failing: https://github.com/elastic/elasticsearch-specification/actions/runs/11064313424/job/30741882039.

This can be tested with:

```bash
make setup
npm install .github/download-artifacts
node .github/download-artifacts/index.js --branch 8.x
git status   # notice that we did change specification/_json_spec, it's just not formatted yet
make contrib
git status   # after formatting, there's no change left
```

### DIFF
--- a/.github/download-artifacts/index.js
+++ b/.github/download-artifacts/index.js
@@ -40,23 +40,21 @@ const downloadedSpec = join(esFolder, 'rest-api-spec', 'api')
 const specFolder = join(__dirname, '..', '..', 'specification', '_json_spec')
 
 async function downloadArtifacts (opts) {
-  if (typeof opts.version !== 'string' && typeof opts.branch !== 'string') {
-    throw new Error('Missing version or branch')
+  if (typeof opts.branch !== 'string') {
+    throw new Error('Missing branch')
   }
 
-  core.info('Checking out spec and test')
+  core.info('Resolving artifact URL')
 
-  core.info('Resolving version')
   let resolved
   try {
-    resolved = await resolve(opts.version || fromBranch(opts.branch), opts.hash)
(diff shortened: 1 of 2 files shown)


.github/workflows/update-rest-api-json.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: Generate output
         run: |
-          SKIP_VERSION_UPDATE=true make contrib
```
pquentin, line 40: This had no effect for some time now.

.github/download-artifacts/index.js
```diff
@@ -172,7 +122,7 @@ async function main (options) {
 }
 
 const options = minimist(process.argv.slice(2), {
-  string: ['id', 'version', 'hash', 'branch']
+  string: ['branch']
```
pquentin, line 125: We only use `--branch` now.

.github/download-artifacts/index.js
```diff
@@ -96,74 +94,26 @@ async function downloadArtifacts (opts) {
...
-    if (!response.ok) {
-      throw new Error(`unexpected response ${response.statusText}`)
-    }
-    const { versions } = await response.json()
-    version = versions.pop()
-  }
-
-  core.info(`Resolving version ${version}`)
-  const response = await fetch(`https://artifacts-api.elastic.co/v1/versions/${version}`)
+async function resolve (branch) {
```
pquentin, line 97: This is largely copied over from clients-flight-recorder.

JoshMock (approved):

PR #2951, opened 2024-09-27
Fix indices.delete_alias response

pquentin:
This regressed between March 25th and April 15th, back when recording examples wasn't stable. Unfortunately, 22 days worth of commits is a lot: https://github.com/elastic/elasticsearch/compare/e9fcb0a0f0f45773248c5258982d6bf2d5517299...d3e0ec157dc0d12d529057ec86170a9e7a5b6b7d. If we find where this is set, we could consider asking the Elasticsearch team to *not* do this. It's not helpful, as `acknowledged` is already set to true.

Another option: I could only change the delete_alias body instead of affecting all APIs that use AcknowledgedResponseBase.


l-trotta:
I can't find this in the sever, [AcknowledgeResponse](https://github.com/elastic/elasticsearch/blob/e77dc6dc1324b00ba2ddf15718686e8bd1bcb3a9/server/src/main/java/org/elasticsearch/action/support/master/AcknowledgedResponse.java#L35C14-L35C34) only has one field. I think the problem is the fact that the body changes when there's an error, so for a normal response it's:

```
{
  "acknowledged": true
}
```

while when there's an error it's:

```
{
  "error": {
    "root_cause": [
      {
        "reason": "unable to put stored script with unsupported lang [painlesss]"
      }
    ],
    "reason": "unable to put stored script with unsupported lang [painlesss]"
  },
  "status": 400
}
```

in the spec we just map the first one, because the second case is the exception case (the server returns 404) so what java does is that it returns the error body in the exception message.

pquentin:
Actually, we also get responses like this one:

```
{
  "api": "indices.delete_alias",
  "file": "/test/free/indices.delete_alias/all_path_options.yml",
  "name": "check delete with prefix* index",
  "origin": "yaml",
  "request": {
    "args": {
      "index": "test_*",
      "name": "alias1"
    }
  },
  "response": {
    "headers": {
      "content-encoding": "gzip",
      "content-length": "62",
    },
    "payload": {
      "acknowledged": true,
      "errors": false
    },
    "statusCode": 200
  }
}
```

l-trotta:
ooh that's why: it's a [subclass](https://github.com/elastic/elasticsearch/blob/e3a91d30e64a1c61a9f6af7403e56d18ceae1426/server/src/main/java/org/elasticsearch/action/admin/indices/alias/IndicesAliasesResponse.java#L35)!  we have to map this

l-trotta:
update: it's more complex than it looks, I'm still working on this. in particular it seems like in case of error it can return detailed information, including a class that represents an ElasticsearchException which I don't think we already have.

l-trotta:
@pquentin sorry I forgot about this, rebased it + added a commit with what I think is the solution

pquentin:
@l-trotta Sorry for not getting back to you sooner here. Should we add `action_results` too? https://github.com/elastic/elasticsearch/pull/106514

pquentin:
Never mind, `action_results` only applies to `indices.update_aliases.

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/4971)|

### Questions ?

PR #2953, opened 2024-09-27
Fix indices.get_alias error response

pquentin:
Our recordings show two different errors paths. The first one is supported:

```json
{
  "api": "indices.get_alias",
  "file": "/test/free/indices.get_alias/10_basic.yml",
  "name": "Getting alias on an non-existent index should return 404",
  "origin": "yaml",
  "request": {
    "args": {
      "index": "non-existent",
      "name": "foo"
    }
  },
  "response": {
    "headers": {
      "content-encoding": "gzip",
      "content-length": "164",
    },
    "payload": {
      "error": {
        "index": "non-existent",
        "index_uuid": "_na_",
        "reason": "no such index [non-existent]",
        "resource.id": "non-existent",
        "resource.type": "index_or_alias",
        "root_cause": [
          {
            "index": "non-existent",
            "index_uuid": "_na_",
            "reason": "no such index [non-existent]",
            "resource.id": "non-existent",
            "resource.type": "index_or_alias",
            "type": "index_not_found_exception"
          }
        ],
        "type": "index_not_found_exception"
      },
      "status": 404
    },
    "statusCode": 404
  }
}
```

But the second one was not supported:

```json
{
  "api": "indices.get_alias",
  "file": "/test/free/indices.delete_alias/all_path_options.yml",
  "name": "check delete with * index",
  "origin": "yaml",
  "request": {
    "args": {
      "name": "alias1"
    }
  },
  "response": {
    "headers": {
      "content-encoding": "gzip",
      "content-length": "70",
    },
    "payload": {
      "error": "alias [alias1] missing",
      "status": 404
    },
    "statusCode": 404
  }
}
```

Not sure if changing `ErrorResponseBase` this way is OK or not.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -65172,7 +65172,14 @@
         "type": "object",
         "properties": {
           "error": {
-            "$ref": "#/components/schemas/_types:ErrorCause"
+            "oneOf": [
+              {
+                "$ref": "#/components/schemas/_types:ErrorCause"
+              },
+              {
+                "type": "string"
+              }
+            ]
           },
           "status": {
             "type": "number"
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -52351,7 +52351,14 @@
         "type": "object",
         "properties": {
           "error": {
-            "$ref": "#/components/schemas/_types:ErrorCause"
+            "oneOf": [
+              {
+                "$ref": "#/components/schemas/_types:ErrorCause"
+              },
+              {
+                "type": "string"
+              }
+            ]
           },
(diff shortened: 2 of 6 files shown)


l-trotta:
this is probably breaking for all static clients generator since it's a new union

flobernd:
Does this only apply to `indices.get_alias`?

pquentin:
Too bad. Can you think of a non breaking way to fix this? The common case is actually the second one, but of course less common than the happy path

pquentin:
@flobernd Sesrching the docs did turn up a few cases, but it seems pretty non standard.

l-trotta:
@pquentin let's bring this up in the next meeting, it's should be easy to map this, but we all have to be aware of the change

pquentin:
Closing as we'll be skipping this test instead.

PR #2956, opened 2024-09-27
7.17 - Fix nodes security response

l-trotta:
If security is disabled all other values are null, so `http` has to be optional as well


pquentin (approved):
Thanks! LGTM.

PR #2959, opened 2024-09-27
Java issues batch 8

l-trotta:
- [877](https://github.com/elastic/elasticsearch-java/issues/877): some optional values in ngram and edge tokenizer (details and server code references in the issue).
- [880](https://github.com/elastic/elasticsearch-java/issues/880): ForceMergeRequest accepts `wait_for_competion`, so the response should be optional to handle both `wait_for_competion` cases.
- [881](https://github.com/elastic/elasticsearch-java/issues/881): added missing `allow_custom_routing` to PutIndexTemplate. [server code](https://github.com/elastic/elasticsearch/blob/7b523568fac465f989910d7f58ad6abae68c05ff/server/src/main/java/org/elasticsearch/cluster/metadata/ComposableIndexTemplate.java#L366) - `failure_store` is internal.
- [883](https://github.com/elastic/elasticsearch-java/issues/883): we use StepKey as both current_step and next_step since they have the same fields, but in next_step `action` and `name` are optional. not super clear from the [server code](https://github.com/elastic/elasticsearch/blob/db632ee3cd355c8f50ec716ef9911f76b76016b7/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/Step.java#L97) (it's all String), but the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-move-to-step.html#ilm-move-to-step-request-body) confirms it (and DevTools test as well). 

Everything should be safe to backport to 8.x and 8.15

### DIFF
--- a/specification/ilm/move_to_step/types.ts
+++ b/specification/ilm/move_to_step/types.ts
@@ -18,7 +18,8 @@
  */
 
 export class StepKey {
-  action: string
-  name: string
+  // action and name are optional in case they are used in next_step
+  action?: string
+  name?: string
   phase: string
 }
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -9643,7 +9643,11 @@
                   "next_step": {
                     "$ref": "#/components/schemas/ilm.move_to_step:StepKey"
                   }
-                }
+                },
+                "required": [
+                  "current_step",
+                  "next_step"
+                ]
               }
             }
           }
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -67313,6 +67313,9 @@
         "properties": {
           "hidden": {
             "type": "boolean"
+          },
+          "allow_custom_routing": {
+            "type": "boolean"
           }
         }
       },
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -66738,10 +66738,7 @@
           "_shards": {
             "$ref": "#/components/schemas/_types:ShardStatistics"
           }
-        },
-        "required": [
-          "_shards"
-        ]
+        }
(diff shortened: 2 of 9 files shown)


specification/ilm/move_to_step/types.ts
```diff
@@ -18,7 +18,8 @@
  */
 
 export class StepKey {
-  action: string
-  name: string
+  // action and name are optional in case they are used in next_step
+  action?: string
+  name?: string
   phase: string
 }
```
pquentin, line 25: Should we make a `NextStepKey` class? Just wondering, really.

pquentin (approved):
Thanks! LGTM. Sorry that I missed it earlier.

specification/ilm/move_to_step/types.ts
l-trotta: was wondering the same, that would be a breaking change though and it would have to wait

pquentin: Agreed! Can you please add it to https://github.com/elastic/elasticsearch-specification/issues/2973?

And feel free to merge this pull request

flobernd: Not 100% related, but juding from the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-move-to-step.html), it seems that both, `current_step` and `next_step` should be required.

We currently model them as optional:

```ts
/**
 * @rest_spec_name ilm.move_to_step
 * @availability stack since=6.6.0 stability=stable
 */
  path_parts: {
    index: IndexName
  }
  body: {
    current_step?: StepKey
    next_step?: StepKey
  }
}
```

flobernd (approved):

specification/ilm/move_to_step/types.ts
l-trotta: good catch! I'll remove the optional

PR #2960, opened 2024-09-27
Allow script to have string or object value

l-trotta:
Originally reported as a [Java Client issue](https://github.com/elastic/elasticsearch-java/issues/876).

PutScript API support either the `source` of the script as a string, or as a Search Template object, which is equivalent to [the whole SearchRequest body](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#create-search-template).

Since the Search Request body is also needed by Multisearch, it was extracted as a separate class in the search package to that both Multisearch and StoredScript can use it (it was also ordered alphabetically).

`source` in StoredScript is now an externally tagged union. (potentially breaking for generators)

Ideally nothing else should be changed in the client codes as a result of this PR except for StoredScript, I'm setting this up as a draft for now, let me know when the generators are ready.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -29894,8 +29894,7 @@
         "type": "object",
         "properties": {
           "source": {
-            "description": "The script source.",
-            "type": "string"
+            "$ref": "#/components/schemas/_types:ScriptSource"
           },
           "id": {
             "$ref": "#/components/schemas/_types:Id"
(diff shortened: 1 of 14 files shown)


pquentin (approved):
Since this does not affect elasticsearch-py, I'm approving.

However since it could affect elasticsearch-dsl-py, I added @miguelgrinberg to the list of reviewers.

miguelgrinberg (approved):
LGTM. And the Python DSL client is not affected by this change.

l-trotta:
thank you both! I'll remember the python dsl for future issues

JoshMock:
Shouldn't cause any issues for the JS generator either. :+1:

flobernd:
The same change should be applied to `Script.source`, `RenderSearchTemplate.source`, `PhraseSuggestCollateQuery.source` (and probably more places).

PR #2961, opened 2024-09-27
Add @doc_tag to override default tag behaviour

lcawl:
Fixes https://github.com/elastic/elasticsearch-specification/issues/2310

This PR adds a @doc_tag modifier so that we can override the default operation-level tags that are based on the namespace.

For example, we can use it to put a "data stream" tag on the create data stream API to match its grouping in the existing docs (https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-data-stream.html) rather than the default "indices" value.


Anaethelion (approved):
LGTM
I've left a code suggestion to avoid using `expect` which simplifies a bit the code.

PR #2965, opened 2024-09-30
Add trace_redact to RedactProcessor

pquentin:
As added in https://github.com/elastic/elasticsearch/pull/113552.

### DIFF
--- a/specification/ingest/simulate/types.ts
+++ b/specification/ingest/simulate/types.ts
@@ -27,8 +27,18 @@ import { ErrorCause } from '@_types/Errors'
 import { DateTime } from '@_types/Time'
 
 export class Ingest {
+  _redact?: Redact
   timestamp: DateTime
   pipeline?: Name
+}
+
+export class Redact {
+  /**
+   * indicates if document has been redacted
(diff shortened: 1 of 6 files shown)


l-trotta (approved):

specification/ingest/simulate/types.ts
```diff
@@ -27,10 +27,20 @@ import { ErrorCause } from '@_types/Errors'
 import { DateTime } from '@_types/Time'
 
 export class Ingest {
+  _redact?: Redact
```
flobernd, line 30: Should we add an availability annotation here as well? It probably makes no sense to have an empty class in pre 8.16 versions e.g. in case this gets backported.

flobernd (approved):

specification/ingest/simulate/types.ts
l-trotta: wait how could this get backported to 8.15?

pquentin: In any case it does make sense to add the annotation at this level I think?

PR #2975, opened 2024-10-02
Add x-model for QueryContainer

lcawl:
This PR replaces occurrences of `x-abbreviated` with the `x-model` extension, which is being implemented by Bump.sh as a work-around for overly deep schema objects.

It also adds this extension to `_types.query_dsl:QueryContainer` since those details are more appropriately discussed in the Query DSL docs, which are linked as `externalDocs`.


szabosteve (approved):
LGTM!

PR #2976, opened 2024-10-02
Add x-model overlays for tokenizer, filter, charfilter schemas

lcawl:
This PR adds the `x-model` extension to tokenizer, filter, and char_filter schemas in the OpenAPI documents, since those are more appropriate described in the tokenizer reference docs.
This matches the handling of those items in the current [analyze API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html) docs, for example.


kilfoyle (approved):
LGTM! 🦏

Issue #2714, opened 2024-07-12
Add stability to OpenAPI output 


lcawl:
## 🚀 Feature Proposal

Relates to https://github.com/elastic/elasticsearch-specification/issues/22

Per https://github.com/elastic/elasticsearch-specification/blob/main/docs/add-new-api.md, the `@availability` tag includes:

> stability: the API stability, one of experimental, beta, stable

However I do not see this reflected in the OpenAPI documents in any way.

Since https://github.com/elastic/elasticsearch-specification/pull/2655 recently added an `x-available-since` qualifier from the `@availability since` details, I'm hoping we can likewise extract the `stability`.

## Motivation

There's currently no automated way to reflect the stability in the docs. We can add it to the descriptions but then folks need to remember to keep them up-to-date.

There's an additional motivation for us docs-wise since the presence of this type of info affects what is flagged as a "breaking change" in Bump.sh, so we'd want to ensure any beta/experimental stuff can be handled appropriately.

### Beta

Per https://docs.bump.sh/help/specification-support/doc-beta/ the Bump.sh tool we use to publish the openAPI documents already provides an `x-beta` extension, so it would just be a matter of adding that `x-beta: true` each time an operation or property requires that qualification.

### Experimental 

Per https://github.com/elastic/docs?tab=readme-ov-file#beta-dev-and-preview-experimental we refer to "experimental" externally as "technical preview" now.  So for this one, I'd suggest adding a `x-technical-preview: true`.   We can then work with Bump.sh to display that extension similarly to how they're handling `x-beta`.

### Stable

IMO this should be the default assumption for all the APIs in our documentation. Therefore I don't think any action is required for this scenario.

### Dev

Per https://github.com/elastic/docs?tab=readme-ov-file#beta-dev-and-preview-experimental there's also an "in development" state but I've never seen it used so unless someone else is aware of a need for it, I think this can be omitted for now.

Issue #2609, opened 2024-06-07
Add x-availability-since to OpenAPI output


lcawl:
The specifications currently include an "availability since" piece of information
(https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#availability) which would be useful information to include in the docs too.

We're in discussion with the Bump.sh folks who'll be hosting our docs to add support for this info in the form of a [specification extension](https://spec.openapis.org/oas/v3.0.3#specification-extensions) such as "x-available-since". Can the openapi generator include that detail in the output?

PR #2989, opened 2024-10-07
Output x-state instead of x-technical-preview and x-available-since

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2714, https://github.com/elastic/elasticsearch-specification/issues/2609

Rather than implementing `x-technical-preview` and `x-available-since`, Bump.sh is going to use `x-state`. This PR updates the openAPI generator to reflect that change.


l-trotta (approved):
LGTM!

PR #2994, opened 2024-10-08
Make Inference API task_settings optional

davidkyle:
`task_settings` are optional when creating an endpoint in the Inference API and may not be returned in the GET response. The spec incorrectly has `task_settings` as required for both creation (PUT) and retrieval (GET), this PR changes that

The Elasticsearch documentation contains an example where an endpoint is created without task_settings ,see [creating the ELSER service](https://www.elastic.co/guide/en/elasticsearch/reference/current/infer-service-elser.html#inference-example-elser).

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -69014,8 +69014,7 @@
         },
         "required": [
           "service",
-          "service_settings",
-          "task_settings"
+          "service_settings"
         ]
       },
       "inference._types:ServiceSettings": {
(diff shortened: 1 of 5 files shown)


pquentin:
Thanks! This pull request makes a response field optional, which is why I asked a review from static client maintainers here.

l-trotta (approved):
no problems for java

Anaethelion (approved):

pquentin (approved):
After discussion with the team, the conclusion is that making a response field optional is OK. User code handling the case where task_settings is set won't change. (Except in Rust, where it won't compile because the value will be wrapped in std::Option. But the Rust client isn't ready yet.)

Making a field optional in requests is actually more of an issue, because that means constructors need to change to include a variant without that field. Before that change, users will be forced to set this field.

But in any case, since this a bug fix to a tech preview API, this pull request is fine. I would guess that today multiple clients would be failing if given a payload without task settings anyway.

PR #2996, opened 2024-10-09
Adds ingest API request and response examples

szabosteve:
## Overview

This PR adds request and response examples to the ingest API.


specification/ingest/put_pipeline/PutPipelineRequestExample2.yaml
```diff
@@ -0,0 +1,11 @@
+summary: Creates an ingest pipeline.
```
lcawl, line 1:
```suggestion
summary: Create an ingest pipeline with metadata.
```

Since there are two examples for this operation, the summaries will appear in a drop-down selector and should help clarify how they're different.

lcawl (approved):
One suggestion, otherwise LGTM

PR #2998, opened 2024-10-09
Xpack settings fixes

l-trotta:
The ml field in NodeInfoXpackSecurity is used in start-local.
A java user [reported](https://github.com/elastic/elasticsearch-java/issues/892) that having an anonymous realm means the realm field is null (and according to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/anonymous-access.html), probably also the token field).

Unfortunately I couldn't find any server side proof, let's see what the validation says.


pquentin (approved):
Thanks! LGTM.

PR #2999, opened 2024-10-09
Fixing snapshot restore response

l-trotta:
This is one of the many endpoints which changes its response depending if `wait_for_completion` was set as true or false in the request. [server code](https://github.com/elastic/elasticsearch/blob/21e3a17e5b70a26c6703590d10bd4305ca81425b/server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/restore/RestoreSnapshotResponse.java#L64)


flobernd (approved):
LGTM! I guess that's the only viable solution for now.

PR #3005, opened 2024-10-10
Update Serverless scope for legacy put_template

picandocodigo:
From @pquentin's research:

~`tasks.get` - This one is not actually `tasks.get` (with endpoint `/_tasks/{task_id}`), it's `tasks.list` (with endpoint `/_tasks/`), [which is INTERNAL](https://github.com/elastic/elasticsearch/blob/c27bc086770b12a71c335e5537318f54cce53a65/server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestListTasksAction.java#L35-L46).~

`indices.put_template` - The /_template URLs refer to legacy index templates, that are not available on Serverless.


specification/tasks/get/GetTaskRequest.ts
```diff
@@ -26,7 +26,7 @@ import { Duration } from '@_types/Time'
  * Returns information about the tasks currently executing in the cluster.
  * @rest_spec_name tasks.get
  * @availability stack since=5.0.0 stability=experimental
- * @availability serverless stability=experimental visibility=public
+ * @availability serverless stability=experimental visibility=private
```
pquentin, line 29:
```suggestion
 * @availability serverless stability=experimental visibility=public
```

`tasks.get` *is* public, but `tasks.list` is not, but this is already correct: https://github.com/elastic/elasticsearch-specification/blob/afcfe48dedd1e40515261117540927d3a90df15b/specification/tasks/list/ListTasksRequest.ts#L29

specification/indices/put_template/IndicesPutTemplateRequest.ts
```diff
@@ -31,7 +31,7 @@ import { Duration } from '@_types/Time'
  * Index templates define settings, mappings, and aliases that can be applied automatically to new indices.
  * @rest_spec_name indices.put_template
  * @availability stack stability=stable
- * @availability serverless stability=stable visibility=public
+ * @availability serverless stability=stable visibility=private
```
pquentin, line 34: For consistency with the other legacy template APIs, can we simply remove the serverless line?

```suggestion
```

pquentin (commented):
Thanks!

specification/tasks/get/GetTaskRequest.ts
picandocodigo: I finally found why this wasn't making sense to me :sweat_smile: 
The Ruby code generator had a bug and `task.get` didn't have `id` as required so I was testing the `tasks.list` path (without an id) :facepalm:

picandocodigo:
Thanks @pquentin, I updated the PR so that it's just removing the scope for serverless in `indices.put_template`.

pquentin (approved):
Thanks! LGTM

PR #3012, opened 2024-10-11
Add proper enums for some `DenseVectorProperty` fields

flobernd:
As titled.

Taken from here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html


pquentin (approved):
Thanks! LGTM.

PR #3015, opened 2024-10-14
Host can be array in info settings network

l-trotta:
As reported [here](https://github.com/elastic/elasticsearch-java/issues/892#issuecomment-2410696170)


flobernd (approved):
LGTM!

PR #3023, opened 2024-10-17
[OpenAPI] Add x-model to specialized queries

lcawl:
This PR adds the `x-model` extension and `externalDocs` links to appropriate pages in the Query DSL reference via overlays.

In the longer term, the addition of these `x-model` extensions ideally should be implemented in the specifications rather than in overlays, per https://github.com/elastic/elasticsearch-specification/issues/3034


szabosteve (approved):
LGTM!

PR #3025, opened 2024-10-17
[Connectors API] Fix advanced snippet value type

jedrazb:
### Changes

Fix type mismatch between definition in ES, in ES we define the advancedSnippet value as `Object`, see [code reference](https://github.com/elastic/elasticsearch/blob/6d039c2b9dfb1dfc1eaac86ecff37b3db6521908/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/filtering/FilteringAdvancedSnippet.java#L40)

This is because this can be either object or array of objects. Changing 

`Dictionary<string, UserDefinedValue>` to `UserDefinedValue`

Shoutout to @l-trotta for spotting the issue! 🏅

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -63729,10 +63729,7 @@
             "$ref": "#/components/schemas/_types:DateTime"
           },
           "value": {
-            "type": "object",
-            "additionalProperties": {
-              "type": "object"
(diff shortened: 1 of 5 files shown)


l-trotta (approved):
LGTM (to be fair it was originally spotted by @Anaethelion)

PR #3028, opened 2024-10-18
Add @ext_doc_id for externalDocs in OpenAPI documents

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2748

This PR copies the existing `@doc_url` and `@doc_id` but is specifically for links that will be used in the OpenAPI output since the others have specific purposes in the clients.
The value specified in `@ext_doc_id` is mapped to a URL in the `table.csv` file (the same one that's used by `@doc_id`) then output as an [External Documentation object](https://spec.openapis.org/oas/v3.0.3#external-documentation-object).

This PR also updates one of the OpenAPI overlays, since it  is now unnecessary. I will clean up the rest of the `externalDocs` updates in the overlays in a separate PR.


l-trotta (approved):
LGTM! heads up @Anaethelion, @flobernd this might break generators since it's a new annotation

PR #3029, opened 2024-10-18
Add specification for query rule test API 

kderusso:
Adds a specification for the query rules test API added in https://github.com/elastic/elasticsearch/pull/114168


specification/query_rules/test/QueryRulesetTestRequest.ts
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.
+ */
+import { RequestBase } from '@_types/Base'
+import { Id } from '@_types/common'
+import { Dictionary } from '@spec_utils/Dictionary'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+
+/**
+ * Creates or updates a query ruleset.
+ * @rest_spec_name query_rules.put_ruleset
```
pquentin, line 26:
```suggestion
 * @rest_spec_name query_rules.test
```

specification/query_rules/test/QueryRulesetTestResponse.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { integer } from '@_types/Numeric'
+import { QueryRulesetMatchedRule } from './types'
```
pquentin, line 21: We recommend keeping classes that are used only once in the same file as the class that uses it - it's allowed as this is not Java!

pquentin (commented):
Thank you for this!

specification/query_rules/test/QueryRulesetTestRequest.ts
kderusso: You caught me in a copy paste 😅

pquentin:
Can you confirm whether there's a Elasticsearch YAML test for API or not?

l-trotta:
I have a question about `match_criteria`: in the documentation it says that:

> Match criteria should match the keys defined in the `criteria.metadata` field of the rule.

so can it only be something like `field_name: field_value`? since criteria.metadata was defined as `string` in the spec. If so then we could use a more specific type?

kderusso:
> Can you please confirm whether there's a Elasticsearch YAML test for API or not?

@pquentin Yes, the yaml REST tests may be found [here](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ent-search/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/entsearch/rules/70_query_rule_test.yml). 

> so can it only be something like field_name: field_value? since criteria.metadata was defined as string in the spec. If so then we could use a more specific type, like Dictionary<string,string>?

@l-trotta We can't use `Dictionary<string,string>` because the Java code supports `Object` such as numeric input.

l-trotta (approved):

PR #3033, opened 2024-10-21
Fix role-related API spec

n1v0lg:
There is a display glitch in the ES serverless API: see live [spec](https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-security-get-role).

where details of several APIs are rendered as API headers. This PR fixes the rendering issue by including the relevant header info in top-level comments for each affected API.

### DIFF
--- a/specification/security/delete_role/SecurityDeleteRoleRequest.ts
+++ b/specification/security/delete_role/SecurityDeleteRoleRequest.ts
@@ -22,6 +22,7 @@ import { Name, Refresh } from '@_types/common'
 
 /**
  * Delete roles API
+ *
  * Removes roles in the native realm.
  * @rest_spec_name security.delete_role
  * @availability stack stability=stable
--- a/specification/security/delete_role/SecurityDeleteRoleRequest.ts
+++ b/specification/security/delete_role/SecurityDeleteRoleRequest.ts
@@ -21,8 +21,7 @@ import { RequestBase } from '@_types/Base'
 import { Name, Refresh } from '@_types/common'
 
 /**
- * Delete roles API
- *
+ * Delete roles API.
  * Removes roles in the native realm.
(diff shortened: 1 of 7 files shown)


slobodanadamovic (approved):
LGTM 

Nice catch!

specification/security/delete_role/SecurityDeleteRoleRequest.ts
```diff
@@ -21,6 +21,8 @@ import { RequestBase } from '@_types/Base'
 import { Name, Refresh } from '@_types/common'
 
 /**
+ * Delete roles API
```
flobernd, line 24: There should be an empty line between the header and the description.

The JSDoc is interpreted as markdown which means that no new paragraph is created without the empty line.

Same for all other places 🙂

n1v0lg: @flobernd thanks! 

Is this what you had in mind: https://github.com/elastic/elasticsearch-specification/pull/3033/commits/4ca237aae7af48fe35926c4614be4437d3ef3d81
?

n1v0lg: Hmm the `description` field is still borked: https://github.com/elastic/elasticsearch-specification/pull/3033/files#diff-4ec6052a151433c851f9e7b60e2ac3a6814b1ac6aa977c6c43bf61cebc976038L8382

flobernd: @n1v0lg Yes, the previous commit looks good! It should affect the ouput as well 😅 (maybe we have a bug here in the spec compiler, if it really does not..).

flobernd (approved):
LGTM! Thank you.

PR #3035, opened 2024-10-21
[OpenAPI] Edit summaries for security APIs

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2635, https://github.com/elastic/elasticsearch-specification/pull/3033

This PR adds/edits summaries for some security APIs. The first phrase is used as the "operation summary" and should basically align with what existed in the table of contents in https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api.html (omitting "API" from the end of the phrase, however). Any subsequent information gets published as the "operation description".

NOTE: I changed the "enable"/"disable" terminology to "activate"/"deactivate" for inclusivity reasons but if that's problematic lmk.

Since there's such a large set of security APIs, I'll defer the remainder to a subsequent PR.

### DIFF
--- a/specification/security/disable_user/SecurityDisableUserRequest.ts
+++ b/specification/security/disable_user/SecurityDisableUserRequest.ts
@@ -21,8 +21,9 @@ import { RequestBase } from '@_types/Base'
 import { Refresh, Username } from '@_types/common'
 
 /**
- * Deactivate users.
- * Deactivate users in the native realm.
+ * Disable users.
+ * 
+ * Disable users in the native realm.
  * @rest_spec_name security.disable_user
  * @availability stack stability=stable
  */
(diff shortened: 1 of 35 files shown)


specification/security/disable_user/SecurityDisableUserRequest.ts
```diff
@@ -21,6 +21,8 @@ import { RequestBase } from '@_types/Base'
 import { Refresh, Username } from '@_types/common'
 
 /**
+ * Deactivate users.
```
n1v0lg, line 24: I'd go with enable/disable here (as well as for the profile routes) since that matches the existing API, ES docs, and its terminology that's been around for a while now. Not a string opinion though.

n1v0lg (approved):
LGTM

specification/security/disable_user/SecurityDisableUserRequest.ts
lcawl: Yes, it make sense to stay consistent for now. I've made that change in dc8609aa4cc030bb67a447c204418f23807ce143

PR #3036, opened 2024-10-21
[OpenAPI] Edit more security API summaries

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2635, https://github.com/elastic/elasticsearch-specification/pull/3033, https://github.com/elastic/elasticsearch-specification/pull/3035

This PR adds/edits summaries for the remainder of the security APIs. The first phrase is used as the "operation summary" and should basically align with what existed in the table of contents in https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api.html (omitting "API" from the end of the phrase, however). Any subsequent information gets published as the "operation description".

### DIFF
--- a/specification/security/put_role/SecurityPutRoleRequest.ts
+++ b/specification/security/put_role/SecurityPutRoleRequest.ts
@@ -29,13 +29,14 @@ import { RequestBase } from '@_types/Base'
 import { Metadata, Name, Refresh } from '@_types/common'
 
 /**
- * Create or update roles API.
- *
- * Create or update roles in the native realm.
+ * Create or update roles.
+ * The role management APIs are generally the preferred way to manage roles in the native realm, rather than using file-based role management.
+ * The create or update roles API cannot update roles that are defined in roles files.
  * @rest_spec_name security.put_role
  * @availability stack stability=stable
  * @availability serverless stability=stable visibility=public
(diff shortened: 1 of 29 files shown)


specification/security/put_role/SecurityPutRoleRequest.ts
```diff
@@ -29,13 +29,14 @@ import { RequestBase } from '@_types/Base'
 import { Metadata, Name, Refresh } from '@_types/common'
 
 /**
- * Create or update roles API.
- *
- * Create or update roles in the native realm.
+ * Create or update roles.
+ * The role management APIs are generally the preferred way to manage roles in the native realm, rather than using file-based role management.
+ * The create or update roles API cannot update roles that are defined in roles files.
```
n1v0lg, line 34: Could we also add a note here that file-based role management is not available in Serverless? It's a bit confusing otherwise.

n1v0lg (approved):
LGTM! Thanks for fixing these. One suggestion around the Put Role spec.

specification/security/put_role/SecurityPutRoleRequest.ts
lcawl: Sure! Added in 304227bbd6cc1e6360b82264ada0dc409ca932f8

PR #3038, opened 2024-10-22
Fix additional property name for `TermsQuery`

flobernd:
This seems to be a copy/paste mistake made by me when I implemented the `behavior_meta` tag.

@swallez @Anaethelion @ezimuel @picandocodigo @miguelgrinberg @l-trotta @pquentin @JoshMock   Please check, if this is a breaking change in any of your clients. Most likely only the static language clients should be affected - if at all.


miguelgrinberg:
I'm not using the key and key and value names, just their types, so for the Python DSL client this is fine 👍

Anaethelion (approved):

JoshMock:
`AdditionalProperty` behaviors are currently ignored by JS codegen. :+1:

picandocodigo (approved):

swallez:
It's not implemented yet in the Java generator, so no problem.

PR #3057, opened 2024-10-23
Adding text_similarity_reranker retriever specification

pmpailis:
Adding missing `text_similarity_reranker` retriever specification


specification/_types/Retriever.ts
```diff
@@ -78,3 +80,16 @@ export class RRFRetriever extends RetrieverBase {
...
+  /** The nested retriever which will produce the first-level results, that will later be used for reranking. */
+  retriever: RetrieverContainer
+  /** This value determines how many documents we will consider from the nested retriever.  */
+  rank_window_size?: integer
+  /** Unique identifier of the inference endpoint created using the inference API. */
+  inference_id?: string
+  /** The text snippet used as the basis for similarity comparison */
+  inference_text?: string
+  /** The document field to be used for text similarity comparisons. This field should contain the text that will be evaluated against the inference_text */
+  field?: string
```
Anaethelion, line 94: I think we're missing `min_score`
```suggestion
  field?: string
  /** Sets a minimum threshold score for including documents in the re-ranked results. Documents with similarity scores below this threshold will be excluded. Note that score calculations vary depending on the model used. */
  min_score?: float
```

Anaethelion (changes requested):
Looks good! I've left a code suggestion as I think a field was missing.

specification/_types/Retriever.ts
pmpailis: Good catch @Anaethelion ! `min_score` has actually been made available for all retrievers through `RetrieverBase`, so updating accordingly!

pmpailis: Updated in 54fd49375ac452a0e6d62a19cbc839a7787829b7

pmpailis:
@Anaethelion  updated and added `min_score` to  base class `RetrieverBase`. Run the tests locally and couldn't find something that is related to these changes, so would appreciate another look whenever you have some time :)

Anaethelion (approved):
LGTM!

PR #3059, opened 2024-10-23
Remove deprecated local parameter from alias APIs

PeteGillinElastic:
This updates the specification to reflect the changes in https://github.com/elastic/elasticsearch/pull/115393.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -10963,9 +10963,6 @@
           },
           {
             "$ref": "#/components/parameters/indices.get_alias#ignore_unavailable"
-          },
-          {
-            "$ref": "#/components/parameters/indices.get_alias#local"
           }
         ],
         "responses": {
@@ -11083,9 +11080,6 @@
           },
           {
             "$ref": "#/components/parameters/indices.exists_alias#ignore_unavailable"
-          },
-          {
-            "$ref": "#/components/parameters/indices.exists_alias#local"
           }
         ],
         "responses": {
@@ -11953,9 +11947,6 @@
           },
           {
             "$ref": "#/components/parameters/indices.get_alias#ignore_unavailable"
-          },
-          {
-            "$ref": "#/components/parameters/indices.get_alias#local"
           }
         ],
         "responses": {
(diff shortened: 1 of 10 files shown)


PeteGillinElastic:
This PR removes the `?local` query parameter from three APIs in the `_json_spec`. It makes this equivalent change in the typescript requests for two of the three. I'm not sure of the best way to do it for the third. In `CatAliasesRequest.ts`, the `Request` interface extends the class `CatRequestBase` which implements `CommonCatQueryParameters` which lists the `local` parameter (although not in a `query_parameters` which I don't quite understand). I'm not sure of the best way to remove the parameter from this API without affecting the twenty-something other places which use that base class. Any advice? (There's also a `@behavior` annotation and I don't know what that means. Sorry, I'm kind of groping around in the dark here!)

Also: For the two index APIs, I had to make parallel changes to the `*Request.ts` and the `*.json` in order to avoid adding validation errors. But for this CAT API, it looks like all of the parameters which are listed in `cat.aliases.json` and also in `CommonCatQueryParameters` are listed in `validation-errors.json`, as if the validation logic isn't recognizing that inheritance. When I remove the local parameter from the JSON without touching the typescript files, it therefore removes that validation error rather than adding a new one. I'm not sure what's going on there.

PeteGillinElastic:
Regarding updating the TS to reflect the change to the `_cat/aliases` API:

 - There are 26 requests which implemented `CommonCatQueryParameters` (via `CatRequestBase`).
 - Before this change, only 10 of those actually have the `local` parameter according to the JSON; the other 16 do not.
 - After the change currently in this PR, those numbers are 9 and 17.

I would tentatively make the argument that, as it stands, this PR is not making things materially worse, and that fixing that wider problem is out of scope for this change. What do you think?

(I did have a go at moving the parameter out of the interface and into the 9 concrete request classes which have the `local` parameter according to the JSON. It created some fairly major changes to the output. This reinforces my feeling that it's out of scope for what I'm doing right now. You can see my attempt at that over [here](https://github.com/elastic/elasticsearch-specification/compare/remove-deprecated-local-from-get-aliases...remove-deprecated-local-from-get-aliases-v2).)

l-trotta (approved):
LGTM, thank you for contributing to the spec! I'll create another PR for 8.x branches to set this parameter as deprecated. As for cat.aliases, I think what you did in the experimental branch is the correct way to handle this, but I agree that it's out of scope, I'll take care of it.

PR #3070, opened 2024-10-23
Add cross cluster security APIs

lcawl:
The create and update cross cluster key APIs were not appearing in the OpenAPI output, so I've created an initial pass at the necessary files per https://github.com/elastic/elasticsearch-specification/blob/main/docs/add-new-api.md

The descriptions are copied from https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api-update-cross-cluster-api-key.html and

### DIFF
--- a/specification/security/_types/Access.ts
+++ b/specification/security/_types/Access.ts
@@ -1,0 +1,31 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { RemoteIndicesPrivileges, ReplicationAccess } from './Privileges'
+
+export class Access {
+  /**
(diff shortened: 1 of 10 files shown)


specification/security/create_cross_cluster_api_key/CreateCrossClusterApiKeyRequest.ts
```diff
@@ -0,0 +1,73 @@
...
+     */
+    expiration?: Duration
+    /**
+     * Arbitrary metadata that you want to associate with the API key.
+     * It supports nested data structure.
+     * Within the metadata object, keys beginning with `_` are reserved for system usage.
+     */
+    metadata?: Metadata
+    /** Specifies the name for this API key. */
+    name?: Name
```
n1v0lg, line 71: `name` is required

n1v0lg (commented):
Thanks for picking this up! I have a high-level comment on the `Access` model -- we want a new `SearchAccess` model there, instead of re-using `RemoteIndicesPrivileges` since those are subtly different.

specification/security/_types/Access.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class Access {
+  /**
+   * A list of indices permission entries for cross-cluster replication.
+   */
+  replication?: ReplicationAccess[]
+  /**
+   * A list of indices permission entries for cross-cluster search.
+   */
+  search?: RemoteIndicesPrivileges[]
```
n1v0lg, line 30: We need a separate model (`SearchAccess`) instead of `RemoteIndicesPrivileges` since there are subtle differences:

- `SearchAccess` does not take a list of privileges
- `SearchAccess` does not take a list of remote clusters

specification/security/create_cross_cluster_api_key/CreateCrossClusterApiKeyRequest.ts
lcawl: Thanks, fixed in d5d06c140a94be606f486d570f1579d734add9eb

specification/security/_types/Access.ts
lcawl: Okay! I've added a separate definition in ef524c3ca51a6bfc382444db537c47e95e9c60f8

n1v0lg (approved):
LGTM!

PR #3092, opened 2024-11-05
Add ip_location ingest processor

pquentin:
As added in https://github.com/elastic/elasticsearch/pull/114548. An alternative to fix the failure is to ignore the new "Test ip_location processor with defaults" test.


joegallo (approved):
LGTM. Thank you!

PR #3096, opened 2024-11-05
"local" parameter only in compatible cat requests

l-trotta:
Followup of #3059, removing `local` from common cat parameters and adding it only to those requests which allow it. 
I can't test it with java because it has its own handcrafted test types, but it seems straightforward enough.

### DIFF
--- a/specification/cat/indices/CatIndicesRequest.ts
+++ b/specification/cat/indices/CatIndicesRequest.ts
@@ -73,5 +73,13 @@ export interface Request extends CatRequestBase {
     pri?: boolean
     /** The unit used to display time values. */
     time?: TimeUnit
+    /**
+     * If `true`, the request computes the list of selected nodes from the
+     * local cluster state. If `false` the list of selected nodes are computed
+     * from the cluster state of the master node. In both cases the coordinating
+     * node will send requests for further information to each selected node.
+     * @server_default false
+     */
+    local?: boolean
   }
 }
--- a/specification/cat/indices/CatIndicesRequest.ts
+++ b/specification/cat/indices/CatIndicesRequest.ts
@@ -73,13 +73,5 @@ export interface Request extends CatRequestBase {
     pri?: boolean
     /** The unit used to display time values. */
     time?: TimeUnit
-    /**
-     * If `true`, the request computes the list of selected nodes from the
-     * local cluster state. If `false` the list of selected nodes are computed
-     * from the cluster state of the master node. In both cases the coordinating
-     * node will send requests for further information to each selected node.
-     * @server_default false
-     */
-    local?: boolean
   }
 }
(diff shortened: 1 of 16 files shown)


specification/cat/nodeattrs/CatNodeAttributesRequest.ts
```diff
@@ -28,4 +28,15 @@ import { CatRequestBase } from '@cat/_types/CatBase'
  * @doc_id cat-nodeattrs
  * @cluster_privileges monitor
  */
-export interface Request extends CatRequestBase {}
+export interface Request extends CatRequestBase {
```
JoshMock, line 31: To reduce code duplication, would it help to add a `CatWithLocalRequestBase` that each of these can extend?

JoshMock (approved):
One question, but if we're comfortable with the duplication this LGTM.

specification/cat/nodeattrs/CatNodeAttributesRequest.ts
l-trotta: sure that works too, I'll do it

pquentin: This requires a change in the compiler: https://github.com/elastic/elasticsearch-specification/blob/48e2d9de9de2911b8cb1cf715e4bc0a2b1f4b827/compiler/src/model/build-model.ts#L347-L349

l-trotta: don't think it's worth modifying the compiler for cat, so if it's okay I'd merge it as is

pquentin (commented):
Thanks! The approach looks good. I don't think _cat/indices supports ?local.

specification/cat/indices/CatIndicesRequest.ts
```diff
@@ -73,5 +73,13 @@ export interface Request extends CatRequestBase {
...
     /** The unit used to display time values. */
     time?: TimeUnit
+    /**
+     * If `true`, the request computes the list of selected nodes from the
+     * local cluster state. If `false` the list of selected nodes are computed
+     * from the cluster state of the master node. In both cases the coordinating
+     * node will send requests for further information to each selected node.
+     * @server_default false
+     */
+    local?: boolean
```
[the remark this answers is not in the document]
l-trotta: it doesn't reject it like the other requests which doesn't support it. I think it just does nothing though, I'm removing it

pquentin (approved):
Thanks! LGTM.

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.x|[](https://github.com/elastic/elasticsearch-specification/pull/3359)|

### Questions ?

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.17|[](https://github.com/elastic/elasticsearch-specification/pull/3360)|

### Questions ?

PR #3107, opened 2024-11-08
Add IP Location APIs

pquentin:
I extracted the rest-api-spec JSON files from https://github.com/elastic/elasticsearch-specification/pull/3019. We'll need a review from the Data Management team before merging.


joegallo (changes requested):
There are some text and naming changes I'd like to see, I'll reach out to you offline.

joegallo:
I think we'll still need to make a change to `Database.ts` -- I'll reach out to you offline.

lcawl:
Thanks for creating this PR! I was looking at how to ensure the work from https://github.com/elastic/elasticsearch/pull/116611 got carried over to the new API docs site and this PR already does a lot of the heavy lifting (we can port the examples over in a subsequent PR).

specification/ingest/_types/Database.ts
```diff
@@ -19,15 +19,21 @@
 
 import { Id, Name } from '@_types/common'
 
-export class DatabaseConfiguration {
+specification / ingest / _types / Database.ts
```
pquentin, line 22: I don't know why this does not break CI, but let's remove it anyway:

```suggestion
```

specification/ingest/_types/Database.ts
```diff
@@ -19,15 +19,21 @@
...
+ */
+export type DatabaseConfiguration = Maxmind | Ipinfo
+
+export class DatabaseConfigurationBase {
   /** The provider-assigned name of the IP geolocation database to download. */
   name: Name
-  /** The configuration necessary to identify which IP geolocation provider to use to download the database, as well as any provider-specific configuration necessary for such downloading.
-   * At present, the only supported provider is maxmind, and the maxmind provider requires that an account_id (string) is configured.
-   */
-  maxmind: Maxmind
```
pquentin, line 28: We do need the `maxmind` key which is gone with this approach.

Also it's OK to have some duplication, eg I'm happy to keep the previous approach but simply add an `ipinfo` field and type, with one less field.

joegallo: Agreed, yeah, I'll take a swing at fixing that -- it wasn't an intentional change in this second version of things, I just wasn't thoughtful enough about how I did it.

specification/ingest/_types/Database.ts
joegallo: Gah, yeah, that was just sloppiness on my part. 91e9c648f2ffdb5184e1ad0f35d0d1af4c684519

specification/ingest/_types/Database.ts
joegallo: What do you think about 31fd86a38351c91a75cd67a0f05f1e0b93ee007c?

specification/ingest/put_ip_location_database/PutIpLocationDatabaseRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+     * @server_default 30s */
+    timeout?: Duration
+  }
+  body: {
+    /** The provider-assigned name of the IP geolocation database to download. */
+    name: Name
+    /** The configuration necessary to identify which IP geolocation provider to use to download the database, as well as any provider-specific configuration necessary for such downloading.
+     * At present, the only supported provider is maxmind, and the maxmind provider requires that an account_id (string) is configured.
+     */
+    maxmind: Maxmind
```
pquentin, line 55: We have the same problem as database here, and the current specification fails on this recording:

```ts
// Test file: /test/free/ingest_geoip/60_ip_location_databases.yml
// Test name: Test adding, getting, and removing ip location databases

import * as T from '../types'

expectAssignable<T.IngestPutIpLocationDatabaseRequest>({
  "body": {
    "ipinfo": {},
    "name": "standard_privacy"
  },
  "id": "my_database_3"
})
```

However, unlike Database, we can't express a body as `TypeA | TypeB`, so we'll have to use optional fields.

specification/ingest/_types/Database.ts
pquentin: This looks great, thanks! Can you please confirm that the GeoIP APIs also accept `ipinfo` databases? I'm asking as `DatabaseConfiguration` is currently shared between the two APIs.

joegallo: Yes, they do.

specification/ingest/put_ip_location_database/PutIpLocationDatabaseRequest.ts
joegallo: Ah, actually, I'm going to unresolve this -- we still need to capture that `ipinfo` database configurations can be created.

joegallo:
@pquentin I've got two more things on this one, one that's hopefully minor, and then one that might be more-than-minor:

1.) Can you help me diagnose and understand the failure on `ingest.put_ip_location_database`? I'm seeing 3/4 for the 'Request' column, and I'm not sure what there is to be done about that, if anything.

2.) The `PUT` needs to be able to handle `ipinfo` database configurations, too, and I didn't realize until just now that that's not currently covered. To make things slightly more difficult, the `PUT` API will not handle `web` and `local` configurations -- those are effectively read-only.

pquentin:
> @pquentin I've got two more things on this one, one that's hopefully minor, and then one that might be more-than-minor: [...]

Sorry for the delay!

So 1. and 2. are the same actually! The failure is the one I mentioned in https://github.com/elastic/elasticsearch-specification/pull/3107/files#r1850637399. I can help you offline to get `make validate api=ingest.put_ip_location_database branch=main` to work if you're interested.

That said, I tried to fix those in my latest commit, please take a look.

Something else we need to do (possibly in a follow-up PR) is to apply the same changes to the existing GeoIP APIs.

specification/ingest/get_ip_location_database/GetIpLocationDatabaseResponse.ts
```diff
@@ -0,0 +1,34 @@
...
+import { EpochTime, UnitMillis } from '@_types/Time'
+
+export class Response {
+  body: { databases: DatabaseConfigurationMetadata[] }
+}
+
+class DatabaseConfigurationMetadata {
+  id: Id
+  version: long
+  modified_date_millis: EpochTime<UnitMillis>
```
l-trotta, line 32: the timestamp can be one of `modified_date` or `modified_date_millis` depending on the `humanReadable`flag. [server code](https://github.com/elastic/elasticsearch/blob/bb8503a2521a31b73360a86c62ccec3047369b26/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/direct/GetDatabaseConfigurationAction.java#L113). if I'm reading the `UNIX_EPOCH_MILLIS_FORMATTER` correctly, the types should be:
```
modified_date_millis?: string
modified_date?: EpochTime<UnitMillis>
```

specification/ingest/get_ip_location_database/GetIpLocationDatabaseResponse.ts
```diff
@@ -0,0 +1,34 @@
...
+import { long } from '@_types/Numeric'
+import { EpochTime, UnitMillis } from '@_types/Time'
+
+export class Response {
+  body: { databases: DatabaseConfigurationMetadata[] }
+}
+
+class DatabaseConfigurationMetadata {
+  id: Id
+  version: long
```
l-trotta, line 31:
```suggestion
  version: VersionNumber
```

specification/ingest/get_ip_location_database/GetIpLocationDatabaseResponse.ts
pquentin:
```suggestion
  modified_date_millis?: EpochTime<UnitMillis>
  modified_date?: EpochTime<UnitMillis>
```

l-trotta (approved):
LGTM!

joegallo (approved):
Ship it!

PR #3113, opened 2024-11-11
Set parameter nodes instead of node_id on task.list

margaretpearce:
Closes https://github.com/elastic/elasticsearch-specification/issues/3112, closes https://github.com/elastic/elasticsearch-specification/pull/3114

Use parameter `nodes` instead of `node_id` on `task.list`.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -33046,10 +33046,7 @@
             "description": "Comma-separated list of node IDs or names used to limit returned information.",
             "deprecated": false,
             "schema": {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
+              "$ref": "#/components/schemas/_types:NodeIds"
             },
             "style": "form"
           },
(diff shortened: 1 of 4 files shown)


pquentin (approved):
Thank you! This is a great catch. Unfortunately none of the Elasticsearch YAML tests that we use for validation set `nodes`, which is why we missed it.

I switched the type from `string[]` to `NodeIds` (which is `string | string[]`):

 1. typed clients could have a specific type for this
 2. to allow for a comma-separated string too, which is the format Elasticsearch accepts (lists in query parameters are comma-separated by clients before sending)

pquentin:
Thank you for your contribution, your change is now available in [elasticsearch-py 8.16.0](https://github.com/elastic/elasticsearch-py/releases/tag/v8.16.0) (and all the other clients that use the specification). Note that elasticsearch-py 8.16.0 added warnings for tech preview APIs and the Task management API is experimental. The [release notes](https://github.com/elastic/elasticsearch-py/releases/tag/v8.16.0) explain how to disable the warning.

PR #3114, opened 2024-11-11
Set parameter nodes instead of node_id on task.list

pquentin:
This is only used to run CI in https://github.com/elastic/elasticsearch-specification/pull/3113.


pquentin (commented):
Thank you! I switched the type from `string[]` to NodeIds (which is `string | string[]`):

 1. typed clients could have a specific type for this
 2. to allow comma-separated strings too, which is the format Elasticsearch accepts

PR #3123, opened 2024-11-12
Support for security remote_cluster and associated privileges 

jakelandis:
This commit adds support for the remote_cluster in the role and role descriptors. 
Additionally:
* adds missing references to remote_indices
* add new cluster privilege monitor_stats
* adds related version information where applicable
* updates references to cluster from string[] to proper enumeration


specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesResponse.ts
```diff
@@ -18,7 +18,8 @@
  */
 
 import { IndexName } from '@_types/common'
+import {ClusterPrivilege, ClusterPrivilegeForRemote} from "@security/_types/Privileges";
 
 export class Response {
-  body: { cluster: string[]; index: IndexName[] }
+  body: { cluster: ClusterPrivilege[]; index: IndexName[]; remote_cluster: ClusterPrivilegeForRemote[] }
```
jakelandis, line 24: I am unsure about changing cluster from string[] to the proper enumeration

This should be the preferred model, where we guide users to only add named privileges from the enum.  However, it is technically possible (but not recommended) to add the action names directly which are just string[]. Other places use the enum and this makes them the same (and the validation still passes).

specification/security/get_role/types.ts
```diff
@@ -18,17 +18,25 @@
...
+  IndicesPrivileges, RemoteClusterPrivileges, RemoteIndicesPrivileges
 } from '@security/_types/Privileges'
 import { RoleTemplate } from '@security/_types/RoleTemplate'
 import { Dictionary } from '@spec_utils/Dictionary'
 import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 import { Metadata } from '@_types/common'
 
 export class Role {
-  cluster: string[]
+  cluster: ClusterPrivilege[]
```
jakelandis, line 30: same question (as above) here for the change from string[] to emum

jakelandis:
Could someone help me to navigate the failures here ? It's my first PR to this repo and all the validation either passed locally or failed in the same manner before my change. Not sure how to interpert the build failure here.

pquentin:
Can you please run `make spec-format-fix` and reopen this pull request from a branch? As the recordings are currently private, we don't support pull requests from forks.

Additionally, the errors are in fact different on main and in your pull request.

## main

```bash
$ make validate api=security.get_builtin_privileges branch=main
Validating endpoints
Validating security.get_builtin_privileges
It looks like security.get_builtin_privileges request has some errors, take a look at the workbench.

../clients-flight-recorder/scripts/types-validator/workbench/a6156899cdec2b75c536ba9ed5a3723b.test-d.ts:96:2
Argument of type '{ cluster: string[]; index: string[]; remote_cluster: string[]; }' is not assignable to parameter of type 'ErrorResponseBase | SecurityGetBuiltinPrivilegesRespo
nse'.
  Object literal may only specify known properties, and '"remote_cluster"' does not exist in type 'ErrorResponseBase | SecurityGetBuiltinPrivilegesResponse'.

✔ 2 out of 2 test request cases are passing.
✖ 1 out of 2 test response cases are passing.
```

In other words, this was failing because `remote_cluster` was not expected in the response.


## This pull request

```bash
$ make validate api=security.get_builtin_privileges branch=main
Validating endpoints
Validating security.get_builtin_privileges
It looks like security.get_builtin_privileges request has some errors, take a look at the workbench.

../clients-flight-recorder/scripts/types-validator/workbench/scripts/types-validator/workbench/47a1eec20d0f6734ef46080eababed2a.test-d.ts:7:79
Argument of type '{ cluster: string[]; index: string[]; }' is not assignable to parameter of type 'ErrorResponseBase | SecurityGetBuiltinPrivilegesResponse'.
  Property 'remote_cluster' is missing in type '{ cluster: string[]; index: string[]; }' but required in type 'SecurityGetBuiltinPrivilegesResponse'.

✔ 2 out of 2 test request cases are passing.
✖ 1 out of 2 test response cases are passing.
```

Here, we're seeing the opposite failure! The spec requires `remote_cluster`, but the test does not have it. This is an old recorded test from the NEST .NET client. I've opened a pull request to remove it and required a review from you.

specification/security/_types/Privileges.ts
```diff
@@ -194,6 +198,21 @@ export enum ClusterPrivilege {
   write_fleet_secrets
 }
 
+/**
+ * The subset of cluster level privileges that can be defined for remote clusters.
+ * @availability stack
+ */
+export enum ClusterPrivilegeForRemote {
```
flobernd, line 205: Let's rename this to `RemoteClusterPrivilege`. I think that matches the existing schema:

- `ClusterPrivilege` (enum) -> `ClusterPrivileges` (class)
- `RemoteClusterPrivilege` (enum) -> `RemoteClusterPrivileges` (class)

specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesResponse.ts
flobernd: This one is the response class. Are we sure that only valid privileges from the enum are returned here? If yes, we should definitely prefer the enum (however, this will be a breaking change for all statically typed language clients, so we have to check with everybody else f irst).

specification/security/get_role/types.ts
flobernd: This type is as well only used in the response of `get_role`, so same situation as above.

specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesResponse.ts
jakelandis: From this API, yes, only values from the enum are ever returned.

specification/security/_types/Privileges.ts
jakelandis: will do

specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesResponse.ts
flobernd: For me it would be ok to introduce a breaking change by changing the return type to the correct enum type. What about the rest @elastic/devtools-team ?

JoshMock:
accidental close! :laughing:

jakelandis:
closing in favor https://github.com/elastic/elasticsearch-specification/pull/3125 to allow CI to work. Will make changes on that PR.

specification/security/get_builtin_privileges/SecurityGetBuiltinPrivilegesResponse.ts
pquentin: See https://github.com/elastic/elasticsearch-specification/pull/3125#discussion_r1842003226 - turns out it's non breaking after all.

PR #3016, opened 2024-10-14
Security APIs: Change `Indices` -> `IndexName[]`

flobernd:
From the documentation, it seems that the security APIs always uses `IndexName[]` instead of `Indices`. The latter type has some special semantics, like e.g. `"_all"` string representation - which is not valid in the context of the security APIs.

Related to:
https://github.com/elastic/elasticsearch-net/issues/8377

Issue #2415, opened 2024-02-19
Introduce alias for `StoredFields`


flobernd:
## Problem

The `Fields` type is currently used wherever a collection of fields is expected. In most places, the server accepts the following values:

- missing / `null`
- `[]` (no elements)
- `["a"]` (single element)
- `["a", "b", ...]` (multiple elements)

Unfortunately, there is a special case in all places where `fields` is used for a `stored_fields` property, a query parameter etc.. In these cases, special semantics are applied and the following values are accepted:

- missing/`null` (selects `_id` and `_type` [deprecated])
- `[]` (selects `_id` and `_type` [deprecated])
- `["a"]` (single element)
- `["a", "b", ...]` (multiple elements)
- `"_none_"` (string; deselects all fields)

While the exact semantics are usually not captured by the specification, this particular case needs to be modeled so that the generators can produce a specific serialization behavior. Elasticsearch does not accept the value `_none_` as an element of an array (`["_none_"]`) but requires it to be encoded as a `string`. On the other hand, all regular places where `fields` is used, explicitly do not accept single element as `string`, but require the use of an array.

## Proposal

Introduce a new alias `type StoredFields = Fields` and use it for all `stored_fields` properties. Generators can decide whether to just reduce the alias to `Fields` or to implement special handling.

This mainly affects the statically typed language clients.

cc @swallez @Anaethelion @picandocodigo @l-trotta @stevejgordon

flobernd:
After giving it a second look, we realized the `Fields` type in t he specification actually supports `Field | Field[]`. The .NET generator incorrectly replaced an instance of just `Field[]` with the `Fields` type.

No specification fix is required at this point.

PR #3127, opened 2024-11-14
Allow index name patterns in Privileges index fields

pquentin:
Motivated by [this YAML test](https://github.com/elastic/elasticsearch/blob/a04fd2668b82e0c4324b2cc8afe03eb9b0113ac2/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/roles/60_bulk_roles.yml#L39-L46) I realized that `index` in various privileges classes can always be an index pattern, ie. a string.

For references, here is how IndexName and Indices are defined:

```ts
export type IndexName = string
export type Indices = IndexName | IndexName[]
```

This fixes `security.bulk_put_role`, as can be tested with `make validate api=security.bulk_put_role branch=main`.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -81966,11 +81966,7 @@
             "$ref": "#/components/schemas/security._types:FieldSecurity"
           },
           "names": {
-            "description": "A list of indices (or index name patterns) to which the permissions in this entry apply.",
-            "type": "array",
-            "items": {
-              "$ref": "#/components/schemas/_types:IndexName"
-            }
+            "$ref": "#/components/schemas/_types:Indices"
           },
           "privileges": {
             "description": "The index level privileges that owners of the role have on the specified indices.",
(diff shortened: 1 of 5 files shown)


flobernd:
That's unfortunate 😓 I recently changed these from `Indices` to `IndexName[]` without noticing it breaks validation..

Context:
https://github.com/elastic/elasticsearch-specification/pull/3016

l-trotta (approved):
LGTM! I think for most static clients this should change nothing since we already simplify this union as an array

flobernd:
Could we please change this types to use `IndexName | IndexName[]` instead of the `Indices` alias for now and leave a comment that links to this issue?

That would give us some time to think about these special types that might carry specific semantics, without breaking the .NET client.

IMO we should model these differences in the Spec to some extend in order to allow generators to produce better code. In this particular case, `Indices` usually supports the `_all` string constant which is why the .NET client treats that in a special way. The security APIs however do not accept this constant. They require the usage of a proper pattern (`*`) instead.

We had similar candidates in the past where the typing is the same, but the semantics are different. I remember e.g. `StoredFields` (#2415). Here we used the same workaround, keeping the special `StoredFields` properties as `Field | Field[]` while using `Fields` for everything else.

pquentin:
Thanks Florian, please take another look.

flobernd (approved):
Thank you! This looks good now 🙂

Issue #2830, opened 2024-08-29
Allow to define union availability on types


flobernd:
For situations like in e.g. #2826 we have to include/exclude certain types from union type aliases.

To support this situation, we should introduce a new annotation `union_availability` which can by places on **types**. The syntax should copy the `availability` syntax.

On `schema.json` level, there should be a new key `union_availabilites` which contains the defined availabilities. The individual entries should use the `Availability` type - just like for the regular `availabilities` dictionary.

Open for discussion:
Should we support the same mechanic for `deprecation`?

cc @pquentin @swallez

Issue #2643, opened 2024-06-24
[validation] Proposal: Disallow `deprecated` and `availability` on types


flobernd:
Successor of #2640

Deprecation and availability information should only apply to endpoints, properties and enum-members, but not to types.

## Proposal

- Remove `BaseType.deprecation`

PR #3145, opened 2024-11-19
Fix `RemoteClusterPrivilege` availability

flobernd:
Availability on types should not be used (see #2643 and #2830).

The PR moves the `@availability` tag from the type itself to the place(s) where the type is used.

Without this change, the .NET generator produced an enum without a member for the `serverless` flavor.

### DIFF
--- a/specification/security/_types/Privileges.ts
+++ b/specification/security/_types/Privileges.ts
@@ -200,7 +200,6 @@ export enum ClusterPrivilege {
 
 /**
  * The subset of cluster level privileges that can be defined for remote clusters.
- * @availability stack
  */
 export enum RemoteClusterPrivilege {
   /**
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks! LGTM. Is there any good way to catch this in the compiler?

flobernd:
@pquentin 

> Thanks! LGTM. Is there any good way to catch this in the compiler?

Yes, using availability/deprecated on types will cause a compiler error in the future when I have time to work on #2643 🙂

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.x|[](https://github.com/elastic/elasticsearch-specification/pull/3411)|

### Questions ?

PR #3162, opened 2024-11-21
Add missing autoscaling timeout parameters

pquentin:
In https://github.com/elastic/elasticsearch/pull/108759, @DaveCTurner added `master_timeout` support for autoscaling APIs. We were also already misssing `timeout` parameters. Note that adding those entries removes all the `autoscaling` entries in `output/schema/validation-errors.json`.


DaveCTurner (approved):
LGTM

PR #3165, opened 2024-11-22
Silence formatting output

pquentin:
Today, `make spec-format-fix` (a dependency of make contrib) nearly prints 2000 lines of output, which isn't useful while making the validation errors less visible.


l-trotta (approved):
appreciated, thanks!

PR #3172, opened 2024-11-22
Add FileSettingsIndicator

prdoyle:
Describes the health indicator added in https://github.com/elastic/elasticsearch/pull/117081.


prdoyle:
Looks like I'm not supposed to use a fork.

PR #3194, opened 2024-11-27
[Nodes.Info] Fix NodeInfoPath data

Anaethelion:
Reported in https://github.com/elastic/go-elasticsearch/issues/911

Turns out the `data` key is in fact a union over a string and an array. [server code for reference.](https://github.com/elastic/elasticsearch/blob/e5bcb0c5b3563044edf96ee9c57f03733cd888a8/server/src/main/java/org/elasticsearch/env/Environment.java#L131)

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -78352,10 +78352,17 @@
             }
           },
           "data": {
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks! LGTM.

PR #3195, opened 2024-11-28
Allow reading OpenAPI files in YAML format

pquentin:
The Kibana OpenAPI file is in YAML format. Since serde-yaml is no longer maintained, I'm using the most popular fork for now, serde-yml.


compiler-rs/openapi_to_clients_schema/src/main.rs
```diff
@@ -26,31 +26,40 @@ fn main() -> anyhow::Result<()> {
...
 
-    let open_api: openapiv3::OpenAPI = serde_ignored::deserialize(json_deser, |path| {
-        unused.insert(path.to_string());
-    })?;
+    let open_api = match Path::new(path).extension() {
+        Some(ext) if ext == "json" => {
+            let deser = &mut serde_json::Deserializer::from_reader(file);
+            serde_ignored::deserialize(deser, |path| {
+                unused.insert(path.to_string());
+            })
```
pquentin, line 43: Unfortunately `serde_json::Deserializer::from_reader()` is &mut, while `serde_yml::Deserializer::from_reader(file)` is not, which as far as I can tell forces duplicating the `deserialize()` call.

compiler-rs/openapi_to_clients_schema/src/main.rs
```diff
@@ -26,31 +26,40 @@ fn main() -> anyhow::Result<()> {
...
+            .map_err(|err| From::from(err))
```
pquentin, line 44: How can I avoid calling `map_err` here? I'm simply trying to convert from `serde:Error` to `anyhow:Error`, but `?` does not work.

compiler-rs/openapi_to_clients_schema/src/main.rs
```diff
@@ -26,31 +26,40 @@ fn main() -> anyhow::Result<()> {
     let subscriber = FmtSubscriber::builder().with_max_level(Level::TRACE).finish();
     tracing::subscriber::set_global_default(subscriber)?;
 
-    // let file = "../../ent-search/swagger/v1/enterprise-search.json";
-    // let file = "../../ent-search/swagger/v1/workplace-search.json";
-    let file = "./openapi_to_clients_schema/fixtures/workplace-search.json";
+    let path = "../../../ent-search/swagger/enterprise-search.json";
```
pquentin, line 29: The file in fixtures/ was empty, so I deleted it and pointed to a more likely way to find it instead.

compiler-rs/openapi_to_clients_schema/src/main.rs
Anaethelion: I would change the logic to read the file into a `String` and change the code to:

```suggestion
            let mut deser = serde_json::Deserializer::from_str(&data);
            serde_ignored::deserialize(&mut deser, |path| {
                unused.insert(path.to_string());
            })
```

Given that data comes from:
```rust
    let mut data = String::new();
    std::fs::File::open(path)?.read_to_string(&mut data)?;
```

Counter intuitive stuff happens with reader and buffers. [from_reader doc](https://docs.rs/serde_json/latest/serde_json/de/fn.from_reader.html)

compiler-rs/openapi_to_clients_schema/src/main.rs
Anaethelion: I don't think you can, but it can be simplified to:
```rust
.map_err(From::from)
```

compiler-rs/openapi_to_clients_schema/src/main.rs
Anaethelion: What about adding an `env` var and keeping this as default?

compiler-rs/openapi_to_clients_schema/src/main.rs
pquentin: Thanks, done in [`17ca650` (#3195)](https://github.com/elastic/elasticsearch-specification/pull/3195/commits/17ca650923eb24e31a620c5b1d86e00617fd9997)

compiler-rs/openapi_to_clients_schema/src/main.rs
pquentin: Thanks, done in [`17ca650` (#3195)](https://github.com/elastic/elasticsearch-specification/pull/3195/commits/17ca650923eb24e31a620c5b1d86e00617fd9997). I did not realize using a string would be much faster.

compiler-rs/openapi_to_clients_schema/src/main.rs
pquentin: I was considering addressing that as a follow-up, using `std::env::args` and then use `cargo run -- fixtures/kibana.serverless.yaml`. As described in https://doc.rust-lang.org/book/ch12-01-accepting-command-line-arguments.html. It feels better than an env var as my shell will provide autocompletion. (And clap seems overkill here.)

pquentin: OK, I ended up doing it in [`8cf465a` (#3195)](https://github.com/elastic/elasticsearch-specification/pull/3195/commits/8cf465a83449971b079b260da3968b3b5583439a). Please tell me what you think.

(I also tried using `Err(anyhow::anyhow!("expected a single argument"))` instead of panicking. It gave a better error, but it also leads to uglier code as I had to use `String` instead of `&str`.)

Anaethelion (approved):
LGTM!

PR #3196, opened 2024-11-29
Add initial version of schema endpoint generation

pquentin:
I'm using a simplified version of the Kibana OpenAPI for one endpoint. Simplifications include:
 * I replaced `additionalProperties: {}` with `additonalProperties: true` manually. (I initially thought it was a Kibana bug, but the two are in fact equivalent: https://swagger.io/docs/specification/v3_0/data-models/dictionaries/#free-form-objects. I intend to fix this later)
 * I only handle the 200 error response, and can't handle `AnyOf` where multiple response are possible.

That said, I'm able to generate a schema.json file (that can be used to generate client) from this simplified OpenAPI representation. This is tested using https://insta.rs/.


Anaethelion (changes requested):
Overall looks good, could you add some comments on the `endpoints.rs` file? There's a lot of pattern matching and its not always obvious what each does.

pquentin:
Sorry for the late reply. Assuming you are referring to `generate_get_request`, I'm not sure what I can add here, since well, I'm following the OpenAPI structure, and already return Err with descriptions in cases that I don't handle.w

Anyway, closing this for now as it's now stale with the work on examples.

PR #3203, opened 2024-12-02
Adding resolve_during_rolling_upgrade and _meta to the deprecation info API response

masseyke:
The resolve_during_rolling_upgrade and _meta fields (added sometime in 7.x) had been missing from the deprecation info api response definition. This also clarifies that the details field in the response is optional.


Anaethelion (approved):
LGTM!

PR #3215, opened 2024-12-03
Improve `Analyzer` definitions and fix various classes

flobernd:
Some analyzers had properties incorrectly marked as "required". As well added documentation for all documented fields.

This PR as well fixes different classes in various namespaces. These discrepancies have been found while working on the request converter. Thanks @l-trotta for double-checking my changes on base of the server code!


l-trotta:
the `version` field is deprecated in every analyzer. [server code](https://github.com/elastic/elasticsearch/blob/ed1e3664ad6c50d2af24b09db51448072764f663/server/src/main/java/org/elasticsearch/index/analysis/Analysis.java#L86)

l-trotta:
definitely not easy to check the code here, because it relies on Lucene classes a lot, but I think everything else looks good!

flobernd:
Thanks a lot for checking @l-trotta 🙂 Do you know which version of the server deprecated the `version` fields?

l-trotta:
@flobernd 7.14.0! https://github.com/elastic/elasticsearch/pull/74073

l-trotta:
@pquentin the validation is failing because it's not finding `sql.types.ts`. we don't have it as a namespace in the spec, where is it getting it from?

flobernd:
I removed the controversal change about the SQL + ESQL parameters for now as we decided to fix this in a separate PR.

specification/ingest/_types/Processors.ts
```diff
@@ -309,7 +310,7 @@ export class ProcessorBase {
   /**
    * Conditionally execute the processor.
    */
-  if?: string
+  if?: Script
```
andrewkroh, line 313: @flobernd I'm curious about whether this change is correct. I thought the `if` conditions for all ingest processors were represented only `string` values (that [hold Painless expressions](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html#conditionally-run-processor)). If I understand correctly, this makes the `if` value into a struct which runs contrary to the pipelines processors I have used. Could you double check this please?

flobernd: Hey @andrewkroh, sure I will double check. 

The driver for this change are some
examples in our REST API documentation which do use raw JSON instead of string to encode the condition.

Script has the same semantics, but maybe we need a new distinct type, if these conditions are no actual scripts.

flobernd: [Documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html#conditionally-run-processor) says:

> You can also specify a [stored script](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-using.html#script-stored-scripts) as the if condition.

The corresponding example is this one:

```json
PUT _ingest/pipeline/my-pipeline
{
  "processors": [
    {
      "drop": {
        "description": "Drop documents that don't contain 'prod' tag",
        "if": { "id": "my-prod-tag-script" }
      }
    }
  ]
}
```

Our current `Script` type is defined like this:

```ts
/**
 * @shortcut_property source
 * */
export class Script {
  /**
   * The script source.
   */
  source?: string
  /**
   * The `id` for a stored script.
   */
  id?: Id
  // ...
}
```

Because of the "shortcut" to `source`, any JSON `string` literal is eqivalent to initializing a `Script` object and setting the `source` property to the value.

For example:

```json
"if": "test"
```

is eqivalent to

```json
"if": {
  "source": "test"
}
```

I just double checked in the Kibana dev console again and both requests (+ the one that uses `id` instead of `source`) do indeed work fine 🙂

andrewkroh: Fantastic. My missing piece of knowledge was this `@shortcut_property`. Thank you!

PR #3221, opened 2024-12-04
[DOCS] Add SQL operation summaries

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/2635

Adds summaries to https://www.elastic.co/docs/api/doc/elasticsearch-serverless/group/endpoint-sql based on https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-apis.html

### DIFF
--- a/specification/sql/clear_cursor/ClearSqlCursorRequest.ts
+++ b/specification/sql/clear_cursor/ClearSqlCursorRequest.ts
@@ -20,6 +20,7 @@
 import { RequestBase } from '@_types/Base'
 
 /**
+ * Clear an SQL search cursor.
  * @rest_spec_name sql.clear_cursor
  * @availability stack since=6.3.0 stability=stable
  * @availability serverless stability=stable visibility=public
--- a/specification/sql/query/QuerySqlRequest.ts
+++ b/specification/sql/query/QuerySqlRequest.ts
@@ -26,6 +26,8 @@ import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { Duration, TimeZone } from '@_types/Time'
 
 /**
+ * Get SQL search results.
+ * Run an SQL request.
  * @rest_spec_name sql.query
  * @availability stack since=6.3.0 stability=stable
  * @availability serverless stability=stable visibility=public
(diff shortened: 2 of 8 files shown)


specification/sql/clear_cursor/ClearSqlCursorRequest.ts
```diff
@@ -20,6 +20,7 @@
 import { RequestBase } from '@_types/Base'
 
 /**
+ * Clear an SQL search cursor.
```
marciw, line 23: Do we have any style guidance for which article to use -- "an SQL" or "a SQL" ? 

It looks like "a SQL" has a slight edge in the es docs, and the [Google dev docs guide](https://developers.google.com/style/word-list#sql) prefers "a" -- but there might be a precedent I don't know about. I can see arguments on both sides of this one!

specification/sql/query/QuerySqlRequest.ts
```diff
@@ -26,6 +26,8 @@ import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { Duration, TimeZone } from '@_types/Time'
 
 /**
+ * Get SQL search results.
+ * Run an SQL request.
```
marciw, line 30: flagging this one in case we decide to go with "a SQL" instead of "an"

specification/sql/translate/TranslateSqlRequest.ts
```diff
@@ -23,6 +23,8 @@ import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { TimeZone } from '@_types/Time'
 
 /**
+ * Translate SQL into Elasticsearch queries.
+ * Translate an SQL search into a search API request containing Query DSL.
```
marciw, line 27: flagging this one in case we decide to go with "a SQL"

marciw (approved):
Minor style quibble to consider but LGTM 👍

specification/sql/clear_cursor/ClearSqlCursorRequest.ts
lcawl: Looks like our current (internal) style guide recommends "an SQL", which I didn't know until you mentioned this.. but good to know it's covered!

PR #3252, opened 2024-12-05
Added missing master timeout

l-trotta:
Removes validation errors in https://github.com/elastic/elasticsearch-specification/pull/3251


pquentin (approved):
Thanks! LGTM. Note that you could also have directly pushed those changes to #3251. Maybe this way helps with backports?

l-trotta:
the problem with pushing directly is that at some point during the night it reloads the changes and I lose the commits

PR #3260, opened 2024-12-06
Consider CommonCatQueryParameters to validate query parameters

pquentin:
The JSON rest-api-spec generally encodes those cat parameters, so this reduces the number of errors.

This reduces the number of validation errors from 329 to 213.


l-trotta (approved):
LGTM! tested locally validating various endponts

PR #3265, opened 2024-12-09
Fix xpack.usage ILM actions

pquentin:
I've been trying to understand today why sometimes 2 responses for xpack.usage validate, and sometimes only one. I don't understand it yet, but I have fixed the `Type 'string[]' has no properties in common with type 'IlmActions'.` issue. Before/after:

```bash
$ make validate api=xpack.usage branch=main | rg --fixed-strings "Type 'string[]' has no properties in common with type 'IlmActions'." | wc -l
    3468
$ git checkout xpack-usage-ilm-action
Switched to branch 'xpack-usage-ilm-action'
$ make validate api=xpack.usage branch=main | rg --fixed-strings "Type 'string[]' has no properties in common with type 'IlmActions'." | wc -l
       0
```

(Plenty of other issues remain, like [configurations](https://github.com/elastic/elasticsearch/blob/b4e852a54be436d8b8036da0a0ec4a472d44524a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/IndexLifecycleFeatureSetUsage.java#L191) is totally missing even though it encodes a lot of information since Elasticsearch 7.15.


specification/ilm/_types/Phase.ts
```diff
@@ -25,10 +25,7 @@ import { Duration, DurationLarge } from '@_types/Time'
 
 export class Phase {
   actions?: Actions
-  /**
-   * @es_quirk output as a millis number in XPack usage stats, which cannot roundtrip with a Duration as it requires a unit.
-   */
-  min_age?: Duration | long
+  min_age?: Duration
```
pquentin, line 28: It turns out ILM and Xpack usage output different representations for `min_age`. It's always a TimeValue, but:

 * ILM APIs [call getStringRep()](https://github.com/elastic/elasticsearch/blob/2e9ff9ae66bff80df6e8a2149fdef818a6efe573/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/Phase.java#L158C51-L158C61) which outputs [units for all non-negative values](https://github.com/elastic/elasticsearch/blob/6bb0799893f4ee4e6afa7094346fbc51d5203538/libs/core/src/main/java/org/elasticsearch/core/TimeValue.java#L341-L354)
 * XPack usage [calls getMillis()](https://github.com/elastic/elasticsearch/blob/b4e852a54be436d8b8036da0a0ec4a472d44524a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/IndexLifecycleFeatureSetUsage.java#L189C61-L189C71) which [returns a long](https://github.com/elastic/elasticsearch/blob/b4e852a54be436d8b8036da0a0ec4a472d44524a/libs/core/src/main/java/org/elasticsearch/core/TimeValue.java#L133-L135) which is why I'm using `DurationValue<UnitMillis>` now.

specification/xpack/usage/types.ts
```diff
@@ -152,6 +151,19 @@ export class HealthStatistics extends Base {
   invocations: Invocations
 }
 
+export class Phase {
+  actions: string[]
```
pquentin, line 155: https://github.com/elastic/elasticsearch/blob/b4e852a54be436d8b8036da0a0ec4a472d44524a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/IndexLifecycleFeatureSetUsage.java#L190

l-trotta: what about the `configuration` field just below?

specification/ilm/_types/Phase.ts
flobernd: Did you forgot to actually add the DurationValue part to the type union or is the diff bugged? I only see a removal of the long variant.

pquentin: DurationValue is in here: https://github.com/elastic/elasticsearch-specification/pull/3265/files#diff-3e3867651e09e047ca056bb1dbcc10583b2128b45b03cd87233d6dda1a2e6e3fR156. I duplicated the `Phase` class in the XPack spec since they're actually different.

specification/xpack/usage/types.ts
pquentin: I chose to not implement it yet, it's long and complicated: https://github.com/elastic/elasticsearch/blob/b4e852a54be436d8b8036da0a0ec4a472d44524a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/IndexLifecycleFeatureSetUsage.java#L228.

flobernd (approved):

PR #3308, opened 2024-12-13
Output the public API only in the stack OpenAPI schema

### LINKED ISSUE #3300: Handling visibility=private
In https://github.com/elastic/elasticsearch-specification/pull/3299 I noticed that one of the operations has `visibility=private`:

```
* @availability stack since=5.4.0 stability=stable visibility=private
 * @availability serverless stability=stable visibility=private
```

It is appropriately omitted from the Serverless OpenAPI document, however it appears in the stack openAPI document: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-validate

Is this disparity because the `filter-for-serverless` make command (https://github.com/elastic/elasticsearch-specification/blob/fb259298355a26881333823a52b9a17b685b8bd6/Makefile#L54) has `filter-by-availability -- --serverless --visibility=public`?  If so, where would be the appropriate place to add a similar filter on the other output file?

swallez:
Keep only public items in the stack OpenAPI export

Fixes #3300


Anaethelion (commented):
What about https://github.com/elastic/elasticsearch-specification/blob/3f5c8469587f0578b224341e748e171d5e45553a/compiler-rs/clients_schema_to_openapi/src/main.rs#L86 ?

We should probably keep these two in sync.

swallez:
> We should probably keep these two in sync.

Good catch. I have refactored the code so that the filtering is done in a single location.

l-trotta (approved):

PR #3313, opened 2024-12-16
Add POST _unified for the inference API

jonathan-buttner:
This PR adds the `_unified` route for the Inference API. This coincides with the elasticsearch PR: https://github.com/elastic/elasticsearch/pull/117589


specification/_json_spec/inference.unified_inference.json
```diff
@@ -0,0 +1,45 @@
+{
+  "inference.unified_inference": {
+    "documentation": {
+      "url": "https://www.elastic.co/guide/en/elasticsearch/reference/master/unified-inference-api.html",
+      "description": "Perform inference using the Unified Schema"
+    },
+    "stability": "stable",
+    "visibility": "public",
+    "headers": {
+      "accept": ["text/event-stream"],
```
jonathan-buttner, line 10: This is what we have for the streaming api here: https://github.com/elastic/elasticsearch-specification/blob/main/specification/_json_spec/inference.stream_inference.json#L10

Not sure if this is correct if it's indicating that the post request accepts an event stream or that the response returns one 🤔

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+  type: string
+  /**
+   * The function definition.
+   */
+  function: CompletionToolFunction
+}
+
+/**
+ * Perform inference on the service using the Unified Schema
+ * @rest_spec_name inference.unified_inference
```
jonathan-buttner, line 158: `unified_inference` seemed to follow `stream_inference`. I'm open to other ideas though.

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+ * @availability stack since=8.18.0 stability=stable visibility=public
```
jonathan-buttner, line 159: @davidkyle Double check that this is what we want

specification/_json_spec/inference.unified_inference.json
pquentin: Both HTTP headers are used in client requests:

 * Accept tells Elasticsearch that returning event streams is OK
 * Content-Type is about the request's body, which is indeed JSON here.

In other words, this is correct, thank you!

specification/inference/unified_inference/UnifiedRequest.ts
pquentin: No strong opinions, this seems OK.

specification/inference/unified_inference/UnifiedRequest.ts
pquentin: This looks good to me.

pquentin (commented):
Thanks! I answered your questions, but did not look at the contents yet. Do we have YAML tests in Elasticsearch for this feature? It would help to validate the requests.

maxhniebergall (commented):
LGTM

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
+/*
```
l-trotta, line 1: we usually try to have only the request class in the corresponding Request file, and all other types we put in the `types` folder above (this is just to make it easier to maintain)

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * The type of the tool call.
+   */
+  type: string
+}
+
+/**
+ * An object representing part of the conversation.
+ */
+export interface Message {
```
l-trotta, line 75: I think this is missing [name](https://github.com/elastic/elasticsearch/blob/7cf28a910e6aa8e5679a73eb54075b6778277b67/server/src/main/java/org/elasticsearch/inference/UnifiedCompletionRequest.java#L158)
```
name?: string
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+   */
+  content: string | Array<ContentObject>
+  /**
+   * The role of the message author.
+   */
+  role: string
+  /**
+   * The tool call that this message is responding to.
+   */
+  tool_call_id?: string
```
l-trotta, line 87:
```suggestion
  tool_call_id?: Id
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+     */
+    messages: Array<Message>
+    /**
+     * The ID of the model to use.
+     */
+    model?: string
+    /**
+     * The upper bound limit for the number of tokens that can be generated for a completion request.
+     */
+    max_completion_tokens?: number
```
l-trotta, line 192:
```suggestion
    max_completion_tokens?: long
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * A sequence of strings to control when the model should stop generating additional tokens.
+     */
+    stop?: Array<string>
+    /**
+     * The sampling temperature to use.
+     */
+    temperature?: number
```
l-trotta, line 200:
```suggestion
    temperature?: float
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+     */
+    tool_choice?: string | CompletionToolChoice
+    /**
+     * A list of tools that the model can call.
+     */
+    tools?: Array<CompletionTool>
+    /**
+     * Nucleus sampling, an alternative to sampling with temperature.
+     */
+    top_p?: number
```
l-trotta, line 212:
```suggestion
    top_p?: float
```

specification/inference/_types/Results.ts
```diff
@@ -88,6 +88,120 @@ export class InferenceResult {
...
+}
+
+/**
+ * Respresents the result format for a completion request using the Unified Inference API.
+ */
+export class UnifiedInferenceResult {
+  /**
+   * A unique identifier for the chat completion
+   */
+  id: string
```
l-trotta, line 186:
```suggestion
  id: Id
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * An object representing part of the conversation.
+ */
+export interface Message {
+  /**
+   * The content of the message.
+   */
+  content: string | Array<ContentObject>
```
l-trotta, line 79: Unions such as this one have to be handled in a separate type for them to be understandable by the static client, so here we need a new type:
```
/**
 * @codegen_names string, object
 */
export type MessageContent = string | Array<ContentObject>
```

```
content: MessageContent
```

specification/inference/unified_inference/UnifiedRequest.ts
```diff
@@ -0,0 +1,214 @@
...
+     */
+    stop?: Array<string>
+    /**
+     * The sampling temperature to use.
+     */
+    temperature?: number
+    /**
+     * Controls which tool is called by the model.
+     */
+    tool_choice?: string | CompletionToolChoice
```
l-trotta, line 204: Same as `content` above, new type needed to define union:
```
/**
 * @codegen_names string, object
 */
export type CompletionToolType = string | CompletionToolChoice
```

```
tool_choice?: CompletionToolType
```

jonathan-buttner:
> @jonathan-buttner I have a question about the response format: is there a way to get it in json format, or is it just text/stream?

Thanks for the review! Currently it's only `text/stream` but in the near future we'd like to add a non-streaming version. Is that possible to do using the same route (.../_unified)? Or do we need to create a new one?

jonathan-buttner:
> Thanks! I answered your questions, but did not look at the contents yet. Do we have YAML tests in Elasticsearch for this feature? It would help to validate the requests.

Thanks for the review! Not yet, let me see if I can add some.

l-trotta:
> > @jonathan-buttner I have a question about the response format: is there a way to get it in json format, or is it just text/stream?
> 
> Thanks for the review! Currently it's only `text/stream` but in the near future we'd like to add a non-streaming version. Is that possible to do using the same route (.../_unified)? Or do we need to create a new one?

the thing is: if the return type is just text then the response body should be just `string`, because the whole body structure is needed just for json, not other types of responses. @pquentin like we were talking about hot threads remember?

jonathan-buttner:
> > > @jonathan-buttner I have a question about the response format: is there a way to get it in json format, or is it just text/stream?
> > 
> > 
> > Thanks for the review! Currently it's only `text/stream` but in the near future we'd like to add a non-streaming version. Is that possible to do using the same route (.../_unified)? Or do we need to create a new one?
> 
> the thing is: if the return type is just text then the response body should be just `string`, because the whole body structure is needed just for json, not other types of responses. @pquentin like we were talking about hot threads remember?

Oh sorry I might have misunderstood. Here's what Postman parses out:

image

Where each one of those messages is valid json:

```
{
    "id": "chatcmpl-AfVcn7De1ibDPKB7iBy83oEhpHxC9",
    "choices": [],
    "model": "gpt-4o-2024-08-06",
    "object": "chat.completion.chunk",
    "usage": {
        "completion_tokens": 6,
        "prompt_tokens": 63,
        "total_tokens": 69
    }
}
```

Do we still want the response to be only text?

pquentin:
Postman makes it look mainly like JSON, but those are [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events). Is this unified API going to be used for multiple providers? OpenAI, Anthropic Claude, and Google Gemini all use server-sent events but then provide really different structures: https://til.simonwillison.net/llms/streaming-llm-apis.

If yes, then clients would need to know what provider this is (in the content-type header?). There would not be a need to map the options in the spec; it's OK to treat it as an opaque string or buffer. Also, I'm afraid that more complete discussions around this will have to wait for next year.

jonathan-buttner:
> Postman makes it look mainly like JSON, but those are [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events). Is this unified API going to be used for multiple providers? OpenAI, Anthropic Claude, and Google Gemini all use server-sent events but then provide really different structures: https://til.simonwillison.net/llms/streaming-llm-apis.
> 
> If yes, then clients would need to know what provider this is (in the content-type header?). There would not be a need to map the options in the spec; it's OK to treat it as an opaque string or buffer. Also, I'm afraid that more complete discussions around this will have to wait for next year.

>  Postman makes it look mainly like JSON, but those are [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).

Yeah that makes sense

> Is this unified API going to be used for multiple providers? OpenAI, Anthropic Claude, and Google Gemini all use server-sent events but then provide really different structures: https://til.simonwillison.net/llms/streaming-llm-apis.

Sorry I should have provided more details when opening the PR. The hope with this API is to provide a consistent request and response schema that will be the same across multiple providers. Internally we'll handle transforming the request and response to match the individual providers and our schema.

When we get a request for Anthropic we'll translate our "unified" schema into a request for Anthropic specifically and then as we get responses from Anthropic we'll translate it back into the format that fits the response schema we're proposing here.

specification/inference/unified_inference/UnifiedRequest.ts
jonathan-buttner: Oops! Thanks for pointing that out, that should be removed. I'll make that change in the relevant places.

l-trotta:
Hey @jonathan-buttner! the team had a discussion about this and we agreed to keep the result as a simple ArrayBuffer, and leave it to the users to parse it using the tools they prefer. This means adding a new type in the Binary.ts file:
```
// Streaming endpoints response
export type StreamResult = ArrayBuffer
```
and using that as the whole response, so UnifiedResponse.ts becomes:
```
export class Response {
  body: StreamResult
}
```
the specific classes modeled only for the response can therefore be deleted. 
Every other comment I made regarding the request are still valid, let me know if there's something I could explain better or if you need any help with this. 
Thank you :)

l-trotta (approved):
LGTM! thank you

PR #3315, opened 2024-12-16
Mark query roles API public in Serverless

slobodanadamovic:
Marks [Query Role API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-query-role.html) as public in Serverless and generates related specification.

Relates to https://github.com/elastic/elasticsearch/pull/118798


slobodanadamovic:
closing in favour of https://github.com/elastic/elasticsearch-specification/pull/3316, since validation is not supported from forks

PR #3391, opened 2024-12-30
Add summary and tag to ml.validate

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/3300

The  `_ml/anomaly_detectors/_validate` [endpoint](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-validate) is currently appearing in the `machine learning` group instead of the `machine learning anomaly detection` group and is missing a summary.

When the referenced issue is fixed, it should not be published but until then I've fixed the doc tag and summary.


davidkyle (approved):
LGTM

Issue #3377, opened 2024-12-19
[DOCS] Add minimal specifications for all Elasticsearch API docs


lcawl:
In order to retire the old API documentation, we need specifications for each API that appears in the old docs unless the API is internal or being removed.  The following APIs do not yet appear in the OpenAPI document:

- [x] [Post event to an analytics collection](https://www.elastic.co/guide/en/elasticsearch/reference/master/post-analytics-collection-event.html)  https://github.com/elastic/elasticsearch-specification/pull/3363
- [x] [Update connector features](https://www.elastic.co/guide/en/elasticsearch/reference/master/update-connector-features-api.html) https://github.com/elastic/elasticsearch-specification/pull/3364
- [x] [Check in connector sync job](https://www.elastic.co/guide/en/elasticsearch/reference/master/check-in-connector-sync-job-api.html) https://github.com/elastic/elasticsearch-specification/pull/3365
- [x] [Claim connector sync job](https://www.elastic.co/guide/en/elasticsearch/reference/master/claim-connector-sync-job-api.html) https://github.com/elastic/elasticsearch-specification/pull/3366
- [x] [Set connector sync job error](https://www.elastic.co/guide/en/elasticsearch/reference/master/set-connector-sync-job-error-api.html) https://github.com/elastic/elasticsearch-specification/pull/3367
- [x] [Set connector sync job stats](https://www.elastic.co/guide/en/elasticsearch/reference/master/set-connector-sync-job-stats-api.html) https://github.com/elastic/elasticsearch-specification/pull/3368
- [x] [Find field structure](https://www.elastic.co/guide/en/elasticsearch/reference/master/find-field-structure.html) https://github.com/elastic/elasticsearch-specification/pull/3346
- [x] [Find messages structure](https://www.elastic.co/guide/en/elasticsearch/reference/master/find-message-structure.html) https://github.com/elastic/elasticsearch-specification/pull/3346
- [x] [Get data stream lifecycle stats](https://www.elastic.co/guide/en/elasticsearch/reference/master/data-streams-get-lifecycle-stats.html) https://github.com/elastic/elasticsearch-specification/pull/3418, https://github.com/elastic/elasticsearch/pull/119472
- [x] [ES|QL async query](https://www.elastic.co/guide/en/elasticsearch/reference/master/esql-async-query-api.html) https://github.com/elastic/elasticsearch-specification/pull/3398
[...]

PR #3396, opened 2024-12-31
Add security.update_settings and security.get_settings

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/3377

This PR adds two missing specifications based on information in https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api-update-settings.html and https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api-get-settings.html

### DIFF
--- a/specification/security/get_settings/SecurityGetSettingsResponse.ts
+++ b/specification/security/get_settings/SecurityGetSettingsResponse.ts
@@ -16,8 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { IndexSettings } from '@indices/_types/IndexSettings'
-import { Dictionary } from '@spec_utils/Dictionary'
+import { SecuritySettings } from '@security/_types/SecuritySettings'
 
 export class Response {
   body: {
@@ -34,8 +33,4 @@ export class Response {
      */
     'security-tokens': SecuritySettings
   }
-}
-
-export class SecuritySettings {
-  index?: IndexSettings
 }
--- a/specification/security/get_settings/SecurityGetSettingsResponse.ts
+++ b/specification/security/get_settings/SecurityGetSettingsResponse.ts
@@ -28,7 +28,7 @@ export class Response {
     /**
      * Settings for the index used to store profile information.
      */
-    'security-profile': Dictionary<string, string>
+    'security-profile': SecuritySettings
     /**
      * Settings for the index used to store tokens.
      */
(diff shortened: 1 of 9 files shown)


specification/security/get_settings/SecurityGetSettingsResponse.ts
```diff
@@ -0,0 +1,41 @@
...
+export class Response {
+  body: {
+    /**
+     * Settings for the index used for most security configuration, including native realm users and roles configured with the API.
+     */
+    security: SecuritySettings
+    /**
+     * Settings for the index used to store profile information.
+     */
+    'security-profile': Dictionary<string, string>
```
l-trotta, line 31: it's all settings - [server code](https://github.com/elastic/elasticsearch/blob/6f43d898fb5eafd317fabe53a1b6d402b56815f9/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/settings/GetSecuritySettingsAction.java#L70)
```suggestion
    'security-profile': SecuritySettings
```

specification/security/update_settings/SecurityUpdateSettingsRequest.ts
```diff
@@ -0,0 +1,61 @@
...
+     * The period to wait for a response.
+     * If no response is received before the timeout expires, the request fails and returns an error.
+     */
+    timeout?: Duration
+  }
+  body: {
+    /**
+     * Settings for the index used for most security configuration, including native realm users and roles configured with the API.
+     */
+    security?: Dictionary<string, string>
```
l-trotta, line 51:
```suggestion
    security?: SecuritySettings
```

specification/security/update_settings/SecurityUpdateSettingsRequest.ts
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Settings for the index used to store profile information.
+     */
+    'security-profile'?: Dictionary<string, string>
+    /**
+     * Settings for the index used to store tokens.
+     */
+    'security-tokens'?: Dictionary<string, string>
```
l-trotta, line 59:
```suggestion
    'security-tokens'?: SecuritySettings
```

specification/security/get_settings/SecurityGetSettingsResponse.ts
flobernd: It uses Kebap case here instead of underscores. That's ... interesting? Very interesting API design choice 😅

I have to remember to update my code generator to support that case.

lcawl:
> after these fixes we'll still have a validation error, not because of this specific API, but it's just because it uses IndexSettings, which is still not mapped 100% correctly

That seems okay to me. I think this PR work is now complete but if there's something still outstanding just lmk!

l-trotta (approved):
LGTM!

PR #3401, opened 2025-01-01
Add search_application.render_query

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/3377

This PR adds a missing specification based on https://www.elastic.co/guide/en/elasticsearch/reference/master/search-application-render-query.html


specification/search_application/render_query/SearchApplicationsRenderQueryRequest.ts
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * The name of the search application to render teh query for.
+     */
+    name: Name
+  }
+  /**
+   * Contains parameters for a search application.
+   */
+  /** @codegen_name search_application */
+  body: SearchApplicationParameters
```
pquentin: This is confusing, but the parameters for [Render Search Application Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-application-render-query.html) and [Put Search Application](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-search-application.html#put-search-application-dictionary-param) are different, so `SearchApplicationParameters` won't work here, which is also why validation is failing.

You can see valid examples here: https://github.com/elastic/clients-flight-recordings/tree/main/elasticsearch/tmp-platinum/search_application.render_query.

Since params are script parameters, I've reused the definition from https://github.com/elastic/elasticsearch-specification/blob/97867a6fc366553b205191b08714bfdacd0ef6d3/specification/_types/Scripting.ts#L87-L90 in [`d2c86e8` (#3401)](https://github.com/elastic/elasticsearch-specification/pull/3401/commits/d2c86e84911b2e117db24308fc96ff28c65099e2).

pquentin (approved):

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.x|[](https://github.com/elastic/elasticsearch-specification/pull/3421)|
|✅|8.17|[](https://github.com/elastic/elasticsearch-specification/pull/3422)|

### Questions ?

specification/search_application/render_query/SearchApplicationsRenderQueryRequest.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * You must have `read` privileges on the backing alias of the search application.
+ * @rest_spec_name search_application.render_query
+ * @availability stack since=8.9.0 stability=experimental visibility=public
+ * @doc_id search-render-query
+ */
+export interface Request extends RequestBase {
+  path_parts: {
+    /**
+     * The name of the search application to render teh query for.
```
flobernd, line 38: Typo: `the`

specification/search_application/render_query/SearchApplicationsRenderQueryResponse.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export class Response {
+  body: {}
```
flobernd, line 21: Q: Do we somewhere track these cases where we only defined a request, but not a proper response?

Doing it like this will suppress the compiler warning about missing request/response.

lcawl: Good point. Does this mean it's preferable to omit the response file entirely when we don't know the contents of the body?

pquentin: > Doing it like this will suppress the compiler warning about missing request/response.

I'm also surprised that `make validate` passed, which is why I did not realize that the body was missing.

> Good point. Does this mean it's preferable to omit the response file entirely when we don't know the contents of the body?

Ideally, yes, but it breaks the OpenAPI transformation right now.

PR #3458, opened 2025-01-07
Convert examples to yaml and add subfolders

lcawl:

This PR moves existing example files to an "examples" subfolder.
It also converts existing example files from json to yaml so that we can comment out some of the content that's generating linting errors.


l-trotta (approved):
looks good, thank you for taking the time to rearrange everything :)

specification/eql/get_status/examples/response/EqlGetStatusResponseExample1.yaml
```diff
@@ -0,0 +1,12 @@
+# summary:
+description: A successful response for getting status information for an async EQL search.
+# type: response
+# response_code: 200
+value: >-
```
szabosteve, line 5: This is the only place in the `value` field where `>` is used as a separator instead of `|`. I don't know if it's intentional or not, and I'm not even sure if it makes a difference at all, but I thought it might be worth pointing it out.

szabosteve (commented):
One question, otherwise LGTM!

PR #3469, opened 2025-01-08
Add endpoint URLs to the API specification (phase 1)

swallez:
The endpoints' URL path templates and http methods are currently missing from the API spec (TypeScript source) and are imported from the json rest-spec when the spec if compiled to `schema.json`. This means that the API spec currently does not contain all the information defining the API.

This PR allows endpoint URLs to be captured in request definitions in the API spec. The format follows that of the json rest-spec and schema.json:

```ts
  urls: [
    {
      /** @deprecated 1.2.3 Use something else */
      path: '/some/path/{part}',
      methods: ["GET", "POST"]
    }
  ]
  path_parts: {
    part: ...
```

This PR also includes a code modification utility to automatically add this information to the API spec from json rest-spec. Only two endpoints have been updated to test and demonstrate this utility. The full update will be done in a subsequent PR so that this one can be backported to other branches.

This PR is split in several commits:
* 775f49d adds the code modification utility that adds URL paths to API spec requests.
* 146d0ed updates the compiler to handle the additional `urls` field, along with validation.
* a463d22 updates two request definitions with the utility mentioned above. This is the result of running
  ```bash
  echo '../specification/_global/clear_scroll/ClearScrollRequest.ts
  ../specification/_global/bulk/BulkRequest.ts' | (cd compiler-rs; cargo run --bin add_url_paths ../output/schema/schema.json | sh)
  ```
  The spec is also compiled, and we can see that the only changed are location information in these two requests because of the added `urls` property.
* 070c15f updates the docs with the new `urls`.


flobernd:
The JS compiler part looks good 🙂

Anaethelion (approved):

PR #3479, opened 2025-01-08
Add SQL API examples

lcawl:

This PR copies examples from https://www.elastic.co/guide/en/elasticsearch/reference/master/sql-apis.html
It also adds missing `doc_id` values, sorts some properties (required properties first, then optional properties alphabetically).
It also edits and augments some descriptions based on what was in the documentation.


specification/sql/query/QuerySqlRequest.ts
```diff
@@ -31,95 +31,117 @@ import { Duration, TimeZone } from '@_types/Time'
...
+     * The maximum number of rows (or entries) to return in one response.
      * @server_default 1000
      */
     fetch_size?: integer
     /**
-     * Elasticsearch query DSL for additional filtering.
-     * @doc_id sql-rest-filtering
-     * @server_default none
+     * If `false`, the API returns an exception when encountering multiple values for a field.
+     * If `true`, the API is lenient and returns the API returns the first value from the array with no guarantee of consistent results.
```
marciw, line 79:
```suggestion
     * If `true`, the API is lenient and returns the first value from the array with no guarantee of consistent results.
```

PR #3529, opened 2025-01-15
Add endpoint URLs to the API specification (phase 2)

swallez:
Follow-up to PR #3524: this adds URL paths to all endpoints. This results in no change in `schema.json` except updated locations.

This is the result of running:

```bash
(cd compiler-rs; find ../specification -name '*Request.ts' | cargo run --bin add_url_paths ../output/schema/schema.json | sh)
```


Anaethelion (approved):
LGTM

PR #3545, opened 2025-01-16
Inference task type endpoints

jonathan-buttner:
This PR makes breaking changes to the client for the inference API. Prior to this PR we had a single endpoint for most task types supported in the inference API: `_inference/<optional_task_type>/<inference id>`. After discussion with @swallez we decided to make the task type required in the URL. This way we could have separate requests and responses for each task type.

This PR does not include another item of work to make well defined `task_settings` for each route. Correct me if I'm wrong, but I don't believe that would be a breaking change? If it is not a breaking change, I think we can defer that work until later.


specification/inference/_types/Results.ts
```diff
@@ -57,13 +65,29 @@ export class TextEmbeddingResult {
...
  * The completion result object
  */
 export class CompletionResult {
   result: string
 }
 
+/**
+ * Defines the completion result.
+ */
+export class CompletionInferenceResult {
```
jonathan-buttner, line 87: I'm open to other ideas for naming the classes. `*Result` was already taken for everything for the nested field which is why I went with `*InferenceResult`.

specification/inference/rerank/RerankRequest.ts
```diff
@@ -0,0 +1,66 @@
...
+     * Specifies the amount of time to wait for the inference request to complete.
+     * @server_default 30s
+     */
+    timeout?: Duration
+  }
+  body: {
+    /**
+     * Query input.
+     */
+    query: string
```
jonathan-buttner, line 55: `query` is required for the rerank task type.

specification/inference/stream_completion/StreamInferenceRequest.ts
```diff
@@ -63,5 +55,9 @@ export interface Request extends RequestBase {
      * NOTE: Inference endpoints for the completion task type currently only support a single string as input.
      */
     input: string | string[]
+    /**
+     * Optional task settings
+     */
+    task_settings?: TaskSettings
```
jonathan-buttner, line 61: Adding this because I think it was missing before.

specification/_json_spec/inference.chat_completion_inference.json
```diff
@@ -0,0 +1,31 @@
...
+    "stability": "stable",
+    "visibility": "public",
+    "headers": {
+      "accept": ["text/event-stream"],
+      "content_type": ["application/json"]
+    },
+    "url": {
+      "paths": [
+        {
+          "path": "/_inference/chat_completion/{inference_id}/_unified",
```
davidkyle, line 16: Shouldn't this be _stream

```suggestion
          "path": "/_inference/chat_completion/{inference_id}/_stream",
```

jonathan-buttner: No, it does have `_unified` at the end. I think technically we could remove it since the client code doesn't need the results to be in a specific format for SSE.

specification/_json_spec/inference.chat_completion_unified.json
```diff
@@ -0,0 +1,31 @@
...
+    "stability": "stable",
+    "visibility": "public",
+    "headers": {
+      "accept": ["text/event-stream"],
+      "content_type": ["application/json"]
+    },
+    "url": {
+      "paths": [
+        {
+          "path": "/_inference/chat_completion/{inference_id}/_unified",
```
davidkyle, line 16:
```suggestion
          "path": "/_inference/chat_completion/{inference_id}/_stream",
```

specification/inference/_types/Results.ts
```diff
@@ -37,6 +37,14 @@ export class SparseEmbeddingResult {
   embedding: SparseVector
 }
 
+/**
+ * The response format for the sparse embedding request.
+ */
+export class SparseEmbeddingInferenceResult {
+  // TODO should we make this optional if we ever support multiple encoding types? So we can make it a variant
+  sparse_embedding: Array<SparseEmbeddingResult>
```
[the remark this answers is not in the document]
davidkyle: For text embeddings, the pattern used in the `InferenceResult` class (also in this file) is to have a different variant for each type.
```
  text_embedding_bytes?: Array<TextEmbeddingByteResult>
  text_embedding_bits?: Array<TextEmbeddingByteResult>
  text_embedding?: Array<TextEmbeddingResult>
```

Sparse would be the same:
```
  sparse_embedding?: Array<SparseEmbeddingResult>
  sparse_embedding_byte?: Array<SparseEmbeddingByteResult>
```

davidkyle (commented):
Please add `chat_completion` to the TaskType enum https://github.com/elastic/elasticsearch-specification/blob/main/specification/inference/_types/TaskType.ts#L27

specification/inference/_types/Results.ts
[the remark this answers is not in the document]
jonathan-buttner: Actually when I make the sparse embedding type a variant, I get an error that indicates there must be multiple fields in the type to be able to leverage the variant type. So I think we can make this change when we need to.

jonathan-buttner:
Created a docs request here to address the docs migration failure: https://github.com/elastic/docs-content/issues/339

davidkyle (approved):
LGTM

pquentin:
Sorry for dropping the ball on this and not reviewing it as we should have. This pull request introduces breaking changes we did not anticipate - we'll need to revisit the approach. We'll come back to you with specific recommendations.

PR #3552, opened 2025-01-20
Remove indices.unfreeze API

pquentin:
This was removed in https://github.com/elastic/elasticsearch/pull/119227, but our sync does not handle removals as it runs `make generate`, and this was breaking generation.


.github/validate-pr/index.js
```diff
@@ -53,7 +53,7 @@ async function run () {
       per_page: 100
     })
     if (data.length > 0) {
-      files.push(...data.map(entry => entry.filename))
+      files.push(...data.filter(entry => entry.status !== 'deleted').map(entry => entry.filename))
```
pquentin, line 56: Validation fails because it tries to validate `indices.unfreeze`. This change should fix it, but it needs to be merged before it has any effect.

PeteGillinElastic (approved):
Thanks!

PR #3563, opened 2025-01-22
Fix typo in overlays

lcawl:
There was a redundant line in one of the overlay files, likely caused by a merge conflict.
It was caught by our linter, so this is just a quick fix.


leemthompo (approved):
🧹

PR #3566, opened 2025-01-22
Add remove_index_blocks param to _create_from

parkertimmins:
Add a new parameter to the request body for `_create_from/{source}/{dest}` api.
Parameter is an optional boolean called `remove_index_blocks`.

Add to ES in https://github.com/elastic/elasticsearch/pull/120548/files

### DIFF
--- a/specification/migrate/create_from/MigrateCreateFromRequest.ts
+++ b/specification/migrate/create_from/MigrateCreateFromRequest.ts
@@ -53,6 +53,7 @@ export class CreateFrom {
   settings_override?: IndexSettings
   /**
    * If index blocks should be removed when creating destination index (optional)
+   * @server_default false
    */
   remove_index_blocks?: boolean
 }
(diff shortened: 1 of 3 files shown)


specification/migrate/create_from/MigrateCreateFromRequest.ts
```diff
@@ -51,4 +51,8 @@ export class CreateFrom {
    * Settings overrides to be applied to the destination index (optional)
    */
   settings_override?: IndexSettings
+  /**
+   * If index blocks should be removed when creating destination index (optional)
+   */
+  remove_index_blocks?: boolean
```
pquentin, line 57:
```suggestion
  /**
   * If index blocks should be removed when creating destination index (optional)
   * @server_default false
   */
  remove_index_blocks?: boolean
```

pquentin (approved):
Thanks! LGTM.

PR #3591, opened 2025-01-24
[DOCS] Upsert documentation clarification

lcawl:
This PR copies over updates made in https://github.com/elastic/elasticsearch/pull/120684

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -38461,7 +38461,7 @@
                     "$ref": "#/components/schemas/_global.search._types:SourceConfig"
                   },
                   "upsert": {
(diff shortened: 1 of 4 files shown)


leemthompo:
You're too fast @lcawl! Might be a few tweaks on original but perhaps @thekofimensah could handle the updates? We mentioned previously that API docs updates need to prioritize working in the spec, rather than the old asciidoc docs.

kilfoyle (approved):
LGTM! 🐫

lcawl:
> You're too fast @lcawl! Might be a few tweaks on original but perhaps @thekofimensah could handle the updates? We mentioned previously that API docs updates need to prioritize working in the spec, rather than the old asciidoc docs.

Sure! @thekofimensah if you want to contribute to this PR or replace it with another, lmk! Or if you want to chat about any of the processes related to this repo, I'm happy to chat too

PR #3611, opened 2025-01-24
[DOCS] Clarify Count API query options

marciw:
Applies changes from https://github.com/elastic/elasticsearch/pull/120564: notes that you can use the `q` parameter or a query in the request body, but not both. Plus other incidental sprucing.


lcawl (approved):
Thanks, LGTM!

PR #3639, opened 2025-01-28
Making async esql responses binary

l-trotta:
The esql async API response format varies depending on the format chosen in the request, also it's not supported when using formats that are not json compatible (so it only supported by json, yaml, cbor, smile). The spec as it is only supports json results, so all the responses should be changed to a binary representation, and the clients will choose which to support. 

In the backports PR the package change for EsqlFormat will be reversed, as it is a breaking change that we can introduce in the next major.


pquentin (approved):
Thanks! LGTM. Do we need approval from other maintainers?

PR #3640, opened 2025-01-28
Correcting the response format for ingest simulate

masseyke:
This is an update to #3400. The simulate ingest response is fairly different from the simulate pipeline response, so I've pulled it into its own object.
Specifically, simulate ingest has an `executed_pipelines` array, and it does not have several fields that the simulate pipeline response has.


specification/simulate/ingest/SimulateIngestResponse.ts
```diff
@@ -17,8 +17,56 @@
...
+import { Stringified } from '@spec_utils/Stringified'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
+import { Id, IndexName, VersionNumber } from '@_types/common'
+import { ErrorCause } from '@_types/Errors'
 
 export class Response {
   body: { docs: SimulateDocumentResult[] }
 }
+
+export class SimulateDocumentResult {
```
l-trotta, line 31: we try to avoid duplicate names since it can break certain clients, could we rename this to something else? like `SimulateIngestDocumentResult`

specification/simulate/ingest/SimulateIngestResponse.ts
```diff
@@ -17,8 +17,56 @@
...
+
+/**
+ * The results of ingest simulation on a single document. The _source of the document contains
+ * the results after running all pipelines listed in executed_pipelines on the document. The
+ * list of executed pipelines is derived from the pipelines that would be executed if this
+ * document had been ingested into _index.
+ *
+ * @behavior_meta AdditionalProperties fieldname=metadata description="Additional metadata"
+ */
+export class DocumentSimulation
```
l-trotta, line 49: same here, coule be `IngestDocumentSimulation`?

specification/simulate/ingest/SimulateIngestResponse.ts
```diff
@@ -17,8 +17,56 @@
...
+   */
+  _index: IndexName
+  /**
+   * JSON body for the document.
+   */
+  _source: Dictionary<string, UserDefinedValue>
+  /**
+   *
+   */
+  _version?: Stringified<VersionNumber>
```
l-trotta, line 67: is `_version` actually nullable? checking the [server code](https://github.com/elastic/elasticsearch/blob/0393e56fa72890d4b46aa5d2b5923a08faf27a69/server/src/main/java/org/elasticsearch/action/ingest/SimulateIndexResponse.java#L92) it seems like it's always added.

masseyke: You're right -- I had just copy/pasted without looking.

masseyke: I also realized that `error` was in the wrong place (it should be within the doc), and `ignored_fields` was missing altogether.

specification/simulate/ingest/SimulateIngestResponse.ts
```diff
@@ -17,8 +17,69 @@
...
+  /**
+   * A list of the names of the pipelines executed on this document.
+   */
+  executed_pipelines: Array<string>
+  /**
+   * A list of the fields that would be ignored at the indexing step. For example, a field whose
+   * value is larger than the allowed limit would make it through all of the pipoelines, but
+   * would not be indexed into Elasticsearch.
+   */
+  ignored_fields?: Array<Dictionary<IgnoredFieldKey, string>>
```
l-trotta, line 71: this dictionary is always just one "field" + string right? we have the type SingleKeyDictionary if so
```suggestion
  ignored_fields?: Array<SingleKeyDictionary<IgnoredFieldKey, string>>
```

specification/simulate/ingest/SimulateIngestResponse.ts
```diff
@@ -17,8 +17,69 @@
...
+   *
+   */
+  _version: Stringified<VersionNumber>
+  /**
+   * A list of the names of the pipelines executed on this document.
+   */
+  executed_pipelines: Array<string>
+  /**
+   * A list of the fields that would be ignored at the indexing step. For example, a field whose
+   * value is larger than the allowed limit would make it through all of the pipoelines, but
```
l-trotta, line 68:
```suggestion
   * value is larger than the allowed limit would make it through all of the pipelines, but
```

specification/simulate/ingest/SimulateIngestResponse.ts
l-trotta: let's leave it like this then

l-trotta (approved):
LGTM!

PR #3643, opened 2025-01-28
Update spec for ESQL async query stop

smalyshev:
Update spec for:
* ESQL async `POST query/stop` API
* is_partial in ESQL query response

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -8922,7 +8922,7 @@
                       "type": "boolean"
                     },
                     "is_partial": {
-                      "description": "Indicates whether the query returned partial result, e.g. because if was stopped.",
+                      "description": "Indicates whether the query returned partial result.\nFor example, this can occur because the query was stopped.",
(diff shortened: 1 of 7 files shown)


JoshMock (approved):
These changes look good to me!

JoshMock:
Pausing before merging to verify with @pquentin about what the missing tests indicates. I'm pretty sure it's fine, but I want to make sure.

smalyshev:
It will probably be non-trivial to test async requests within the context of the spec, since they require some special timing, so I am not surprised the spec does not test them. That said, please tell me if anything needs to be added here.

pquentin (changes requested):
Can we please make the body binary as done in https://github.com/elastic/elasticsearch-specification/pull/3639? The current representation isn't usable for typed client. (The current behavior of "return structured response, unless it's ready, then you get binary data" is unfortunate.)

smalyshev:
@pquentin updated, please review

pquentin (approved):
Thanks! LGTM.

PR #3654, opened 2025-01-29
[DOCS] Clarify remote/info and resolve/cluster descriptions

lcawl:
Relates to https://github.com/elastic/elasticsearch/pull/118993

This PR updates the remote info and resolve cluster specifications to (1) copy some missing descriptions from the docs (https://www.elastic.co/guide/en/elasticsearch/reference/master/cluster-remote-info.html and https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-resolve-cluster-api.html) and (2) incorporate the changes from https://github.com/elastic/elasticsearch/pull/118993

NOTE:

1. I noticed that the examples have been removed from the old docs, so if we need to do the same in these docs, let me know!
2. It looks like `cluster_credentials`, which is mentioned in https://www.elastic.co/guide/en/elasticsearch/reference/master/cluster-remote-info.html, is missing from this specification.

### Preview

The descriptions contain some sections, which I checked work in Bump.sh like this:

image

They also contain some `> info` callouts and external links, which look like this:

image

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -4334,7 +4334,10 @@
           "cluster"
         ],
         "summary": "Get remote cluster information",
-        "description": "Get all of the configured remote cluster information.\nThis API returns connection and endpoint information keyed by the configured remote cluster alias.",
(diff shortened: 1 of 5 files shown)


lcawl:
Replaced by https://github.com/elastic/elasticsearch-specification/pull/3655

PR #3657, opened 2025-01-29
Use current URLs for doc_id values during freeze

lcawl:
This PR removes the `{branch}` variable from the URLs in `table.csv` since the main/master URLs will stop working in the near future, whereas "current" will continue to work until we get the documentation migrated to its new home.


marciw (approved):
LGTM

PR #3688, opened 2025-02-03
Update inference.update.json with the correct verb

davidkyle:
The Inference Update API is a `PUT` API not `POST`

The mistake is probably due to the same error in the Elasticsearch repo which has been fixed in https://github.com/elastic/elasticsearch/pull/121048


szabosteve (approved):
LGTM!

PR #3691, opened 2025-02-03
Add include_source_on_error to bulk API

pquentin:
This is essentially https://github.com/elastic/elasticsearch-specification/pull/3681 but just for bulk for now. I added the `_json_spec` change and ran `make contrib` to ensure that this was not adding a new validation error.

This was added to Elasticsearch in https://github.com/elastic/elasticsearch/pull/120725.


mosche:
why is this not automated @pquentin ? We can't expect to ever keep this up to date in so many different places

mosche:
I don't think is is a feasible path forward

pquentin:
We discussed this offline - the long-term plan is that the specification is the source of truth, which means that, yes, for now, we need to port changes from the rest-api-spec.

flobernd (approved):

mosche (commented):
Sorry @pquentin, just realised my review was still pending 🤦 
Actually did index, create and update locally, but honestly was rather amazed about manually maintaining even stuff like specLocation 😮

pquentin:
I only manually updated the specification/ directory, the rest was generated by `make contrib`. But if you don't we have automation that will fix it, no worries!

PR #3692, opened 2025-02-03
[DOCS] Convert final examples from JSON to YAML

lcawl:
This PR converts more examples from JSON to YAML to address linting errors related to invalid fields in those files.
It also comments out some of those lingering fields from existing example files.


kilfoyle (approved):

kilfoyle (approved):
LGTM! 🐙

PR #3705, opened 2025-02-05
Simplifying ingest doc sim response

l-trotta:
@masseyke Since it's only used in a response then the enum `field` can simply be a string, since user won't have to set it. (also having an enum as dictionary key breaks the java generator). Followup of #3640


masseyke (approved):
Oops, sorry. LGTM.

PR #3721, opened 2025-02-05
[Connector API][Doc] Add info about required privileges

jedrazb:
Add info about required privileges to call Connector API endpoints


leemthompo (approved):
LGTM with a wording change, tagged Lisa for review just because not sure if any other files need to be edited.

PR #3725, opened 2025-02-06
Bump nvm version

pquentin:
It may seem like a small change, but 0.39.1 was released more than three years ago.


flobernd:
We could probably as well bump the node version in nvmrc to at least 22.

PR #3727, opened 2025-02-06
Fix server default for include_source_on_error

mosche:
Fix server default for include_source_on_error


pquentin (approved):
Thanks! LGTM.

PR #3732, opened 2025-02-06
Add failure_store enum to bulk responses

Anaethelion:
As per:
* https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/action/bulk/BulkItemResponse.java
* https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/action/bulk/IndexDocFailureStoreStatus.java

### DIFF
--- a/specification/_global/bulk/types.ts
+++ b/specification/_global/bulk/types.ts
@@ -48,6 +48,7 @@ export class ResponseItem {
    * The HTTP status code returned for the operation.
    */
   status: integer
+  failure_store?: FailureStoreStatus
   /**
    * Additional information about the failed operation.
    * The property is returned only for failed operations.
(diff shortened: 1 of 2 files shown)


specification/_global/bulk/types.ts
```diff
@@ -82,6 +83,13 @@ export class ResponseItem {
   get?: InlineGet<Dictionary<string, UserDefinedValue>>
 }
 
+export enum FailureStoreStatus {
+  not_applicable_or_unknown,
```
l-trotta, line 87: seems like the server will never send `not_applicable_or_unknown`: https://github.com/elastic/elasticsearch/blob/85acf41d2848203cd5320f6b35cc5e2dcf2cf0bf/server/src/main/java/org/elasticsearch/action/bulk/IndexDocFailureStoreStatus.java#L105

Anaethelion: I included because its mentioned in the exception [just below](https://github.com/elastic/elasticsearch/blob/85acf41d2848203cd5320f6b35cc5e2dcf2cf0bf/server/src/main/java/org/elasticsearch/action/bulk/IndexDocFailureStoreStatus.java#L116) in case we wanted to model that in the future.

I'm fine removing it for now. wdyt?

l-trotta: got it, let's leave it then

l-trotta (approved):

PR #3781, opened 2025-02-18
[Inference API] Adds chunking_settings to PUT inference API

szabosteve:
## Overview

This PR adds the [`chunking_settings`](https://www.elastic.co/guide/en/elasticsearch/reference/8.17/infer-service-anthropic.html#infer-service-anthropic-api-request-body) parameter to the inference API spec. The parameter was introduced in 8.16.


pquentin (approved):
I have not checked that every property is correct, but the way the change was made is correct.

jonathan-buttner (approved):

PR #3801, opened 2025-02-20
Ignore errors from private APIs

pquentin:


l-trotta (approved):

swallez:
@pquentin we should do it differently, and test these private namespaces when we check that requests and response exist.

That way, if/when these requests and responses are added, they will go through all other checks instead of being silently skipped.

PR #3810, opened 2025-02-20
Revert "Uppercase enum source mode"

pquentin:
Reverts elastic/elasticsearch-specification#3676

We decided offline to keep all enums as lowercase, except multi word enums that can use camelCase. The clients need to accept any case for enums.


l-trotta:
I'll update this as soon as I have working case insensitive deserialization in java

l-trotta (approved):
java client ready. should we go over all the instances of non lowercase enums in the spec?

PR #3811, opened 2025-02-20
Add missing index parameter to _cat/recovery

pquentin:
While it is redundant with the path part with the same name, it is indeed accepted.


l-trotta (approved):
had to double check first, it's legit :')

l-trotta:
^ manually backporting these

PR #3841, opened 2025-02-25
100% validation security APIs

l-trotta:
There were just 3 endpoint with errors:

- security.get_role: `description` was missing. [server code](https://github.com/elastic/elasticsearch/blob/ec82c24a8722e7c4b50392567bafef7d96051849/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/RoleDescriptor.java#L80C26-L80C37)
- security.get_user_privileges: we were missing two optional "remote" fields. [server code](https://github.com/elastic/elasticsearch/blob/ec82c24a8722e7c4b50392567bafef7d96051849/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/GetUserPrivilegesResponse.java#L45) for `remote_indices` and `remote_cluster`
- security.update_settings: `auto_expand_replicas` is one of that index properties that allows to use `null` to reset the value to the default, so I thought we could use the new WithNullValue. please don't ask for the server proof :')

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -32429,6 +32429,12 @@
                         "type": "string"
                       }
                     },
+                    "remote_cluster": {
+                      "type": "array",
+                      "items": {
+                        "$ref": "#/components/schemas/security._types:RemoteClusterPrivileges"
+                      }
+                    },
                     "global": {
                       "type": "array",
                       "items": {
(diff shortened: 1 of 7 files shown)


flobernd (approved):
LGTM!

Linter seems to be unhappy.

l-trotta:
just noticed 2 more api errors, will fix those too before merging

PR #3847, opened 2025-02-26
fix 'runtime' property type in DynamicTemplate class

fbaligand:
Currently, in DynamicTemplate class, the "runtime" property has "Property" type.
This PR aims to fix that by referencing "RuntimeField" type.


fbaligand:
Thanks for the merge @pquentin !

pquentin:
Merci pour la contribution !

This will be released in version 8.18 and above of the Elasticsearch clients.

fbaligand:
Super!

PR #3877, opened 2025-03-05
Support subobjects: auto

pquentin:
This fixes validation for `cluster.put_component_template`.

### DIFF
--- a/specification/_types/mapping/TypeMapping.ts
+++ b/specification/_types/mapping/TypeMapping.ts
@@ -48,7 +48,7 @@ export class TypeMapping {
   _source?: SourceField
   runtime?: Dictionary<string, RuntimeField>
   enabled?: boolean
-  subobjects?: boolean
+  subobjects?: Subobjects
   /**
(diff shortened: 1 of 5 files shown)


specification/_types/mapping/TypeMapping.ts
```diff
@@ -59,3 +59,16 @@ export class TypeMapping {
 export class DataStreamTimestamp {
   enabled: boolean
 }
+
+/**
+ * @es_quirk This enum is a boolean that evolved into a tri-state enum. True and False have
```
flobernd, line 64: Q: Out of curiosity, does ES accept `true` and `false` as string too? This seems to be the case for the other enums evolved from a boolean.

flobernd (approved):
LGTM!

specification/_types/mapping/TypeMapping.ts
l-trotta: [it does](https://www.elastic.co/guide/en/elasticsearch/reference/current/boolean.html)

PR #3901, opened 2025-03-06
Add search_inference_id property

sergey-tihon:
Fixes/unblock: https://github.com/elastic/elasticsearch-net/issues/8467

Add property mentioned in docs https://www.elastic.co/guide/en/elasticsearch/reference/current/semantic-text.html#semantic-text-params but missing in the specification.

Closes https://github.com/elastic/elasticsearch-specification/pull/3923

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -64690,8 +64690,7 @@
           }
         },
         "required": [
-          "type",
-          "inference_id"
+          "type"
         ]
       },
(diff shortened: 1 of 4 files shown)


Mikep86 (changes requested):
We need to break this into two separate PRs. The `search_inference_id` was added in 8.16 (see #113051), but `inference_id` was made optional in 8.17.

Mikep86 (approved):
We discussed internally and we can merge this PR and make a separate one only for `search_inference_id` for 8.16

PR #3909, opened 2025-03-07
Fix `dynamic_templates` type

flobernd:
As titled. Thanks for checking the server code @l-trotta .


l-trotta (approved):
looks good! not backporting to 8.17 since it's a breaking change for java

PR #3925, opened 2025-03-10
Fix `queue.max_bytes` field in logstash pipeline settings spec

kaiwenleee:
Logstash `PipelineSettings` is using `queue.max_bytes` [since at least 6.0](https://www.elastic.co/guide/en/logstash/6.0/logstash-settings-file.html) rather than `queue.max_bytes.number` & `queue.max_bytes.units`. This is also shown [in the example](https://github.com/elastic/elasticsearch-specification/blob/ae86eb8d9eb3459b3bafa408e66c07471d29e78f/specification/logstash/put_pipeline/examples/request/LogstashPutPipelineRequestExample1.yaml#L18).
Updating the spec so that elasticsearch clients can work correctly.

ref: https://github.com/elastic/elasticsearch/pull/92651


kaiwenleee:
Hello, can any maintainer take a look at the PR, or is there something I'm missing? I did look at the contribution instruction but don't see a clear gap, except that it's not possible to do the validation part since [`clients-flight-recorder`](https://github.com/elastic/clients-flight-recorder) seems to be closed source now.

l-trotta:
hey @kaiwenleee, thanks for contributing! validation is not starting since it doesn't work for forks, I'll open a mirror PR just to start validation.

l-trotta:
so, turns out we're missing tests for this API ^^" but it looks correct, also [this server test](https://github.com/elastic/elasticsearch/blob/ae33eaabdba054a4182e2ae7b0117ace06d79d9d/x-pack/plugin/logstash/src/javaRestTest/java/org/elasticsearch/xpack/test/rest/LogstashSystemIndexIT.java#L188) confirms it. LGTM, thanks again!

l-trotta (approved):

PR #3930, opened 2025-03-11
Fix overlay for inference stream

lcawl:
This PR fixes the following error that occurs when `make overlay-docs` runs:

```
 Error: Error opening file "elasticsearch-specification/specification/inference/stream_inference/examples/request/StreamInferenceRequestExample1.yaml" 
 ›   ENOENT: no such file or directory...
```

I believe this is due to files that moved in https://github.com/elastic/elasticsearch-specification/pull/3545
This is the type of overlay fragility that will be fixed by https://github.com/elastic/elasticsearch-specification/pull/3737


kilfoyle (approved):
LGTM! 🥕

PR #3935, opened 2025-03-12
Use path alias in imports

Anaethelion:
Level the imports declaration with the rest of the spec.
Makes the type tree compatible with the current `typescript-go` state for testing purposes.


pquentin (approved):
Thanks! I tried to remove `suppressImplicitAnyIndexErrors` in the past as it's flagged by Zed but it was generating errors. Good riddance :)

PR #3940, opened 2025-03-12
Adds EIS POST inference details

szabosteve:
## Overview

This PR adds the EIS POST inference API use case.


specification/inference/post_eis/PostEisRequest.ts
```diff
@@ -0,0 +1,58 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import type { Request as RequestChatCompletion } from '../chat_completion_unified/UnifiedRequest'
+import { Id } from '@_types/common'
+
+
+export type OmittedChatCompletion = Omit<RequestChatCompletion, 'urls' | 'path_parts'>;
```
jonathan-buttner, line 24: @swallez @pquentin we're trying to override the `urls` and `path_parts`, we're getting errors when trying to run `make generate`. Is this a valid solution?

Technically we don't really have to do this, I'm just curious if it's possible. `UnifiedRequest` will likely be identical to the `Request` interface we're defining in this file. We are trying to rename the `inference_id` path part to `eis_inference_id` just to make the documentation a little clearer.

jonathan-buttner: An option that looks like it's working is:

```
import type { Request as RequestChatCompletion } from '../chat_completion_unified/UnifiedRequest'

export interface Request extends RequestChatCompletion {}
```

specification/_json_spec/inference.post.eis_chat_completion.json
```diff
@@ -0,0 +1,35 @@
...
+      "accept": ["application/json"],
+      "content_type": ["application/json"]
+    },
+    "url": {
+      "paths": [
+        {
+          "path": "/_inference/chat_completion/{eis_inference_id}/_stream",
+          "methods": ["POST"],
+          "parts": {
+            "task_type": {
```
jonathan-buttner, line 19: I think we can remove the `task_type` object since we've defined `chat_completion` in the URL.

specification/inference/post_eis_chat_completion/PostEisChatCompletionRequest.ts
```diff
@@ -0,0 +1,58 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import type { Request as RequestChatCompletion } from '../chat_completion_unified/UnifiedRequest'
+import { Id } from '@_types/common'
+
+
+export type OmittedChatCompletion = Omit<RequestChatCompletion, 'urls' | 'path_parts'>;
```
jonathan-buttner, line 24: Doesn't look like this is going to work, so either copying the whole `UnifiedRequest` or creating an interface that contains the fields that are shared between `UnifiedRequest` and this file excluding `urls` and `path_parts`.

l-trotta:
Made some changes:

- changed api def from inference.post.eis_chat_completion to inference.post_eis_chat_completion
- created RequestChatCompletionBase to be extended, made Eis and UnifiedRequeste extend it
- removed the path parameter {task_type} since this endpoint will be specific for eis chat completion

output/schema/validation-errors.json
```diff
@@ -56,6 +56,12 @@
       ],
       "response": []
     },
+    "inference.post_eis_chat_completion": {
+      "request": [
+        "Request: url path '/_inference/{task_type}/{eis_inference_id}/_stream' not found in the json spec"
```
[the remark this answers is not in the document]
l-trotta: forgot to regenerate output after path fix ^^" also needed to fix the json spec accordingly, pushing a fix now

jonathan-buttner (approved):

PR #3926, opened 2025-03-10
Add the index mode to the output of get-data-stream API.

dakrone:
Relates to https://github.com/elastic/elasticsearch/pull/122486

Closes #3958

PR #3959, opened 2025-03-14
Move contributing instructions to CONTRIBUTING.md

pquentin:
As suggested by @dakrone in https://github.com/elastic/elasticsearch-specification/pull/3926#issuecomment-2711705345. The information was mostly in the pull request template, but that's arguably a bit too late.


l-trotta (approved):

pquentin:
Thank you for the suggestion. Changing `make` to `make setup && make install` could be interesting, but we probably need to make it fast to detect an existing install. For now, I have decided to merge what I had since it's a step in the right direction.

PR #3975, opened 2025-03-17
Deprecate local param in indices.get_mapping

pquentin:
This was [deprecated in 7.8](https://github.com/elastic/elasticsearch/commit/0b041cccd894f01d723fb2979f70c1cf279700a6), but we don't copy this information from the JSON spec, which is a bug.

### DIFF
--- a/specification/indices/get_mapping/IndicesGetMappingRequest.ts
+++ b/specification/indices/get_mapping/IndicesGetMappingRequest.ts
@@ -71,6 +71,7 @@ export interface Request extends RequestBase {
     ignore_unavailable?: boolean
     /**
      * If `true`, the request retrieves information from the local node only.
+     * @deprecated 7.8.0 This parameter is a no-op and field mappings are always retrieved locally.
      * @server_default false
      */
     local?: boolean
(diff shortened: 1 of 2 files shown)


flobernd (approved):

specification/indices/get_mapping/IndicesGetMappingRequest.ts
```diff
@@ -71,6 +71,7 @@ export interface Request extends RequestBase {
     ignore_unavailable?: boolean
     /**
      * If `true`, the request retrieves information from the local node only.
+     * @deprecated 7.8.0 This parameter is a no-op and field mappings are always retrieved locally.
```
nielsbauman, line 74: Unfortunately, the PR you linked that deprecated this parameter in `rest-api-spec` was incorrect - see https://github.com/elastic/elasticsearch/pull/122922 for more context. The `local` param was deprecated in https://github.com/elastic/elasticsearch/pull/55014 for the GET _field_ mapping request, not the GET mapping request. I see the `local` param is not present in https://github.com/elastic/elasticsearch-specification/blob/31b8bb299cf705d9c963a2dfb736c603f47c2425/specification/indices/get_field_mapping/IndicesGetFieldMappingRequest.ts, so we don't have to do anything there. But `field` should be removed from this line and `7.8.0` should be either removed or updated to `9.1.0` - I implemented the actual deprecation in https://github.com/elastic/elasticsearch/pull/122921/files#diff-759c8fb59bd55bbebf840f520608fabbbc549918cd9121ed754cb629ce5c86c1.

PR #4056, opened 2025-03-24
Fix ml.get_trained_model_stats response

pquentin:
We were using a [RoutingInfo](https://github.com/elastic/elasticsearch/blob/eceb178cd73dce7f78021d6859b7ac74ea2d7431/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/RoutingInfo.java#L25-L30) when it's really a [RoutingStateAndReason](https://github.com/elastic/elasticsearch/blob/eceb178cd73dce7f78021d6859b7ac74ea2d7431/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/RoutingStateAndReason.java#L23-L26).

### DIFF
--- a/specification/ml/_types/TrainedModel.ts
+++ b/specification/ml/_types/TrainedModel.ts
@@ -189,7 +189,7 @@ export class TrainedModelDeploymentNodesStats {
   /** The number of inference requests that were not processed because the queue was full. */
   rejection_execution_count?: integer
   /** The current routing state and reason for the current routing state for this allocation. */
-  routing_state: TrainedModelAssignmentRoutingTable
+  routing_state: TrainedModelAssignmentRoutingStateAndReason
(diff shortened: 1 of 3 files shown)


specification/ml/_types/TrainedModel.ts
```diff
@@ -189,7 +189,7 @@ export class TrainedModelDeploymentNodesStats {
   /** The number of inference requests that were not processed because the queue was full. */
   rejection_execution_count?: integer
```
l-trotta, line 190: typo! should be `rejected_execution_count`

l-trotta:
```suggestion
  rejected_execution_count?: integer
```

l-trotta (approved):

PR #4064, opened 2025-03-24
OpenAI inference request refactor + doc fix

l-trotta:
Avoiding exporting classes from Request files, also `chunking_settings` is only available for text embedding (not sure if that's the right way to write it, feel free to rewrite!)


pquentin:
I fixed the immediate conflict, but I suspect this refactor should apply to Azure OpenAI too: https://github.com/elastic/elasticsearch-specification/blob/main/specification/inference/put_azureopenai/PutAzureOpenAiRequest.ts

l-trotta:
@pquentin found out after creating this that we want to wait to actually fix the docs, right now we want them to match the original ones, so this will be merged in the following releases

Issue #4081, opened 2025-03-24
Support for multiple media types


lcawl:
## 🚀 Feature Proposal

Add a method to specify the supported medias type for each request and response body.
If multiple media types are supported, we'll need a method (such as folder or file names?) to indicate which type each example pertains to.

## Motivation

In OpenAPI documents, you can have request and response bodies that support various media types per https://spec.openapis.org/oas/v3.0.3#media-types
For example, the cat API (e.g [cat aliases](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-alias.html) returns `text/plain` responses by default and you can use the `format` query parameter to get back JSON, YAML, etc.

It does not seem like we currently have a way to represent these media types in the current specifications.
I am converting the `text/plain` examples to `json` to work with the existing system, but it would be good to be able to specify the media type explicitly.

## Example

When we were manually curating the examples, it looked like this (i.e. the cat aliases API responses could have both JSON and text formats:

```
    cat.aliases#200:
      description: ''
      content:
        application/json:
          schema:
            type: array
            items:
              "$ref": "#/components/schemas/cat.aliases:AliasesRecord"
        text/plain:
          schema:
            type: string
          examples:
            catAliasesResponseExample1:
              "$ref": "#/components/examples/CatAliasesResponseExample1"
```

PR #4082, opened 2025-03-25
[DOCS] Convert cat API examples to json format

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/issues/4081

This PR updates the cat API examples to use the `formt=json` query parameter, since they were using the default text string format, which wasn't being correctly represented in the OpenAPI output.


kilfoyle (approved):
LGTM! 🍾

PR #4095, opened 2025-03-25
Fixing inference trained models

l-trotta:
Reported in https://github.com/elastic/elasticsearch-java/issues/964, adding various missing fields and classes around TrainedModels. Also fixing a bug with InferenceChunkingSettings where it was wrongly extending another class.


mariuszpala:
I see this issue in 8.17.4 SDK version, in which version precisely will it be fixed?

PR #4142, opened 2025-03-27
Add max_concurrent_shard_requests to open PIT API

pquentin:
Kibana ran into this missing parameter: https://github.com/elastic/kibana/pull/213375/files#diff-a742bab88e72c2e7784c7fdfba52ea4c1f45c0939854beaf400a5fd191a92b31R56

It was added https://github.com/elastic/elasticsearch/pull/96782 and is defined in https://github.com/elastic/elasticsearch/blob/36c14bf3a56f85177f5ce60b2f839b6b35236bdc/server/src/main/java/org/elasticsearch/action/search/RestOpenPointInTimeAction.java#L52-L58 

I can leave out the long -> integer changes if they are going to break statically typed clients. Or not backport all the way to 8.18.

### DIFF
--- a/specification/_global/msearch/MultiSearchRequest.ts
+++ b/specification/_global/msearch/MultiSearchRequest.ts
@@ -19,7 +19,7 @@
 
 import { RequestBase } from '@_types/Base'
 import { ExpandWildcards, Indices, Routing, SearchType } from '@_types/common'
-import { long } from '@_types/Numeric'
+import { integer, long } from '@_types/Numeric'
 import { RequestItem } from './types'
 
 /**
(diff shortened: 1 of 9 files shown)


specification/_global/msearch/MultiSearchRequest.ts
```diff
@@ -101,13 +101,14 @@ export interface Request extends RequestBase {
...
+     * Defaults to `max(1, (# of data nodes * min(search thread pool size, 10)))`.
      */
-    max_concurrent_searches?: long
+    max_concurrent_searches?: integer
     /**
      * Maximum number of concurrent shard requests that each sub-search request executes per node.
      * @server_default 5
      */
-    max_concurrent_shard_requests?: long
+    max_concurrent_shard_requests?: integer
```
pquentin, line 111: I noticed we used `long` by mistake here, while it's an int: https://github.com/elastic/elasticsearch/blob/36c14bf3a56f85177f5ce60b2f839b6b35236bdc/server/src/main/java/org/elasticsearch/action/search/SearchRequest.java#L88.

specification/_global/msearch/MultiSearchRequest.ts
```diff
@@ -101,13 +101,14 @@ export interface Request extends RequestBase {
     include_named_queries_score?: boolean
     /**
      * Maximum number of concurrent searches the multi search API can execute.
+     * Defaults to `max(1, (# of data nodes * min(search thread pool size, 10)))`.
      */
-    max_concurrent_searches?: long
+    max_concurrent_searches?: integer
```
pquentin, line 106: This is also an integer: https://github.com/elastic/elasticsearch/blob/36c14bf3a56f85177f5ce60b2f839b6b35236bdc/server/src/main/java/org/elasticsearch/rest/action/search/RestMultiSearchAction.java#L115-L117

specification/_global/msearch/MultiSearchRequest.ts
```diff
@@ -101,13 +101,14 @@ export interface Request extends RequestBase {
     include_named_queries_score?: boolean
     /**
      * Maximum number of concurrent searches the multi search API can execute.
+     * Defaults to `max(1, (# of data nodes * min(search thread pool size, 10)))`.
```
pquentin, line 104: Can't use `@server_default` as it expects a number

flobernd (approved):
LGTM!

While `long` -> `int` is technically a breaking change, nobody would notice in the end.

pquentin:
After discussing this offline, I extracted the breaking changes to #4171. Please take another look.

l-trotta (approved):

PR #4151, opened 2025-03-27
Add missing codegen names

flobernd:
As titled.


flobernd:
@pquentin Thanks for the quick review.

I used `data` mainly when there is a `ArrayBuffer` response (for ESQL we had used `data` before, so I adapted to this). `result` is the name I used in all places where the response body contains a rather unspecific type like e.g. `AcknowledgeResponseBase`.

l-trotta:
@flobernd honestly I'd rather keep this for 9, 31 breaking changes is too many

PR #4153, opened 2025-03-27
Inference refactor

l-trotta:
Moving all exported classes in the CommonTypes file, avoiding enum name duplication, adding @codegen_names to responses


flobernd (approved):
LGTM!

As discussed, we should replace the weird 1-element `ServiceType` enums with string `LiteralValue` properties when all generators support this.

PR #4182, opened 2025-04-01
adding rescore_vector to dense vector mapping definition options

benwtrent:
This goes into the 9.1 specification.

New `rescore_vector` API for dense_vector field mapping under index_options.


l-trotta (approved):
LGTM, thanks!

PR #4184, opened 2025-04-02
OpenAPI: use ':' instead of '#' to separate parts of a component path

### LINKED ISSUE #4183: OpenAPI export should not use # characters in $ref
This is causing issues in tools like Apigee and Visual Studio Code OpenAPI SwaggerUI plugin

Image

This behavior can also be reproduced in https://editor-next.swagger.io/

swallez:
Some tools do not accept a `#` in component path identifiers. This PR changes it to a `:` which is already used to separate namespace and name in type names.

Fixes #4183

Docs team: does this have any impact on documentation generation?


lcawl (approved):
I verified that none of the overlays are affected by this change and the docs output LGTM

PR #4189, opened 2025-04-02
Add ESQL result type definitions

swallez:
Adds the API spec for the JSON output of ES|QL query (other formats cannot be represented here).


l-trotta:
@flobernd @JoshMock @Anaethelion any issue if we backport this to 8.19? I think most of us are still using the old binary result

flobernd:
Fine for me @l-trotta !

PR #4195, opened 2025-04-03
Add new write load metrics to `IndexingStats`

PeteGillinElastic:
This adds the fields added in
https://github.com/elastic/elasticsearch/pull/125521 and https://github.com/elastic/elasticsearch/pull/124652 to the spec.

### DIFF
--- a/specification/_types/Stats.ts
+++ b/specification/_types/Stats.ts
@@ -159,6 +159,8 @@ export class IndexingStats {
   index_failed: long
   types?: Dictionary<string, IndexingStats>
   write_load?: double
+  recent_write_load?: double
+  peak_write_load?: double
 }
 
 export class MergesStats {


dakrone (approved):
LGTM, thanks Pete!

dakrone (changes requested):
Actually, I believe you need to `make generate` the output changes also

dakrone:
@pquentin would it be possible to have something like `make generate && git add …` as part of a git precommit hook?

PeteGillinElastic:
> @pquentin would it be possible to have something like `make generate && git add …` as part of a git precommit hook?

I believe that there's a bot which periodically comes along and commits any needed output changes. But it's not part of the same PR. (I've done `make contrib` and pushed it anyway. Which I think does slightly more than `make generate` does.)

dakrone (approved):
LGTM still :)

pquentin:
@dakrone As @PeteGillinElastic mentioned, it's OK not to run those commands, as CI will notice if there's an issue and commit the output/ files. A pre-commit hook is an interesting idea, are you suggesting that I add that script to the repository and explain in CONTRIBUTING.md how to install it?

dakrone:
Ah okay, I hadn't realized that CI would do the commit for it. As for the pre-commit, what you described is what I was suggesting, but I don't think it's necessary unless we have to have the `make` run prior to merging (and we don't, since CI does it.)

PR #4199, opened 2025-04-05
OpenAPI: update path separator, output examples as json

swallez:
Follow-up to #4184 - using the `:` was wrong, as [`$ref` paths must be RFC-3986 compliant](https://spec.openapis.org/oas/v3.1.1.html#reference-object), and `:` is a reserved char that must be escaped. Furthermore, identifiers in `components` are [further restricted](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-5) and must match ^[a-zA-Z0-9\.\-_]+$`

This PR also outputs example values as JSON when they can be successfully parsed instead of just outputting them as strings, as [required by the spec](https://spec.openapis.org/oas/v3.1.1.html#example-object).


swallez:
@lcawl I added you as reviewer: the change in the path separator shouldn't be an issue, but we have to validate that examples as JSON objects (instead of plain strings) doesn't break the docs.

lcawl (approved):
LGTM

PR #4260, opened 2025-04-10
ESQL: Add query monitoring API

GalLalouche:
Follow up to https://github.com/elastic/elasticsearch/pull/124832, adding endpoints for list and get query APIs.


specification/esql/get_query/GetQueryRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { RequestBase } from '@_types/Base'
+import { Id } from '@_types/common'
+
+/**
+ * Get a specific running ES|QL query information.
+ * Returns an object extended information about a running ES|QL query.
+ *
+ * @rest_spec_name esql.get_query
+ * @cluster_privileges monitor_esql
```
pquentin, line 28: Can we please also add `@availability` annotations? Here's an example:

```
 * @availability stack since=8.11.0 stability=stable visibility=public
 * @availability serverless stability=stable visibility=public
```

Documentation here: https://github.com/elastic/elasticsearch-specification/blob/55d1444f19ce33e5a7a318e51ccc638a7901959c/docs/modeling-guide.md#availability. If it's not available in Serverless, you can leave it out.

specification/esql/list_queries/ListQueriesRequest.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import { RequestBase } from '@_types/Base'
+
+/**
+ * Get running ES|QL queries information.
+ * Returns an object containing IDs and other information about the running ES|QL queries.
+
+ * @rest_spec_name esql.list_queries
+ * @cluster_privileges monitor_esql
```
pquentin, line 27: The comment about annotations applies here too.

pquentin (commented):
Excellent work, thank you!

specification/esql/list_queries/ListQueriesRequest.ts
GalLalouche: Done.

specification/esql/get_query/GetQueryRequest.ts
GalLalouche: Done.

specification/esql/list_queries/ListQueriesResponse.ts
```diff
@@ -0,0 +1,36 @@
...
+  start_time_millis: long
+  running_time_nanos: long
+  query: string
+}
+
+export class Response {
+  body: {
+    queries: Dictionary<TaskId, Body>
+  }
+}
```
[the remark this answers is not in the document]
flobernd: No need for `codegen_name` here since this looks like a `PropertiesBody`.

It would only be required if the body looked like this:

```ts
export class Response {
  body: Dictionary<TaskId, Body>
}
```

pquentin (approved):
Thanks! LGTM.

PR #4261, opened 2025-04-10
Add support for chunking_settings in semantic_text fields 

kderusso:
Updates the specification to include `chunking_settings` for `semantic_text` fields which were added in https://github.com/elastic/elasticsearch/pull/121041

### DIFF
--- a/specification/_types/mapping/ChunkingSettings.ts
+++ b/specification/_types/mapping/ChunkingSettings.ts
@@ -1,3 +1,22 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 import { integer } from "@_types/Numeric"
 
 export class ChunkingSettings {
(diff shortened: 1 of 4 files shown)


specification/_types/mapping/ChunkingSettings.ts
```diff
@@ -0,0 +1,49 @@
...
+   * defined as more than half of the max_chunk_size. Required for `word`
+   * type chunking settings.
+   */
+  overlap?: integer
+
+  /**
+   * The number of overlapping sentences allowed in chunks. Required for `sentence` type chunking settings.
+   */
+  sentence_overlap?: 0 | 1
+}
```
pquentin, line 44: We have a similar class for the Inference API; should they be the same, or are they also distinct in Elasticsearch?

https://github.com/elastic/elasticsearch-specification/blob/1447a82cfebb60bce3530401a70f86cc50babc1d/specification/inference/_types/Services.ts#L60-L89

pquentin (commented):
Thank you for this! Before we review in more depth, I'm wondering if this should be consolidated with `InferenceChunkingSettings` or not.

specification/_types/mapping/ChunkingSettings.ts
kderusso: Oh nice catch! 

The only concern that I have with combining them, is that we currently require all required settings to be sent in, with no defaults applied. So if you select `word` based chunking and don't send in both `max_chunk_size` and `overlap` you will get an error. This made sense when we wrote it because you're already putting in custom overrides, though I suppose we could revisit this in the future. 

Based on this specification, though, `max_chunk_size` and `overlap` would both be optional due to default values, and using this would break. 

WDYT the best course of action is? Maybe put in a comment why they're separate and keep them separate for now?

pquentin: Right, let's keep them separate. To ensure they keep using the same property names, we can use `OverloadOf`: https://github.com/elastic/elasticsearch-specification/blob/main/docs/behaviors.md#overloadof. With that, we don't need a comment!

kderusso: TIL, thanks!

pquentin: Thanks for making the changes. However, were the parameters different? You used to have `type` and it's now `strategy`.

kderusso: That would have been a bug 🤦‍♀️ Here's the PR with the example, strategy is correct: https://github.com/elastic/elasticsearch/pull/121041

pquentin (approved):
Thanks! LGTM.

Issue #4251, opened 2025-04-09
Several missing token filter types


JoshMock:
An issue [was opened on the JS client repo](https://github.com/elastic/elasticsearch-js/issues/2726) noting the missing `cjk_bigram` token filter type. I went through [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-tokenfilters.html) and identified several missing token filter types:

- `apostrophe`
- `arabic_normalization`
- `cjk_bigram`
- `cjk_width`
- `classic`
- `decimal_digit`
- `flatten_graph`
- `german_normalization`
- `hindi_normalization`
- `indic_normalization`
- `keyword_repeat`
- `min_hash`
- `persian_normalization`
- `phonetic`
- `scandinavian_folding`
- `scandinavian_normalization`
- `serbian_normalization `
- `sorani_normalization`

Assigning to myself, as I'd like to at least solve for the one reported on my repo, but it's a good opportunity to scan through the spec and knock out several fixes at once.

PR #4291, opened 2025-04-17
Several improvements to token filter types

JoshMock:
An [issue](https://github.com/elastic/elasticsearch-js/issues/2726) on the JS repo was opened about a missing `cjk_bigram` token filter type. I did an audit of the token filter docs and found a few missing types, so I:

- Added the missing types
- Added some missing values to enums
- Moved a couple types into their relevant plugin files
- Created a few base types for token filters that share most of the same parameters
- Added docstrings to as many parameters as I could find
- Removed a few parameters not mentioned in the docs

Fixes https://github.com/elastic/elasticsearch-specification/issues/4251
Fixes https://github.com/elastic/elasticsearch-js/issues/2726


l-trotta:
@JoshMock I'm having some doubts about this one, I think while it's good to have the full list of stopwords, some of the specific tokenizers only accept part of it? For example, [this is the code](https://github.com/elastic/elasticsearch/blob/45d321d91b905816e3ce289b2328883848b7559f/plugins/analysis-kuromoji/src/main/java/org/elasticsearch/plugin/analysis/kuromoji/JapaneseStopTokenFilterFactory.java#L29) for the JapaneseStopTokenFilter, which only supports `_japanese_`

JoshMock:
@l-trotta Good catch. I updated the stopwords types in 5fc7dbe3fffa67ad0e935962c13cb46298723792.

l-trotta (approved):

specification/_types/analysis/analyzers.ts
```diff
@@ -56,7 +56,7 @@ export class FingerprintAnalyzer {
    *
    * @server_default _none_
    */
-  stopwords?: StopWords
+  stopwords?: StopWordLanguage | string[]
```
flobernd, line 59: This will cause inconveniences for users in statically typed languages.

Could we please continue using the `StopWords` type here?

`StopWords` is currently defined as:

```ts
export type StopWords = string | string[]
```

Changing that to:

```ts
export type StopWords = StopWordLanguage | string[]
```

while keeping this file "as is", should do the trick.

What do you think?

flobernd (commented):
Thank you @JoshMock 🙂 Just a minor comment.

specification/_types/analysis/analyzers.ts
JoshMock: @flobernd I think I got what you mean. Made a change in db8c13039816597ff87429c8aafe25f25384e890. Let me know if that looks better to you.

flobernd: Perfect now! Thank you @JoshMock

PR #4292, opened 2025-04-17
Use correct status values for simulated ingest pipeline processor results

JoshMock:
Fixes <https://github.com/elastic/elasticsearch-js/issues/2647>


l-trotta (approved):
[server side proof](https://github.com/elastic/elasticsearch/blob/45d321d91b905816e3ce289b2328883848b7559f/server/src/main/java/org/elasticsearch/action/ingest/SimulateProcessorResult.java#L39), LGTM

Issue #4288, opened 2025-04-17
Rename to ml stats, not usage


sophiec20:
<img width="501" alt="Image" src="https://github.com/user-attachments/assets/cda0deb3-fb00-44ea-a11f-da573b0f2841" />

`GET /_ml/anomaly_detectors/_stats` and `GET /_ml/datafeeds/_stats` have been renamed to `Get datafeeds usage info` and `Get anomaly detection jobs usage info`.

We use the language to say "get stats". This is common SDH language.
The response contains stats. 
The legacy docs called them "Get statistics".

It should be called "stats" or "statistics". It's confusing to relabel them as usage.

see https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-get-datafeed-stats

sophiec20:
cc @valeriy42

PR #4299, opened 2025-04-23
[DOCS] Renames ML get usage info APIs back to get stats

szabosteve:
## Overview

Closes https://github.com/elastic/elasticsearch-specification/issues/4288

This PR renames the GET datafeed, anomaly detection, and data frame analytics usage info APIs to GET stats, as this was the name before switching to the new system.


valeriy42 (approved):
LGTM

PR #4325, opened 2025-04-29
Phase optional lifecycle explain

l-trotta:
Reported in https://github.com/elastic/elasticsearch-java/issues/999, [server code](https://github.com/elastic/elasticsearch/blob/e222e5c636dfbd3e0ac68916c6a6cda7a191b275/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/IndexLifecycleExplainResponse.java#L91)


flobernd (approved):
LGTM!

PR #4344, opened 2025-05-06
Add cat node -h enumerated values

lcawl:
This PR adds the enumerated values for the `-h` parameter in the cat nodes API.
They are copied from https://www.elastic.co/guide/en/elasticsearch/reference/8.18/cat-nodes.html and implemented in a similar manner to the machine learning cat APIs, such as https://github.com/elastic/elasticsearch-specification/blob/main/specification/cat/ml_data_frame_analytics/CatDataFrameAnalyticsRequest.ts


szabosteve (approved):
LGTM

specification/cat/_types/CatBase.ts
```diff
@@ -555,8 +556,463 @@ export enum CatDfaColumn {
    */
   version
 }
+
+export enum CatNodeColumn {
```
pquentin, line 560: Thank you for this! Note however that this enum should be marked as non-exhaustive, as values like `heap.*` are allowed. This is done by adding `/** @non_exhaustive */` as a comment. See https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#enum for details.

pquentin: Fixed in https://github.com/elastic/elasticsearch-specification/pull/4373

PR #4371, opened 2025-05-14
Add x-codeSamples to generated OpenAPI

miguelgrinberg:
**NOTE**: this is not only a proof of concept but also a work in progress. DO NOT MERGE!

This change adds the `x-codeSamples` extension to the endpoints that have at least one example. A `Console` listing is added using the standard Kibana DevConsole syntax. Not implemented yet, but the idea is to run this DevConsole example through the request converter to generate versions of the example in all the supported languages.


compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -251,6 +251,43 @@ pub fn add_endpoint(
...
 
+        // add the x-state extension for availability
+        let mut extensions = crate::availability_as_extensions(&endpoint.availability);
+
+        // add the x-codeSamples extension
+        if !request_examples.is_empty() {
+            let mut code_samples = vec![];
+            if let Some((_, first_example)) = request_examples.clone().first() {
+                if let Some(example) = first_example.as_item() {
+                    if let Some(description) = example.description.clone() {
```
pquentin, line 262: Could we use `summary` when `description` is not set?

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -251,6 +251,43 @@ pub fn add_endpoint(
...
+                        for r in description.split('`') {
+                            if r.starts_with("GET ")
+                                || r.starts_with("POST ")
+                                || r.starts_with("PUT ")
+                                || r.starts_with("DELETE ")
+                                || r.starts_with("HEAD ")
+                            {
+                                request = Some(String::from(r));
+                                break;
+                            }
```
pquentin, line 273: Do we have examples that don't start with a method? QUERY will be a thing soon!

pquentin (commented):
Thank you! This will vastly improve the docs experience.

I'm sure the many `if let Some` can be improved, but at least the code is readable. Left two questions.

compiler-rs/clients_schema_to_openapi/src/paths.rs
miguelgrinberg: This is really a hack. I hope we don't have to do this. The problem is that the examples that were added to this repository are not clean Dev Console examples, so I'm forced to reconstruct the DevConsole syntax.

compiler-rs/clients_schema_to_openapi/src/paths.rs
miguelgrinberg: This feature does not need summaries or descriptions, all I'm doing here is parsing the description text with the hope that somewhere within it I can find the method and URL of the request. The only reason I'm using `description` is because that is how these examples are parsed.

The proper thing to do would be to expand these examples to include this information in a easier format, at which point there wouldn't be a need to use descriptions or summaries for anything.

miguelgrinberg:
I'm continuing the work in #4445 which is opened from a branch to allow validation to run in CI.

PR #4377, opened 2025-05-15
Enforce globally unique type names

pquentin:


specification/watcher/_types/Input.ts
```diff
@@ -84,9 +84,6 @@ export class HttpInputRequestDefinition {
   scheme?: ConnectionScheme
   url?: string
 }
-
-export class Input {}
-
```
pquentin, line 89: The new check caught this unused type.

flobernd (approved):

PR #4383, opened 2025-05-19
Update Inference specification for Hugging Face's completion and chat completion tasks

Jan-Kazlouski-elastic:
This PR is for changes to specification caused by https://github.com/elastic/elasticsearch/pull/127254:

**Extended Task Support:**

- [x] Added completion and chat_completion tasks to the list of supported Hugging Face tasks.

**Model Requirements for Chat Tasks:**

- [x] Updated documentation to describe specific requirements for using chat_completion and completion tasks, including model compatibility with the OpenAI API format and usage guidelines for serverless vs. dedicated endpoints.

**New Configuration Parameters:**

- [x] Introduced optional model_id field in Hugging Face service settings, applicable to completion and chat_completion tasks.

**Rate Limit Clarifications:**

- [x] Updated rate_limit documentation to clarify default behavior and guidance for tuning based on deployment specifics.

**Documentation Fixes:**

- [x] Corrected typos in existing text_embedding request examples.

**Additional actions**

- [x] Signed the CLA

- [x] Executed make contrib


specification/inference/_types/CommonTypes.ts
```diff
@@ -882,16 +882,27 @@ export class HuggingFaceServiceSettings {
...
    * The URL endpoint to use for the requests.
+   * For `completion` and `chat_completion` tasks, endpoint must be compatible with the OpenAI API format and include `v1/chat/completions`.
    */
   url: string
+  /**
+   * The name of the HuggingFace model to use for the inference task.
+   * For `completion` and `chat_completion` tasks, this field is optional but may be required for certain models — particularly when using serverless inference endpoints.
+   * For the `text_embedding` task, this field is not required and will be ignored if provided.
+   */
+  model_id?: string
```
jonathan-buttner, line 900: @lcawl @szabosteve In the previous docs system we created separate sections for the settings per task type. Is there a way to do that in the new docs system? What I mean is, it'd probably be easier to go to the hugging face doc page and click on a section to see all fields that are applicable for a particular task type rather than having to go through all fields and look at the text to see if it applies.

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
```diff
@@ -44,6 +47,15 @@ import { Id } from '@_types/common'
  * * `e5-small-v2`
  * * `multilingual-e5-base`
  * * `multilingual-e5-small`
+ *
+ * For Elastic's `chat_completion` and `completion` tasks:
```
jonathan-buttner, line 51: @lcawl @szabosteve Similar to my previous comment, would it make sense to create a put request per task type? I'm not sure if that would work or not.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@redocly/cli": "^1.34.1",
+    "@redocly/cli": "^1.34.3",
```
jonathan-buttner, line 3: Do we need this change?

package-lock.json
```diff
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
-        "@redocly/cli": "^1.34.1",
+        "@redocly/cli": "^1.34.3",
```
jonathan-buttner, line 8: Was this file supposed to change?

jonathan-buttner (changes requested):
Thanks for the PR, I left a few suggestions.

package.json
Jan-Kazlouski-elastic: This change is being made automatically by running pre commit set of tasks. This value is being incremented along with different changes committed over the time, so ignoring the change made by pre commit set of tasks would have to have a reason behind it.

package-lock.json
Jan-Kazlouski-elastic: As answered above change is being made automatically by running pre commit set of tasks. If we don't have reason for ignoring changes made by pre commit set of tasks - I'd keep it.

jonathan-buttner: Yep sounds good, thanks for explaining.

specification/inference/_types/CommonTypes.ts
```diff
@@ -882,16 +882,27 @@ export class HuggingFaceServiceSettings {
...
    * This setting helps to minimize the number of rate limit errors returned from Hugging Face.
-   * By default, the `hugging_face` service sets the number of requests allowed per minute to 3000.
+   * By default, the `hugging_face` service sets the number of requests allowed per minute to 3000 for all supported tasks.
+   * Hugging Face does not publish a universal rate limit — actual limits may vary.
+   * It is recommended to adjust this value based on the capacity and limits of your specific deployment environment.
    */
   rate_limit?: RateLimitSetting
   /**
    * The URL endpoint to use for the requests.
+   * For `completion` and `chat_completion` tasks, endpoint must be compatible with the OpenAI API format and include `v1/chat/completions`.
```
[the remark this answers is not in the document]
jonathan-buttner: What is `OpenAI mode`? I searched hugging face to see if I couldn't find it. Are you referring to the request format that the inference API requires?

specification/inference/_types/CommonTypes.ts
szabosteve: @jonathan-buttner I think there is a solution to this issue, but I need to investigate a bit. I don't want to block this PR, so please go ahead and merge if you find it good, and I'll fix these when I have a working solution.

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
```diff
@@ -29,13 +29,16 @@ import { Id } from '@_types/common'
 /**
  * Create a Hugging Face inference endpoint.
  *
- * Create an inference endpoint to perform an inference task with the `hugging_face` service.
+ * Creates an inference endpoint to perform an inference task with the `hugging_face` service.
```
szabosteve: I suggest changing it back to be consistent with the rest of the endpoint docs.
```suggestion
 * Create an inference endpoint to perform an inference task with the `hugging_face` service.
```

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
szabosteve: @jonathan-buttner It is certainly possible from a technical standpoint. However, we already have a lot of similar endpoints under the inference namespace. If we had multiple pages for an integration type (one for each task type the integration supports), the confusion might be even worse. It would also affect the left-side navigation.
I understand your point, and I think it's a valid concern. I just don't think creating multiple pages for the same integration would be the best solution to address it. I'll talk to others, too, and try to come up with a solution soon. WDYT?

jonathan-buttner: Yeah I agree having multiple pages would likely make it worse. I'm not sure what a good solution would be. Maybe if it's possible to have multiple expandable section within a single page or something. I think the underlying issue is that it's hard to determine what fields are applicable for a single task type without having to read all the text for all the fields. Maybe we could leverage examples to show the fields that should go with each task type.

specification/inference/_types/CommonTypes.ts
```diff
@@ -886,16 +886,27 @@ export class HuggingFaceServiceSettings {
...
    * This setting helps to minimize the number of rate limit errors returned from Hugging Face.
-   * By default, the `hugging_face` service sets the number of requests allowed per minute to 3000.
+   * By default, the `hugging_face` service sets the number of requests allowed per minute to 3000 for all supported tasks.
+   * Hugging Face does not publish a universal rate limit — actual limits may vary.
+   * It is recommended to adjust this value based on the capacity and limits of your specific deployment environment.
    */
   rate_limit?: RateLimitSetting
   /**
    * The URL endpoint to use for the requests.
+   * For `completion` and `chat_completion` tasks, the deployed model must be compatible with the Hugging Face Chat Completion interface (https://huggingface.co/docs/inference-providers/en/tasks/chat-completion#conversational-large-language-models-llms). OpenAI mode must be enabled, and the endpoint URL must include `v1/chat/completions`.
```
jonathan-buttner: > OpenAI mode must be enabled

What is OpenAI mode?

Can we remove that portion of the sentence?

```suggestion
   * For `completion` and `chat_completion` tasks, the deployed model must be compatible with the Hugging Face Chat Completion interface (https://huggingface.co/docs/inference-providers/en/tasks/chat-completion#conversational-large-language-models-llms). The endpoint URL must include `v1/chat/completions`.
```

specification/inference/_types/CommonTypes.ts
jonathan-buttner: Just to give an example, if a user includes `model_id` for a `text_embedding` task type request, the request will fail. If the user includes `model_id` for `completion` or `chat_completion` the request will succeed.

So the field is required depending on the model, but normally optional when using `completion` or `chat_completion`. For `text_embedding` the field is invalid and will result in an error.

specification/inference/_types/CommonTypes.ts
```diff
@@ -886,16 +886,27 @@ export class HuggingFaceServiceSettings {
...
    */
   url: string
+  /**
+   * The name of the HuggingFace model to use for the inference task.
+   * For `completion` and `chat_completion` tasks, this field is optional but may be required for certain models — particularly when using serverless inference endpoints.
+   * For the `text_embedding` task, this field is not required and will be ignored if provided.
```
jonathan-buttner:
```suggestion
   * For the `text_embedding` task, this field should not be included or the request will result fail.
```

I tried this locally:

```
{
    "error": {
        "root_cause": [
            {
                "type": "status_exception",
                "reason": "Configuration contains settings [{model_id=id}] unknown to the [hugging_face] service"
            }
        ],
        "type": "status_exception",
        "reason": "Configuration contains settings [{model_id=id}] unknown to the [hugging_face] service"
    },
    "status": 400
}
```

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
```diff
@@ -44,6 +47,15 @@ import { Id } from '@_types/common'
  * * `e5-small-v2`
  * * `multilingual-e5-base`
  * * `multilingual-e5-small`
+ *
+ * For Elastic's `chat_completion` and `completion` tasks:
+ * The selected model must support the `Text Generation` task and expose OpenAI API. HuggingFace supports both serverless and dedicated endpoints for `Text Generation`. When creating dedicated endpoint select the `Text Generation` task.
+ * After the endpoint is initialized (for dedicated) or ready (for serverless), ensure it supports the OpenAI API and includes `/v1/chat/completions` part in URL. Then, copy the full endpoint URL for use.
+ * Recommended models for `chat_completion` and `completion` tasks:
+ *
+ * * `Mistral-7B-Instruct-v0.2`
```
jonathan-buttner, line 56: @szabosteve should we include the full URL link to these models?

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: image
OpenAI mode is the toggle that needs to be enabled in order for endpoint to use OpenAI API.
I think that having this info is useful for the customer. Not every model has this toggle - so not every model can be used for elastic inference.
Do you still want me to remove it?

specification/inference/_types/CommonTypes.ts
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Sorry, missed your question regarding OpenAI mode. OpenAI mode is what I call the toggle that needs to be enabled in order for endpoint to use OpenAI API.
Not every model has this toggle - so not every model can be used for elastic inference.
image

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: Should have tested this myself before claiming that it would be ignored.
Good catch. Fixed.

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
Jan-Kazlouski-elastic: FYI Using of "Create" vs "Creates" is not consistent across the endpoint docs. For Amazon and Mistral it is "Creates" and it was taken as example. Thought it made more sense describing "what it does", but now I see that it is invitation to "do it".
Changed it since for the most of the providers it is "Create".

szabosteve: @Jan-Kazlouski-elastic Thanks! I'll fix Amazon and Mistral.

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
szabosteve: I would rather not include the full URLs. Currently, we can provide only one link per description; otherwise, the generator complains.

specification/inference/_types/CommonTypes.ts
jonathan-buttner: Ah I hadn't seen that before.

Can you try disabling it and see if a chat completion request using the inference API still works? I wonder if that only controls the UI in hugging face.

Can you give me an example of a model that has it?

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
jonathan-buttner: Ah ok

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: >Can you try disabling it and see if a chat completion request using the inference API still works? 

After disabling it on UI - it still works, but url that is presented to the client doesn't contain `v1/chat/completions`. So basically switching it on and off doesn't keep endpoint from processing OpenAI payloads if they are sent to the `/v1/chat/completions`, but hides the `/v1/chat/completions` section of the URL and provides different payload example. So if OpenAI mode is turned off or not there at all for a specific model - then full URL with `/v1/chat/completions` is hidden - client can't see/use the URL that must be used for integration.
Client might have prior knowledge on how url should look like and we're saying that it must contain `/v1/chat/completions`, but it is not safe to imply that customer will understand this section must be added on top of the regular url. Specially if model doesn't support OpenAI payload and attempt to include `/v1/chat/completions` will be made and error will be returned because model doesn't support it.

I guess we could make it clearer by telling customer that absence of `/v1/chat/completions` on running model page can be caused by OpenAI mode being turned off. And enabling this mode if this toggle is present when url doesn't contain `/v1/chat/completions` can lead to it being shown. But since this mode always been turned **ON** by default for me I think this issue has really low risks. BUT! Presence of this toggle is one of the signs that model is capable of processing OpenAI type payloads and can be used for integration. So for me personally having this note is making us safer. 
But you mentioning that you hadn't seen it before and obviously running successful tests in the past makes me wonder if it is something local to only some of the models.

UI when OpenAI mode is
On:
image

Off:
image

>I wonder if that only controls the UI in hugging face.

Yes it just controls UI.
>Can you give me an example of a model that has it?

https://huggingface.co/microsoft/Phi-3-mini-128k-instruct

jonathan-buttner: Ah I see. I wonder if it's be more helpful to put this information and an example picture like you have somewhere else in the docs. @szabosteve what do you think?

My suggestion is to would be to have the text as something like:

```
For `completion` and `chat_completion` tasks, the deployed model must be compatible with the Hugging Face Chat Completion interface (https://huggingface.co/docs/inference-providers/en/tasks/chat-completion#conversational-large-language-models-llms). The endpoint URL must include `v1/chat/completions`. To determine if the model supports the Hugging Face Chat Completion Interface and to access the correct URL follow the information here.
```

We'd have a link or something to either another page or a different section of the page that explains that the deployment should have a toggle for OpenAI. Then to get the correct URL they should enable the toggle and ensure the URL ends with `v1/chat/completions`.

szabosteve: I think the best would be to present this information somehow here. I'm afraid that linking to another docs page from the reference for such a low-level detail would be hiding information. I suggest something like this:

```suggestion
   * For `completion` and `chat_completion` tasks, the deployed model must be compatible with the Hugging Face Chat Completion interface (see the linked external documentation for details). The endpoint URL for the request must include `/v1/chat/completions`.
  * If the model supports the OpenAI Chat Completion schema, a toggle should appear in the interface. Enabling this toggle doesn't change any model behavior, it reveals the full endpoint URL needed (which should include `/v1/chat/completions`) when configuring the inference endpoint in Elasticsearch. If the model doesn't support this schema, the toggle may not be shown.
  * @ext_doc_id huggingface-chat-completion-interface
```

And then add the following to the [table.csv](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv):
```
huggingface-chat-completion-interface, https://huggingface.co/docs/inference-providers/en/tasks/chat-completion#conversational-large-language-models-llms
```
It will provide a link with the text `External documentation` at the end of the description, it makes the docs a bit more readable.

@jonathan-buttner What do you think?

jonathan-buttner: Yep that looks good. Thanks!

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
Jan-Kazlouski-elastic: Thank you @szabosteve

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: Applied change proposed by @szabosteve

specification/inference/_types/CommonTypes.ts
```diff
@@ -888,16 +888,29 @@ export class HuggingFaceServiceSettings {
...
    * The URL endpoint to use for the requests.
+   * For `completion` and `chat_completion` tasks, the deployed model must be compatible with the Hugging Face Chat Completion interface (see the linked external documentation for details). The endpoint URL for the request must include `/v1/chat/completions`.
+   * If the model supports the OpenAI Chat Completion schema, a toggle should appear in the interface. Enabling this toggle doesn't change any model behavior, it reveals the full endpoint URL needed (which should include `/v1/chat/completions`) when configuring the inference endpoint in Elasticsearch. If the model doesn't support this schema, the toggle may not be shown.
+   * @ext_doc_id huggingface-chat-completion-interface
    */
   url: string
+  /**
+   * The name of the HuggingFace model to use for the inference task.
+   * For `completion` and `chat_completion` tasks, this field is optional but may be required for certain models — particularly when using serverless inference endpoints.
+   * For the `text_embedding` task, this field should not be included or the request will result fail.
```
jonathan-buttner:
```suggestion
   * For the `text_embedding` task, this field should not be included. Otherwise, the request will fail.
```

specification/inference/put_hugging_face/PutHuggingFaceRequest.ts
```diff
@@ -44,6 +47,15 @@ import { InferenceChunkingSettings } from '@inference/_types/Services'
  * * `e5-small-v2`
  * * `multilingual-e5-base`
  * * `multilingual-e5-small`
+ *
+ * For Elastic's `chat_completion` and `completion` tasks:
+ * The selected model must support the `Text Generation` task and expose OpenAI API. HuggingFace supports both serverless and dedicated endpoints for `Text Generation`. When creating dedicated endpoint select the `Text Generation` task.
+ * After the endpoint is initialized (for dedicated) or ready (for serverless), ensure it supports the OpenAI API and includes `/v1/chat/completions` part in URL. Then, copy the full endpoint URL for use.
+ * Recommended models for `chat_completion` and `completion` tasks:
```
jonathan-buttner, line 54: @szabosteve should we say something like "known supported models"?

jonathan-buttner (changes requested):
Looking good.

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: Done.

jonathan-buttner (approved):

PR #4391, opened 2025-05-21
Adds custom InferenceEndpointInfo class for JinaAI

szabosteve:
## Overview

The Response Attributes section of the reference documentation currently displays the `task_types` defined in the `TaskType` enum used by the `InferenceEndpointInfo` class. However, `InferenceEndpointInfo` is not specific to any particular inference API integration — it lists _all_ available task types, not just those relevant to the specific integration being documented (see screenshot below).

Create_an_JinaAI_inference_endpoint___Elasticsearch_API_documentation__v9_

[Link to docs](https://www.elastic.co/docs/api/doc/elasticsearch/v9/operation/operation-inference-put-jinaai)

The screenshot shows the Response Attributes section of the JinaAI integration documentation. JinaAI supports only two task types: `rerank` and `text_embedding`, which are correctly listed under path parameters.

This PR introduces a new `TaskTypeJinaAi` enum that includes only the relevant task types and references it in the newly added `InferenceEndpointInfoJinaAI` class. As a result, the documentation will now accurately display only the supported task types for this integration — even in the Response Attributes section.


l-trotta (approved):
LGTM!

PR #4400, opened 2025-05-21
Initial prototype of an ESLint-based validator of the Elasticsearch specification

JoshMock:
As titled. :rocket:


flobernd:
Looks very promising @JoshMock ! Very good work.

I will try to write some rules next week.

It's probably worth experimenting with that foundation a bit in order to identify recurring patterns for whose it might be worth providing high level helper functions / abstractions for as part of the spec validator "SDK".

JoshMock:
> It's probably worth experimenting with that foundation a bit in order to identify recurring patterns for whose it might be worth providing high level helper functions / abstractions for as part of the spec validator "SDK".

I agree, and would be happy to add that, but I think it would require some discussion to make sure I (or whoever implements it) am correctly targeting the correct problems. It's probably a lot easier to for someone on the static language side to list those problems, because many of them we don't run into as often on the dynamic side. :smile: 

@flobernd Would it be worth merging this as-is so we can iterate on it, or do you think more SDK tooling is a hard requirement?

flobernd (approved):

flobernd:
I just approved!

Issue #4277, opened 2025-04-14
Unique summaries per path


lcawl:
## 🚀 Feature Proposal

Have a maximum of one specification per unique path.
Alternatively, have a method for specifying a unique operation summary for each unique path so that this information can be used in the OpenAPI document output.

NOTE: This might eventually overlap with a desire for something like [x-tagGroups](https://redocly.com/docs-legacy/api-reference-docs/specification-extensions/x-tag-groups), where we could group similar endpoints and affect the navigation such that tags have a hierarchy. That's not yet supported in our publishing system, however.

## Motivation

Currently we can have multiple paths represented in a single specification, for example:

https://github.com/elastic/elasticsearch-specification/blob/main/specification/nodes/stats/NodesStatsRequest.ts
```
 urls: [
    {
      path: '/_nodes/stats'
      methods: ['GET']
    },
    {
      path: '/_nodes/{node_id}/stats'
      methods: ['GET']
    },
    {
      path: '/_nodes/stats/{metric}'
      methods: ['GET']
    },
    {
      path: '/_nodes/{node_id}/stats/{metric}'
      methods: ['GET']
    },
    {
      path: '/_nodes/stats/{metric}/{index_metric}'
      methods: ['GET']
    },
    {
      path: '/_nodes/{node_id}/stats/{metric}/{index_metric}'
      methods: ['GET']
    }
  ]
```

Since they're all covered in a single specification, they all share the same operation summary and description in the OpenAPI output, which makes it harder to tell which one to use:

![Image](https://github.com/user-attachments/assets/4d3c8473-7601-47a1-893e-9b3dfe26ee79)

## Example

Since we only had a single page in the old docs for content like this (e.g. https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-stats.html) we don't have an example that shows how the description *should* vary, but the work required to explain those difference is worth the effort IMO.

DaveCTurner:
> we don't have an example 

[...]

PR #4415, opened 2025-05-26
[OpenAPI] Merge multiple paths in a single operation

swallez:
Fixes https://github.com/elastic/elasticsearch-specification/issues/4277

Adds an option in the OpenAPI export to group the many operations resulting from endpoints with multiple paths in a single operation. This works around the current issue where there are many duplicated operations in the nav bar and in the documentation.

A "main path" is chosen to become the path of the single OpenAPI operation that is produced. This is the endpoint's longest path, which will show all path parameters, be them required or optional. All paths are then listed at the top of the description (including the main one), see rendering below.

**WIP: do not merge this PR now.** This option must be configurable, as it is destructive from an OpenAPI schema perspective since it removes some operations.

---

Search endpoint rendered by Bump with this PR:

image


lcawl:
> **WIP: do not merge this PR now.** This option must be configurable, as it is destructive from an OpenAPI schema perspective since it removes some operations.

I see 179 operations where this new text appears in the OpenAPI document, so it's a pretty impactful change. It's not something we ideally do manually, but I think it will be necessary to confirm that (1) none of the operations that have been removed are ones that are affected by [overlays](https://github.com/elastic/elasticsearch-specification/tree/main/docs/overlays) (or if they are, that the overlay is moved to the appropriate consolidated operation), and (2) none of the operations that have been removed are target URLs (e.g. from https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv, https://github.com/elastic/kibana/blob/main/src/platform/packages/shared/kbn-doc-links/src/get_doc_links.ts, or I guess even from docs in general... which would imply that in addition to cleaning up their usage we'll likely need to create redirects just in case we don't catch all usages).

swallez:
> IMO if it make it more usable in general to just accomplish this list via simple markdown

The generated markup uses the CSS classes used by Bump in the main path description to provide a consistent rendering.

I've added a commit that removes the `<operation-summary>` that doesn't have any benefit in this context. The remaining `<div><span>` still make it valid Markdown and will be rendered correctly even outside of Bump. Now can also produce plain Markdown if you prefer.

> I see 179 operations where this new text appears in the OpenAPI document, so it's a pretty impactful change

Yes. As we already discussed, it cannot _replace_ the current OpenAPI file. It has to be a new one, dedicated to the docs pipeline.

> none of the operations that have been removed are target URLs

So I understand we're good on that front?

Some background: when we generate one operation for each method+path, the first one (as found in schema.json) has a "plain" (no suffix) identifier, and the others have a counter suffix (`-1`, `-2`, etc). When merging multiple path, all these suffixed operations are removed, and we only have the unsuffixed operation.

lcawl:
> Some background: when we generate one operation for each method+path, the first one (as found in schema.json) has a "plain" (no suffix) identifier, and the others have a counter suffix (`-1`, `-2`, etc). When merging multiple path, all these suffixed operations are removed, and we only have the unsuffixed operation.

That's great and should make it easier to check we don't have any of those suffixed URLs used elsewhere.
Looping in @georgewallace and @szabosteve so they can weigh in on the outstanding questions around whether it's preferrable to use plain markdown for the list of alternatives and when/if we want to do redirects and check for impact to overlays, link services, etc per https://github.com/elastic/elasticsearch-specification/pull/4415#issuecomment-2910631551

swallez:
I rebased on `main` where openapi generation now accepts command-line arguments (see PR #4437).

We now have an additional `make transform-to-openapi-for-docs` that has a different set of parameters, in particular the one added by this PR to merge operations.

This new make target stores is result in `output/openapi/elasticsearch-openapi-docs.json`, which **has not been added to the repository**. So we must run it manually (for now) to feed Bump and enjoy a streamlined navigation.

@szabosteve, @georgewallace can you give your opinion on outputting html with classes that will keep a consistent rendering in Bump vs just plain markdown? My opinion is that html with classes is fine as this is feature that specifically targets document rending with Bump for now, and we can always revisit it later if needed.

georgewallace:
@swallez I think for now HTML is fine. I am unsure on if bump can render markdown into html in their process. For cleaniness and ease of reading we should probably look at markdown in the future but no need to change at the moment.

lcawl:
> We now have an additional `make transform-to-openapi-for-docs` that has a different set of parameters, in particular the one added by this PR to merge operations.
> This new make target stores is result in `output/openapi/elasticsearch-openapi-docs.json`, which **has not been added to the repository**. So we must run it manually (for now) to feed Bump and enjoy a streamlined navigation.

I presume we can update the `make transform-to-openapi-for-docs` ought to generate an `elasticsearch-openapi-serverless-docs.json` too, right? And then update the `make overlay-docs` and `make lint-docs` to target those new files as well. If you want me to add those changes to this PR via a commit lmk!

Anaethelion (approved):

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/4550)|

### Questions ?

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.18|[](https://github.com/elastic/elasticsearch-specification/pull/4551)|

### Questions ?

Issue #4382, opened 2025-05-19
Missing retriever types


flobernd:
We are currently missing the `rescorer` and `linear` retriever types.

See:
https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers

Related to:
https://github.com/elastic/elasticsearch-net/issues/8528

l-trotta:
we're also missing `pinned` and `random rank`

sergey-tihon:
> we're also missing pinned and random rank

why they are not mentioned in docs? https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers

l-trotta:
I think they're only available through plugins

leemthompo:
@Samiul-TheSoccerFan @Mikep86 do you folks have a corresponding issue for this? I'm trying to track down various open threads about improving the retrievers docs

PR #4420, opened 2025-05-28
Search endpoint fixes

l-trotta:
Slowly increasing validation score for `search`.

- `reason` in [ErrorCause](https://github.com/elastic/elasticsearch-specification/blob/3004c087458578beda8be1c9c8d9c4d9f887f6ba/specification/_types/Errors.ts#L31) can be "null", not just missing (source: I trust the recordings for this one)
- in [GeoGridQuery](https://github.com/elastic/elasticsearch-specification/blob/3004c087458578beda8be1c9c8d9c4d9f887f6ba/specification/_types/query_dsl/geo.ts#L99) `geogrid` is actually called `geotile` (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/query/GeoGridQueryBuilder.java#L80))
- Missing RescorerRetriever, LinearRetriever, PinnedRetriever in [RetrieverContainer](https://github.com/elastic/elasticsearch-specification/blob/3004c087458578beda8be1c9c8d9c4d9f887f6ba/specification/_types/Retriever.ts#L31) (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/server/src/main/java/org/elasticsearch/search/retriever/RescorerRetrieverBuilder.java#L35))
- `ruleset_ids` in [RuleRetriever](https://github.com/elastic/elasticsearch-specification/blob/3004c087458578beda8be1c9c8d9c4d9f887f6ba/specification/_types/Retriever.ts#L106) can also accept string (not going to check it's always like this with string arrays)
- [RuleQuery](https://github.com/elastic/elasticsearch-specification/blob/2c9bffa8a577010bae6791aabb046c1ad2be7442/specification/_types/query_dsl/specialized.ts#L401) has both `ruleset_ids` and `ruleset_id` (which is the same just in a string format) (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/rules/RuleQueryBuilder.java#L361))
- Missing IntervalRange and IntervalRegexp in [IntervalsContainer](https://github.com/elastic/elasticsearch-specification/blob/2c9bffa8a577010bae6791aabb046c1ad2be7442/specification/_types/query_dsl/fulltext.ts#L85) (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/server/src/main/java/org/elasticsearch/index/query/IntervalsSourceProvider.java#L752))
- ~~[WeightedTokensQuery](https://github.com/elastic/elasticsearch-specification/blob/2c9bffa8a577010bae6791aabb046c1ad2be7442/specification/_types/query_dsl/WeightedTokensQuery.ts#L28)'s `tokens` is actually a list of WeightedToken objects, but since the field is a sting let's make it a list of Single Key Dictionaries (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/search/WeightedToken.java#L19))~~
- ~~[HighlightField](https://github.com/elastic/elasticsearch-specification/blob/2c9bffa8a577010bae6791aabb046c1ad2be7442/specification/_global/search/_types/highlighting.ts#L154) is also a Single Key Dictionary (one or many) (source: [server code](https://github.com/elastic/elasticsearch/blob/83a13b9cc4d71c7a0c695bf378eca781c05cac5a/server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/HighlightBuilder.java#L335))~~

~~The last 2 are breaking changes for static clients so I'm wondering if we should even introduce them just for the sake of validation, since it's not really broken, but it currently does not comply 100% with what the server can accept.~~


l-trotta:
> 2349/2401 	2385/2385

before it was 2316/2401 2384/2385

flobernd (approved):
LGTM from my side.

I would as well be fine with the breaking changes.

Btw: I think there is another retriever missing (the `LinearRetriever`).

flobernd:
Related to #4382

l-trotta:
splitting this in 2 so that the additive changes can be pushed to 8.x/9.0 and the breaking changes to 9.1

specification/_types/Retriever.ts
```diff
@@ -39,13 +39,60 @@ export class RetrieverContainer {
...
+
+export class InnerRetriever {
+  retriever: RetrieverContainer
+  weight: float
+  normalizer: ScoreNormalizer
+}
+
+export enum ScoreNormalizer {
+  none,
+  minmax
```
ioanatia, line 90: can I ask to please add `l2_norm` here? 🥺 
it was just merged this week cc @mridula-s109 

EDIT: we should only add this for 9.1/8.19

flobernd: Hi @ioanatia , we have to add the `l2_norm` in a separate PR since this one should as well be in 8.18.

l-trotta: thanks for noticing this! I'll add this in another PR since l2_norm will only be available in 8.19 and 9.1

Issue #4386, opened 2025-05-20
[OpenAPI] Publish `availability.since` in more contexts


lcawl:
## 🚀 Feature Proposal

Currently we support the `@availability` property as specified in https://github.com/elastic/elasticsearch-specification/blob/main/docs/add-new-api.md, including the `since` detail. However, we only output the `since` detail in the OpenAPI document when `stability` is `stable`.

Related to https://github.com/elastic/kibana/issues/221056 I think we should reconsider including that `since` detail in more contexts. For example, as stated in that Kibana issue:

- if `stability` = `stable`, `since` should be mandatory and `x-state` should be "Generally available; added in xxx" where xxxx is the `since` value.
- if `stablility` = `experimental` and `since` exists, `x-state` should be "Technical preview; added in xxxx` (or "Technical preview since xxx"). Otherwise it can just be "Technical preview".
- if `stability` = `beta` and `since` exists, `x-state` should be "Beta; added xxxx" (or "Beta since xxx"). Otherwise it can continue to just be "Beta".

I think ideally `since` would be required in most scenarios. IMO `since` is not necessary in the Serverless-specific OpenAPI document, where only "Generally available", "Technical preview", and "Beta" are necessary (unless there is a need to start displaying date values in those cases). The exact terms for these lifecycle states should align with what is used in the narrative documentation frontmatter.

## Motivation

This helps clarify situations where users are looking at the "main" branch of the documentation and can't tell when a new API was introduced (irrespective of its lifecycle state). The same can also occur once we have both 9.1.x and 9.0.x publishing on the "v9" branch of the documentation.

## Example

For example, [GetQueryRequest.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/esql/get_query/GetQueryRequest.ts) has the following specification:

```
 * @availability stack since=9.1.0 stability=experimental visibility=public
 * @availability serverless stability=experimental visibility=public
```

Currently that appears in both OpenAPI documents as follows:

```
        "x-state": "Technical preview"
```

However, the proposed change would mean that the `elasticsearch-serverless-openapi.json` is unchanged but the `elasticsearch-openapi.json` should instead have:

```
  "x-state": "Technical preview; added in 9.1.0"
```

flobernd:
I very much like the idea to support additive lifecycles!

If you don't mind, here is a modified proposal for the syntax that feels slightly more consistent to me:

```ts
 * @availability stack visibility=public, since=9.3.0 stability=ga
[...]

Issue #4356, opened 2025-05-06
Update OpenAPI generator to include parameter- and property-level availability


lcawl:
priority: high
effort: low?

## 🚀 Feature Proposal

Some specifications include `@availability` details at the API, parameter, and property level. However only the operation-level information is currently included in the OpenAPI documents. The generator should be updated to put that information into the `x-state` at all levels.

## Motivation

Since we are currently publishing only the "current" release of the API docs for each major version, it's useful to show as much info as possible about when parameters and properties were added/deprecated/discontinued etc.

I did a quick test to verify that if the `x-state` information exists at the parameter level, it will be displayed in Bump.sh:

![Image](https://github.com/user-attachments/assets/787a45ad-e37a-41a0-9d28-cead174447b3)

## Example

For example in 
https://github.com/elastic/elasticsearch-specification/blob/main/specification/_global/field_caps/FieldCapabilitiesRequest.ts 
it has `@availability` details at the operation-level:

```
...
 * @rest_spec_name field_caps
 * @availability stack since=5.4.0 stability=stable
 * @availability serverless stability=stable visibility=public
...
```

... as well as at the parameter level:

```
...
    /**
     * A comma-separated list of filters to apply to the response.
     * @availability stack since=8.2.0
     * @availability serverless
     */
    filters?: string
```

In the [generated OpenAPI files](https://github.com/elastic/elasticsearch-specification/tree/main/output/openapi) the details appear at the operation-level:

```
    "/_field_caps": {
      "get": {
...
        "x-state": "Added in 5.4.0"
      },
```

... but are missing from the parameter level:

```
      "field_caps-filters": {
        "in": "query",
        "name": "filters",
        "description": "A comma-separated list of filters to apply to the response.",
        "deprecated": false,
        "schema": {
          "type": "string"
        },
        "style": "form"
      },
```

lcawl:
I think there might be some circumstances where this functionality is not working yet.
For example, `FieldCapabilitiesRequest.ts` contains:

```ts
    /**
     * A comma-separated list of filters to apply to the response.
     * @availability stack since=8.2.0
     * @availability serverless
     */
    filters?: string
```

...And I see the availability details now in the `schema.json`, for example:

```json
        {
          "availability": {
            "serverless": {},
            "stack": {
              "since": "8.2.0"
            }
          },
          "description": "A comma-separated list of filters to apply to the response.",
          "name": "filters",
          "required": false,
          "type": {
[...]

PR #4313, opened 2025-04-28
[OpenAPI] Lift enum member descriptions in property descriptions

swallez:
The OpenAPI spec doesn't support adding descriptions to enumeration members. This PR fixes that by having the OpenAPI converter lift enum member descriptions in the descriptions of properties of an enum type.

It also handles aliases, enum arrays and lenient arrays (e.g. `Foo | Foo[]`).

<details><summary>Rendering in bump.sh</summary>

image

</details><details><summary>Enumerations with no descriptions are left unchanged</summary>

image

</details>

Caveat: for unary values (i.e. not an array or lenient array), bump.sh will still render its list below the descriptions added by this PR. This could be avoided by removing the enumeration values from the OpenAPI spec (and keep them only in the docs), but this would remove some information from that OpenAPI spec.

<details><summary>Example</summary>

image

</details>

PR #4431, opened 2025-05-30
Openapi: adding more info in x-state, adding x-state info to fields and params

l-trotta:
Closes https://github.com/elastic/elasticsearch-specification/issues/4386, Closes https://github.com/elastic/elasticsearch-specification/issues/4356.
Couple of doubts:

- ~~This would replace "x-beta" with "x-state" in the Beta case, and change the field type from a boolean to a string, is this okay?~~
- ~~Does this also have to be configurable, or can it be the standard output?~~


compiler-rs/clients_schema_to_openapi/src/main.rs
```diff
@@ -151,6 +151,16 @@ mod tests {
         .run()
     }
 
+    #[test]
```
l-trotta, line 154: ignore this, I use it for quick testing, will remove it before merging

lcawl:
> * This would replace "x-beta" with "x-state" in the Beta case, and change the field type from a boolean to a string, is this okay?

There is info about the differences between these extensions in https://docs.bump.sh/help/specification-support/doc-badges/. The benefit of using x-beta was that it affected what was deemed a "breaking change" in the automated changelog. However, given that we didn't have the same consideration for tech preview / experimental states, IMO it's not worth trying to have *both* the `x-beta` and `x-state` applied. Rather, I think the latter is the richer and more consistent option.

> * Does this also have to be configurable, or can it be the standard output?

Since this PR affects only a [specification extension](https://spec.openapis.org/oas/latest.html#specification-extensions), in my opinion it should be something that's ignorable by any other tools or processes that consume these OpenAPI documents. That is to say, IMO it can be standard output.

l-trotta:
@lcawl fixed the serverless output!
@swallez right now what happens if there's no flavor set in the configuration is that availability gets completely skipped, because if it does both stack end serverless we'll have the same bug as before where the first info gets overwritten. Should we have "stack" as default flavor if none is set?

l-trotta:
@lcawl @swallez ready for review!

compiler-rs/clients_schema_to_openapi/src/lib.rs
```diff
@@ -103,7 +104,7 @@ pub fn convert_expanded_schema(model: &IndexedModel, config: &Configuration) ->
                 continue;
             }
         }
-        paths::add_endpoint(endpoint, &mut tac, &mut openapi.paths)?;
+        paths::add_endpoint(endpoint, &mut tac, &mut openapi.paths, &config.flavor)?;
```
swallez: The config is part of `tac`, I added it in #4313. This effectively makes `TypesAndComponents` more than types and components. I'll open a PR to rename it to something more "kitchensink-esque" like `Context`.

compiler-rs/clients_schema_to_openapi/src/lib.rs
```diff
@@ -149,30 +150,37 @@ fn info(model: &IndexedModel) -> openapiv3::Info {
...
-        for (_, availability) in avails {
-            if let Some(stability) = &availability.stability {
-                match stability {
+        if let Some(flav) = flavor {
+            if let Some(availability) = avails.get(flav) {
+                let Availability {since,stability,..} = &availability;
+                let stab = stability.clone().unwrap_or(Stability::Stable);
+                let mut since_str = "".to_string();
+                if let Some(since) = since {
+                    since_str = "; Added in ".to_string() + since;
```
swallez: The `+` operator on `String` is rarely used in Rust. Prefer the more compact and more efficient `since_str = format!("; Added in {since}");`

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.0|[](https://github.com/elastic/elasticsearch-specification/pull/4584)|

### Questions ?

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/4585)|

### Questions ?

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.18|[](https://github.com/elastic/elasticsearch-specification/pull/4586)|

### Questions ?

Issue #4303, opened 2025-04-23
Security privileges are missing from Elasticsearch API documentation


bytebilly:
Security privileges that are needed to call a specific Elasticsearch API are not documented, even if it looks like they're supported (https://github.com/elastic/elasticsearch-specification/pull/717). This information is important to create roles with the right privileges.

For example, looking at the [Create pipeline API "old" documentation](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/put-pipeline-api.html) we can find

> **Prerequisites**
If the Elasticsearch security features are enabled, you must have the `manage_pipeline`, `manage_ingest_pipelines`, or `manage` cluster privilege to use this API.

but looking at the ["new" documentation page](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-ingest-put-pipeline), there is no mention of that.

lcawl:
The support for this type of information in the specification was originally added per https://github.com/elastic/elasticsearch-specification/issues/713 and there's more info in https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#index_privileges and https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#cluster_privileges.  However, it looks like the "create pipeline" specification is missing that info: https://github.com/elastic/elasticsearch-specification/blob/main/specification/ingest/put_pipeline/PutPipelineRequest.ts

IMO this brings to light three things that need to be resolved:

1. Ensure that teams are aware of those fields and are adding them when they create / update the specifications.
2. Update the OpenAPI generator to include that information (e.g. append it to the operation description). I thought we had an issue already open for that, but didn't find one in a quick search.
3. Clarify the guidelines about how to encode alternative privileges. As in this pipeline case, it says "manage_pipeline, manage_ingest_pipelines, or manage cluster privileges", which you could technically encode as `@cluster_privileges manage_pipeline, manage_ingest_pipelines, manage` but I think we'd want to have a clear way of denoting a case where all three, for example, are required.
4. IMO this should also be revisited as to how/if it encompasses Serverless. Per https://www.elastic.co/docs/reference/elasticsearch/security-privileges there are privileges that are unique to Serverless, therefore a new `@cluster_privileges_serverless` and `@index_privileges_serverless` might be required (or some other variation for denoting deployment-specific variations).

bytebilly:
Thanks Lisa for taking a look at that!

[...]

PR #4434, opened 2025-05-30
Openapi: adding privileges

l-trotta:
Closes https://github.com/elastic/elasticsearch-specification/issues/4303.

There was no desired output in the original issue, so I'm not sure if this is the format expected. Let me know and I'll change it!


compiler-rs/clients_schema_to_openapi/src/lib.rs
```diff
@@ -176,5 +176,14 @@ pub fn availability_as_extensions(availabilities: &Option<Availabilities>) -> In
         }
     }
 
+    if let Some(privs) = privileges {
+        if privs.index.len()>0 {
+            result.insert("index-privileges".to_string(),serde_json::Value::String(privs.index.join(",")));
+        }
+        if privs.cluster.len()>0 {
+            result.insert("cluster-privileges".to_string(),serde_json::Value::String(privs.cluster.join(",")));
```
lcawl:
```suggestion
            result.insert("x-cluster-privileges".to_string(),serde_json::Value::String(privs.cluster.join(",")));
```

compiler-rs/clients_schema_to_openapi/src/lib.rs
```diff
@@ -176,5 +176,14 @@ pub fn availability_as_extensions(availabilities: &Option<Availabilities>) -> In
         }
     }
 
+    if let Some(privs) = privileges {
+        if privs.index.len()>0 {
+            result.insert("index-privileges".to_string(),serde_json::Value::String(privs.index.join(",")));
```
lcawl: Per https://spec.openapis.org/oas/v3.0.3#specification-extensions if we go the route of publishing these as specification extensions, they need to start with `x-`.
```suggestion
            result.insert("x-index-privileges".to_string(),serde_json::Value::String(privs.index.join(",")));
```

lcawl:
My two cents is that of the three options I mentioned in https://github.com/elastic/elasticsearch-specification/issues/4303#issuecomment-2836840039, "an addendum tacked onto the beginning or end of the operation's "description" field" is likely the best current option since it aligns with what Kibana's doing. Otherwise I think if we went the extension route we'd need a more open-ended structure since the privileges in the Kibana and Cloud API docs would need different options than "cluster-privileges" and "index-privileges".

georgewallace:
Coming into this one a little late but I do agree with @lcawl if we already have a team doing this a different way then we should follow suit to stay consistent. I do think long term keeping these as separate fields makes the most sense though.  @lcawl is this what you are talking about?
Screenshot 2025-06-03 at 8 37 59 PM

l-trotta:
@lcawl @georgewallace updated so that privileges are now in the description! please help me with the output format because I'm really not a fan of this one ^^"

georgewallace:
@lcawl curious your thoughts but if we follow suite with what Kibana is doing would be something like

```
[Required authorization] index-privilege: monitor, cluster-privilege: monitor
```

lcawl:
Could we format it into more complete phrases? For example:

```
## Required authorization

 * Index privileges: `blah`, `blah`, `blah`
 * Cluster privileges: `blah`
 ```

Or if there's concern about including markdown, then maybe:

```
Required authorization includes: `blah`, `blah`, `blah` index privileges and `blah` cluster privileges.
 ```

georgewallace:
@lcawl I like the first one. It doesnt necessarily match with what Kibana is doing ([Get an agent action status](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-fleet-agents-agentid-upgrade)) but I think that is ok as your first one likes much better. The Operation summary name is already an H2, for this sake should it be an h3 instead? I see inconsistencies on the pages now so H2 should be fine but just in case.

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -439,6 +446,24 @@ fn split_summary_desc(desc: &str) -> SplitDesc{
...
 }
 
+fn add_privileges(privileges: &Option<Privileges>) -> Option<String>{
+    if let Some(privs) = privileges {
+        let mut result = "\n ##Required authorization\n".to_string();
+        if privs.index.len()>0 {
+            result = result + "* Index privileges: " + &privs.index.iter()
+                .map(|a| {"`".to_string() + a + "`"})
+                .collect::<Vec<String>>().join(",");
+        }
```
swallez: We can write this in a more concise and idiomatic way. However you need to add the [itertools](https://docs.rs/itertools/latest/itertools/) dependency to `clients_schema_to_openapi/Cargo.toml` by adding `itertools = { workspace = true }` (the version is defined in the top-level `Cargo.toml`).

And then import `itertools::Itertools` in this file.

```suggestion
        if !privs.index.is_empty() {
            result += "* Index privileges: ";
            result += &privs.index.iter()
                .map(|a| format!("`{a}`"))
                .join(",");
        }
```

compiler-rs/clients_schema_to_openapi/src/paths.rs
```diff
@@ -439,6 +446,24 @@ fn split_summary_desc(desc: &str) -> SplitDesc{
...
+        if privs.cluster.len()>0 {
+            result = result + " * Cluster privileges: " + &privs.cluster.iter()
+                .map(|a| {"`".to_string() + a + "`"})
+                .collect::<Vec<String>>().join(",");
+        }
```
swallez: Same here :-)

PR #4438, opened 2025-06-02
Revert "[Backport 9.0] Adds new `embedding_type` values to PUT Cohere inference endpoint"

szabosteve:
Reverts elastic/elasticsearch-specification#4432

### DIFF
--- a/specification/inference/_types/CommonTypes.ts
+++ b/specification/inference/_types/CommonTypes.ts
@@ -585,8 +585,6 @@ export class CohereServiceSettings {
   api_key: string
   /**
    * For a `text_embedding` task, the types of embeddings you want to get back.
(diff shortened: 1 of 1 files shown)


flobernd:
@szabosteve What's the reason for reverting this change? Is it 9.1+ only? Just asking because I released a new client version today and it seems to contain this change 😅

szabosteve:
@flobernd Yes, it's a 9.1 (and 8.19) change, so it's not released yet.
This is the corresponding issue: https://github.com/elastic/docs-content/issues/310
Which refers to the dev PR: https://github.com/elastic/elasticsearch/pull/120751, the tags on that PR also state 9.1 and 8.19.

flobernd (approved):

PR #4452, opened 2025-06-04
Adds integration and task type list to PUT inference endpoint docs

szabosteve:
## Overview

This PR completes the PUT inference endpoint docs by adding the integration list with the supported task types to the description.


pquentin (approved):
Thanks! LGTM.

PR #4454, opened 2025-06-04
Remove Experimental Labels for SparseVectorQuery Token Pruning for GA

markjhoy:
A small PR to remove the `stability=experimental` tags for the pruning configuration items for the `sparse_vector` query type.


output/schema/schema.json
```diff
@@ -91495,12 +91492,9 @@
...
         {
           "availability": {
-            "serverless": {
-              "stability": "experimental"
-            },
+            "serverless": {},
             "stack": {
-              "since": "8.15.0",
-              "stability": "experimental"
+              "since": "8.15.0"
```
kderusso: Did you automatically regenerate these files, or hand edit them? I think we still need the stack version information and I'm surprised if it got wiped out automatically.

markjhoy: I followed the CONTRIBUTING instructions and ran `make contrib` -- should the `since` label be updated to 8.19?

kderusso: I don't think so, because it's still available in previous APIs? Maybe we can get help from the clients team to confirm this is correct.

pquentin: It depends. If the API did not change since 8.15, then it's fine, otherwise we should switch to 8.19.

pquentin: (Also, sorry for the 2 weeks delay! Asking devtools-team for a review can help with faster review time, as there are many changes here owned by the docs team which we don't review.)

markjhoy: > It depends. If the API did not change since 8.15, then it's fine, otherwise we should switch to 8.19.

I think we're good unless there's a change I don't know about (@kderusso ?)

markjhoy:
NOTE - do not merge until https://github.com/elastic/elasticsearch/pull/128854 is merged

kderusso (approved):
LGTM, thanks!

PR #4470, opened 2025-06-05
Add l2 norm to linear retriever

l-trotta:
Addition to #4420, but only for 9.1/8.19. `l2_norm` was recently added as option for ScoreNormalizer in LinearRetriever https://github.com/elastic/elasticsearch/pull/128504


pquentin (approved):
Thanks! LGTM.

PR #4474, opened 2025-06-06
Add configuration for sorenlouv/backport tool

pquentin:
Elastic engineers commonly use https://github.com/sorenlouv/backport to perform backports. I've been using it in the elasticsearch-specification repository for months now.

Since pull request forks are not accepted in this repository, committing the configuration is helpful to avoid backports being pushed to forks. It will mean that active branches will have to be maintained here, but that's okay. I've had this locally for a while now, and I update it whenever I notice any missing branches.

(I'm backporting this change because it's easy to do, but it's not needed as backports are done from main.)

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -1,0 +1,6 @@
+{
+  "repoOwner": "elastic",
+  "repoName": "elasticsearch-specification",
+  "targetBranchChoices": ["9.0", "8.19", "8.18", "8.17"],
+  "fork": false
+}


flobernd (approved):
Thanks! Let's switch to the corresponding GHA as well. Tested this in the .NET generator repo and it's way better than the current action.

Can do that change in a follow up. If you prefer to do it yourself or to include it in this PR, just a quick note: the backport prefix setting can't contain whitespaces which means we have to rename our labels to "backport-x.x" instead of "backport x.x".

pquentin:
Oh, wow, I had no idea there was a GitHub action. That's great to hear! Happy to do this as a follow-up, as it's a larger change.

Regarding the whitespace issue, I think we can change `branchLabelMapping` from the default `^backport-to-(.+)$` to `^backport (.+)$`.

flobernd:
@pquentin Sure. I think the action automatically reads the config that you added in this PR.

In the .NET generator repository I don't have a config yet, but I used the provided option on the action:
https://github.com/sorenlouv/backport-github-action/blob/929f69d04adbc196d982e60f02837b6cc00b3129/action.yml#L7-L9

The default value does not seem to contain a regex:

```yaml
auto_backport_label_prefix: backport-to-
```

I tried with:

```yaml
auto_backport_label_prefix: 'backport '
```

but this was not correctly recognized when I did an initial test:

```json
"results": [
    {
      "targetBranch": " 9.0",
      "status": "handled-error",
      "error": {
        "name": "BackportError",
        "errorContext": {
          "code": "invalid-branch-exception",
          "branchName": " 9.0"
        }
      }
    }
  ]
```

It parsed the prefix incorrectly and assumed the whitespace to be part of the branch name. That's probably something that we could fix upstream or ask to get fixed upstream, but I didn't had time to check that any further so I just used `backport-` which works perfectly fine.

If we are lucky, we don't have to use this option in the GHA at all and the tool will read it from the config file.

Issue #4443, opened 2025-06-03
Related product metadata per operation


georgewallace:
**Please read this entire template before posting any issue. If you ignore these instructions
and post an issue here that does not follow the instructions, your issue might be closed,
locked, and assigned the `invalid` label.**

It's not uncommon that somebody already opened an issue or in the best case it's already fixed but not merged. That's the reason why you should [search](https://github.com/elastic/elasticsearch-specification/labels/enhancement) at first before submitting a new one.

## 🚀 Feature Proposal

To support our site and doumentation search we need the ability to tag each page of our documentation with a product label from the following list. https://elastic.github.io/docs-builder/syntax/frontmatter/#products

## Motivation

Adding this metadata will allow us to add a META header tag to each page so that our site search ran by the web team can pick it up and add the facet to search and allow customers filtering content on e.g elasticsearch, to only see content related to Elasticsearch. 

## Example

We would use the product extension to populate the meta class elastic, product, name with the products that are related. An example of that tag can be seen on an existing piece of documentation. 
<img width="715" alt="Image" src="https://github.com/user-attachments/assets/1598cf7c-f07c-4d93-9787-8b0e5f3b29eb" />

georgewallace:
I can't find the earlier issue created for this so opening a new one. 

the following are the Distinct tags found in the Elasticsearch OpenAPI spec 

```
{'health_report', 'security', 'ilm', 'data stream', 'ccr', 'inference', 'cluster', 'enrich', 'ml data frame', 'transform', 'license', 'ml', 'fleet', 'document', 'tasks', 'rollup', 'graph', 'cat', 'migration', 'xpack', 'features', 'ml anomaly', 'esql', 'snapshot', 'watcher', 'shutdown', 'sql', 'searchable_snapshots', 'text_structure', 'info', 'ml trained model', 'ingest', 'search', 'slm', 'indices', 'autoscaling', 'logstash', 'query_rules', 'script', 'eql', 'search_application', 'analytics', 'synonyms', 'connector'}
```

One potential solution would be to define a mapping of these tag values to values from the Products list shared earlier.

georgewallace:
After further ideation with Sylvain, the idea would be

1. Add into the OpenAPI generation (for docs) logic to add a property to each operation that adds a extension product, `x-product-feature` that's value is `elasticsearch` as all calls within the elasticsearch specification relate to elasticsearch
[...]

PR #4511, opened 2025-06-10
[Openapi] Product metadata in extension

l-trotta:
Closes https://github.com/elastic/elasticsearch-specification/issues/4443


l-trotta:
@georgewallace I have hardcoded the table that will be in a file just to output a quick preview, if the output is okay then I'll proceed with the implementation!

georgewallace:
Output looks good. Thanks @l-trotta 
cc: @szabosteve

l-trotta:
~~@swallez not sure about how to pass the file, should it be another cli input? or just a constant since it won't change? should I worry about file system separators?~~ we've decided to hardcode it since it won't change

PR #4514, opened 2025-06-10
add language client examples to YAML files

miguelgrinberg:
This change adds all the language variations to YAML example files. There is a new command:

    make generate-language-examples

to refresh all YAML files with up to date examples for all available languages (currently Ruby, PHP, JS, Python and curl).


lcawl (approved):
I spot-checked a few examples and tested a curl example and overall LGTM.

One thing to consider, since we're using some variables like this:

```
curl -X GET -H "Authorization: ApiKey $ELASTIC_API_KEY" "http://localhost:9200/_alias"
```
Should we use them for the hostname and port number too? It took me a few minutes to notice that it was "http" instead of "https" while I was testing.

miguelgrinberg:
@lcawl ah yes, I'm doing this all the time and I think I'm now sort of blind wrt this type of detail. I can change the generated examples to use env vars as well. Thanks!

Makefile
```diff
@@ -52,6 +54,7 @@ transform-expand-generics: ## Create a new schema with all generics expanded
 	@npm run transform-expand-generics --prefix compiler
 
 transform-to-openapi: ## Generate the OpenAPI definition from the compiled schema
+	@make generate-language-examples
```
l-trotta: I think we should just keep it in make generate, if the schema json isn't updated then the openapi output won't change

l-trotta:
should we write somewhere that now `make generate` also requires to install dependencies from /docs/examples? I'm also assuming some CI will break because of this, but I cannot anticipate which ones so we can add the install commands later.

miguelgrinberg:
@l-trotta I have added the `npm install` for `docs/examples` to the `make setup` command ([here](https://github.com/elastic/elasticsearch-specification/blob/06d960106197f8737123181dcad1d71e23452dc4/Makefile#L47)). Isn't that sufficient for everything to continue working as before. This setup installs dependencies on a bunch of other directories too.

Makefile
miguelgrinberg: Good catch. This should only be done in generate, as you say. Thanks!

PR #4517, opened 2025-06-10
Fix AsyncQueryRequest - those parameters should be in the body

smalyshev:
`keep_alive`, `keep_on_completion` and `wait_for_completion_timeout` are body parameters for `AsyncQueryResuest`.


pquentin (approved):
Thanks! LGTM.

Note that YAML tests would have helped detect this earlier.

PR #4543, opened 2025-06-12
Dictionary allow only string as key lint

### LINKED ISSUE #4510: [validation] Dictionary key type should resolve to string
Still debating if Enums should be acceptable (would need to check if the enum is string only)

l-trotta:
Adds linter rule for dictionary + fix rule-breaking type.
TaskId is currently a `string | integer` union, but I can find no evidence of it being a number anywhere. 
sources:

- [flight recorder 1](https://github.com/elastic/clients-flight-recorder/blob/6ad96f07e62ffb4639dc31cb406004dbbe4632d9/recordings/nest-tests/reindex_rethrottle/Document/Multiple/ReindexRethrottle/ReindexRethrottleApiTests/ReindexRethrottleApiTests__ClientUsage_b__21_2_2525__response.json)
- [flight recorder 2](https://github.com/elastic/clients-flight-recorder/blob/6ad96f07e62ffb4639dc31cb406004dbbe4632d9/recordings/nest-tests/reindex_rethrottle/Document/Multiple/ReindexRethrottle/ReindexRethrottleApiTests/ReindexRethrottleApiTests__ClientUsage_b__21_3_2526__response.json)
- [server code](https://github.com/elastic/elasticsearch/blob/4cda8c2dcdac1f8a1701337bc2c1ec663343d35c/server/src/main/java/org/elasticsearch/tasks/TaskId.java#L103)

the format is actually `nodeId:taskId` which is a string.
Closes #4510


l-trotta:
@flobernd @Anaethelion this is not breaking for java since `string | integer` unions are simplified to string, checking just in case it is breaking for your clients

validator/eslint-plugin-es-spec.js
```diff
@@ -17,11 +17,13 @@
  * under the License.
  */
 import singleKeyDict from './rules/single-key-dictionary-key-is-string.js'
+import Dict from './rules/dictionary-key-is-string.js'
 import invalidNodeTypes from './rules/invalid-node-types.js'
 
 export default {
   rules: {
     'single-key-dictionary-key-is-string': singleKeyDict,
+    'dictionary-key-is-string': Dict,
```
flobernd: Nit: code style.

flobernd (approved):
LGTM!

Anaethelion (approved):

PR #4576, opened 2025-06-13
[DOCS] Add link to stored fields example

marciw:
Related to elastic/docs-projects#302 and https://github.com/elastic/elasticsearch/pull/129380

✅ requisite example page already merged https://github.com/elastic/elasticsearch/pull/129380

- Add example link to `stored_fields` parameter description
- Fix unrelated typos


specification/_global/get/GetRequest.ts
```diff
@@ -163,8 +163,9 @@ export interface Request extends RequestBase {
      * A comma-separated list of stored fields to return as part of a hit.
      * If no fields are specified, no stored fields are included in the response.
      * If this field is specified, the `_source` parameter defaults to `false`.
-     * Only leaf fields can be retrieved with the `stored_field` option.
-     * Object fields can't be returned;​if specified, the request fails.
+     * Only leaf fields can be retrieved with the `stored_fields` option.
+     * Object fields can't be returned; if specified, the request fails.
+     * For an example, refer to [Retrieve stored fields](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrieve-stored-fields).
```
szabosteve: Sorry, I didn't mention it earlier, we can use `ext_doc_id` here. Please add the next line to the [table.csv](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv) file as part of this PR, so we can keep track the links we are adding:
`retrieve-stored-fields,https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrieve-stored-fields`
and then refer to the ID as the suggestions shows below, please.
```suggestion
     * @ext_doc_id retrieve-stored-fields
```

marciw: I didn't think this was a good case for ext_doc_id because it's a section-specific thing that's not likely to be reused. Or is it required?

szabosteve: No problem if it's not reused. The spec has a lot of `ext_doc_id` in param level, which might never be reused elsewhere. We try to avoid adding markdown links to the API docs description – at least when we can avoid it – because they are not tracked, and it makes them much harder to redirect. As a description field has only one ext_doc_id, sometimes we forced to add MD links to descriptions if there is already an external doc link in it, but otherwise, it's better to use the `ext_doc_id`. I know it's far from ideal, so for this reason, we'd like to expand the external documentation property, so we could add more links to a single object. But that's something for the future. LMKWYT, please.

marciw:
> I left one suggestion.

@szabosteve Thanks! fixed and ready for your re-review 🙏

szabosteve (approved):
Thank you very much, LGTM!

marciw:
Closing because that's easier and faster than resolving some dizzying conflicts 😵  

redoing in https://github.com/elastic/elasticsearch-specification/pull/4640

PR #4542, opened 2025-06-12
Move _reindex API examples to ES reference documentation

charlotte-hoblik:
This PR moves the Reindex API examples to the Elasticsearch reference documentation, as part of the approach to unify API reference documentation.

DO NOT MERGE BEFORE https://github.com/elastic/elasticsearch/pull/129341

Connected to: 
- https://github.com/elastic/docs-projects/issues/302
- https://github.com/elastic/elasticsearch/pull/129341

PR #4580, opened 2025-06-16
add Java language examples

l-trotta:
Followup of #4514, adding java client examples for the docs.
Out of 652 examples, 388 were generated by the java-request-converter, while the others are skipped for a number of reasons (APIs not supported by the client, unions the client cannot deserialize, classes that don't support json deserialization directly). 
Out of those 388 examples, 12 had to be fixed manually (because of missing string escapes or missing/wrong brackets).

Unlike the other language examples, these won't be regenerated each time the openapi files are updated, since they rely on the experimental java-request-converter, so they will be generated locally when needed, using the new new `generate-language-examples-with-java` make command.


compiler-rs/clients_schema/src/lib.rs
```diff
@@ -506,6 +506,7 @@ pub struct SchemaExample {
     pub value: Option<String>,
     pub external_value: Option<String>,
     pub alternatives: Option<Vec<ExampleAlternative>>,
+    pub alternative_java: Option<Vec<ExampleAlternative>>,
```
swallez: Probably missing some context: why isn't that part of `alternatives`, since `ExampleAlternative` has a `language` property?

miguelgrinberg: I discussed this with Laura earlier today. We are not going to use a separate entry for Java, we can add it to the same list used by the other languages.

I have some changes in #4582 that she needs for this to work. Once my change is merged she'll be able to improve this and integrate Java fully with the other languages.

docs/examples/generate-language-examples.js
```diff
@@ -44,6 +46,50 @@ async function generateLanguages(example) {
...
+
+    const {status, stdout, stderr} = await java.run(args);
+    if (status) {
+      console.log(stderr);
+      console.log(JSON.stringify(javaReqs));
+    }
+    else {
+      const alternative_java = [];
+      alternative_java.push({
+        language: "java",
```
miguelgrinberg: Should it be "Java", capitalized? This is exactly what the docs are going to show in the dropdown.

output/openapi/elasticsearch-openapi.json
```diff
@@ -30975,6 +30975,66 @@
                   "summary": "Reindex multiple sources",
                   "description": "Run `POST _reindex` to reindex from multiple sources. The `index` attribute in source can be a list, which enables you to copy from lots of sources in one request. This example copies documents from the `my-index-000001` and `my-index-000002` indices.\n",
                   "value": "{\n  \"source\": {\n    \"index\": [\"my-index-000001\", \"my-index-000002\"]\n  },\n  \"dest\": {\n    \"index\": \"my-new-index-000002\"\n  }\n}"
+                },
+                "ReindexRequestExample10": {
```
miguelgrinberg: Why are all these new reindex examples part of this PR?

output/schema/schema.json
```diff
@@ -25965,6 +25985,10 @@
             {
               "code": "curl -X POST -H \"Authorization: ApiKey $ELASTIC_API_KEY\" -H \"Content-Type: application/json\" -d '{\"query\":{\"term\":{\"user.id\":\"kimchy\"}},\"max_docs\":1}' \"$ELASTICSEARCH_URL/my-index-000001/_delete_by_query\"",
               "language": "curl"
+            },
+            {
+              "code": "client.deleteByQuery(d -> d\n\t.index(\"my-index-000001\")\n\t.maxDocs(1L)\n\t.query(q -> q\n\t\t.term(t -> t\n\t\t\t.field(\"user.id\")\n\t\t\t.value(FieldValue.of(\"kimchy\"))\n\t\t)\n\t)\n);\n",
```
miguelgrinberg, line 25990: It looks like your code generator outputs tabs for indentation. I do not know how bump.sh interprets tabs when it renders code. I think it would be safer to produce spaces, if you can.

output/openapi/elasticsearch-openapi.json
l-trotta: good catch, they have been [recently removed](https://github.com/elastic/elasticsearch-specification/pull/4542), but I reintroduced them here due to rebasing, I'll delete them

PR #4597, opened 2025-06-17
Add GET snapshots state query parameter

JeremyDahlgren:
Adds documentation for the SnapshotState enum and state query parameter added in elasticsearch PR [128635](https://github.com/elastic/elasticsearch/pull/128635).


specification/snapshot/get/SnapshotGetRequest.ts
```diff
@@ -147,6 +148,14 @@ export interface Request extends RequestBase {
...
      */
     sort?: SnapshotSort
+    /**
+     * Only return snapshots with a state found in the given comma-separated list of snapshot states.
+     * The default is all snapshot states.
+     * @server_default All snapshot states
+     * @availability stack since=9.1.0
+     * @availability serverless
+     */
+    state?: SnapshotState
```
pquentin: For better user experience, comma-separated values in query parameters are exposed as lists in Elasticsearch clients, so we should allow `SnapshotState` but also `SnapshotState[]`, as shown below. I had to remove the default value line too because of this error: `Error: Default values in unions can only be configured for instance_of types, you are using union_of`.

```suggestion
    /**
     * Only return snapshots with a state found in the given comma-separated list of snapshot states.
     * The default is all snapshot states.
     * @availability stack since=9.1.0
     * @availability serverless
     */
    state?: SnapshotState | SnapshotState[]
```

With this change, validation passes for `snapshot.get`, as you will see in the GitHub comment.

JeremyDahlgren: Thanks for the details here, I updated the entry per your suggestion.

pquentin (approved):
Thanks! LGTM.

PR #4488, opened 2025-06-06
Add rerank task type specification

timgrein:
Adding `rerank` task type specification for the `elastic` provider

Closes https://github.com/elastic/elasticsearch-specification/pull/4609

PR #4609, opened 2025-06-18
Eis rerank task type

pquentin:
This is a trick to run CI on https://github.com/elastic/elasticsearch-specification/pull/4488.


pquentin:
I just realized that `EisTaskType` is unused. Not sure what the fix is here.

PR #4617, opened 2025-06-19
Update variants API guidelines

pquentin:


api-design-guidelines/data-modelling.md
```diff
@@ -169,7 +169,7 @@ The second is using **external variants** where the inner object is wrapped with
 }
 ```
 
-When choosing between these two possibilities **favor using external variants** as it removes the requirement to buffer key-value pairs until the internal variant property is found. Using external variants also improves traversability of the API (ie auto-complete) as properties can be anticipated without waiting for the discriminant property.
+Internal tagging is a common way to identify variants and should usually be preferred. However, it may have performance implications when used to deserialize large objects in strongly-typed languages. In that case, external tagging should be considered instead, as it removes the requirement to buffer key-value pairs until the internal variant property is found. Using external tagging also improves traversability of the API (ie auto-complete) as properties can be anticipated without waiting for the discriminant property.
```
Anaethelion, line 172:
```suggestion
Internal tagging is a common way to identify variants and should be preferred. External tagging could be considered, as it removes the requirement to buffer key-value pairs until the internal variant property is found. However external tagging forces the schema to be similar to what aggregations uses, also known as `typed_keys`, the field would then be a map where the key abides by the `name#type` layout.
```

swallez: As I said in our conversation, let's **not mention `typed_keys`** which is an ugly hack because tagging was initially forgotten there and they did not want to use internal tagging (for performance reasons?) or add an additional wrapping object (that would have broken compatibility). This approach should never be seen in any API.

A correct way of externally tagging aggregations would have been :

```json
"aggregations": {
  "agg-1": {
      "avg": {
         "value": 42
      }
   }
}
```

And instead of that we get:

```json
"aggregations": {
  "agg-1#avg": {
      "value": 42
   }
}
```

swallez (approved):
LGTM. Let's not talk about typed-keys ;-)

api-design-guidelines/data-modelling.md
Anaethelion: I'm providing feedback on API guidelines that are still within the realm of the specification. I agree its a hack, still it exists. We shouldn't put it under the rug.

I intended to explain the implications of using what **we support** as external_tag.

Just saying `external tagging should be considered instead` doesn't feel enough.

Anaethelion: This discussion highlighted some confusion around how the specification deals with and identify `typed_keys` vs the true external tagging.

We've come to the agreement that the guidelines are fine as is and the specification is going to me modified to highlight the difference!

Anaethelion (approved):

api-design-guidelines/data-modelling.md
flobernd, line 172:
```suggestion
Internal tagging is a common way to identify variants and should usually be preferred. However, it may have performance implications when used to deserialize large objects in strongly-typed languages and the discriminator property is not in the beginning of the object. In that case, external tagging should be considered instead, as it removes the requirement to buffer key-value pairs until the internal variant property is found. Using external tagging also improves traversability of the API (ie auto-complete) as properties can be anticipated without waiting for the discriminant property.
```

PR #4619, opened 2025-06-19
[compiler] Rename external to typed_keys_quirk

Anaethelion:
This PR changes the variant name `external` to `typed_keys_quirk` to better highlight its specific use.

The real external tagging being `container` resolving types by the field name and able to be non-exhaustive.

The internal naming `external_tag` remains intact for backward compatibility with existing generators.

### DIFF
--- a/docs/modeling-guide.md
+++ b/docs/modeling-guide.md
@@ -334,6 +334,8 @@ An example of internal variants are the type mapping properties.
 
 #### typed_keys_quirk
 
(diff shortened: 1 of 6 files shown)


docs/modeling-guide.md
```diff
@@ -332,7 +332,7 @@ type FooOrBar = Foo | Bar
 
 An example of internal variants are the type mapping properties.
 
-#### External
+#### typed_keys_quirk
```
swallez, line 335: Let's make it very clear that it should not be used for new APIs

```suggestion
#### typed_keys_quirk

**Note**: this feature exists because of some early Elasticsearch APIs where tagging was forgotten, and added after the fact using this quirk to avoid breaking compatibility. **It should not be used for new APIs.**
```

PR #4632, opened 2025-06-19
Fixing cluster stats response

l-trotta:
From 0/27 validated responses to 27/27. Providing server side proof for each of these changes would be a lot, most of these changes are adding the `in_bytes` (or not in_bytes) alternative for fields and making both optional.
New classes:
- `GlobalOrdinalsStats` -> [server code](https://github.com/elastic/elasticsearch/blob/0dba16384ef51064b079987ba45887d8437f7f2c/server/src/main/java/org/elasticsearch/index/fielddata/FieldDataStats.java#L136)
- `ClusterSnapshotStats` -> [server code](https://github.com/elastic/elasticsearch/blob/0dba16384ef51064b079987ba45887d8437f7f2c/server/src/main/java/org/elasticsearch/cluster/ClusterSnapshotStats.java)
- `SearchUsageStats` -> [server code](https://github.com/elastic/elasticsearch/blob/0dba16384ef51064b079987ba45887d8437f7f2c/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/SearchUsageStats.java)
- `DenseVectorStats` -> [server code](https://github.com/elastic/elasticsearch/blob/0dba16384ef51064b079987ba45887d8437f7f2c/server/src/main/java/org/elasticsearch/index/shard/DenseVectorStats.java#L31)
- `SparseVectorStats` -> [server code](https://github.com/elastic/elasticsearch/blob/0dba16384ef51064b079987ba45887d8437f7f2c/server/src/main/java/org/elasticsearch/index/shard/SparseVectorStats.java#L24)

Since it was broken before I think this can all be backported safely

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -215,6 +215,7 @@ export type ExpandWildcards = ExpandWildcard | ExpandWildcard[]
 
 /**
  * Health status of the cluster, based on the state of its primary and replica shards.
+ * @non_exhaustive
  */
 export enum HealthStatus {
   // ES will send this enum as upper or lowercase depending on the APIs
--- a/specification/cluster/stats/types.ts
+++ b/specification/cluster/stats/types.ts
@@ -28,7 +28,8 @@ import {
   SegmentsStats,
   StoreStats
 } from '@_types/Stats'
-import { Duration, DurationValue, UnitMillis } from '@_types/Time'
+import { DateFormat, Duration, DurationValue, UnitMillis } from '@_types/Time'
+import { IndexingPressureMemory } from '@nodes/_types/Stats'
 import { Dictionary } from '@spec_utils/Dictionary'
 
 export class ClusterFileSystem {
@@ -37,15 +38,29 @@ export class ClusterFileSystem {
    * Depending on operating system or process-level restrictions, this number may be less than `nodes.fs.free_in_byes`.
    * This is the actual amount of free disk space the selected Elasticsearch nodes can use.
    */
-  available_in_bytes: long
+  available_in_bytes?: long
+  /**
+   * Total number of bytes available to JVM in file stores across all selected nodes.
+   * Depending on operating system or process-level restrictions, this number may be less than `nodes.fs.free_in_byes`.
+   * This is the actual amount of free disk space the selected Elasticsearch nodes can use.
+   */
+  available?: ByteSize
+  /**
+   * Total number, in bytes, of unallocated bytes in file stores across all selected nodes.
+   */
+  free_in_bytes?: long
   /**
    * Total number of unallocated bytes in file stores across all selected nodes.
    */
-  free_in_bytes: long
+  free?: ByteSize
   /**
    * Total size, in bytes, of all file stores across all selected nodes.
    */
-  total_in_bytes: long
+  total_in_bytes?: long
+  /**
+   * Total size of all file stores across all selected nodes.
+   */
+  total?: ByteSize
 }
 
 export class ClusterIndicesShardsIndex {
@@ -89,6 +104,12 @@ export class ClusterIndices {
   fielddata: FielddataStats
   /** Contains statistics about the query cache of selected nodes. */
   query_cache: QueryCacheStats
+  /**
+   * Holds a snapshot of the search usage statistics.
+   * Used to hold the stats for a single node that's part of a ClusterStatsNodeResponse, as well as to
+   * accumulate stats for the entire cluster and return them as part of the ClusterStatsResponse.
+   */
+  search: SearchUsageStats
   /** Contains statistics about segments in selected nodes. */
   segments: SegmentsStats
   /** Contains statistics about indices with shards assigned to selected nodes. */
@@ -104,6 +125,44 @@ export class ClusterIndices {
    * @doc_id analyzer-anatomy
    */
   versions?: IndicesVersions[]
+  /**
+   * Contains statistics about indexed dense vector
+   */
+  dense_vector?: DenseVectorStats
+  /**
+   * Contains statistics about indexed sparse vector
+   */
+  sparse_vector?: SparseVectorStats
+}
+
+export class SearchUsageStats {
+  total: long
+  queries: Dictionary<Name, long>
+  rescorers: Dictionary<Name, long>
+  sections: Dictionary<Name, long>
+  retrievers: Dictionary<Name, long>
+}
+
+export class DenseVectorStats {
+  value_count: long
+  off_heap?: DenseVectorOffHeapStats
+}
+
+export class SparseVectorStats {
+  value_count: long
+}
+
+export class DenseVectorOffHeapStats {
+  total_size_bytes?: long
+  total_size?: ByteSize
+  total_veb_size_bytes?: long
+  total_veb_size?: ByteSize
+  total_vec_size_bytes?: long
+  total_vec_size?: ByteSize
+  total_veq_size_bytes?: long
+  total_veq_size?: ByteSize
+  total_vex_size_bytes?: long
+  total_vex_size?: ByteSize
 }
 
 export class FieldTypesMappings {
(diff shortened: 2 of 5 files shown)


l-trotta:
nodes.stats will also be fixed, but this one and [the other one](https://github.com/elastic/elasticsearch-specification/pull/4636) need to be merged first

specification/_types/common.ts
```diff
@@ -215,6 +215,7 @@ export type ExpandWildcards = ExpandWildcard | ExpandWildcard[]
 
 /**
  * Health status of the cluster, based on the state of its primary and replica shards.
+ * @non_exhaustive
```
pquentin: If I remove this, I get two failures where the health was "unavailable", both in CCS tests (because we don't support CCS with the current test runner). Should we just add `unavailable` to the list and remove the `non_exhaustive` annotation?

I'm also not sure about the backwards-compatibility implications.

l-trotta:
> In `SegmentStats`, I'm seeing (existing) fields that I would not expect:
> 
>     * index_writer_max_memory_in_bytes?: long
> 
>     * stored_memory?: ByteSize
nice catch, removing those 
> 
> In `StatsResponseBase`:
> 
>     * `status` appears to be optional
> 
>     * should we add [ESQL metrics](https://github.com/elastic/elasticsearch/blob/c94bd1cf248bff8e37ea7dc652d19eacfe8fef05/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/ClusterStatsResponse.java#L162-L166)?

status is optional yes, while `_esql` we already have, it's in `ccs`
> 
> In `ClusterFileSystem`, there are [other optional fields](https://github.com/elastic/elasticsearch/blob/c94bd1cf248bff8e37ea7dc652d19eacfe8fef05/server/src/main/java/org/elasticsearch/monitor/fs/FsInfo.java#L172-L222)

adding them, thank you!

specification/_types/common.ts
l-trotta: the reason I went for `@non_exhaustive` is that there are two HealthStatus enums in the server code:

- [ClusterHealthStatus](https://github.com/elastic/elasticsearch/blob/c94bd1cf248bff8e37ea7dc652d19eacfe8fef05/server/src/main/java/org/elasticsearch/cluster/health/ClusterHealthStatus.java): in theory can be only red, green and yellow, but yaml tests show that it can also be "unavailable", mapped with [HealthStatus](https://github.com/elastic/elasticsearch-specification/blob/bd228bbfc7c81a6e8a19b61eff0e517a7278456e/specification/_types/common.ts#L219) in the spec
- [HealthStatus](https://github.com/elastic/elasticsearch/blob/c94bd1cf248bff8e37ea7dc652d19eacfe8fef05/server/src/main/java/org/elasticsearch/health/HealthStatus.java): can be red, green, yellow or unknown, mapped with [IndicatorHealthStatus](https://github.com/elastic/elasticsearch-specification/blob/bd228bbfc7c81a6e8a19b61eff0e517a7278456e/specification/_global/health_report/types.ts#L25) in the spec

In the spec we mostly use HealthStatus while the second one is mapped with the IndicatorHealthStatus enum, which is only used in the `health_report` endpoint. Since I didn't want to introduce breaking changes by renaming  IndicatorHealthStatus and using it to replace HealthStatus when needed, I was thinking of just using `@non_exhaustive` since I'm not even sure where `unavailable` comes from, but I agree it's not 100% correct. 

Should we introduce the breaking changes for 9.1 to fix every variation of health status?

l-trotta: after an internal discussion we decided to avoid adding an unnecessary `@non_exhaustive` and just add `unavailable` and `unknown` to both, will update the PR

pquentin: We discussed this offline. Given that:

* going for two `HealthStatus` enums is breaking and
* adding `@non_exhaustive` is breaking in .NET (even though there are some ways to make it not that breaking) and
* we think it's possible to see `unavailable` in practice

We decided to add `unvailable` to the relevant enum instead.

specification/cluster/stats/types.ts
```diff
@@ -104,6 +136,44 @@ export class ClusterIndices {
...
    */
   versions?: IndicesVersions[]
+  /**
+   * Contains statistics about indexed dense vector
+   */
+  dense_vector?: DenseVectorStats
+  /**
+   * Contains statistics about indexed sparse vector
+   */
+  sparse_vector?: SparseVectorStats
```
pquentin: Those stats are not optional

specification/cluster/stats/types.ts
```diff
@@ -104,6 +136,44 @@ export class ClusterIndices {
...
+  total_size?: ByteSize
+  total_veb_size_bytes?: long
+  total_veb_size?: ByteSize
+  total_vec_size_bytes?: long
+  total_vec_size?: ByteSize
+  total_veq_size_bytes?: long
+  total_veq_size?: ByteSize
+  total_vex_size_bytes?: long
+  total_vex_size?: ByteSize
 }
```
pquentin: Thanks! I'm assuming you haven't seen https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/index/shard/DenseVectorStats.java#L140-L142 do anything in practice?

specification/nodes/_types/Stats.ts
```diff
@@ -196,6 +196,10 @@ export interface PressureMemory {
    * Number of indexing requests rejected in the replica stage.
    */
   replica_rejections?: long
+  primary_document_rejections?: long
+  limit?: ByteSize
+  limit_in_bytes?: long
```
pquentin: There's an `endObject()` before `limit_in_bytes`, suggesting it's not at the same level: https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/index/stats/IndexingPressureStats.java#L342-L343

pquentin (commented):
In `ClusterIndices`, [mappings is optional](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/ClusterStatsIndices.java#L173-L175), as well as [analysis](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/ClusterStatsIndices.java#L176-L178)

In FieldTypesMappings, [total_field_count](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/MappingStats.java#L346-L347) and [total_deduplicated_field_count](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/MappingStats.java#L349-L349) are long.

In FieldTypes, [indexed_vector_count](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/server/src/main/java/org/elasticsearch/action/admin/cluster/stats/DenseVectorFieldStats.java#L50-L50) is a long. However, should we call it `DenseVectorFieldStats`? `FieldStats` is a different child class in the Java code.


All fields built with `humanReadableField` are marked optional, but I don't understand why when [reading the code](https://github.com/elastic/elasticsearch/blob/c8e719544a9228c6b87a0d23100d478411101656/libs/x-content/src/main/java/org/elasticsearch/xcontent/XContentBuilder.java#L1160-L1174).

specification/cluster/stats/types.ts
l-trotta: ah thanks I forgot about this one! I'm not sure how to map it since it's all custom field values, let's discuss it with the team

l-trotta: added it!

l-trotta:
@pquentin thanks for the review! I think I addressed everything

l-trotta:
nodes.stats is next

pquentin (approved):
Thanks! LGTM.

PR #4636, opened 2025-06-20
Fixing nodes info response

l-trotta:
From 1/112 validated responses to 112/112. Details:
- Missing fields in `NodeInfo` -> [server code
](https://github.com/elastic/elasticsearch/blob/b2acda9a9fe150cb643ff07da93d15d00a9c58a3/server/src/main/java/org/elasticsearch/action/admin/cluster/node/info/NodesInfoResponse.java#L67)
- `transport.ignore_deserialization_errors` is said to be only used in unit tests, so I'm setting it as private -> [server code](https://github.com/elastic/elasticsearch/blob/b2acda9a9fe150cb643ff07da93d15d00a9c58a3/server/src/main/java/org/elasticsearch/transport/TcpTransport.java#L104)
- In `NodeJvmInfo`, `bundled_jdk` is not actually there, there's only `using_bundled_jdk` -> [server code](https://github.com/elastic/elasticsearch/blob/b2acda9a9fe150cb643ff07da93d15d00a9c58a3/server/src/main/java/org/elasticsearch/monitor/jvm/JvmInfo.java#L408)


specification/nodes/info/types.ts
```diff
@@ -59,6 +62,8 @@ export class NodeInfo {
   transport?: NodeInfoTransport
   /** Host and port where transport HTTP connections are accepted. */
   transport_address: TransportAddress
+  transport_version?: VersionNumber
```
pquentin: This one looks mandatory to me: https://github.com/elastic/elasticsearch/blob/b2acda9a9fe150cb643ff07da93d15d00a9c58a3/server/src/main/java/org/elasticsearch/action/admin/cluster/node/info/NodesInfoResponse.java#L65. Also, the validation still passes with that change.

```suggestion
  transport_version: VersionNumber
```

(First appeared in 8.9 or before)

l-trotta:
> In `NodesInfo`:
> 
>     * It looks like `remote_cluster_server` is missing? See [NodesInfoResponse.java](https://github.com/elastic/elasticsearch/blob/b2acda9a9fe150cb643ff07da93d15d00a9c58a3/server/src/main/java/org/elasticsearch/action/admin/cluster/node/info/NodesInfoResponse.java#L118C34-L120) and [RemoteClusterServerInfo.java](https://github.com/elastic/elasticsearch/blob/5b9b0339ed809787b88e66d1062b6e2e167715d4/server/src/main/java/org/elasticsearch/transport/RemoteClusterServerInfo.java#L34-L47).
> 
>     * `network` was removed in 2.0: [Remove network stats & info elasticsearch#12054](https://github.com/elastic/elasticsearch/pull/12054)

added and removed, thank you! it's likely that fields like `remote_cluster_server` are missing from other responses like nodes.cluster, I'll merge this first and then go on with fixing other apis.

PR #4638, opened 2025-06-20
Add 8.18 documentation URLs to schema and OpenAPI

### LINKED ISSUE #4369: Add mapping to old docs
## 🚀 Feature Proposal

Add at least another column to the [table.csv](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv) that enables us to link back to the old documentation (i.e. for versions that were generated from Asciidoc). For example, if an API existed in the old V7 and V8 docs, there might be two additional columns of URLs. If an API was introduced in V9, there would be no entries in those columns.

## Motivation

There is a request to be able to navigate to the old documentation from the new OpenAPI-based documentation.

## Example

For each specification that has a `@doc_id` entry (ideally all of them), if there are links to these new columns in the `table.csv`, output it in an extension that we can use to show those links in Bump.sh

miguelgrinberg:
Fixes #4369 

The table.csv file now has a third column with the 8.18 documentation URL, if available. The mapping that was used comes from [this spreadsheet](https://docs.google.com/spreadsheets/d/10PF_1GJkMK-EdnV1FitiBh5R-8PAqii9chGsokzmY8E/edit?gid=0#gid=0).

These 8.18 docs URLs are imported in the new `extLegacyDocUrl` field of the schema. The OpenAPI generator then uses it to generate a `x-legacyUrl` extension inside the `Documentation` object. The name `x-legacyUrl` is made up by me, we can change it if there is a better suggestion, and in general, if there is a better word than `legacy` to use for this information let me know and I'll replace it.


swallez (approved):
LGTM. Left a nitpick.

Issue #4599, opened 2025-06-17
[validation] Check if @shortcut and @aliases are supported correctly... together


l-trotta:
Here's SpanTermQuery, where `value` is both a shortcut and has an alias (yes it's correct):

```
/**
 * @shortcut_property value
 * @ext_doc_id query-dsl-span-term-query
 */
export class SpanTermQuery extends QueryBase {
  /** @aliases term */
  value: FieldValue
}
```

All flight recorder tests fail with this one, saying: 
```
Type '{ _name: string; boost: number; value: string; }' is not assignable to type 'FieldValue | QueryDslSpanTermQuery | undefined'.
```
and
```
Property 'term' is missing in type '{ _name: string; boost: number; value: string; }' but required in type 'QueryDslSpanTermQuery'.
```

So it seems like the combination of shortcut and aliases breaks both

pquentin:
The second error applies to all aliased mandatory properties. They're just added to the class and of course they can't all be mandatory.

https://github.com/elastic/elasticsearch-specification/blob/cdaa4ef3c12e1cfe6a16a607951a0311ed8ba93a/typescript-generator/src/index.ts#L203-L209

pquentin:
We discussed this offline. There's a way to handle aliases in TypeScript: https://stackoverflow.com/questions/44425344/typescript-interface-with-xor-barstring-xor-cannumber. But it can quickly get messy when there are multiple aliases (either on multiple fields, or for fields with multiple aliases). It could actually impair debugging.

PR #4652, opened 2025-06-23
Fixing indices create partially

l-trotta:
The missing validated tests are either:
- wrong tests IMO, will provide a fix in the flight recorder
- new features ([bbq_ivf is still very experimental](https://github.com/elastic/elasticsearch/pull/128631), since even the name could change I don't think it should be added yet)
- validation bugs

Details:
- `IcuCollationTokenFilter`'s fields are actually snakeCase and not camel_case -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/plugins/analysis-icu/src/main/java/org/elasticsearch/plugin/analysis/icu/IcuCollationTokenFilterFactory.java#L136)
- `HunspellTokenFilter`: `locale` has two aliases, and if it's set as mandatory as it was before, this means that the generator will generate 3 mandatory fields. For now I'm setting this as optional, but maybe we should support this as a feature?
- `PatternReplaceTokenFilter` was missing the `flags` field -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/modules/analysis-common/src/main/java/org/elasticsearch/analysis/common/PatternReplaceTokenFilterFactory.java#L36)
- In `EdgeNGramTokenizer`, token_chars can also be expressed as a csv, so making it a `string | enum` union
- `RankVectorProperty` was missing -> [server code](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/rank-vectors/src/main/java/org/elasticsearch/xpack/rank/vectors/mapper/RankVectorsFieldMapper.java)
- `FlattenedProperty` was missing `time_series_dimensions` -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/server/src/main/java/org/elasticsearch/index/mapper/flattened/FlattenedFieldMapper.java#L117)
- `AggregateMetricDoubleProperty` was missing `ignore_malformed` -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateMetricDoubleFieldMapper.java#L107)
- `GeoPointProperty` was missing `time_series_metric` -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/server/src/main/java/org/elasticsearch/index/mapper/GeoPointFieldMapper.java#L137)
- `Storage` in `IndexSettings` was missing `stats_refresh_interval`, but I also noticed that in the code and yaml tests there was no trace of the `type` field, nor the `StorageType` enum, so maybe we can remove that. -> [server code](https://github.com/elastic/elasticsearch/blob/ded666ec664dbc0b5cc64b95a52ec919b4e38cae/server/src/main/java/org/elasticsearch/index/store/Store.java#L142)


specification/_types/analysis/token_filters.ts
```diff
@@ -274,7 +274,7 @@ export class HunspellTokenFilter extends TokenFilterBase {
   dictionary?: string
   /** Locale directory used to specify the `.aff` and `.dic` files for a Hunspell dictionary.
    * @aliases lang, language */
-  locale: string
+  locale?: string
```
pquentin: > HunspellTokenFilter: locale has two aliases, and if it's set as mandatory as it was before, this means that the generator will generate 3 mandatory fields. For now I'm setting this as optional, but maybe we should support this as a feature?

Which generator? I think we should discuss this before making it wrongly optional

specification/_types/analysis/token_filters.ts
```diff
@@ -486,6 +487,14 @@ export class FlattenGraphTokenFilter extends TokenFilterBase {
   type: 'flatten_graph'
 }
 
+export class BengaliNormalizationTokenFilter extends TokenFilterBase {
+  type: 'bengali_normalization'
+}
+
+export class BrazilianNormalizationTokenFilter extends TokenFilterBase {
+  type: 'brazilian_stem'
+}
```
pquentin:
```suggestion
export class BrazilianStemTokenFilter extends TokenFilterBase {
  type: 'brazilian_stem'
}
```

specification/_types/mapping/core.ts
```diff
@@ -213,7 +213,14 @@ export class RankFeaturesProperty extends PropertyBase {
   type: 'rank_features'
 }
 
+export class RankVectorProperty extends PropertyBase {
+  type: 'rank_vectors'
+  element_type?: RankVectorElementType
+  dims?: integer
+}
+
```
pquentin, line 225: This is a technical preview feature, how can we specify it?

specification/indices/_types/IndexSettings.ts
```diff
@@ -534,44 +534,15 @@ export class SlowlogTresholdLevels {
 }
 
 export class Storage {
-  type: StorageType
```
pquentin: I can see it here: https://github.com/elastic/elasticsearch/blob/1ad6dbc5991b77ce1b05d5015e1f5ac758e1fcb4/server/src/main/java/org/elasticsearch/index/IndexModule.java#L427-L432

pquentin (commented):
Great job, thank you! In `TokenFilterDefinition`, `persian_stem` is missing. I think it's a newish stemmer.

specification/_types/analysis/token_filters.ts
l-trotta: ah sorry wrong wording, I mean the ts compiler. if you look at https://github.com/elastic/elasticsearch-specification/blob/cdaa4ef3c12e1cfe6a16a607951a0311ed8ba93a/output/typescript/types.ts#L4804 you'll see the generated output

specification/_types/mapping/core.ts
l-trotta: could this work? since type is the internal tag  
```
export class RankVectorProperty extends PropertyBase {
  /**
   * @availability stack since=8.18.0 stability=experimental
   */
  type: 'rank_vectors'
  element_type?: RankVectorElementType
  dims?: integer
}
```

pquentin: Actually, I don't think that is possible today, see https://github.com/elastic/elasticsearch-specification/issues/2830. Leaving this for future work then.

pquentin: Can we add a comment at least?

specification/_types/analysis/token_filters.ts
pquentin: Right `aliases` seem to be a recipe for problem, as their generation is pretty naive. I added a comment to https://github.com/elastic/elasticsearch-specification/issues/4599#issuecomment-3012973813 to mention it. In the meantime, making this optional just because we can't handle it still seems wrong.

l-trotta: okay, reverting then

pquentin (approved):
Thanks! LGTM.

PR #4684, opened 2025-06-25
Add index block removal API

nielsbauman:
Follow-up of https://github.com/elastic/elasticsearch/pull/129128


specification/indices/add_block/IndicesAddBlockRequest.ts
```diff
@@ -49,7 +50,7 @@ export interface Request extends RequestBase {
     /**
      * The block type to add to the index.
      */
-    block: IndicesBlockOptions
+    block: keyof IndexSettingBlocks
```
nielsbauman: I'm not optimistic that this will work since I couldn't find any other occurrences, but I figured I'd give it a try anyway.

pquentin: Indeed, that's not supported - we need an explicit type for the clients to rely on. Additionally, removing the enum is breaking in `indices.add_block`, so we'll have to add `read_only_allow_delete` in here.

I'll ask the rest of the team how we think about moving *that* to `_types`.

pquentin: Confirmed: we need to duplicate the enum for now. Please add a comment above both of them to mention that we should keep them in sync. Not ideal, but better than a breaking change.

nielsbauman: I just realized `read_only_allow_delete` is for internal use only, so we don't want to include it here. I've made some updates to the enum. Let me know what you think.

pquentin: Thanks! After fixing lint and adding the doc id to the table, I encountered a new problem with my suggested approach: we had two `IndicesBlockOptions` and that's also a source of issues in some clients. (This is also the reason why I had to rename `IndicesBlockStatus`.)

I discussed with other client maintainers and we decided to actually do the breaking change by moving the type `indices/_types`.

nielsbauman: Thank you for the fixes!

specification/indices/add_block/IndicesAddBlockResponse.ts
```diff
@@ -23,11 +23,11 @@ export class Response {
   body: {
     acknowledged: boolean
     shards_acknowledged: boolean
-    indices: IndicesBlockStatus[]
+    indices: AddIndicesBlockStatus[]
   }
 }
 
-export class IndicesBlockStatus {
+export class AddIndicesBlockStatus {
```
flobernd, line 30: Thinking about future API extensions: If there is the chance for a `get_index_block_status` endpoint in the future, it would probably make sense to have this type in the `_types` namespace.

nielsbauman: I don't think that's super relevant, as an API like that will always return which type of block exists on the index, rather than just a boolean.

specification/indices/remove_block/examples/response/IndicesAddBlockResponseExample1.yaml
```diff
@@ -0,0 +1,12 @@
+# summary: ''
```
l-trotta: this file should be named IndicesRemoveBlockResponseExample1.yaml

specification/indices/remove_block/IndicesRemoveBlockResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class Response {
+  body: {
+    acknowledged: boolean
+    indices: RemoveIndicesBlockStatus[]
+  }
+}
+
+export class RemoveIndicesBlockStatus {
+  name: IndexName
```
l-trotta, line 31: looking at the [server code](https://github.com/elastic/elasticsearch/blob/5df3a06d49e3c484cde41f9ea2ef3aaf40cdc641/server/src/main/java/org/elasticsearch/action/admin/indices/readonly/RemoveIndexBlockResponse.java#L116) for this it seems like it's missing the optional `exception` field, also `unblocked` should be optional

specification/indices/remove_block/IndicesRemoveBlockRequest.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Remove an index block.
+ *
+ * Remove an index block from an index.
+ * Index blocks limit the operations allowed on an index by blocking specific operation types.
+ * @rest_spec_name indices.remove_block
+ * @availability stack since=9.1.0 stability=stable
+ * @availability serverless stability=stable visibility=public
+ * @doc_id index-block-remove
```
georgewallace, line 33: Are there any [index_privileges](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#index_privileges) or [cluster_privileges](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#cluster_privileges) we need to specify? these will show up in the docs.

specification/indices/remove_block/examples/response/IndicesAddBlockResponseExample1.yaml
nielsbauman: Good catch, thanks!

specification/indices/remove_block/IndicesRemoveBlockRequest.ts
nielsbauman: Good question, I'm not familiar with privileges myself, so I'll ask around.

nielsbauman: I added the `manage` index privilege. We'll have to add it to `IndicesAddBlockRequest` as well, but I'll do that in a follow-up PR.

nielsbauman:
@pquentin I'm not sure how to resolve the conflicts with `main`. Could you take care of that, perhaps?

PR #4721, opened 2025-06-30
[8.19] Update Inference specification for Watsonx's completion and chat completion tasks

Evgenii-Kazannik:
Backport https://github.com/elastic/elasticsearch-specification/commit/fe2f14c719f43e43c0173891f97b6c1d7d3119a1
from https://github.com/elastic/elasticsearch-specification/pull/4505


Evgenii-Kazannik:
Hi Quentin,
Since we moved to version 9.2 we do not have an equivalent 8.x version thus as it seems we don't need to backport
I'll close this PR.
Thank you
@pquentin

Issue #4628, opened 2025-06-19
[OpenAPI] URL resolution is always current instead of branch


lcawl:
## 🐛 Bug

The usage of the `table.csv` file to resolve URLs for `@doc_id` and `@ext_doc_id` was implemented in https://github.com/elastic/elasticsearch-specification/issues/714 and https://github.com/elastic/elasticsearch-specification/issues/2748.

The table supports the use of `{branch}` as a variable in the URL, for example:

```
data-stream-lifecycle,https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/data-stream-lifecycle.html
```

However, it seems like that's being resolved to `current` every time, instead of the appropriate branch. For example, check out https://github.com/elastic/elasticsearch-specification/pull/4627 which ought to be linking to https://www.elastic.co/guide/en/elasticsearch/reference/8.18/data-stream-lifecycle.html but the OpenAPI document contains:

```
        "summary": "Get data stream lifecycles",
        "description": "Get the data stream lifecycle configuration of one or more data streams.",
        "externalDocs": {
          "url": "https://www.elastic.co/guide/en/elasticsearch/reference/current/data-stream-lifecycle.html"
        },
```

Current is inappropriate in this case, since it'll redirect to the V9 docs and therefore lead our users to the wrong site.

## Solution

It seems like the problem is related to this code in [schema.rs](https://github.com/elastic/elasticsearch-specification/blob/main/compiler-rs/clients_schema_to_openapi/src/schemas.rs):

```
pub fn convert_external_docs(&self, obj: &impl clients_schema::ExternalDocument) -> Option<ExternalDocumentation> {
        // FIXME: does the model contain resolved doc_id?
        obj.ext_doc_url().map(|url| {
            let branch: &str = self
                .model
                .info
                .as_ref()
                .and_then(|i| i.version.as_deref())
                .unwrap_or("current");
            ExternalDocumentation {
                description: None,
                url: url.trim().replace("{branch}", branch),
                extensions: Default::default(),
            }
        })
    }
```

If I hack the `schema.json` file to explicitly add a `version` value, like this:

```
  "_info": {
    "version": "8.18",
...
  },
```

... then the URLs are correct in the OpenAPI output when I run the `make transform-to-openapi` command:

```
        "summary": "Get data stream lifecycles",
        "description": "Get the data stream lifecycle configuration of one or more data streams.",
        "externalDocs": {
[...]

PR #4722, opened 2025-06-30
[OpenApi] Add branch option to cli

l-trotta:
Closes https://github.com/elastic/elasticsearch-specification/issues/4628.
This doesn't affect main or 9.x, it's only an old docs system issue, but I'll port this everywhere for consistency.
In the old docs the branch name in `ext_docs_url` (`externalDocs` in openapi)  is a `{branch}` placeholder that gets resolved in the rust openapi converter by getting the correct branch value from the schema model. The problem is that the branch/version doesn't exist in the schema model anymore, because it was causing problems with PRs, so right now it always defaults to "current".

This PR gets the version for `externalDocs` from a CLI argument instead. This is a temporary solution, since the whole logic will get rewritten when the repo is transfered. 

So no changes for branches main, 9.1, 9.0, while for 8.18 and 8.19 it will be possible to run:
```
npm run transform-to-openapi -- --schema output/schema/schema.json --flavor stack --output output/openapi/elasticsearch-openapi.json --branch 8.19
```

and fix the url issue. 

Output preview for 8.19: https://github.com/elastic/elasticsearch-specification/blob/backport-4722-to-8.19/output/openapi/elasticsearch-openapi.json

### DIFF
--- a/compiler-rs/clients_schema/src/lib.rs
+++ b/compiler-rs/clients_schema/src/lib.rs
@@ -1003,7 +1003,6 @@ pub struct UrlTemplate {
 pub struct ModelInfo {
     pub title: String,
     pub license: License,
-    pub version: Option<String>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


lcawl (approved):
This sounds great, thanks! Once it's merged, I'll create a separate PR to add that option to the makefile for the 8.19 and 8.18 branches and update our release processes in case they need to be updated (e.g. right now the "8.19" branch has to point to 8.x docs but that will change once we have live 8.19 docs)

miguelgrinberg (approved):

PR #4731, opened 2025-07-01
 Add new search load metrics to SearchStats

drempapis:
This adds the `recent_search_load` field, as introduced in https://github.com/elastic/elasticsearch/pull/128660, to the spec.

Closes https://github.com/elastic/elasticsearch-specification/pull/4739


pquentin (approved):
Thank you! This is timely as we're working on fixing the node stats API in https://github.com/elastic/elasticsearch-specification/pull/4632 but did not catch this one. Next time, please open a PR from a branch, this is why the CI was failing. I just added that to instructions here: https://github.com/elastic/elasticsearch-specification/pull/4738

drempapis:
Thank you @pquentin

PR #4751, opened 2025-07-01
Adding specification for the new data stream mappings API

masseyke:
This adds specifications for the new data stream mappings APIs.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -8581,7 +8581,7 @@
               },
               "examples": {
                 "IndicesPutDataStreamMappingsRequestExample1": {
-                  "summary": "Change a data stream setting",
+                  "summary": "Change a data stream mapping",
                   "description": "This is a request to add or modify two fields in a mapping on a data stream.\n",
                   "value": "{\n   \"properties\":{\n      \"field1\":{\n         \"type\":\"ip\"\n      },\n      \"field3\":{\n         \"type\":\"text\"\n      }\n   }\n}"
                 }
@@ -52644,9 +52644,7 @@
         },
         "required": [
           "name",
-          "applied_to_data_stream",
-          "mappings",
-          "effective_mappings"
+          "applied_to_data_stream"
         ]
       },
       "indices.put_data_stream_settings.UpdatedDataStreamSettings": {
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -12813,8 +12813,8 @@ export interface IndicesPutDataStreamMappingsUpdatedDataStreamMappings {
   name: IndexName
   applied_to_data_stream: boolean
   error?: string
-  mappings: MappingTypeMapping
-  effective_mappings: MappingTypeMapping
+  mappings?: MappingTypeMapping
+  effective_mappings?: MappingTypeMapping
 }
 
 export interface IndicesPutDataStreamOptionsRequest extends RequestBase {


specification/indices/get_data_stream_mappings/examples/request/IndicesGetDataStreamMappingsRequestExample1.yaml
```diff
@@ -0,0 +1,24 @@
...
+      response = client.indices.get_data_stream_mappings(
+        name: "my-data-stream"
+      )
+  - language: PHP
+    code: |-
+      $resp = $client->indices()->getDataStreamSettings([
+          "name" => "my-data-stream",
+      ]);
+  - language: curl
+    code: 'curl -X GET -H "Authorization: ApiKey $ELASTIC_API_KEY" "$ELASTICSEARCH_URL/_data_stream/my-data-stream/_mappings"'
```
pquentin: All the alternatives are generated. However a summary (and description if needed) would help.

specification/indices/put_data_stream_mappings/examples/request/IndicesPutDataStreamMappingsRequestExample1.yaml
```diff
@@ -0,0 +1,82 @@
...
+                    "type" => "text"
+                 ]
+              ]
+          ],
+      ]);
+  - language: curl
+    code:
+      'curl -X PUT -H "Authorization: ApiKey $ELASTIC_API_KEY" -H "Content-Type: application/json" -d
+      ''{"index.lifecycle.name":"new-test-policy","index.number_of_shards":11}''
+      "$ELASTICSEARCH_URL/_data_stream/my-data-stream/_mappings"'
```
pquentin: Sorry that you took the time to do this, but it's generated.

specification/indices/get_data_stream_mappings/examples/200_response/IndicesGetDataStreamMappingsResponseExample1.yaml
```diff
@@ -0,0 +1,37 @@
+summary: Get data stream settings on a data stream
+description: >
+  This is a response to `GET /_data_stream/my-data-stream/_settings` where my-data-stream that has two settings set. The
+  `effective_settings` field shows additional settings that are pulled from its template.
+# type: response
+# response_code: 200
```
pquentin: I think this can be uncommented. Not too familiar with examples to be honest. (I do appreciate that you worked on them!)

```suggestion
type: response
response_code: 200
```

specification/indices/put_data_stream_mappings/IndicesPutDataStreamMappingsRequest.ts
```diff
@@ -0,0 +1,76 @@
...
+import { Duration } from '@_types/Time'
+
+/**
+ * Update data stream mappings.
+ *
+ * This API can be used to override mappings on specific data streams. These overrides will take precedence over what
+ * is specified in the template that the data stream matches. The mapping change is only applied to new write indices
+ * that are created during rollover after this API is called. No indices are changed by this API.
+ * @rest_spec_name indices.put_data_stream_mappings
+ * @availability stack stability=stable visibility=public
```
lcawl: I presume this is true here too:
```suggestion
 * @availability stack since=9.1.0 stability=stable visibility=public
```

specification/indices/put_data_stream_mappings/examples/request/IndicesPutDataStreamMappingsRequestExample1.yaml
masseyke: Ah, I had wondered if that was the case, but didn't see a way to generate them.

specification/indices/put_data_stream_mappings/IndicesPutDataStreamMappingsRequest.ts
masseyke: Actually, 9.2.0. But I'll add that. Thanks.

specification/indices/get_data_stream_mappings/examples/200_response/IndicesGetDataStreamMappingsResponseExample1.yaml
masseyke: Hmm I copied all of these from the examples I had done for the simulate ingest API. Those also have this commented out. I probably need to fix those as well. (And I had probably copied those from somewhere else that also needs to be fixed.)

specification/indices/put_data_stream_mappings/IndicesPutDataStreamMappingsResponse.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  applied_to_data_stream: boolean
+  /**
+   * A message explaining why the mappings could not be applied to the data stream.
+   */
+  error?: string
+  /**
+   * The mappings that are specfic to this data stream that will override any mappings from the matching index template.
+   */
+  mappings: TypeMapping
```
pquentin: Is `mappings` optional? Validation fails because there's no `mapping` field in this recording: https://github.com/elastic/clients-flight-recordings/blob/6b75a4d9630fe9d31fede2921e54e29f2944a34a/elasticsearch/indices.put_data_stream_mappings/088c21bc700471d61299699c6fe16612.json#L20-L25.

masseyke: Oh you are right. I'm not sure that was the wisest decision on my part, but it is optional. I've updated the spec.

PR #4774, opened 2025-07-03
add server default values to properties in generated OpenAPI

### LINKED ISSUE #3773: Default values missing from OpenAPI output
Priority: high
effort: low?

## Request

The OpenAPI generator should include default values in the output files.

NOTE: If there are circunstances where the default value might differ in the Serverless context, should the `@server_default` also be changed to be context-aware (e.g. similar to the `@availability` option)?

## Motivation

The specifications contain `@server_default` values, but they're not being propagated to the OpenAPI output.

In Kibana API documents, those default fields are present in the OpenAPI, for example:

```
      parameters:
        - description: The number of rules to return per page.
          in: query
          name: per_page
          required: false
          schema:
            default: 10
            minimum: 0
            type: number
```

... which then is displayed in our docs like this (https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-alerting-rules-find):

Image

## Example

For example, looking at https://github.com/elastic/elasticsearch-specification/blob/main/specification/esql/query/QueryRequest.ts, which has `@server_default` values set like this:

```
    /**
     * When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`
     * object with information about the clusters that participated in the search along with info such as shards
     * count.
     * @server_default false
     */
    include_ccs_metadata?: boolean
```

I can see that same information carried into the specification files, e.g. `serverDefault` appears for `include_ccs_metadata` in https://raw.githubusercontent.com/elastic/elasticsearch-specification/refs/heads/main/output/schema/schema-serverless.json

```
          {
            "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.",
            "name": "include_ccs_metadata",
            "required": false,
            "serverDefault": false,
            "type": {
              "kind": "instance_of",
              "type": {
                "name": "boolean",
                "namespace": "_builtins"
              }
            }
          }
        ]
```

However, it's missing from the OpenAPI output. For example:

```
 "include_ccs_metadata": {
                    "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.",
                    "type": "boolean"
                  }
```

It should appear in a `default` field, as described in https://spec.openapis.org/oas/v3.0.3#properties

miguelgrinberg:
Fixes #3773 

This change adds the `default` field to properties in the exported OpenAPI files, when there is a `serverDefault` for the property in schema.json.


lcawl (approved):
Output LGTM, thanks!

PR #4797, opened 2025-07-07
Adds missing required cluster privilege information

kosabogi:
This PR adds required cluster privilege information to the [Check index templates API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-exists-index-template). 

Related issue: https://github.com/elastic/docs-content/issues/2040


leemthompo (approved):
Looks legit to me

PR #4772, opened 2025-07-03
add a description field for external doc links

miguelgrinberg:
Fixes #4225 

This change adds a fourth column to `table.csv` where a description for an external documentation link can be given. This description is included in schema.json, and later transferred to the `externalDocs` object in the generated OpenAPI files.

Note that I have not added any descriptions, the 4th column in table.csv is empty for all rows in this PR.

PR #4841, opened 2025-07-08
[9.1] Adds new 'none' and 'recursive' chunking strategies to Inference APIs

kosabogi:
This PR adds the new 'none' and 'recursive' chunking strategies to Inference APIs.
Related issue: https://github.com/elastic/developer-docs-team/issues/308

### DIFF
--- a/specification/_doc_ids/table.csv
+++ b/specification/_doc_ids/table.csv
@@ -85,6 +85,7 @@ ccr-put-follow,https://www.elastic.co/docs/api/doc/elasticsearch/operation/opera
 ccr-resume-auto-follow-pattern,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ccr-resume-auto-follow-pattern,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/ccr-resume-auto-follow-pattern.html,
 ccs-network-delays,https://www.elastic.co/docs/solutions/search/cross-cluster-search#ccs-network-delays,,
 ccs-privileges,https://www.elastic.co/docs/deploy-manage/remote-clusters/remote-clusters-cert#remote-clusters-privileges-ccs,,
+chunking-strategies,https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#chunking-strategies,
 clean-up-snapshot-repo,https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore/self-managed#snapshots-repository-cleanup,,
 clear-repositories-metering-archive-api,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-nodes-clear-repositories-metering-archive,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/clear-repositories-metering-archive-api.html,
 clear-scroll-api,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-clear-scroll,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/clear-scroll-api.html,
--- a/specification/inference/_types/Services.ts
+++ b/specification/inference/_types/Services.ts
@@ -314,7 +314,7 @@ export class InferenceChunkingSettings {
    * - `max_chunk_size`
    * - either `separators` or`separator_group`
    *
-   * Learn more about different chunking strategies in the External documentation.
+   * Learn more about different chunking strategies in the linked documentation.
    * @server_default sentence
    * @ext_doc_id chunking-strategies
    */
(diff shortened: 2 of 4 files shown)


leemthompo (approved):
LGTM

Not sure if we could use an `@ext_doc_id` here? We do now have the ability to customize the link text so it doesn't just read _External documentation_.

kosabogi:
> LGTM
> 
> Not sure if we could use an `@ext_doc_id` here? We do now have the ability to customize the link text so it doesn't just read _External documentation_.

Yes, I just learned that we can add external doc links for parameters too. It's on my list to update this PR along with other chunking reference updates. Thanks so much for your review!

specification/_doc_ids/table.csv
```diff
@@ -85,6 +85,7 @@ ccr-put-follow,https://www.elastic.co/docs/api/doc/elasticsearch/operation/opera
 ccr-resume-auto-follow-pattern,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ccr-resume-auto-follow-pattern,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/ccr-resume-auto-follow-pattern.html,
 ccs-network-delays,https://www.elastic.co/docs/solutions/search/cross-cluster-search#ccs-network-delays,,
 ccs-privileges,https://www.elastic.co/docs/deploy-manage/remote-clusters/remote-clusters-cert#remote-clusters-privileges-ccs,,
+chunking-strategies,https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#chunking-strategies,
```
leemthompo, line 88: I think we need to add a `Description` field here to avoid the default link text?

`Chunking strategies` probably works

Sorry if you've already planned to do this :)

specification/inference/_types/Services.ts
```diff
@@ -288,8 +288,35 @@ export class InferenceChunkingSettings {
...
+  separators: string[]
+  /**
+   * The chunking strategy: `sentence`, `word`, `none` or `recursive`.
+   *
+   *  * If `strategy` is set to `recursive`, you must also specify:
+   *
+   * - `max_chunk_size`
+   * - either `separators` or`separator_group`
+   *
+   * Learn more about different chunking strategies in the External documentation.
```
leemthompo:
```suggestion
   * Learn more about different chunking strategies in the linked documentation.
```

Let's steer clear of "External documentation".

Or alternatively could just use more descriptive link text in the csv file per my [other comment](https://github.com/elastic/elasticsearch-specification/pull/4841#discussion_r2209769511) and we might not even need this prose?

If have time, maybe experiment with both approaches and see how they look.

specification/_doc_ids/table.csv
kosabogi: I tried this a few times, but it didn’t work, unfortunately. 

I added a description to the table.csv file like this:

`chunking-strategies,https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#chunking-strategies,https://www.elastic.co/guide/en/kibana/8.18/inference-endpoints.html,Chunking strategies documentation`

However, I didn’t see any results in the elasticsearch-openapi.json file, neither the previous URL nor the link description:

```
"strategy": {
  "externalDocs": {
    "url": "https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#chunking-strategies"
  },
  "description": "The chunking strategy: `sentence`, `word`, `none` or `recursive`.\n\n * If `strategy` is set to `recursive`, you must also specify:\n\n- `max_chunk_size`\n- either `separators` or `separator_group`\n\nLearn more about different chunking strategies in the External documentation.",
  "default": "sentence",
  "type": "string"
} 
```

To understand what’s going on, I tested the exact same case that @lcawl referenced here:
https://github.com/elastic/elasticsearch-specification/pull/4772#issuecomment-3033377294

t worked perfectly for me too. The OpenAPI file was updated exactly as Lisa described in her comment.

My only guess is that the example Lisa tested wasn't for a parameter-level `externalDocs` link, but rather for a general API description, which might be treated differently by the tooling?
Could that explain the difference? Any ideas? Or am I missing something here?

leemthompo: images

leemthompo: > My only guess is that the example Lisa tested wasn't for a parameter-level externalDocs link, but rather for a general API description, which might be treated differently by the tooling?
Could that explain the difference? Any ideas? Or am I missing something here?

That might be it, I think you can only have one ext-id per API too IIRC, don't know if there's a previous one defined for this API.

There's as many exceptions as there are rules I fear 😠

dan-rubinstein (approved):

specification/_doc_ids/table.csv
kosabogi: I also know it that way  that you can have only one `ext-id` - but here, the `ext-id` seems to work, what doesn’t work is the unique link text 🤔
@lcawl, do you happen to have any idea why this might not be working?

PR #4846, opened 2025-07-09
Adds input_type parameter to POST inference docs at the root level.

szabosteve:
## Overview

Related to https://github.com/elastic/developer-docs-team/issues/306

This PR adds the `input_type` parameter to the body parameters section of the POST inference API docs.

### DIFF
--- a/specification/inference/inference/InferenceRequest.ts
+++ b/specification/inference/inference/InferenceRequest.ts
@@ -83,6 +83,19 @@ export interface Request extends RequestBase {
      */
     input: string | Array<string>
     /**
+     * Specifies the input data type. Possible values include:
+     * * `SEARCH`
+     * * `INGEST`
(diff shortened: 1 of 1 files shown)


specification/inference/inference/InferenceRequest.ts
```diff
@@ -82,6 +82,19 @@ export interface Request extends RequestBase {
      * > Inference endpoints for the `completion` task type currently only support a single string as input.
      */
     input: string | Array<string>
+    /**
+     * Specifies the input data type. Possible values include:
```
davidkyle: We need to mention that this setting only applies to the `text_embedding` task type

```suggestion
     * Specifies the input data type for the text embedding model. The `input_type` parameter only applies to Inference Endpoints with the `text_embedding` task type. Possible values include:
```

PR #4849, opened 2025-07-09
Add Cluster/Allocation/Explain Query Param Example

joshua-adams-1:
---
### Change Details

Extends the documentation for the `cluster/allocation/explain` API after query parameter support was added in https://github.com/elastic/elasticsearch/pull/129342

- The query parameters were defined
- A second request example was provided, highlighting how to pass all parameters in the URL string rather than the request body

---
### Release

To be merged after https://github.com/elastic/elasticsearch/pull/129342

---
Github Issue - https://github.com/elastic/elasticsearch/issues/127028
Jira Ticket - ES-11865

### DIFF
--- a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
+++ b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
@@ -25,7 +25,7 @@ import { Duration } from '@_types/Time'
 /**
  * Explain the shard allocations.
  * Get explanations for shard allocations in the cluster.
- * This API accepts the current_node, index, primary and shard parameters via the request body, or via query parameters, but not via both at the same time.
+ * This API accepts the current_node, index, primary and shard parameters in the request body or in query parameters, but not in both at the same time.
  * For unassigned shards, it provides an explanation for why the shard is unassigned.
  * For assigned shards, it provides an explanation for why the shard is remaining on its current node and has not moved or rebalanced to another node.
  * This API can be very useful when attempting to diagnose why a shard is unassigned or why a shard continues to remain on its current node when you might expect otherwise.
--- a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
+++ b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
@@ -25,6 +25,7 @@ import { Duration } from '@_types/Time'
 /**
  * Explain the shard allocations.
  * Get explanations for shard allocations in the cluster.
+ * This API accepts the current_node, index, primary and shard parameters via the request body, or via query parameters, but not via both at the same time.
  * For unassigned shards, it provides an explanation for why the shard is unassigned.
  * For assigned shards, it provides an explanation for why the shard is remaining on its current node and has not moved or rebalanced to another node.
  * This API can be very useful when attempting to diagnose why a shard is unassigned or why a shard continues to remain on its current node when you might expect otherwise.
--- a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
+++ b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
@@ -46,6 +46,22 @@ export interface Request extends RequestBase {
   ]
   query_parameters: {
     /**
+     * Specifies the node ID or the name of the node to only explain a shard that is currently located on the specified node.
+     */
+    current_node?: string
+    /**
+     * Specifies the name of the index that you would like an explanation for.
+     */
+    index?: IndexName
+    /**
+     * If true, returns explanation for the primary shard for the given shard ID.
+     */
+    primary?: boolean
+    /**
+     * Specifies the ID of the shard that you would like an explanation for.
+     */
+    shard?: integer
+    /**
      * If true, returns information about disk usage and shard sizes.
      * @server_default false
      */
--- a/specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
+++ b/specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
@@ -1,5 +1,5 @@
 summary: Query Parameters
-method_request: GET _cluster/allocation/explain
+method_request: GET _cluster/allocation/explain?index=my-index-000001&shard=0&primary=false&current_node=my-node
 description: >
   Run `GET _cluster/allocation/explain?index="my-index-000001"&shard=0&primary=false&current_node="my-node` to get an explanation for a shard's current allocation. No parameters are required in the request body
 # type: request
--- a/specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
+++ b/specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
@@ -41,8 +41,7 @@ alternatives:
           "body" => [],
       ]);
   - language: curl
-    code:
-      'curl -X GET -H "Authorization: ApiKey $ELASTIC_API_KEY"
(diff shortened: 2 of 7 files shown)


specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
     }
   ]
   query_parameters: {
+    /**
+     * Specifies the node ID or the name of the node to only explain a shard that is currently located on the specified node.
```
lcawl, line 49: Is this what is meant?
```suggestion
     * Explain a shard only if it is currently located on the specified node name or node ID.
```

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
...
+     */
+    current_node?: string
+    /**
+     * Specifies the name of the index that you would like an explanation for.
```
lcawl, line 53:
```suggestion
     * The name of the index that you would like an explanation for.
```

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -25,6 +25,7 @@ import { Duration } from '@_types/Time'
 /**
  * Explain the shard allocations.
  * Get explanations for shard allocations in the cluster.
+ * This API accepts the current_node, index, primary and shard parameters via the request body, or via query parameters, but not via both at the same time.
```
lcawl: We try to avoid using the term "via"
```suggestion
 * This API accepts the current_node, index, primary and shard parameters in the request body or in query parameters, but not in both at the same time.
```

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
...
+    /**
+     * Specifies the node ID or the name of the node to only explain a shard that is currently located on the specified node.
+     */
+    current_node?: string
+    /**
+     * Specifies the name of the index that you would like an explanation for.
+     */
+    index?: IndexName
+    /**
+     * If true, returns explanation for the primary shard for the given shard ID.
```
lcawl, line 57: Does this mean that the `shard` parameter is required when `primary` is specified? If so, we should mention that in the description.
```suggestion
     * If true, returns an explanation for the primary shard for the specified shard ID.
```

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
...
+     */
+    primary?: boolean
+    /**
+     * Specifies the ID of the shard that you would like an explanation for.
```
lcawl, line 61:
```suggestion
     * An identifier for the shard that you would like an explanation for.
```
Alternatively, maybe "... that you want explained".

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
     }
   ]
   query_parameters: {
+    /**
+     * Specifies the node ID or the name of the node to only explain a shard that is currently located on the specified node.
+     */
+    current_node?: string
```
l-trotta, line 51: we prefer using aliases rather than primitive types when possible
```suggestion
    current_node?: NodeId
```

specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
```diff
@@ -0,0 +1,53 @@
+summary: Query Parameters
+method_request: GET _cluster/allocation/explain
```
l-trotta: this is incorrect, all the parameters need to be added to the method_request field for the automation to generate the examples correctly
```suggestion
method_request: GET _cluster/allocation/explain?index=my-index-000001&shard=0&primary=false&current_node=my-node
```

specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
```diff
@@ -0,0 +1,53 @@
...
+description: >
+  Run `GET _cluster/allocation/explain?index="my-index-000001"&shard=0&primary=false&current_node="my-node` to get an explanation for a shard's current allocation. No parameters are required in the request body
+# type: request
+value: |-
+  {}
+alternatives:
```
l-trotta, line 8: you can delete everything below the `value` field, as language examples are automatically generated

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
joshua-adams-1: I took the description verbatim from the existing [API docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-allocation-explain), see the definition for the `current_node` parameter. I don't like the existing wording myself but figured it had already passed through a review to have been published. Happy to change if you agree

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
joshua-adams-1: Happy to change! I had copied the parameter definitions from later in the file, so will change there as well. Can I ask what the benefit of this is? And does this apply to other primitives, like `integer`, `boolean` etc?

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
joshua-adams-1: Again, happy to change but will change below as well. 

Side question, in this case the same parameters are defined as both `query` and `body` parameters. Should they be defined twice, or is there a way to define once and use twice? (So that their descriptions are guaranteed to remain synced?)

specification/cluster/allocation_explain/examples/request/ClusterAllocationExplainRequestExample2.yaml
joshua-adams-1: Did not know that! Thank you, that's very clever

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
l-trotta: benefit is that some clients (like .net) use these aliases to provide shortcuts/helpers or for documentation purposes, it applies to most strings, most "time" related numbers, versions numbers and others (`boolean` is the only one without an alias I think)

joshua-adams-1:
Thank you both for the reviews. The majority of the comments regard the parameter descriptions. I copied these verbatim from later in the file, link [here](https://github.com/elastic/elasticsearch-specification/blob/main/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts#L63-L80):
```
body: {
    /**
     * Specifies the node ID or the name of the node to only explain a shard that is currently located on the specified node.
     */
    current_node?: string
    /**
     * Specifies the name of the index that you would like an explanation for.
     */
    index?: IndexName
    /**
     * If true, returns explanation for the primary shard for the given shard ID.
     */
    primary?: boolean
    /**
     * Specifies the ID of the shard that you would like an explanation for.
     */
    shard?: integer
  }
```
Two questions:
1. Since I copied the descriptions, I am happy to change them in light of the review, but assume I should also change the original?
2. Is it correct to have the same parameter definitions multiple times in a file? I couldn't find an API in the same position as us, facilitating parameters to be passed both as query parameters but also in the request body

PR #4851, opened 2025-07-09
Fixed a badly formatted example

miguelgrinberg:
This particular example was not correctly parsed by the request converter. I have fixed it to match what the description of the example says.

### DIFF
--- a/specification/indices/put_settings/examples/request/indicesPutSettingsRequestExample3.yaml
+++ b/specification/indices/put_settings/examples/request/indicesPutSettingsRequestExample3.yaml
@@ -1,7 +1,7 @@
 summary: Update index analysis
 method_request: PUT /my-index-000001/_settings
 description:
-  To add an analyzer, you must close the index (`POST /my-index-000001/_open`), define the analyzer, then reopen the
+  To add an analyzer, you must close the index (`POST /my-index-000001/_close`), define the analyzer, then reopen the
   index (`POST /my-index-000001/_open`).
 # type: request
 value: |-


specification/indices/put_settings/examples/request/indicesPutSettingsRequestExample3.yaml
```diff
@@ -2,36 +2,41 @@ summary: Update index analysis
...
-
-  \        \"tokenizer\":\"whitespace\"
-
-  \      }
-
-  \    }
-
-  \  }
-
+value: |-
```
miguelgrinberg, line 7: We don't really have a way to define examples that issue multiple requests, so I'm putting the 2nd and 3rd requests in the `value` key.

lcawl: This does not seem like a valid OpenAPI example to me. Linters that check the request body example against the schema (we currently have those linting rules turned off) would fail due to those extra API requests in the `value`. I think we need to stick to one API request body per yaml file.

miguelgrinberg: The current version of this example has two requests. And the `method_request` key in this example file is set to the "close index" endpoint, which does not take a body, and for that reason fails parsing in the request converter.

If we have to use one request per example, then maybe it is best to remove this example?

FYI, this is how the current two-request version looks in the official docs:

Screenshot 2025-07-10 at 10 40 44 AM

lcawl: I've pushed a commit eb281adb8c11d4ee7b0607a0576a45910a5785cf to put that info in the description instead. WDYT?

specification/indices/put_settings/examples/request/indicesPutSettingsRequestExample3.yaml
```diff
@@ -1,51 +1,54 @@
 summary: Update index analysis
-method_request: POST /my-index-000001/_close
-description: To add an analyzer, you must close the index, define the analyzer, then reopen the index.
+method_request: PUT /my-index-000001/_settings
+description:
+  To add an analyzer, you must close the index (`POST /my-index-000001/_open`), define the analyzer, then reopen the
```
miguelgrinberg:
```suggestion
  To add an analyzer, you must close the index (`POST /my-index-000001/_close`), define the analyzer, then reopen the
```

specification/indices/put_settings/examples/request/indicesPutSettingsRequestExample3.yaml
miguelgrinberg: Looks good save a minor mistake in one of the URLs. Thanks!

Let me know if this looks good now.

lcawl: Great, thanks!

PR #4860, opened 2025-07-10
[Openapi] Change product feature output

l-trotta:
Changing the "product feature" property to something bump can support.


l-trotta:
@miguelgrinberg yes exactly, they don't want to support `x-product-feature` because it's too specific

l-trotta:
manually backporting

PR #4896, opened 2025-07-14
[9.1] Updates 'model_id' and 'input_type' as required parameters for Cohere V2

kosabogi:
This PR updates the `model_id` and `input_type` parameters to be required for the Cohere V2 inference endpoint starting from 9.1.

I also removed the default value for `model_id`, assuming that it no longer has one since it's now required. @davidkyle  could you please confirm if this assumption is correct?


davidkyle (approved):
LGTM

davidkyle:
> I also removed the default value for model_id, assuming that it no longer has one since it's now required. @davidkyle could you please confirm if this assumption is correct?

This is correct. Thanks @kosabogi

PR #4939, opened 2025-07-16
`cluster/allocation/explain` Query Parameters

joshua-adams-1:
---
### Change Details

Extends the documentation for the `cluster/allocation/explain` API after query parameter support was added in https://github.com/elastic/elasticsearch/pull/129342

- The query parameters are defined, and the descriptions are updated based on comments from https://github.com/elastic/elasticsearch-specification/pull/4849
- A second request example was provided, highlighting how to pass all parameters in the URL string rather than the request body

---
### Release

To be merged after https://github.com/elastic/elasticsearch/pull/129342

---
Github Issue - https://github.com/elastic/elasticsearch/issues/127028
Jira Ticket - ES-11865

### DIFF
--- a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
+++ b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
@@ -18,13 +18,14 @@
  */
 
 import { RequestBase } from '@_types/Base'
-import { IndexName } from '@_types/common'
+import { IndexName, NodeId } from '@_types/common'
 import { integer } from '@_types/Numeric'
 import { Duration } from '@_types/Time'
 
 /**
  * Explain the shard allocations.
  * Get explanations for shard allocations in the cluster.
+ * This API accepts the current_node, index, primary and shard parameters in the request body or in query parameters, but not in both at the same time.
  * For unassigned shards, it provides an explanation for why the shard is unassigned.
  * For assigned shards, it provides an explanation for why the shard is remaining on its current node and has not moved or rebalanced to another node.
  * This API can be very useful when attempting to diagnose why a shard is unassigned or why a shard continues to remain on its current node when you might expect otherwise.
@@ -45,6 +46,22 @@ export interface Request extends RequestBase {
   ]
   query_parameters: {
     /**
+     * The name of the index that you would like an explanation for.
+     */
+    index?: IndexName
+    /**
+     * An identifier for the shard that you would like an explanation for.
+     */
+    shard?: integer
+    /**
+     * If true, returns an explanation for the primary shard for the specified shard ID.
+     */
+    primary?: boolean
+    /**
+     * Explain a shard only if it is currently located on the specified node name or node ID.
+     */
+    current_node?: NodeId
+    /**
      * If true, returns information about disk usage and shard sizes.
      * @server_default false
      */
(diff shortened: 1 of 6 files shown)


specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
```diff
@@ -44,6 +45,22 @@ export interface Request extends RequestBase {
...
+     */
+    index?: IndexName
+    /**
+     * An identifier for the shard that you would like an explanation for.
+     */
+    shard?: integer
+    /**
+     * If true, returns an explanation for the primary shard for the specified shard ID.
+     */
+    primary?: boolean
```
pquentin, line 59: Validation currently fails because you have YAML tests where `primary` can be a string. Is it really needed?

https://github.com/elastic/elasticsearch/blob/f135998b118d4e58922300cf942506ae25cd7ab1/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cluster.allocation_explain/10_basic.yml#L297-L299

pquentin:
Looking more closely, we have YAML tests allowing different types in multiple places. Here is the full list of changes needed to make the validation pass. This won't give a good developer experience. I don't recommend applying this diff unless there's a good reason. 

```diff
diff --git a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
index 50cda8f21..5bffd3ce4 100644
--- a/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
+++ b/specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
@@ -48,15 +48,15 @@ export interface Request extends RequestBase {
     /**
      * The name of the index that you would like an explanation for.
      */
-    index?: IndexName
+    index?: IndexName | integer
     /**
      * An identifier for the shard that you would like an explanation for.
      */
-    shard?: integer
+    shard?: integer | string
     /**
      * If true, returns an explanation for the primary shard for the specified shard ID.
      */
-    primary?: boolean
+    primary?: boolean | string
     /**
      * Explain a shard only if it is currently located on the specified node name or node ID.
      */
@@ -65,12 +65,12 @@ export interface Request extends RequestBase {
      * If true, returns information about disk usage and shard sizes.
      * @server_default false
      */
-    include_disk_info?: boolean
+    include_disk_info?: boolean | string
     /**
      * If true, returns YES decisions in explanation.
      * @server_default false
      */
-    include_yes_decisions?: boolean
+    include_yes_decisions?: boolean | string
     /**
      * Period to wait for a connection to the master node.
      * @server_default 30s
@@ -85,11 +85,11 @@ export interface Request extends RequestBase {
     /**
      * An identifier for the shard that you would like an explanation for.
      */
-    shard?: integer
+    shard?: integer | string
     /**
      * If true, returns an explanation for the primary shard for the specified shard ID.
      */
-    primary?: boolean
+    primary?: boolean | string
     /**
      * Explain a shard only if it is currently located on the specified node name or node ID.
      */

```

specification/cluster/allocation_explain/ClusterAllocationExplainRequest.ts
joshua-adams-1: According to the [ElasticSearch API](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/api-conventions#_boolean_values), it is documented that we facilitate strings to be passed for boolean values. I was adding YAML Rest Tests to ensure that if a user is to do so, we correctly parse this and proceed as expected. 

A separate question, why is validation here subjected to whether or not a YAML rest test exists / does not exist? What changes if I was to delete the test?

pquentin (approved):
Thanks! LGTM. Validation passes but this will be fixed by removing some YAML tests from Elasticsearch.

PR #4031, opened 2025-03-19
Stop reporting missing endpoints as errors

pquentin:
We have other ways of identifying them.

PR #4944, opened 2025-07-17
Ignore human overloading error in xpack.info request

pquentin:

### DIFF
--- a/compiler/src/validation-errors.ts
+++ b/compiler/src/validation-errors.ts
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+const IGNORED_ERRORS = [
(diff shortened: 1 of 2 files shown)


compiler/src/validation-errors.ts
```diff
@@ -54,10 +62,8 @@ export class ValidationErrors {
     let count = 0
     const logArray = function (errs: string[], prefix = ''): void {
       for (const err of errs) {
-        if (err !== 'Missing request & response') {
```
pquentin, line 57: This is no longer needed since https://github.com/elastic/elasticsearch-specification/pull/4031

l-trotta (approved):

PR #4953, opened 2025-07-17
Update Inference specification for Azure AI Studio rerank task

Evgenii-Kazannik:
Update specification related to [Add-Azure-AI-Foundry-Rerank-support](https://github.com/Evgenii-Kazannik/elasticsearch/tree/Add-Azure-AI-Foundry-Rerank-support)


l-trotta (approved):

l-trotta:
is this just for 9.1? is there a server pr reference?

output/openapi/elasticsearch-openapi.json
```diff
@@ -88848,6 +88854,14 @@
           "user": {
             "description": "For a `text_embedding` task, specify the user issuing the request.\nThis information can be used for abuse detection.",
             "type": "string"
+          },
+          "return_documents": {
+            "description": "For a `rerank` task, return doc text within the results.",
+            "type": "boolean"
+          },
+          "top_n": {
+            "description": "For a `rerank` task, the number of the top N documents that should be returned.",
```
dan-rubinstein: Minor: Can we instead use the description we use for some of our other services with the same value `For a `rerank` task, the number of most relevant documents to return.\nIt defaults to the number of the documents.`. I think it's potentially a bit more clear and provides a bit of extra information on what happens when the value is not set.

Same applies for the other descriptions for this value.

Evgenii-Kazannik: That's better. Thanks. Updated with this description

output/openapi/elasticsearch-openapi.json
```diff
@@ -88881,6 +88887,14 @@
           "user": {
             "description": "For a `text_embedding` task, specify the user issuing the request.\nThis information can be used for abuse detection.",
             "type": "string"
+          },
+          "return_documents": {
+            "description": "For a `rerank` task, return doc text within the results.",
+            "type": "boolean"
+          },
+          "top_n": {
+            "description": "For a `rerank` task, the number of most relevant documents to return.\\nIt defaults to the number of the documents.",
```
dan-rubinstein: Minor comment, does this need to be `\\n` or should it just be `\n`?

Evgenii-Kazannik: Thank you, Dan.
I removed a special character in order the json to be generated correctly

dan-rubinstein (approved):

PR #4981, opened 2025-07-21
[DOCS]: Clarify GET / response on Serverless

charlotte-hoblik:
This PR:
- clarifies the response when a user runs `GET /`
- updates the existing example response
- adds a new example specific to serverless deployments.

On Serverless, this endpoint will always return 8.11.0 as the version number and it is expected. Serverless offers automatic upgrades. Version is no longer something the user manages.

Based on: https://github.com/elastic/docs-content/issues/1960

### DIFF
--- a/specification/_global/info/RootNodeInfoRequest.ts
+++ b/specification/_global/info/RootNodeInfoRequest.ts
@@ -22,6 +22,7 @@ import { RequestBase } from '@_types/Base'
 /**
  * Get cluster info.
  * Get basic build, version, and cluster information.
+ * ::: In Serverless, this API is retained for backward compatibility only. Some response fields, such as the version number, do not reflect the actual runtime state.
  * @rest_spec_name info
  * @availability stack stability=stable
  * @availability serverless stability=stable visibility=public
--- a/specification/_global/info/examples/response/RootNodeInfoResponseExample2.yaml
+++ b/specification/_global/info/examples/response/RootNodeInfoResponseExample2.yaml
@@ -1,0 +1,19 @@
+# summary: ''
+description: A successful response from `GET /` on Serverless.
+# type: response
+# response_code: 200
+value:
+  name: serverless
+  cluster_name: my_test_serverless_cluster
+  cluster_uuid: 8xx0pi24Squnf4PFDOAtwg
+  version:
+    number: 8.11.0
+    build_flavor: serverless
+    build_type: docker
+    build_hash: 00000000
+    build_date: 2023-10-31
+    build_snapshot: false
+    lucene_version: 9.7.0
+    minimum_wire_compatibility_version: 8.11.0
+    minimum_index_compatibility_version: 8.11.0
+  tagline: 'You Know, for Search'
--- a/specification/_types/Base.ts
+++ b/specification/_types/Base.ts
@@ -110,6 +110,10 @@ export class ElasticsearchVersionInfo {
   minimum_wire_compatibility_version: VersionString
   /**
    * The Elasticsearch version number.
+   * 
+   * ::: In Serverless deployments, this value is always `8.11.0`, regardless of the actual runtime version.
+   * This static value exists for backward compatibility with legacy clients.
+   * Serverless environments are automatically upgraded, so this value can be safely ignored.
    */
   number: string
 }
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -13651,7 +13651,7 @@
           "info"
         ],
         "summary": "Get cluster info",
-        "description": "Get basic build, version, and cluster information.\n\n## Required authorization\n\n* Cluster privileges: `monitor`\n",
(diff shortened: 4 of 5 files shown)


charlotte-hoblik:
@leemthompo One question: I used the `:::` syntax in my changes to highlight notes (screenshot attached). It renders really nicely in the built docs, but I haven’t seen anyone else use it, so I’m not sure if it’s officially supported or just working by coincidence.

Could you let me know if it’s safe to use? I’d prefer to avoid relying on something that might break later.
Screenshot 2025-07-21 at 10 28 27
Screenshot 2025-07-21 at 10 28 36

specification/_types/Base.ts
```diff
@@ -110,6 +110,10 @@ export class ElasticsearchVersionInfo {
   minimum_wire_compatibility_version: VersionString
   /**
    * The Elasticsearch version number.
+   *
+   * ::: In Serverless deployments, this value is always `8.11.0`, regardless of the actual runtime version.
+   * This static value exists for backward compatibility with legacy clients.
+   * Serverless environments are automatically upgraded, so this value can be safely ignored.
```
leemthompo:
```suggestion
   * ::: IMPORTANT: For Serverless deployments, this static value is always `8.11.0` and is used solely for backward compatibility with legacy clients.
   *  Serverless environments are versionless and automatically upgraded, so this value can be safely ignored.
```

Just an edit for clarity and concision, and making it explicit that serverless is versionless

> the actual runtime version.

removed this because potentially confusing

specification/_global/info/RootNodeInfoRequest.ts
```diff
@@ -22,6 +22,7 @@ import { RequestBase } from '@_types/Base'
 /**
  * Get cluster info.
  * Get basic build, version, and cluster information.
+ * ::: In Serverless, this API is retained for backward compatibility only. Some response fields, such as the version number, do not reflect the actual runtime state.
```
leemthompo:
```suggestion
 * ::: In Serverless, this API is retained for backward compatibility only. Some response fields, such as the version number, should be ignored.
```

This might be clearer and more direct.

leemthompo (commented):
Thanks @charlotte-hoblik! Just suggested a rewording for clarity. On the `:::` syntax, I have no idea TBH, I think @flobernd probably knows about this. 😄 

I'll let @ppf and @rjernst weigh in on this too

leemthompo:
@charlotte-hoblik you'll need to run your make generate commands again to make sure the output files are also updated

Would be nice to have screenshots of the examples too for reviewers :)

specification/_global/info/examples/response/RootNodeInfoResponseExample2.yaml
```diff
@@ -0,0 +1,19 @@
+# summary: ''
+description: A successful response from `GET /` on Serverless.
```
leemthompo: You can also add as much information to the description field as you like to provide context

ppf2: ++ I see that the most important details are in comment blocks in code today. Let's move that into the description field so that it will actually show up in the API doc details?

```
  /**
   * The Elasticsearch version number.
   *
   * ::: IMPORTANT: For Serverless deployments, this static value is always `8.11.0` and is used solely for backward compatibility with legacy clients.
   *  Serverless environments are versionless and automatically upgraded, so this value can be safely ignored.
   */
```

Also, in the description field, I will suggest in addition to noting "safely ignored", it will be helpful to explicitly call out that Serverless are automatically upgraded to the _most current version_.  Thx!

leemthompo: The Typescript comments actually are used to generate the OpenAPI documents! But the example yaml files can also have extended descriptions

This is the Elasticsearch api docs workflow schematic:

 
api-docs-generation-pipeline

leemthompo: > it will be helpful to explicitly call out that Serverless are automatically upgraded to the _most current version_. 

@ppf I wonder if this wording might confuse people into thinking they're running the current/latest stack version?

ppf2: True, we can leave it out. I think automatically upgraded already suggests it's upgrading to the latest version that is appropriate for Serverless :D

charlotte-hoblik: Screenshot 2025-07-22 at 15 36 34
Screenshot 2025-07-22 at 15 37 35

Thanks for the feedback @leemthompo @ppf2  I’ve added the same clarification to the new Serverless example as well.
Let me know if anything else needs tweaking.

ppf2: When the user clicks on the version OBJECT item:

image

Is it going to show something like the following?

```
  /**
   * The Elasticsearch version number.
   *
   * ::: IMPORTANT: For Serverless deployments, this static value is always `8.11.0` and is used solely for backward compatibility with legacy clients.
   *  Serverless environments are versionless and automatically upgraded, so this value can be safely ignored.
   */
```

leemthompo: Correct!

leemthompo (approved):
LGTM, thanks!

specification/_global/info/examples/response/RootNodeInfoResponseExample2.yaml
ppf2: Awesome! Thx all!

PR #5020, opened 2025-07-22
Update inference API specification to include new Llama Service

Jan-Kazlouski-elastic:
This PR adds changes to specification caused by https://github.com/elastic/elasticsearch/pull/130092. It doesn't include the changes from PR: https://github.com/elastic/elasticsearch/pull/131827 to add openai text embedding support since that isn't merged yet.

**Additional actions**
- [x] Signed the CLA
- [x] Executed make contrib

### DIFF
--- a/specification/inference/_types/CommonTypes.ts
+++ b/specification/inference/_types/CommonTypes.ts
@@ -1560,7 +1560,7 @@ export class LlamaServiceSettings {
   /**
    * The URL endpoint of the Llama stack endpoint.
    * URL must contain:
-   * * For `text_embedding` task - `/v1/openai/v1/embeddings`.
+   * * For `text_embedding` task - `/v1/inference/embeddings`.
    * * For `completion` and `chat_completion` tasks - `/v1/openai/v1/chat/completions`.
    */
   url: string
@@ -1574,25 +1574,9 @@ export class LlamaServiceSettings {
    */
   model_id: string
   /**
-   * A valid API key for accessing Llama stack endpoint that is going to be sent as part of Bearer authentication header.
-   * This field is optional because Llama stack doesn't provide authentication by default.
-   *
-   * IMPORTANT: You need to provide the API key only once, during the inference model creation.
-   * The get inference endpoint API does not retrieve your API key.
-   * After creating the inference model, you cannot change the associated API key.
-   * If you want to use a different API key, delete the inference model and recreate it with the same name and the updated API key.
-   */
-  api_key?: string
-  /**
    * For a `text_embedding` task, the maximum number of tokens per input before chunking occurs.
    */
   max_input_tokens?: integer
-  /**
-   * For a `text_embedding` task, the number of dimensions the resulting output embeddings must have.
-   * It is supported only in `text-embedding-3` and later models. If it is not set by user, it defaults to the model returned dimensions.
-   * If model returns embeddings with a different number of dimensions, error is returned.
-   */
-  dimensions?: integer
   /**
    * For a `text_embedding` task, the similarity measure. One of cosine, dot_product, l2_norm.
    */
@@ -1602,14 +1586,6 @@ export class LlamaServiceSettings {
    * By default, the `llama` service sets the number of requests allowed per minute to 3000.
    */
   rate_limit?: RateLimitSetting
-}
-
-export class LlamaTaskSettings {
-  /**
-   * For a `completion` or `text_embedding` task, specify the user issuing the request.
-   * This information can be used for abuse detection.
-   */
-  user?: string
 }
 
 export enum LlamaTaskType {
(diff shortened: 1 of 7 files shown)


specification/inference/put_llama/PutLlamaRequest.ts
```diff
@@ -0,0 +1,79 @@
...
+  LlamaTaskType
+} from '@inference/_types/CommonTypes'
+import { InferenceChunkingSettings } from '@inference/_types/Services'
+
+/**
+ * Create a Llama inference endpoint.
+ *
+ * Create an inference endpoint to perform an inference task with the `llama` service.
+ * @rest_spec_name inference.put_llama
+ * @availability stack since=9.2.0 stability=stable visibility=public
```
Jan-Kazlouski-elastic, line 35: @jonathan-buttner could you please check if this 9.2.0 version is correctly set here. I assume it is, but want to be sure.

Jan-Kazlouski-elastic:
@jonathan-buttner since we didn't have backport for Llama integration, I added skip-backport label in order to skip backport label github action check

specification/inference/_types/CommonTypes.ts
```diff
@@ -1377,6 +1377,68 @@ export enum JinaAITextEmbeddingTask {
...
+  /**
+   * The URL endpoint of the Llama stack endpoint.
+   * URL must contain:
+   * * For `text_embedding` task - `/v1/inference/embeddings`.
+   * * For `completion` and `chat_completion` tasks - `/v1/openai/v1/chat/completions`.
+   */
+  url: string
+  /**
+   * The name of the model to use for the inference task.
+   * Refer to the Llama downloading models documentation for different ways of getting list of available models and downloading them.
```
jonathan-buttner:
```suggestion
   * Refer to the Llama downloading models documentation for different ways of getting a list of available models and downloading them.
```

specification/inference/_types/CommonTypes.ts
```diff
@@ -1377,6 +1377,68 @@ export enum JinaAITextEmbeddingTask {
...
+  /**
+   * A valid API key for accessing Llama stack endpoint that is going to be sent as part of Bearer authentication header.
+   * This field is optional because Llama stack doesn't provide authentication by default.
+   *
+   * IMPORTANT: You need to provide the API key only once, during the inference model creation.
+   * The get inference endpoint API does not retrieve your API key.
+   * After creating the inference model, you cannot change the associated API key.
+   * If you want to use a different API key, delete the inference model and recreate it with the same name and the updated API key.
+   */
+  api_key?: string
```
jonathan-buttner: I should have mentioned this on the elasticsearch PR. When did you need to supply an API key? In my testing of running the stack locally I didn't need to supply one 🤔 

I was running it like this:

```
PUT _inference/text_embedding/llama-text-embedding
{
    "service": "llama",
    "service_settings": {
        "url": "http://localhost:8321/v1/inference/embeddings",
        "model_id": "all-MiniLM-L6-v2"
    }
}
```

specification/inference/put_llama/PutLlamaRequest.ts
jonathan-buttner: Yep this is correct 👍

jonathan-buttner (commented):
Looking good, left a few questions

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: It is good point to discuss. Glad you're bringing that up.
Llama Stack doesn't have built in authorization check by default, so It is possible to use it without providing any tokens. Specially when testing with Ollama locally.
However for me it is doubtful that users are going to use llama stack without auth 100% of the time so I added this api_key parameter as option for clients that would want to set up bearer auth. Authentication Configuration is described in Distribution Overview's Server Configuration section of official Llama Stack guide.
https://llama-stack.readthedocs.io/en/latest/distributions/configuration.html#authentication-configuration 
I haven't investigated it in depth but I think it is safe to assume that providing ability to send bearer token pretty much covers security concerns.

specification/inference/_types/CommonTypes.ts
Jan-Kazlouski-elastic: Thanks. Fixed.

specification/inference/_types/CommonTypes.ts
jonathan-buttner: Did we test that the api key functionality works with our llama integration? Like did we configure llama so that it requires an api key and test to see if it returns failures if the api key isn't included in the request?

Jan-Kazlouski-elastic: No, unfortunately we did not test that. Llama doesn't provide out of the box auth support. We'd have to configure it and host it, that would take more time. 
We're sending Authorization Bearer token if it is provided in service settings. Do you want to have this tested?

Jan-Kazlouski-elastic:
@jonathan-buttner I removed mentions of OpenAI API for text_embeddings, "user" field an dimensions (because in current implementation those are not being sent to the model)
You can review this PR

PR #5052, opened 2025-07-29
Expand docs for `include_defaults`

DaveCTurner:
This parameter is trappy, its notion of "default" includes the values
from some arbitrary node's `elasticsearch.yml` file which means they
aren't really defaults. But then again the real default values would be
even less useful in the (common) case that some setting is overridden on
all nodes.

### DIFF
--- a/specification/cluster/get_settings/ClusterGetSettingsRequest.ts
+++ b/specification/cluster/get_settings/ClusterGetSettingsRequest.ts
@@ -45,7 +45,12 @@ export interface Request extends RequestBase {
      */
     flat_settings?: boolean
     /**
-     * If `true`, returns default cluster settings from the local node.
+     * If `true`, also returns default values for all other cluster settings, reflecting the values
+     * in the `elasticsearch.yml` file of one of the nodes in the cluster. If the nodes in your
+     * cluster do not all have the same values in their `elasticsearch.yml` config files then the
+     * values returned by this API may vary from invocation to invocation and may not reflect the
+     * values that Elasticsearch uses in all situations. Use the `GET _nodes/settings` API to
+     * observe the settings on individual nodes in your cluster.
      * @server_default false
      */
     include_defaults?: boolean


specification/cluster/get_settings/ClusterGetSettingsRequest.ts
```diff
@@ -45,7 +45,12 @@ export interface Request extends RequestBase {
      */
     flat_settings?: boolean
     /**
-     * If `true`, returns default cluster settings from the local node.
+     * If `true`, also returns default values for all other cluster settings, reflecting the values
+     * in the `elasticsearch.yml` file of one of the nodes in the cluster. If the nodes in your
+     * cluster do not all have the same values in their `elasticsearch.yml` config files then the
+     * values returned by this API may vary from invocation to invocation and may not reflect the
+     * values that Elasticsearch uses in all situations. Use the `GET _nodes/settings` API to
```
DiannaHohensee, line 52: I'm having trouble finding a `_nodes/settings` API?

DiannaHohensee (commented):
This is quite the labyrinth of code -- old code I'm guessing. Do we have a test someplace for this behavior, per chance?

specification/cluster/get_settings/ClusterGetSettingsRequest.ts
DaveCTurner: Bizarrely it's the `GET /_nodes/{nodeId}` path from `RestNodesInfoAction`, see how we compute `metricsOrNodeIds`.

DaveCTurner:
Yeah this is all really old behaviour, and mostly in the REST layer where the test coverage is pretty light. There's some YAML tests that use this functionality but I don't see anything that really tests it very well.

specification/cluster/get_settings/ClusterGetSettingsRequest.ts
```diff
@@ -45,7 +45,12 @@ export interface Request extends RequestBase {
...
+     * observe the settings on individual nodes in your cluster.
```
DiannaHohensee:
```suggestion
     * fetch the settings for each individual node in your cluster.
```

specification/cluster/get_settings/ClusterGetSettingsRequest.ts
DiannaHohensee: Should we say `_nodes/{nodeId}/info` instead??

I tracked down a couple of commits that suggest `GET _nodes/settings` actually works as an endpoint: https://github.com/elastic/elasticsearch/commit/f7db7eb99e00bb0cf6d9c69b864183463fb6d67e (adds the code you mentioned, `metricsOrNodeIds`) and https://github.com/elastic/elasticsearch/commit/ad50afbec8c2d619a33fdb591361d5d2ad4c766d. I can't find any documentation around the API, though. There's technically a test `RemoteClusterSecurityRestIT#testNodesInfo` that uses `GET _nodes/settings`. I gather that `GET /_nodes/settings` will return `GET /_nodes/_all/settings`

I think the _nodes/* API lacks both documentation and testing -- like a user can't look up what `GET /_nodes/settings` does and find an answer. I'm good with committing this as is, but it'd be great to file a JIRA ticket to address testing and docs.

DaveCTurner: `GET _nodes/{nodeId}/info` isn't a valid endpoint. [Here are the nodes info API docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-nodes-info) - particularly `GET /_nodes/{metric}`. We seem to have lost the list of valid metrics, but `info` isn't one of them.

DaveCTurner: I opened https://github.com/elastic/elasticsearch-specification/issues/5094

DiannaHohensee: Ohhh, it's [`GET _nodes/{nodeId}/info/{metrics}`](https://github.com/elastic/elasticsearch/blob/58f050a38bcba4dc2aa2888516292db19224c37a/server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestNodesInfoAction.java#L50) -- I didn't have it quite right. Though not in [that documentation](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-nodes-info) you linked, anyway... 🤨

DaveCTurner: What a mess :) but yeah `GET _nodes/_all/info/settings` is a longer (but more cohesive) way of getting the same output as `GET _nodes/settings`

PR #5061, opened 2025-07-30
Use strings instead of integer ids for simulate examples

dakrone:
These examples were being rendered with the `_id` as an integer, however, that is invalid and we only allow strings for integers.

### DIFF
--- a/specification/simulate/ingest/examples/request/SimulateIngestRequestExample1.yaml
+++ b/specification/simulate/ingest/examples/request/SimulateIngestRequestExample1.yaml
@@ -7,11 +7,11 @@ description:
 # type: request
 value:
   docs:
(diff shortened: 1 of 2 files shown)


masseyke (approved):
Looks like our prettier config wants those to be single quotes, but otherwise LGTM. Thanks for fixing it.
```
npx prettier --config compiler/.prettierrc.json --log-level warn --write specification/simulate/ingest/examples/request/SimulateIngestRequestExample* 
```

dakrone:
The CI fails with `Validation is not supported from forks`, so I guess I should wait until someone verifies this manually?

l-trotta:
no need to run validation on examples, I'll just merge it

PR #5073, opened 2025-08-04
Remove `number_of_shards` setting from clone API example

PeteGillinElastic:
This setting is allowed only if the number of shards matches that of the source index, and that's the behaviour anyway, so it is needlessly confusing to include it here.


specification/indices/clone/examples/request/indicesCloneRequestExample1.yaml
```diff
@@ -6,12 +6,6 @@ description: >
 # type: request
 value: "{
 
-  \  \"settings\": {
```
dakrone, line 9: Since the description says "The API accepts `settings` and `aliases` parameters for the target index", can we keep an example of updating the settings here? Perhaps we can use `index.refresh_interval` instead of the primary shards one?

PeteGillinElastic: Yeah, that's a good idea. Done. (I had considered doing something like this, but I couldn't think of any settings that one would be likely to specify while cloning. I would guess that most clones are "clean" copies, though I could be wrong.)

dakrone (approved):
LGTM, thanks!

PR #5129, opened 2025-08-12
Generate a docs for properties that are a simple reference

### LINKED ISSUE #5082: [OpenAPI] Parameter descriptions are missing from the generated OpenAPI file while present in the ts files
## Problem

Certain parameter descriptions present in the TypeScript files are not displayed in the API docs, and they are also not included in the OpenAPI spec.

## Example

This is how the generated docs look like for the [Cohere inference endpoint docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-cohere#operation-inference-put-cohere-body-application-json-service):

Image

And this is the description in the [TS file](https://github.com/elastic/elasticsearch-specification/blob/f1534e287bbb5e6b2c3d423558af5730d0634e3e/specification/inference/put_cohere/PutCohereRequest.ts#L74):

```
The type of service supported for the specified task type. In this case, `cohere`.
```
`service` is an enum and its value is `cohere`.

It seems the description is stripped down and only the enum value is displayed as the description is missing in the OpenAPI spec.

```
"service": {
                    "$ref": "#/components/schemas/inference._types.CohereServiceType"
```

and this is the `CohereServiceType`:

```
"inference._types.CohereServiceType": {
        "type": "string",
        "enum": [
          "cohere"
        ]
      },
```

Would it be possible to also include the description of the parameter in the OpenAPI file, so it can be displayed in the generated documentation?

swallez:
Open API 3.0 doesn't allow attaching docs and additional properties to a `$ref` type definition. Because of this (and this was a forgotten `TODO`) many property descriptions were dropped in the OpenAPI output.

This PR fixes this by using a common/recommended workaround consisting in using a single `allOf` inline schema around the reference.

Fixes #5082.

@szabosteve can you verify that the doc generation pipeline correctly picks up this change?

Example (taken from the diff):

```json
"chunking_settings": {
    "$ref": "#/components/schemas/inference._types.InferenceChunkingSettings"
}
```

Becomes:
```json
"chunking_settings": {
    "externalDocs": {
      "url": "https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#infer-chunking-config"
    },
    "description": "The chunking configuration object.",
    "allOf": [
     {
        "$ref": "#/components/schemas/inference._types.InferenceChunkingSettings"
      }
    ]
}
```


szabosteve:
Thank you so much for this fix, @swallez! I asked the folks responsible for the docs generation process to verify if they can pick this change up correctly. I'll get back to you as soon as I have an answer.

szabosteve:
@swallez The folks at Bump pushed a modification on their side to support the new input. Just checked and your fix works fine! Thank you very much for your work, it significantly improves the API docs quality!

Screenshot 2025-08-25 at 10 59 49

Parameter descriptions like `sevice` and `input_type`, for example, were empty before your fix!

szabosteve (approved):
LGTM!

PR #5138, opened 2025-08-13
Updates the method of ping from HEAD to OPTIONS 

picandocodigo:
This was proposed in https://github.com/elastic/elasticsearch/issues/97613#issuecomment-1633891769.
Opening this as a Draft to check with the rest of the Devtools team about this and see if it still makes sense and if all clients support the `OPTIONS` method.

Let me know, otherwise we can close this PR and the original issue.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -14141,7 +14141,7 @@
           }
         ]
       },
-      "head": {
+      "options": {
         "tags": [
           "cluster"
         ],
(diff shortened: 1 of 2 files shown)


l-trotta:
np for java

pquentin:
It's also fine for Python. Actually `ping` is custom there, so I can make the changes independently of the specification.

Notice how this adds an error: we'll need to add it to https://github.com/elastic/elasticsearch/blob/main/rest-api-spec/src/main/resources/rest-api-spec/api/ping.json first. It's also fine to allow both `HEAD` and `OPTIONS`, and the generators can choose the best version as they see fit.

picandocodigo:
Thanks!
Opened the PR in Elasticsearch:
https://github.com/elastic/elasticsearch/pull/133122

picandocodigo:
Not moving forward with this after discussing with the team.

PR #5141, opened 2025-08-15
Adding effective-mapping to the ingest simulate response

masseyke:
The `effective_mapping` field was added to the simulate ingest response in https://github.com/elastic/elasticsearch/pull/132833. This adds it to the specification.


pquentin (approved):
Thanks! LGTM.

PR #5143, opened 2025-08-17
Use `Sort` object in "delete by query"

ramikg:
The delete by query API's `sort` property supports the `Sort` object:
```
"sort": {
  "timestamp": {
    "order": "desc",
    "unmapped_type": "date"
  }
}
```
~~But the property's type does not reflect this fact.~~

~~This PR changes the type of `sort` from `string[]` to the more general `Sort` object (as used by the search API):~~

But the body is missing the sort parameter, this PR adds it as `Sort` object.

### DIFF
--- a/specification/_global/delete_by_query/DeleteByQueryRequest.ts
+++ b/specification/_global/delete_by_query/DeleteByQueryRequest.ts
@@ -31,6 +31,7 @@ import { float, long } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
 import { Operator } from '@_types/query_dsl/Operator'
 import { SlicedScroll } from '@_types/SlicedScroll'
+import { Sort } from '@_types/sort'
 import { Duration } from '@_types/Time'
 
 /**
@@ -310,5 +311,9 @@ export interface Request extends RequestBase {
      * Slice the request manually using the provided slice ID and total number of slices.
      */
     slice?: SlicedScroll
+    /**
+     * A sort object that specifies the order of deleted documents.
+     */
+    sort?: Sort
   }
 }


l-trotta:
Hello @ramikg, thank you for you interest in contributing to the spec! 
That `sort` field in [DeleteByQueryRequest](https://github.com/elastic/elasticsearch-specification/blob/c1938dec3238a702616e61a38bd7a887e277ca31/specification/_global/delete_by_query/DeleteByQueryRequest.ts#L261) is a query param and cannot be an object, if you check SearchRequest you can see that it has the [same field](https://github.com/elastic/elasticsearch-specification/blob/3a98596611b8de508741bc86a81e78a414e0d719/specification/_global/search/SearchRequest.ts#L384). 
One thing we have to do for sure is making them the same, so `sort` in DeleteByQueryRequest should also be a `string | string[]`. But I'm not finding any evidence that DeleteByQueryRequest has a body field `sort`, did you manage to reproduce it?

ramikg:
Oops, I've missed that the `sort` property I've modified is under `query_parameters`.
I've restored the query parameter to its original form.

I tried using the `sort` property (exactly as provided above) in the body and verified that Elasticsearch actually parses it (e.g. wrong `order` or `unmapped_type` values resulted in an error).
I've checked this in Elasticsearch versions 9.1.2 & 8.19.2.

l-trotta:
@ramikg thanks for checking! I've done a bit of digging and this seems to be part of a larger issue, where `delete by query` actually supports all parameters supported by `search request`. let's merge this for now, then we'll see what to do about the rest :D

l-trotta (approved):

ramikg:
Thank you!

By the way, this seems to be the case for "update by query" as well (and perhaps for additional APIs).

l-trotta:
We've investigated this a bit more and found out that the `sort` body parameter actually does nothing in this context, and moreover the `sort` query parameter isn't actually supported. I'm reverting this and deprecating the existing query param, sorry for the confusion!

ramikg:
> We've investigated this a bit more and found out that the `sort` body parameter actually does nothing in this context, and moreover the `sort` query parameter isn't actually supported. I'm reverting this and deprecating the existing query param, sorry for the confusion!

Could you please elaborate?

I may be missing something, but in my tests it seems that the `sort` body parameter is effective. For example:
```
POST test/_bulk
{ "index": {} }
{ "amount": 1 }
{ "index": {} }
{ "amount": 2 }
{ "index": {} }
{ "amount": 3 }
{ "index": {} }
{ "amount": 4 }
{ "index": {} }
{ "amount": 5 }

POST test/_delete_by_query
{
  "query": {
    "match_all": {}
  },
  "sort": {
    "amount": {
      "order": "asc",
      "unmapped_type": "long"
    }
  },
  "max_docs": 2
}
```
After running the commands above, the remaining amounts were {3, 4, 5}, and then when I deleted again with `"order": "desc"` then I was left with just 3.
(I've also tried several other experiments.)

l-trotta:
@ramikg ah that's how! sorry I was testing with `size`, not considering `max_docs`. reverting again ^^"

PR #5149, opened 2025-08-18
Add tip about availability since across majors

lcawl:
Relates to https://github.com/elastic/elasticsearch-specification/pull/5148

This PR adds a sentence to `add-new-api.md` to clarify recommended behaviour for the `@availability since` property when an API is added to multiple major versions.


pquentin:
We'll definitely have to be careful about this, thanks. @leemthompo should we add this to https://www.elastic.co/docs/extend/contribute/api-docs/organize-annotate too?

leemthompo:
@pquentin good idea 👍 opened https://github.com/elastic/docs-content/pull/2618 to replicate

PR #5150, opened 2025-08-19
[DOCS] Update repo branch in API docs intro

lcawl:
This PR fixes the [detail in the Elasticsearch API docs](https://www.elastic.co/docs/api/doc/elasticsearch/v9/#topic-documentation-source-and-versions) that indicates which branch it's derived from. It incorrectly indicates that the branch used for the v9 API docs is main but it is in fact 9.1.


leemthompo (approved):
LGTM2

PR #5170, opened 2025-08-20
Generalize comment on monitoring reindex tasks

jxie-1:
In serverless, only GET is supported for the task management API, so users can't cancel a task.

I think this is the simplest fix, as monitoring the reindex task is probably the most relevant part for troubleshooting. I can also make the description more granular if that's a better approach


specification/_global/reindex/ReindexRequest.ts
```diff
@@ -68,7 +68,7 @@ import { Destination, Source } from './types'
  *
  * It's recommended to reindex on indices with a green status. Reindexing can fail when a node shuts down or crashes.
  * * When requested with `wait_for_completion=true` (default), the request fails if the node shuts down.
- * * When requested with `wait_for_completion=false`, a task id is returned, which can be used via the task management API to monitor, debug, or cancel the task. The task may disappear or fail if the node shuts down.
+ * * When requested with `wait_for_completion=false`, a task id is returned, which can be used via the task management API to monitor the task. The task may disappear or fail if the node shuts down.
```
leemthompo:
```suggestion
 * * When requested with `wait_for_completion=false`, a task id is returned, for use with the task management APIs. The task may disappear or fail if the node shuts down.
```

leemthompo: totally optional idea, just for consideration if wanted to be even more generic :)

jxie-1: Sounds good, thanks for the suggestion

PR #5199, opened 2025-08-26
Run custom eslint rules as part of CI

pquentin:
Right now, all rules pass, but this needs to be enforced:

- by setting the severity to `error`
- and running lint as part of `make contrib`, which runs in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 SHELL := /bin/bash
 
-validate: ## Validate a given endpoint request or response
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
Makes sense! I thought `make validate` had the same effect in CI, but I guess that's not the case.

pquentin:
Right! `make validate` is only used by developers. Validation in CI happens with https://github.com/elastic/elasticsearch-specification/tree/main/.github/validate-pr

PR #5221, opened 2025-08-28
Clarify docs on repo analysis params

DaveCTurner:
Makes it explicit that for realistic experiments you need to leave many
of the parameters alone: for instance setting `?concurrency=1` is not a
realistic experiment.

Also adjusts the wording slightly to use the imperative voice more.

PR #5247, opened 2025-09-02
Clarify docs on repo analysis params

DaveCTurner:
Makes it explicit that for realistic experiments you need to leave many
of the parameters alone: for instance setting `?concurrency=1` is not a
realistic experiment.

Also adjusts the wording slightly to use the imperative voice more.

### DIFF
--- a/specification/snapshot/repository_analyze/SnapshotAnalyzeRepositoryRequest.ts
+++ b/specification/snapshot/repository_analyze/SnapshotAnalyzeRepositoryRequest.ts
@@ -143,12 +143,13 @@ export interface Request extends RequestBase {
   query_parameters: {
     /**
      * The total number of blobs to write to the repository during the test.
-     * For realistic experiments, you should set it to at least `2000`.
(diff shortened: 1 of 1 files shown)


DaveCTurner:
Replaces https://github.com/elastic/elasticsearch-specification/pull/5221 since we apparently need PRs to come from branches in the main repo here.

szabosteve (approved):
LGTM, thank you!

pquentin:
Thanks for opening it using a repository branch! Have you considered changing the Elasticsearch defaults to be more realistic?

DaveCTurner:
That's [answered in the docs](https://github.com/elastic/elasticsearch-specification/blob/22a822872c2d27f238c49cd92beff9d6092c26a2/specification/snapshot/repository_analyze/SnapshotAnalyzeRepositoryRequest.ts#L34):

> The default values for the parameters are deliberately low to reduce the impact of running an analysis inadvertently and to provide a sensible starting point for your investigations.

PR #5256, opened 2025-09-03
Document new index mode filter for _resolve/index

smalyshev:
See: https://github.com/elastic/elasticsearch/pull/133616


leemthompo:
@smalyshev @quux00 a little FYI on API docs, we added some contribution guidelines recently here: https://www.elastic.co/docs/extend/contribute/api-docs/

They are heavily Elasticsearch-oriented, including a quickstart for working locally and previewing changes.

Reach out in Slack if you think these are helpful, or if we should boost the comms around it :)

PR #5259, opened 2025-09-05
Add project_routing query param to _search and _async_search

quux00:
Adds the project_routing query param to the _search and _async_search endpoints
but only in serverless behind the feature flag "serverless.cross_project.enabled". 
This param is not available in stateful ("stack").

### DIFF
--- a/specification/_global/search/SearchRequest.ts
+++ b/specification/_global/search/SearchRequest.ts
@@ -228,7 +228,7 @@ export interface Request extends RequestBase {
      *  _alias:_origin
      *  _alias:*pr*
      * Supported in serverless only.
-     * @availability serverless
+     * @availability serverless stability=stable visibility=feature_flag feature_flag=serverless.cross_project.enabled
      */
     project_routing?: ProjectRouting
     /**
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -263,9 +263,6 @@
             "$ref": "#/components/parameters/async_search.submit-preference"
           },
           {
-            "$ref": "#/components/parameters/async_search.submit-project_routing"
-          },
-          {
             "$ref": "#/components/parameters/async_search.submit-request_cache"
           },
           {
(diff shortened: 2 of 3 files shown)


l-trotta:
@quux00 thanks! waiting for the [server side](https://github.com/elastic/elasticsearch/pull/134240) PR to be merged. just to be sure, is the only value allowed `_alias:something`? would it make sense to have the user only input the customizable part of the string?

specification/_global/search/SearchRequest.ts
```diff
@@ -219,6 +220,17 @@ export interface Request extends RequestBase {
...
     pre_filter_shard_size?: long
+    /**
+     * Specifies a subset of projects to target for a search using project metadata
+     * tags in a subset Lucene syntax. Allowed Lucene queries: the _alias tag
+     * and a single value (possible wildcarded). Examples:
+     *  _alias:my-project
+     *  _alias:_origin
+     *  _alias:*pr*
+     * Supported in serverless only.
+     * @availability serverless
```
quux00: I see these on another PR:
```
@availability serverless stability=stable visibility=public
```

Do I need to add those? What are the other possible values?

l-trotta: they are optional, but nice to have. possible values:

- stability: stable, experimental, unknown
- visibility: public, private, feature_flag

in case of visibility=feature_flag you can also add an optional `feature_flag={name of the feature flag}`.
lastly there's `since`, which indicates (for stack) the version where the endpoint/field was introduced. 
complete example: https://github.com/elastic/elasticsearch-specification/blob/3a98596611b8de508741bc86a81e78a414e0d719/specification/_global/search/SearchRequest.ts#L389

quux00: Got it thanks. I'll add that in the next push.

quux00:
@l-trotta 

> is the only value allowed `_alias:something`? would it make sense to have the user only input the customizable part of the string?

For now that is the only allowed entry, but in the future we will allow arbitrary key/value pairs with AND/OR/NOT booleans and boolean groupings, using Lucene query syntax, so I don't think we want additional constraints in the spec.

pquentin:
@JoshMock There's #5277 too!

quux00:
> s this all the APIs we know of for now that will support CPS project routing rules? If not, the sooner we get those into the spec the better, as they'll be necessary to support the middleware Kibana needs.

No, there are several more. Those PR(s) will be out later this week.

PR #5265, opened 2025-09-08
Fix Level string value

charlotte-hoblik:
This PR fixes an error in the documentation: changes the vales of the Level string from `cluster, indices or shards` to `node, indices or shards`
Connected to: https://github.com/elastic/docs-content/issues/1258


pquentin:
We discussed this with @l-trotta and @Anaethelion, and we decided to introduce a NodeStatsLevel enum since using `Level` is wrong and breaks users.

PR #5270, opened 2025-09-08
Feature/update authenticate specs

ankit--sethi:
Updating API spec for the `/_security/_authenticate` API to add a couple of new fields, `internal` and `managed_by` that are applicable to API Keys

### DIFF
--- a/specification/security/authenticate/SecurityAuthenticateResponse.ts
+++ b/specification/security/authenticate/SecurityAuthenticateResponse.ts
@@ -38,18 +38,17 @@ export class Response {
      * @availability serverless
      */
     token?: Token
-
   }
 }
 
 export class AuthenticateApiKey {
   id: Id
   name?: Name
-  managed_by: ManagedBy
+  managed_by: AuthenticateApiKeyManagedBy
   internal: boolean
 }
 
-export enum ManagedBy {
+export enum AuthenticateApiKeyManagedBy {
   cloud,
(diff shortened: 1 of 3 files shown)


specification/security/authenticate/SecurityAuthenticateResponse.ts
```diff
@@ -44,4 +44,11 @@ export class Response {
 export class AuthenticateApiKey {
   id: Id
   name?: Name
+  managed_by: AuthenticateApiKeyManagedBy
```
ankit--sethi, line 47: slightly verbose name to avoid a naming conflict that comes up when using `ManagedBy`:
```
ManagedBy is present in multiple namespaces: indices._types and security.authenticate
```

pquentin: Do you we have plans to add `managed_by` to other APIs? If yes it should be added under `specification/security` so that it's not specific to this API.

ankit--sethi: I'll raise this internally but to my knowledge, this is specific to this api currently. Will refactor if we come back here again!

pquentin: Right, thank you! Actually, a refactor here would be a breaking change for some clients that expose the specification hierarchy, which is why I asked.

PR #5274, opened 2025-09-10
Add the `input_type` param to the Perform text embedding inference API

davidkyle:
The `input_type` parameter is used in the text embedding Inference API. 

It's documented in the general inference API docs https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-inference but not in the specific text embedding inference doc  https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-text-embedding.

For this PR I've just copied and pasted the input_type details from https://github.com/elastic/elasticsearch-specification/blob/main/specification/inference/inference/InferenceRequest.ts#L97

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -14780,6 +14780,10 @@
                       }
                     ]
                   },
+                  "input_type": {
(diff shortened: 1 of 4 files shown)


davidkyle:
Thanks for adding the backport labels @pquentin

PR #5277, opened 2025-09-10
[ML] Add project_routing

prwhelan:
Add project_routing to Transform and Anomaly Detection API.


pquentin:
Unlike https://github.com/elastic/elasticsearch-specification/pull/5259, this is also accepted in stack. Is that intended?

prwhelan:
We are no longer planning to support projectRouting explicitly during tech preview - we'll revisit this for GA

PR #5292, opened 2025-09-15
Adds project_routing query param to _msearch Elasticsearch API

quux00:
Corresponds to https://github.com/elastic/elasticsearch/pull/134741


specification/_global/msearch/types.ts
```diff
@@ -37,6 +43,7 @@ export class MultisearchHeader {
   ignore_unavailable?: boolean
   index?: Indices
   preference?: string
+  project_routing?: ProjectRouting
```
quux00, line 46: I'm limiting this PR to just msearch because I want to highlight this addition.

I'm a little dubious as I don't really understand what this MultisearchHeader thingy is doing. 
Here's why I'm confused:

First, this list only partially matches the list of query_params that _search supports and is only a partial list of query_params that _msearch supports.  

Second, the comment says "used to change the subsequent search body request". But these query params don't go in the search body.  For example, you can't put `ignore_unavailable` in the search body. It only goes in the query-param section. Same for "routing" and the others.

So I don't know what this list is for.

JoshMock: Much like a bulk request, where the body is NDJSON with instructions given on one line for how to handle the data in the following line, every other row in an msearch call is either a header or a search query. All of this is [documented here](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-msearch#operation-msearch-body-application-json).

In the clients (and thus in the spec), the `header` rows are typed as `MultisearchHeader`, and the `search` rows are typed as `SearchRequestBody`. In the clients, they are represented as an array called `searches`:

```
{
  "searches": [
    MultiSearchHeader,
    SearchRequestBody,
    MultiSearchHeader,
    SearchRequestBody
  ]
}
```

As far as I am aware, `project_routing` would go in the header lines, which you've done here, so I think your change is correct.

Does that help?

JoshMock (approved):

JoshMock (approved):

Issue #3551, opened 2025-01-17
Common query parameters omitted from OpenAPI


lcawl:
There are common query parameters that are applicable to all the CAT APIs, for example the `verbose`, `format`, `sort`, `help` etc covered in https://www.elastic.co/guide/en/elasticsearch/reference/master/cat.html#common-parameters

I can see that they're defined in the specification in https://github.com/elastic/elasticsearch-specification/blob/2397e3139d2a318d31c49b1af01db2142889df4b/specification/_spec_utils/behaviors.ts#L90
However, they don't show up in the OpenAPI output.

For example, in https://github.com/elastic/elasticsearch-specification/blob/main/output/openapi/elasticsearch-openapi.json the "cat aliases" operation shows only the parameters that are defined explicitly in https://github.com/elastic/elasticsearch-specification/blob/main/specification/cat/aliases/CatAliasesRequest.ts

```
 "operationId": "cat-aliases",
        "parameters": [
          {
            "$ref": "#/components/parameters/cat.aliases#expand_wildcards"
          },
          {
            "$ref": "#/components/parameters/cat.aliases#master_timeout"
          }
        ],
```

Do we need to add them to all the request definitions? Or is there a simple fix to get the `CommonCatQueryParameters` to resolve?

lcawl:
Per discussion with @swallez "The common parameters are indeed not present in the OpenAPI output. The idea was to mimic what was in the doc and avoid having them shown for every API endpoint, for readability purposes. Now the bad side effect is that they actually disappear entirely...We can indeed include them in the OpenAPI output but... if they're linked to requests, they will be expanded everywhere. That's a lot of noise, that was rightfully avoided in the old docs"

Some possible solutions are:

- Get the common parameters added to the specification then pursue display improvements such as collapsing parameter lists
- Put the common parameters in the tag description, so they'll appear in the introduction to the "cat" grouping of APIs, similar to how it was handled in the old docs. IMO this is less ideal since it's hiding pertinent details in a description, but it would be feasible if necessary especially if https://github.com/elastic/elasticsearch-specification/issues/2984 is addressed.

DaveCTurner:
Just to add that the `?bytes`, `?size` and `?time` parameters are also totally missing in the new docs - I'll update the OP to reflect this too.

PR #5297, opened 2025-09-16
Add common cat API parameters for unit rendering

DaveCTurner:
Describes the `?bytes=`, `?size=` and `?time=` parameters which are
accepted by all the `GET _cat/...` APIs.


DaveCTurner:
Relates https://github.com/elastic/elasticsearch-specification/issues/3551

DaveCTurner:
Ah sorry yes my usual automation doesn't work here because of the no-forks thing. I'll try again, sec.

PR #5298, opened 2025-09-16
Add common cat API parameters for unit rendering

DaveCTurner:
Describes the `?bytes=`, `?size=` and `?time=` parameters which are
accepted by all the `GET _cat/...` APIs.

### DIFF
--- a/specification/_spec_utils/behaviors.ts
+++ b/specification/_spec_utils/behaviors.ts
@@ -97,6 +97,28 @@ export interface CommonCatQueryParameters {
    * @server_default false
    */
   v?: boolean
+  /**
+   * Sets the units for columns that contain a byte-size value.
+   * Note that byte-size value units work in terms of powers of 1024. For instance `1kb` means 1024 bytes, not 1000 bytes.
+   * If omitted, byte-size values are rendered with a suffix such as `kb`, `mb`, or `gb`, chosen such that the numeric value of the column is as small as possible whilst still being at least `1.0`.
+   * If given, byte-size values are rendered as an integer with no suffix, representing the value of the column in the chosen unit.
+   * Values that are not an exact multiple of the chosen unit are rounded down.
+   */
+  bytes?: Bytes
+  /**
+   * Sets the units for columns that contain a size value which is not a byte-size value.
+   * If omitted, size values are rendered with a suffix such as `k`, `m`, or `g`, chosen such that the numeric value of the column is as small as possible whilst still being at least `1.0`.
+   * If given, size values are rendered as an integer with no suffix, representing the value of the column in the chosen unit.
+   * Values that are not an exact multiple of the chosen unit are rounded down.
+   */
+  size?: '' | 'k' | 'm' | 'g' | 't' | 'p'
+  /**
+   * Sets the units for columns that contain a time duration.
+   * If omitted, time duration values are rendered with a suffix such as `ms`, `s`, `m` or `h`, chosen such that the numeric value of the column is as small as possible whilst still being at least `1.0`.
+   * If given, time duration values are rendered as an integer with no suffix.
+   * Values that are not an exact multiple of the chosen unit are rounded down.
+   */
+  time?: TimeUnit
 }
 
 /**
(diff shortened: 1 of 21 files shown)


specification/_spec_utils/behaviors.ts
```diff
@@ -97,6 +97,28 @@ export interface CommonCatQueryParameters {
...
+   * Values that are not an exact multiple of the chosen unit are rounded down.
+   */
+  bytes?: Bytes
+  /**
+   * Sets the units for columns that contain a size value which is not a byte-size value.
+   * If omitted, size values are rendered with a suffix such as `k`, `m`, or `g`, chosen such that the numeric value of the column is as small as possible whilst still being at least `1.0`.
+   * If given, size values are rendered as an integer with no suffix, representing the value of the column in the chosen unit.
+   * Values that are not an exact multiple of the chosen unit are rounded down.
+   */
+  size?: '' | 'k' | 'm' | 'g' | 't' | 'p'
```
DaveCTurner: Not sure what to do with this one - it doesn't seem to be defined anywhere yet.

specification/shutdown/delete_node/ShutdownDeleteNodeRequest.ts
```diff
@@ -51,11 +51,11 @@ export interface Request extends RequestBase {
      * Period to wait for a connection to the master node. If no response is received before the timeout expires, the request fails and returns an error.
      * @server_default 30s
      */
-    master_timeout?: TimeUnit
+    master_timeout?: Duration
```
DaveCTurner, line 54: This (and similar changes) looks like it was a mistake, it should have been `Duration` all along right?

DaveCTurner:
It's telling me to run `prettier` but I do not have this tool.

Also this seems problematic:

```
Error: ../output/typescript/types.ts(8097,18): error TS2430: Interface 'CatMlTrainedModelsRequest' incorrectly extends interface 'CatCatRequestBase'.
  Types of property 'size' are incompatible.
    Type 'number' is not assignable to type '"" | "p" | "g" | "m" | "k" | "t"'.
Error: ../output/typescript/types.ts(9032,18): error TS2430: Interface 'CatTransformsRequest' incorrectly extends interface 'CatCatRequestBase'.
  Types of property 'size' are incompatible.
    Type 'number' is not assignable to type '"" | "p" | "g" | "m" | "k" | "t"'.
make: *** [Makefile:12: generate] Error 2
```

The `?size` parameter is indeed overloaded for these APIs. Not sure what to do about that.

specification/_spec_utils/behaviors.ts
pquentin: I would add a new `UnitSize` enum in `_types/common.ts` next to `Bytes`.

specification/shutdown/delete_node/ShutdownDeleteNodeRequest.ts
pquentin: Yes, thank you!

pquentin:
> It's telling me to run `prettier` but I do not have this tool.

`make contrib` usually takes care of it, but it currently fails due to the `size` issue, so I've used `make spec-format-fix` instead.

> The `?size` parameter is indeed overloaded for these APIs. Not sure what to do about that.

The last time we had such an issue, we decided to remove the affected parameter from the common cat parameters and explicitly place it in each API where it made sense. Doing this will be easier than fighting the TypeScript compiler which correctly errors out here.

DaveCTurner:
Ok I looked more deeply and I think the `?size` parameter is a complete no-op, so I've just dropped it here.

Edit: also opened https://github.com/elastic/elasticsearch/pull/134871 to drop the `SizeValue` concept entirely.

pquentin (approved):
Thank you! LGTM.

DaveCTurner:
Thanks Quentin. Do I need to wait for https://github.com/elastic/elasticsearch/pull/134870 to land and then adjust things here, or is it ok to merge this first?

pquentin:
It's OK to merge this first. We have automation to sync the rest-api-spec from Elasticsearch to the specification that will remove the errors. (I did check that the changes were the same in both places.)

PR #5309, opened 2025-09-17
Add ingest field access pattern to specification

jbaiera:
Adds the new field access pattern to the specification. I didn't update any example code because this is not really expected to be a common option to set.

Relates to:
* https://github.com/elastic/elasticsearch/pull/133270
* https://github.com/elastic/docs-content/pull/2994

Closes #5310

### DIFF
--- a/specification/ingest/_types/Pipeline.ts
+++ b/specification/ingest/_types/Pipeline.ts
@@ -73,6 +73,13 @@ export class Pipeline {
    * @availability serverless
    */
   modified_date_millis?: EpochTime<UnitMillis>
+  /**
+   * Controls how processors in this pipeline should read and write data on a document's source.
+   * @server_default classic
+   * @availability stack since=9.2.0
+   * @availability serverless
+   */
+  field_access_pattern?: FieldAccessPattern
 }
 
 // Unused .. but let's keep it for now
@@ -103,4 +110,9 @@ export class PipelineConfig {
    * Processors run sequentially in the order specified.
    */
   processors: ProcessorContainer[]
+}
+
+export enum FieldAccessPattern {
+  classic,
+  flexible,
 }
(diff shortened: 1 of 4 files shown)


specification/ingest/_types/Pipeline.ts
```diff
@@ -73,6 +73,13 @@ export class Pipeline {
    * @availability serverless
    */
   modified_date_millis?: EpochTime<UnitMillis>
+  /**
+   * Controls how processors in this pipeline should read and write data on a document's source.
+   * @server_default classic
+   * @availability stack since=9.2.0
+   * @availability serverless
+   */
+  field_access_pattern?: FieldAccessPattern
```
pquentin, line 82: Sorry, the `ingest-common` module tests are not part of our validation, so we don't have the results of `340_flexible_access_pattern.yml`. The `Pipeline` class is used for three things:

 * `ingest.get_pipeline` response
 * `ingest.simulate` request
 * `simulate.ingest` request

Does it make sense in those three cases?

jbaiera:
> but next time please open this from a branch instead of a fork

Ah, my apologies! Didn't realize that was a restriction. Will do in the future. Would you like me to reopen it as a branch?

specification/ingest/_types/Pipeline.ts
jbaiera: Yes, anything that specifies a pipeline definition (that is, any configuration map that will be parsed as a pipeline) can specify this property. It is also returned as an optional field when querying it with `get_pipeline`. It is mirrored on the put pipeline request because that request object is effectively a copy of the `Pipeline` class.

pquentin (approved):
Thanks! LGTM.

No need to open from a branch as no new changes are planned.

Issue #5215, opened 2025-08-27
Generate Elasticsearch rest-api-spec from the Elasticsearch specification


pquentin:
To this day, https://github.com/elastic/elasticsearch/tree/main/rest-api-spec is maintained by hand by Elasticsearch developers. It currently serves two purposes:

 * Generate queries in YAML tests
 * Generate the Ruby 8.x / PHP 8.x / PHP 9.x clients.

However, we don't want to maintain two specifications. For the most part, the Elasticsearch specification is a superset of the rest-api-spec. We have been working hard for the past year to remove differences:

<img width="1200" height="600" alt="Image" src="https://github.com/user-attachments/assets/ca5794ba-9c60-48b7-b4ec-827b2bfd3b2f" />

So we should now strive to generate rest-api-spec for the others.

## Details

* Some APIs are *not* in the Elasticsearch specification, either because they're still experimental and behind a feature flag, or just internal.
* They are many differences in default values. They are only used for docs, but it would still be valuable to not lose this information.
* Some information is implicit from rest-api-spec. For example, `async_search.submit` has `"body": {}` but we would generate the more explicit `"body": {"required": false}`
* Mapping complex types to string/list/object isn't that easy
* Some availability/visibility information is out-of-date in the rest-api-spec
* Documentation URLs are out of date

The above is just an initial list, I only looked at a small percentage of the 560+ APIs we have.

pquentin:
Visibility/Stability PRs:

- [x] https://github.com/elastic/elasticsearch/pull/133814
- [x] https://github.com/elastic/elasticsearch/pull/133819
- [x] https://github.com/elastic/elasticsearch/pull/133912
- [x] https://github.com/elastic/elasticsearch/pull/133914
- [x] https://github.com/elastic/elasticsearch/pull/133915
- [x] https://github.com/elastic/elasticsearch/pull/133981
- [x] https://github.com/elastic/elasticsearch-specification/pull/5235
- [x] https://github.com/elastic/elasticsearch/pull/136642

pquentin:
Required request body PRs:

- [x] https://github.com/elastic/elasticsearch-specification/pull/5271
- [x] https://github.com/elastic/elasticsearch-specification/pull/5465
- [x] https://github.com/elastic/elasticsearch-specification/pull/5469
- [x] https://github.com/elastic/elasticsearch/pull/134425
- [x] https://github.com/elastic/elasticsearch/pull/136356

pquentin:
[...]

PR #5312, opened 2025-09-18
Remove '| string' in query parameters

pquentin:
The `| string` isn't meaningful in those two cases and this helps #5215. I'm skipping backports since this is a breaking change, though.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -8330,6 +8330,70 @@
         ]
       }
     },
+    "/_create_from/{source}/{dest}": {
+      "put": {
+        "tags": [
+          "migration"
+        ],
+        "summary": "Create an index from a source index",
+        "description": "Copy the mappings and settings from the source index to a destination index while allowing request settings and mappings to override the source values.",
+        "operationId": "indices-create-from",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/indices.create_from-source"
+          },
+          {
+            "$ref": "#/components/parameters/indices.create_from-dest"
+          }
+        ],
+        "requestBody": {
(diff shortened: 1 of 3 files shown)


specification/ml/get_overall_buckets/MlGetOverallBucketsRequest.ts
```diff
@@ -139,7 +139,7 @@ export interface Request extends RequestBase {
     /**
      * Refer to the description for the `overall_score` query parameter.
      */
-    overall_score?: double | string
+    overall_score?: double
```
pquentin, line 142: While this one isn't a query parameter, it feels like a straight-up bug.

Parsing is done in https://github.com/elastic/elasticsearch/blob/eb3bc2f2929209461802884ea01f77d3b3fcf96a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/GetOverallBucketsAction.java#L73

pquentin:
At least I have the reason why this was done: YAML tests!
 * full_id is [sent as a boolean](https://github.com/elastic/elasticsearch/blob/eb3bc2f2929209461802884ea01f77d3b3fcf96a/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cat.nodes/10_basic.yml#L103), but gets [recorded as a string somehow](https://github.com/elastic/clients-flight-recordings/blob/20cbc61113f0832a4f98a8de20a3b6437f1b23ab/elasticsearch/cat.nodes/b009d4a35ac0e026c4ed560290ac93b0.json#L9): I attempted a fix in https://github.com/elastic/clients-flight-recorder/pull/1376
 * overall_score is a YAML test bug: https://github.com/elastic/elasticsearch/pull/134999

pquentin:
Merging! There's one failure left on `cat.nodes`, but it will be fixed with the next clients-flight-recorder run.

PR #5322, opened 2025-09-19
[Transform] Specify use_point_in_time in settings

prwhelan:
`use_point_in_time` has been in the docs since before 8.18

Resolve https://github.com/elastic/elasticsearch/issues/134841

### DIFF
--- a/specification/transform/_types/Transform.ts
+++ b/specification/transform/_types/Transform.ts
@@ -130,6 +130,12 @@ export class Settings {
    * @server_default 500
    */
   max_page_search_size?: integer
+  /**
+   * Specifies whether the transform checkpoint will use the Point In Time API while searching over the source index.
+   * @doc_id point-in-time-api
+   * @server_default true
+   */
+  use_point_in_time?: boolean
 
   /**
    * If `true`, the transform runs in unattended mode. In unattended mode, the transform retries indefinitely in case
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -22396,6 +22396,7 @@ export interface TransformSettings {
   deduce_mappings?: boolean
   docs_per_second?: float
   max_page_search_size?: integer
+  use_point_in_time?: boolean
   unattended?: boolean
 }
 


specification/transform/_types/Transform.ts
```diff
@@ -130,6 +130,12 @@ export class Settings {
    * @server_default 500
    */
   max_page_search_size?: integer
+  /**
+   * Specifies whether the transform checkpoint will use the Point In Time API while searching over the source index.
+   * @ext_doc_id point-in-time-api
```
pquentin, line 139: @elastic/developer-docs Is this a correct usage of `ext_doc_id`?

This Transform API can be configured to use the PIT API, and here `point-in-time-api` links to the point-in-time API docs.

specification/transform/_types/Transform.ts
```diff
@@ -130,6 +130,12 @@ export class Settings {
    * @server_default 500
    */
   max_page_search_size?: integer
+  /**
+   * Specifies whether the transform checkpoint will use the Point In Time API while searching over the source index.
```
pquentin, line 134: What happens if you set it to false? It uses the scroll API?

specification/transform/_types/Transform.ts
kosabogi: Hey @pquentin, the `@ext_doc_id` looks fine here. I see it’s already in the `table.csv` file as well, so it should be good. What we usually do to make sure the links appear correctly is test them locally by running these (you should have the bump CLI installed):

```
make generate
make transform-to-openapi
bump preview output/openapi/elasticsearch-openapi.json
```

specification/transform/_types/Transform.ts
prwhelan: Nah it just uses a search request without PIT: https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/TransformIndexer.java#L1154

specification/transform/_types/Transform.ts
prwhelan: Seems to have worked as intended:
image
And it correctly links to https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-open-point-in-time

leemthompo: FYI we could add a description field to the link definition in https://github.com/elastic/elasticsearch-specification/blob/34dfe8bd8f76046b98fbfbd7e07c85b9c1db55cc/specification/_doc_ids/table.csv#L521

and this would be render nicer link text than just `External documentation` 💅

specification/transform/_types/Transform.ts
leemthompo: ideally we'd explain the pros/cons differences of each approach, or at least outline _why_ you'd want to use the PIT, rather than just the how :)

specification/transform/_types/Transform.ts
prwhelan: Do you have an example of that?

specification/transform/_types/Transform.ts
prwhelan: should that go on the PIT page? in either case, we'd need someone familiar with PIT to go into that detail, I'm just trying to show what users can configure rather than keeping the setting hidden

specification/transform/_types/Transform.ts
leemthompo: this could be as simple as `Point in Time API`

leemthompo: can see an example in this PR: https://github.com/elastic/elasticsearch-specification/pull/5331/files#diff-afcc28b0721ba019d194536c1ff7cb6117ac7db92b094fbc8723ec371771385d

pquentin: I agree that it would be nice to name this "Open a point in time API" instead of "External documentation."

specification/transform/_types/Transform.ts
pquentin: I think what I missed is that transforms use composite aggregations, which offer pagination without using PIT. However, transforms use PIT by default, as it returns more correct results. That said, as it can be expensive if many PITs are open at any given time, this can be disabled. Is my understanding correct?

This is not about explaining what PIT does (especially given that we have a link), but making it clear what the setting does. For example, `max_page_search_size` is clear here!

leemthompo: > This is not about explaining what PIT does (especially given that we have a link), but making it clear what the setting does.

right Quentin said it much clearer than me 😄

specification/transform/_types/Transform.ts
prwhelan: It didn't seem to work, but perhaps I did it wrong:
image

All the others from that example PR still show `External documentation` for me so :shrug:

pquentin:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.0|[](https://github.com/elastic/elasticsearch-specification/pull/5463)|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/5464)|
|❌|8.18|Conflict resolution was aborted by the user|

### Manual backport
To create the backport manually run:
```
backport --pr 5322
```

### Questions ?

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.18|[](https://github.com/elastic/elasticsearch-specification/pull/5467)|

### Questions ?

PR #5323, opened 2025-09-19
Add serverless refresh interval

shainaraskas:
The stack and serverless elasticsearch APIs have a different default minimum refresh interval. reflecting this in the API docs for `/{index}/_refresh` to reduce confusion.

https://bump.sh/preview/5d9553f5-3c52-433b-8b03-84c09359bc57/operation/operation-indices-refresh

@elastic/developer-docs let me know if I'm breaking the rules with my approach / missed something special about adding serverless-specific details.

related to https://github.com/elastic/docs-content/issues/1643

~assuming this doesn't need to be backported because it is serverless-only. let me know if that's wrong because it's a shared description~


specification/indices/refresh/IndicesRefreshRequest.ts
```diff
@@ -28,6 +28,8 @@ import { ExpandWildcards, Indices } from '@_types/common'
  * By default, Elasticsearch periodically refreshes indices every second, but only on indices that have received one search request or more in the last 30 seconds.
  * You can change this default interval with the `index.refresh_interval` setting.
  *
+ * In Elastic Cloud Serverless, the default refresh interval is 5 seconds.
```
pquentin: Is it always 5 seconds, or do you also need to receive at least one search request in the last 30 seconds?

pquentin (commented):
Thanks! Regarding backporting, it could go either way, but since it's shared I'd backport. It will also reduce the likelihood of future conflicts.

specification/indices/refresh/IndicesRefreshRequest.ts
shainaraskas: verified with @/fcofdez (vacationing) that the search idle this describes does not apply in serverless. I updated the line to reflect this:

> In Elastic Cloud Serverless, the default refresh interval is 5 seconds across all indices.

pquentin (approved):
Thanks! LGTM.

Issue #5299, opened 2025-09-16
Add include_cps_response to ES|QL query spec


luigidellaquila:
With https://github.com/elastic/elasticsearch/pull/134446 we are adding a new query parameter to ES|QL request: `include_cps_response`

The behavior is the same as `include_ccs_response`, but it will be available only on Serverless, in the scope of Cross-Project Search.

quux00:
Example of how I recently added some specs that are CPS specific:

https://github.com/elastic/elasticsearch-specification/pull/5259/files#diff-4114e335296798822015a064d77c5c6d14c797a89c76a0d08b63dcf5d4028651R231

I think the `include_cps_response` support is also behind this feature flag, so you'll probably want to do the same.

PR #5328, opened 2025-09-22
ES|QL: add include_execution_metadata to the query spec

luigidellaquila:
Adding `include_execution_metadata` parameter to ES|QL query specification.
See https://github.com/elastic/elasticsearch/pull/134446

Closes: https://github.com/elastic/elasticsearch-specification/issues/5299

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -6509,7 +6509,12 @@
                     }
                   },
                   "include_ccs_metadata": {
-                    "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.",
+                    "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.\nThis is a synonym for `include_execution_metadata`, the two parameters can be used interchangeably but not together.",
+                    "default": false,
+                    "type": "boolean"
+                  },
+                  "include_execution_metadata": {
+                    "description": "When set to `true` and performing a cross-cluster or cross-project query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.\nThis is a synonym for `include_ccs_metadata`, the two parameters can be used interchangeably but not together.",
                     "default": false,
                     "type": "boolean"
                   }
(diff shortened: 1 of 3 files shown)


l-trotta:
thanks for the contribution @luigidellaquila! considering `include_ccs_metadata` and `include_execution_metadata` are synonyms and share the exact same description, I think it makes sense to add `include_execution_metadata` as an alias of the existing `include_ccs_metadata`: 
```ts
...
 * @server_default false
 * @aliases include_execution_metadata
 */
include_ccs_metadata?: boolean
```

Also, should this be in `specification/esql/async_query/AsyncQueryRequest.ts` too?

luigidellaquila:
Thanks for checking @l-trotta
Since from now on we want `include_execution_metadata` to be the preferred way, do you think it makes sense to do it the other way around, ie. have `include_ccs_metadata` as an alias for `include_execution_metadata`? Or is it unusual since `include_ccs_metadata` was already there?

Or maybe we should keep both with the respective `@alias` definitions?

output/openapi/elasticsearch-serverless-openapi.json
```diff
@@ -6509,7 +6509,12 @@
...
-                    "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.",
+                    "description": "When set to `true` and performing a cross-cluster query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.\nThis is a synonym for `include_execution_metadata`, the two parameters can be used interchangeably but not together.",
+                    "default": false,
+                    "type": "boolean"
+                  },
+                  "include_execution_metadata": {
+                    "description": "When set to `true` and performing a cross-cluster or cross-project query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.\nThis is a synonym for `include_ccs_metadata`, the two parameters can be used interchangeably but not together.",
```
idegtiarenko: I am not very familiar with this project, but since we have a distinct file for serverless, should the regular one not include `include_execution_metadata`?

l-trotta: if `include_execution_metadata` has not been marked as only available in stack or serverless, then it will be added to both files

luigidellaquila: `include_execution_metadata` will be available in stateful as well, so I guess it's fine to have a single definition with 

```
 * @availability stack ...
 * @availability serverless
```

luigidellaquila:
OK, I went for just adding `@aliases` to the existing parameters. 
@l-trotta in case you want to have another look

l-trotta:
@luigidellaquila thanks for the changes! ~~I'll just quickly check with the team because I need to make sure this won't be a breaking change when `include_ccs_metadata` gets removed.~~ LGTM!

luigidellaquila:
Thanks all! Merging

PR #5335, opened 2025-09-24
`CompositeAggregation.sources` expects an array of single-key objects

flobernd:
As titled. Also marks `CompositeAggregationSource` as a container since the aggregations are mutually exclusive.

This fix as well greatly improves the quality of the generated code for the .NET client (see [#8704](https://github.com/elastic/elasticsearch-net/issues/8704)).

Server code:
https://github.com/elastic/elasticsearch/blob/a59c182f9f7e9d1bf3d6eecbc0e44f24ff91d053/server/src/main/java/org/elasticsearch/search/aggregations/bucket/composite/CompositeValuesSourceParserHelper.java#L78

No backport, since this is a breaking change for statically typed languages.

### DIFF
--- a/specification/_types/aggregations/bucket.ts
+++ b/specification/_types/aggregations/bucket.ts
@@ -148,6 +148,9 @@ export class CompositeAggregation extends BucketAggregationBase {
   sources?: Array<SingleKeyDictionary<string, CompositeAggregationSource>>
 }
 
+/**
+ * @variants container
+ */
 export class CompositeAggregationSource {
   /**
    * A terms aggregation.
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thank you, great catch! In addition to the server code, this is [nicely documented](https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation#_value_sources) and [there's a YAML test for that](https://github.com/elastic/elasticsearch/blob/32ff026c8a08a1af53ce2e1bf984735a7e12931e/modules/aggregations/src/yamlRestTest/resources/rest-api-spec/test/aggregations/composite.yml#L806-L836).

LGTM.

ianschmaltz:
Thanks @flobernd and team for the clarification and fix.

We were exactly hitting this bug while migrating from NEST to `Elastic.Clients.Elasticsearch` (v9.1.7) and really appreciate the fast response. We'll keep testing with the suggested workaround until the next client release includes this fix officially.

Thanks again for the support and transparency!

PR #5336, opened 2025-09-24
Update availability documentation in GraphExploreRequest

georgewallace:
Per https://github.com/elastic/elasticsearch/commit/487b217afe200028d1f27a25546e29cf8160983f the Graph API was removed from serverless scope . So the spec should be updated to reflect that.


georgewallace:
@javanna per your PR, I am updating the ES OpenAPI spec to update our documentation.

pquentin (approved):
Thanks! LGTM.

Added a 9.1 backport for good measure even though this is only for serverless.

PR #5370, opened 2025-09-30
Fix cluster.put_component_template API

pquentin:

### DIFF
--- a/specification/indices/_types/IndexState.ts
+++ b/specification/indices/_types/IndexState.ts
@@ -22,6 +22,7 @@ import { TypeMapping } from '@_types/mapping/TypeMapping'
 import { DataStreamLifecycle } from '@indices/_types/DataStreamLifecycle'
 import { Dictionary } from '@spec_utils/Dictionary'
 import { Alias } from './Alias'
+import { DataStreamOptionsTemplate } from './DataStreamOptions'
 import { IndexSettings } from './IndexSettings'
 
 export class IndexState {
@@ -31,6 +32,11 @@ export class IndexState {
   /** Default settings, included when the request's `include_default` is `true`. */
   defaults?: IndexSettings
   data_stream?: DataStreamName
+  /**
+   * @availability stack since=8.18.0 stability=stable
+   * @availability serverless stability=stable
+   */
+  data_stream_options?: DataStreamOptionsTemplate
   /**
    * Data stream lifecycle applicable if this is a data stream.
    * @availability stack since=8.11.0 stability=stable
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -60504,6 +60504,14 @@
               }
             ]
           },
+          "data_stream_options": {
+            "x-state": "Generally available",
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/indices._types.DataStreamOptionsTemplate"
+              }
+            ]
+          },
           "lifecycle": {
             "description": "Data stream lifecycle applicable if this is a data stream.",
             "x-state": "Generally available",
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -11981,6 +11981,7 @@ export interface IndicesIndexState {
   settings?: IndicesIndexSettings
   defaults?: IndicesIndexSettings
   data_stream?: DataStreamName
+  data_stream_options?: IndicesDataStreamOptionsTemplate
   lifecycle?: IndicesDataStreamLifecycle
 }
 


specification/indices/_types/IndexState.ts
```diff
@@ -31,6 +32,11 @@ export class IndexState {
   /** Default settings, included when the request's `include_default` is `true`. */
   defaults?: IndexSettings
   data_stream?: DataStreamName
+  /**
+   * @availability stack since=8.18.0 stability=stable
+   * @availability serverless stability=stable
+   */
+  data_stream_options?: DataStreamOptionsTemplate
```
nielsbauman: I don't think this is the right place to add this field. I see the class `IndexState` is indeed used for the `template` field of component templates here: https://github.com/elastic/elasticsearch-specification/blob/8bb614b7de9b18efa78595f575c02b142f369de0/specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts#L87-L90
But it's also used in the `GET {index}` API response: https://github.com/elastic/elasticsearch-specification/blob/8bb614b7de9b18efa78595f575c02b142f369de0/specification/indices/get/IndicesGetResponse.ts#L26
I'd say that the name `IndexState` matches most with the second use case, and that maybe at the time its fields happened to match the `template` field of the `PUT _component_template` API, but they don't match anymore; the `data_stream` field is not valid in the `template` field for component templates (or any kind of templates for that matter, it's only valid in the root body of index templates, see [here](https://github.com/elastic/elasticsearch-specification/blob/69c9eee0cb67d05a2e7b58348dccd84a572541d4/specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts#L103)).

I see that [00be10a](https://github.com/elastic/elasticsearch-specification/commit/00be10ac5ace4211ff41c0c640eb9b6ecaf10f07) did try to add the `data_stream_options` field in some places, but it doesn't look 100% correct to me. Let me have a look and try to correct some things.

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
```diff
@@ -20,7 +20,7 @@
 import { RequestBase } from '@_types/Base'
 import { Metadata, Name, VersionNumber } from '@_types/common'
 import { Duration } from '@_types/Time'
-import { IndexState } from '@indices/_types/IndexState'
+import { IndexTemplateMapping } from '@indices/put_index_template/IndicesPutIndexTemplateRequest'
```
nielsbauman, line 23: This import isn't particularly pretty. I'll leave it up to you whether this is fine or if we should extract the `IndexTemplateMapping` class to a dedicated file. If we're extracting it (which I'm in favor of), we should probably also rename it.

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
```diff
@@ -87,7 +87,7 @@ export interface Request extends RequestBase {
     /**
      * The template to be applied which includes mappings, settings, or aliases configuration.
      */
-    template: IndexState
+    template: IndexTemplateMapping
```
nielsbauman, line 91: Both component templates and composable/index templates define the `template` field. In the ES code, they both point to the same class and are thus parsed in the same way, so we can/should use the same class in the spec too. (just explaining why I made this change)

specification/indices/_types/IndexTemplate.ts
```diff
@@ -145,5 +145,5 @@ export class IndexTemplateSummary {
    * @availability stack since=8.19.0 stability=stable
    * @availability serverless stability=stable
    */
-  data_stream_options?: DataStreamOptionsTemplate | null
+  data_stream_options?: DataStreamOptions
```
nielsbauman, line 148: According to the description of `DataStreamOptionsTemplate`:
> Data stream options template contains the same information as DataStreamOptions but allows them to be set explicitly to null.

Which means it's only relevant for `PUT` requests, and this `IndexTemplateSummary`/`IndexTemplate` is only used in retrieval APIs. That's why I changed it to `DataStreamOptions`.

specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
```diff
@@ -178,4 +179,9 @@ export class IndexTemplateMapping {
    * @availability serverless stability=stable
    */
   lifecycle?: DataStreamLifecycle
+  /**
+   * @availability stack since=8.19.0 stability=stable
```
nielsbauman, line 187: While the options are technically available in `8.18.0` (which is the version you originally used in your changes), the other data stream option changes (from [00be10a](https://github.com/elastic/elasticsearch-specification/commit/00be10ac5ace4211ff41c0c640eb9b6ecaf10f07)) all specify `8.19.0`, which was probably done because the failure store feature flag (currently the only field in data stream options) was only removed in `8.19.0`, so I'm going for consistency here.

nielsbauman (commented):
I reverted your commit and pushed some other changes; I hope you don't mind :). My intention with my commit is to indicate what the spec should look like for these changes. I'll let it up to you if you want to make code style changes, extract parts, etc. I ran `make contrib` right before pushing, so I'm hoping the generated files should be ok like this. Let me know what you think!

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
pquentin: Yes, we should move to a shared place.  Options include:

 1. `specification/indices/_types`
 2. `specification/cluster/_types`
 3. `specification/_types`

Not sure which is best. Probably `cluster` since it's where `Template` lives in the Elasticsearch code. Unfortunately, moving a class is a breaking change, so that will be 9.3 only.

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
pquentin: Using the same class as Elasticsearch is fine and intended!

specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
```diff
@@ -181,4 +182,9 @@ export class IndexTemplateMapping {
...
+   * @availability serverless stability=stable
+   */
+  data_stream_options?: DataStreamOptionsTemplate
```
pquentin: According to this failing test

```ts
// Test file: /test/free/data_stream/230_data_stream_options.yml
// Test name: Test resetting data stream options in template composition

import * as T from '../types'

expectAssignable<T.IndicesPutIndexTemplateRequest>({
  "body": {
    "composed_of": [
      "my-component-template"
    ],
    "index_patterns": [
      "no-data-stream"
    ],
    "template": {
      "data_stream_options": null,
      "settings": {
        "number_of_replicas": 0
      }
    }
  },
  "name": "my-valid-template"
})
```

We need this change:

```suggestion
  data_stream_options?: DataStreamOptionsTemplate | null
```

pquentin (commented):
Thank you for this great work, and sorry for the huge delay.

There's one failing test left for `cluster.put_component_template`:

```ts
// Test file: /test/free/cluster.component_template/10_basic.yml
// Test name: Add data stream lifecycle with downsampling

import * as T from '../types'

expectAssignable<T.ClusterPutComponentTemplateRequest>({
  "body": {
    "template": {
      "lifecycle": {
        "data_retention": "10d",
        "downsampling": [
          {
            "after": "1d",
            "fixed_interval": "5m"
          }
        ],
        "downsampling_method": "last_value"
      }
    }
  },
  "name": "test-lifecycle"
})
```

The issue is on line 12:

> Property 'rounds' is missing in type '{ after: string; fixed_interval: string; }[]' but required in type 'IndicesDataStreamLifecycleDownsampling'.


Also, are we missing `created_date` and `modified_date` in the `cluster.put_component_template` body? [Server code](https://github.com/elastic/elasticsearch/blob/e6ee2fee1833e697a3def14d1fd216a04246d0c7/server/src/main/java/org/elasticsearch/cluster/metadata/ComponentTemplate.java#L55-L62)

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
nielsbauman: Yep, moving it to `specification/cluster/_types` works for me :+1:. I'm having trouble coming up with a good name though... `IndexTemplate` is already used for retrieving templates. Is there some kind of naming convention for types that are used for PUT requests?

nielsbauman:
@pquentin regarding the failing test related to downsampling, I think @gmarouli's https://github.com/elastic/elasticsearch-specification/pull/5593 should fix that.

nielsbauman:
> Also, are we missing created_date and modified_date in the cluster.put_component_template body?

Nope, we don't allow specifying those in the API: https://github.com/elastic/elasticsearch/blob/7cb125be9e971dee0fbd8b0c8d2d0d796b126d39/server/src/main/java/org/elasticsearch/action/admin/indices/template/put/PutComponentTemplateAction.java#L82-L93
Only ES itself can set those fields.

nielsbauman:
I noticed another API that used `IndexTemplate` instead of `IndexTemplateMapping` incorrectly. I modifed it in this PR, but I'm also ok with extracting that change to another PR, but I'll let that up to you.

specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
```diff
@@ -181,4 +182,9 @@ export class IndexTemplateMapping {
    * @availability serverless stability=stable
    */
   lifecycle?: DataStreamLifecycle
```
nielsbauman, line 185: Actually, this `lifecycle` field in the `template` object can also explicitly be set to `null`, and the same holds for all the fields inside `DataStreamLifecycle`. Do you want to address that in this PR or a follow-up one?

specification/indices/simulate_index_template/IndicesSimulateIndexTemplateRequest.ts
```diff
@@ -68,5 +68,5 @@ export interface Request extends RequestBase {
     include_defaults?: boolean
   }
   /** @codegen_name index_template */
-  body?: IndexTemplate
+  body?: IndexTemplateMapping
```
pquentin: This causes two new failures with `composed_of` that `indices.simulate_index_template` supports. Here's one:

```ts
// Test file: /test/free/indices.simulate_index_template/10_basic.yml
// Test name: Simulate index template specifying a new template

import * as T from '../types'

expectAssignable<T.IndicesSimulateIndexTemplateRequest>({
  "body": {
    "composed_of": [
      "ct"
    ],
    "index_patterns": "test*",
    "priority": 15,
    "template": {
      "aliases": {
        "test_alias": {}
      },
      "settings": {
        "index": {
          "blocks": {
            "write": true
          }
        }
      }
    }
  },
  "name": "test"
})
```

specification/indices/put_index_template/IndicesPutIndexTemplateRequest.ts
pquentin: Happy to do this as a follow-up, as long as this is before 9.3, to avoid breaking those APIs twice.

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
pquentin: No naming convention, I'm not sure what to use. Is keeping `IndexTemplateMapping` that bad? It will reduce breakage for clients that don't care about the namespace of types.

specification/indices/simulate_index_template/IndicesSimulateIndexTemplateRequest.ts
pquentin: It looks like `indices.simulate_index_template` accepts a [ComposableIndexTemplate](https://github.com/elastic/elasticsearch/blob/b344a152e411c33185ccd1bfdca0d752918537a3/server/src/main/java/org/elasticsearch/rest/action/admin/indices/RestSimulateTemplateAction.java#L53C52-L53C75) object, while `cluster.put_component_template` accepts a [ComponentTemplate](https://github.com/elastic/elasticsearch/blob/b344a152e411c33185ccd1bfdca0d752918537a3/server/src/main/java/org/elasticsearch/rest/action/admin/indices/RestPutComponentTemplateAction.java#L61C42-L61C59).

nielsbauman: Ahh yeah I was mistaken indeed. The current value of `IndexTemplate` is not really correct, because it contains things like `created_date_millis` which you're not allowed to specify in the request, but it's not a disaster. I'll remove that commit, thanks for flagging.

pquentin:
Thanks for updating the branch @gmarouli! We have one failure left related to downsampling, with `downsampling_method` missing: https://github.com/elastic/elasticsearch/pull/137023. This can be treated in a different PR.

nielsbauman:
@pquentin, @gmarouli already opened https://github.com/elastic/elasticsearch-specification/pull/5594 for that :)

pquentin:
Thanks! We merged this one, which indeed fixed the issue (and fixed three other APIs along the way!)

The only item left is https://github.com/elastic/elasticsearch-specification/pull/5370#discussion_r2392533795

specification/cluster/put_component_template/ClusterPutComponentTemplateRequest.ts
nielsbauman: I mean, I don't think the name itself really makes any sense, but if it avoids breakage, I'm fine with leaving it as is. Then let's just stick to moving the class. Could you take care of that perhaps? :pray:

pquentin: Sure! So I did it locally, which made me realize that we already have a case of an endpoint doing this:

https://github.com/elastic/elasticsearch-specification/blob/6566f6962beceaa27c50ac31100ac22f62aea822/specification/indices/simulate_template/IndicesSimulateTemplateRequest.ts#L25C10-L25C30

So let's merge without doing this for now, and I'll ask to @l-trotta what's the best thing to do here, as the location of types mostly affects the Java client.

pquentin:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.2|[](https://github.com/elastic/elasticsearch-specification/pull/5803)|
|✅|9.1|[](https://github.com/elastic/elasticsearch-specification/pull/5804)|
|❌|8.19|Conflict resolution was aborted by the user|

### Manual backport
To create the backport manually run:
```
backport --pr 5370
```

### Questions ?

pquentin:
8.19 backport currently blocked on https://github.com/elastic/elasticsearch-specification/pull/5807

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/5833)|

### Questions ?

PR #5371, opened 2025-09-30
Add age_in_millis to ILM Explain Response

pquentin:
As done in https://github.com/elastic/elasticsearch/pull/128866

### DIFF
--- a/specification/ilm/explain_lifecycle/types.ts
+++ b/specification/ilm/explain_lifecycle/types.ts
@@ -19,7 +19,13 @@
 
 import { IndexName, Name, VersionNumber } from '@_types/common'
 import { integer } from '@_types/Numeric'
-import { DateTime, Duration, EpochTime, UnitMillis } from '@_types/Time'
+import {
+  DateTime,
+  Duration,
+  DurationValue,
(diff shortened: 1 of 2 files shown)


specification/ilm/explain_lifecycle/types.ts
```diff
@@ -30,6 +36,11 @@ export class LifecycleExplainManaged {
   action_time_millis?: EpochTime<UnitMillis>
   /* `lifecycle_date` as a duration */
   age?: Duration
+  /**
+   * @availability stack since=9.2.0 stability=stable
+   * @availability serverless stability=stable
```
nielsbauman: ILM is not available as a whole in serverless. Not sure what value this should have, but it shouldn't be serverless.

nielsbauman (approved):
LGTM with one small comment regarding availability. Thanks!

specification/ilm/explain_lifecycle/types.ts
pquentin: Oh, of course, thanks! I fixed it because it was wrong. Thankfully, it never made it to the serverless OpenAPI file.

PR #5399, opened 2025-10-03
Add deprecation notice for max_size rollover condition

lukewhiting:
Add deprecation notice for the `max_size` condition used for rollovers and ILM policies

Relates to: elastic/elasticsearch#130737


pquentin (approved):
Thanks! LGTM.

PR #5401, opened 2025-10-03
Allow hex-encoded byte query vectors

pquentin:
Added in https://github.com/elastic/elasticsearch/pull/105393. Fixes the following test:

```ts
// Test file: /test/free/search.vectors/170_knn_search_hex_encoded_byte_vectors.yml
// Test name: Knn search with hex string for byte field

import * as T from '../types'

expectAssignable<T.SearchRequest>({
  "body": {
    "knn": {
      "field": "my_vector_byte",
      "k": 10,
      "num_candidates": 100,
      "query_vector": "400ae2"
    },
    "size": 3
  },
  "index": "knn_hex_vector_index",
  "typed_keys": true
})
```


pquentin:
This will be handled as part of the more general work to accept base64 vectors.

Issue #5372, opened 2025-09-30
Specification fixes for 9.2


pquentin:
As part of the 9.2 feature freeze, I'd like your help to get the following changes in the specification. They're not all 9.2 specific.

## Search Relevance

New features

- [x] https://github.com/elastic/elasticsearch/pull/131058
  - https://github.com/elastic/elasticsearch-specification/pull/5386
- [x] https://github.com/elastic/elasticsearch/pull/135306
  - https://github.com/elastic/elasticsearch-specification/pull/5389 
- [x] https://github.com/elastic/elasticsearch/pull/119183
- [x] Bonus: https://github.com/elastic/elasticsearch/pull/111851 (added in 8.16)

More specific failures

- [x] Allow [chunking_settings: null](https://github.com/elastic/elasticsearch/blame/e0c41324c2eaa80cac94ad0febb8ea906b839c4b/x-pack/plugin/inference/src/yamlRestTest/resources/rest-api-spec/test/inference/25_semantic_text_field_mapping_chunking.yml#L193-L202) and [chunking_settings: {}](https://github.com/elastic/elasticsearch/blame/e0c41324c2eaa80cac94ad0febb8ea906b839c4b/x-pack/plugin/inference/src/yamlRestTest/resources/rest-api-spec/test/inference/25_semantic_text_field_mapping_chunking.yml#L211-L220) (added in https://github.com/elastic/elasticsearch/pull/121041). Not sure why they are accepted though to be honest. We could also skip the relevant tests.
  - https://github.com/elastic/elasticsearch-specification/pull/5408
- [x] ChunkRescorer: ["Nested" chunk_rescorer](https://github.com/elastic/elasticsearch/blob/e0c41324c2eaa80cac94ad0febb8ea906b839c4b/x-pack/plugin/inference/src/yamlRestTest/resources/rest-api-spec/test/inference/70_text_similarity_rank_retriever.yml#L816-L819), missing [chunk_size](https://github.com/elastic/elasticsearch/blob/e0c41324c2eaa80cac94ad0febb8ea906b839c4b/x-pack/plugin/inference/src/yamlRestTest/resources/rest-api-spec/test/inference/70_text_similarity_rank_retriever.yml#L749)
[...]

PR #5408, opened 2025-10-03
Support explicit null chunking_settings in semantic_text field mappings

kderusso:
Related to https://github.com/elastic/elasticsearch-specification/issues/5372


pquentin (approved):
Thanks! LGTM.

Merging without backporting for now, but will discuss them with the team later this week.

PR #5412, opened 2025-10-06
Move examples for completion task of OpenAI and Amazon Bedrock to the dedicated sections

Jan-Kazlouski-elastic:
Fixing a mix-up between OpenAI and Amazon Bedrock examples for completion task


pquentin (approved):
Thanks! LGTM.

PR #5418, opened 2025-10-06
Add cartesian_bounds, cartesian_centroid and change_point aggregations

l-trotta:
-  Missing cartesian_bounds aggregation: [Request](https://github.com/elastic/elasticsearch/blob/ae6d18037953609d58a2069121e37b484eba5c87/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/metrics/CartesianBoundsAggregationBuilder.java#L29), [Response](https://github.com/elastic/elasticsearch/blob/609e8059eb4949b17aef69d62720ee27ace0b637/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/metrics/InternalCartesianBounds.java#L27)
- Missing cartesian_centroid aggregation: [Request](https://github.com/elastic/elasticsearch/blob/ae6d18037953609d58a2069121e37b484eba5c87/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/metrics/CartesianCentroidAggregationBuilder.java#L30), [Response](https://github.com/elastic/elasticsearch/blob/638c86c5579e4561bcb17721c81a052fcf080437/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/metrics/InternalCartesianCentroid.java#L22)
- Missing change_point aggregation: [Request](https://github.com/elastic/elasticsearch/blob/19257125b19ff05a6b014f37693f9417ca579626/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/changepoint/ChangePointAggregationBuilder.java#L28), [Response](https://github.com/elastic/elasticsearch/blob/609e8059eb4949b17aef69d62720ee27ace0b637/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/changepoint/InternalChangePointAggregation.java#L22)

These should all be safe to backport.


specification/_types/aggregations/AggregationContainer.ts
```diff
@@ -183,13 +186,31 @@ export class AggregationContainer {
    * @ext_doc_id search-aggregations-metrics-cardinality-aggregation
    */
   cardinality?: CardinalityAggregation
+  /**
+   * A metric aggregation that computes the spatial bounding box containing all values for a Point or Shape field.
+   * @ext_doc_id search-aggregations-metrics-cartesian-bounds-aggregation
```
pquentin, line 191: I believe you need an entry in specification/_doc_ids/table.csv for `ext_doc_id` to fully work. This also applies to `change_point` and `cartesian_centroid`.

specification/_types/aggregations/AggregationContainer.ts
```diff
@@ -183,13 +186,31 @@ export class AggregationContainer {
...
    * @ext_doc_id search-aggregations-bucket-categorize-text-aggregation
    * @availability stack stability=experimental
    * @availability serverless stability=experimental
    */
   categorize_text?: CategorizeTextAggregation
+  /**
+   * A sibling pipeline that detects, spikes, dips, and change points in a metric.
+   * Given a distribution of values provided by the sibling multi-bucket aggregation,
+   * this aggregation indicates the bucket of any spike or dip and/or the bucket at which
+   * the largest change in the distribution of values, if they are statistically significant.
```
pquentin, line 210: As someone who has used this aggregation in the past for performance regression detection, I'd like to mention an important caveat: 22 points are needed.

```suggestion
   * the largest change in the distribution of values, if they are statistically significant.
   * There must be at least 22 bucketed values. Fewer than 1,000 is preferred.
```

specification/_types/aggregations/Aggregate.ts
```diff
@@ -369,6 +385,68 @@ export class MultiBucketBase
...
+/** @variant name=indeterminable */
+export class Indeterminable {
+  reason: string
+}
+
+/** @variant name=non_stationary */
+export class NonStationary {
+  p_value: double
+  r_value: double
+  trend: string
```
pquentin, line 437: We could decide to make this an enum with `increasing` and `decreasing` values: https://github.com/elastic/elasticsearch/blob/414972d5c7f65fec11c9ac2b4f6c0baaf456abdc/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/changepoint/ChangeDetector.java#L482

pquentin (commented):
Great work, thank you! I especially like the careful specification for the IMO tricky change point aggregation.

specification/_types/aggregations/Aggregate.ts
l-trotta: in this case I was referring to https://github.com/elastic/elasticsearch/blob/f9c72c17d18859e3511bd17fbff759ebe965b812/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/changepoint/ChangeType.java#L254, which is what's used by the response

pquentin: Right, I noticed `trend` is a string in the server code, but in practice, there are only two values. Well, it's an implementation detail, so it's likely better not to rely on it. You're right.

l-trotta:
manually backported 8.19 https://github.com/elastic/elasticsearch-specification/pull/5439

PR #5424, opened 2025-10-06
Add specification for certificate_identity field for cross-cluster ap…

gmjehovich:
Add certificate_identity field to API key specs

Updates TypeScript specs for create/update cross-cluster API key endpoints and ApiKey response type to include certificate_identity field.

### DIFF
--- a/specification/security/_types/ApiKey.ts
+++ b/specification/security/_types/ApiKey.ts
@@ -100,6 +100,13 @@ export class ApiKey {
    * @availability serverless
    */
   access?: Access
+    /**
+     * The certificate identity associated with a cross-cluster API key.
+     * Restricts the API key to connections authenticated by a specific TLS certificate.
+     * Only applicable to cross-cluster API keys.
+     * @availability stack since=9.3.0
+     */
+    certificate_identity?: string
   /**
    * The profile uid for the API key owner principal, if requested and if it exists
    * @availability stack since=8.14.0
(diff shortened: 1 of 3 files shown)


gmjehovich:
Wondering if I need to make a new separate CertificateIdentity type, or if keeping it as string like I have now is fine.

Also, I didn't add anything to the query spec- Are we going to support querying by cert_identity?

Finally, I didn't include the Serverless availability tag, is this correct?

pquentin:
Thank you for this! Can you please open this from a branch to ensure CI runs? As mentioned in CONTRIBUTING.md.

> Wondering if I need to make a new separate CertificateIdentity type, or if keeping it as string like I have now is fine.

If it's an enum, it would make sense to add a type. Can you please link to the corresponding Elasticsearch PR so that I can look?

> Also, I didn't add anything to the query spec- Are we going to support querying by cert_identity?

What is the "query spec"?

> Finally, I didn't include the Serverless availability tag, is this correct?

If it's not meant for serverless than yes it is correct!

jfreden:
>Wondering if I need to make a new separate CertificateIdentity type, or if keeping it as string like I have now is fine.

Since this can be any string (a regex) I think keeping it as String makes sense. 

@pquentin this is the PR: https://github.com/elastic/elasticsearch/pull/134604

>Also, I didn't add anything to the query spec- Are we going to support querying by cert_identity?

No, we should not support that yet so makes sense to leave that out. 

>Finally, I didn't include the Serverless availability tag, is this correct?

Yes, that's correct. We currently do not support this in serverless

gmjehovich:
Closing. Opened new PR from a branch instead of a fork: https://github.com/elastic/elasticsearch-specification/pull/5446

PR #5425, opened 2025-10-06
Add `api_key` param to remote reindex

PeteGillinElastic:
This reflects the change made in
https://github.com/elastic/elasticsearch/pull/135949.

### DIFF
--- a/specification/_global/reindex/types.ts
+++ b/specification/_global/reindex/types.ts
@@ -125,13 +125,21 @@ export class RemoteSource {
    */
   host: Host
   /**
-   * The username to use for authentication with the remote host.
+   * The username to use for authentication with the remote host (required when using basic auth).
    */
   username?: Username
   /**
-   * The password to use for authentication with the remote host.
+   * The password to use for authentication with the remote host (required when using basic auth).
    */
   password?: Password
+  /**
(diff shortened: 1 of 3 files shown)


specification/_global/reindex/types.ts
```diff
@@ -125,13 +125,21 @@ export class RemoteSource {
...
    */
   username?: Username
   /**
-   * The password to use for authentication with the remote host.
+   * The password to use for authentication with the remote host (required when using basic auth).
    */
   password?: Password
+  /**
+   * The API key to use for authentication with the remote host (as an alternative to basic auth when the remote cluster is in Elastic Cloud).
+   * (It is not permitted to set this and also to set an Authorization header via {@link headers}.)
```
pquentin: While `@link` is [supported in JSDoc](https://jsdoc.app/tags-inline-link), I don't think it does anything here, especially in OpenAPI docs. Is this good enough?

```suggestion
   * (It is not permitted to set this and also to set an Authorization header via `headers`.)
```

pquentin (commented):
Thank you for opening this early!

specification/_global/reindex/types.ts
PeteGillinElastic: Yeah, absolutely. You're the expert! The thing being referenced is only a couple of lines above, so the reader shouldn't need the cross-reference.

pquentin (approved):

PR #5445, opened 2025-10-07
available_processors field in CAT Nodes

piotrsulkowski-elastic:
Cat nodes REST action will return an extra field:
available_processors

Part of #134692


pquentin:
Can you please open this PR from a branch, instead of a fork? See CONTRIBUTING.md for details.

PR #5454, opened 2025-10-09
Add default values from rest-api-spec

pquentin:


pquentin:
(Backports are a bit weird with this one because I backported the 9.2 commit instead of the main commit. But the result is the same.)

PR #5471, opened 2025-10-10
Remove top-level `size` from `ReindexRequest`

PeteGillinElastic:
This was removed from ES in 8.0.0 (see
https://github.com/elastic/elasticsearch/pull/43373).

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -23460,9 +23460,6 @@
                       }
                     ]
                   },
-                  "size": {
-                    "type": "number"
-                  },
                   "source": {
                     "description": "The source you are copying from.",
                     "allOf": [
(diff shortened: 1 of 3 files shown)


pquentin:
Thanks! Is it silently accepted, or does the server refuse it? I'd like to feel less bad about the backports :)

PeteGillinElastic:
> Thanks! Is it silently accepted, or does the server refuse it? I'd like to feel less bad about the backports :)

The server has thrown an error since 8.0.0, based on my reading of the PRs.

PeteGillinElastic:
Hi @pquentin, can you advise on the backporting here? I think this falls under

> If the API is usable, but fix is on the request side -> no backport, unless the API is partially usable and the fix unlocks a missing feature that has no workaround

because the API is absolutely usable, the only issue is that clients may provide an option that would result in a 400. But I may be interpreting that wrongly.

pquentin:
Since this option is unusable, backporting is fine!

PeteGillinElastic:
Okay, backporting is also fine with me.

Are you okay to approve this PR, or should I ask someone from my team?

PeteGillinElastic:
Do backports in this repo need review?
 - https://github.com/elastic/elasticsearch-specification/pull/5480
 - https://github.com/elastic/elasticsearch-specification/pull/5481

Issue #5451, opened 2025-10-08
[9.2] Add new parameters to inference PUT API for the rerank task type


kosabogi:
Related dev issue: https://github.com/elastic/elasticsearch/pull/130485

Starting in version 9.2, the [PUT _inference/{task_type}/{elasticsearch_inference_id}](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-elasticsearch)
 API with the `rerank` task type introduces two new parameters in `service_settings`:

- `long_document_strategy`
- `max_chunks_per_doc`

These parameters define how the reranker handles long documents during inference.

### `long_document_strategy`

Type: string (optional)

Controls the strategy used for processing long documents.

**Possible values:**

- `truncate` (default): Processes only the beginning of each document.
- `chunk`: Splits long documents into smaller parts (chunks) before inference.

To enable chunking, this value must be set to "chunk".

### `max_chunks_per_doc`

Type: integer (optional)

Limits how many chunks per document are sent for inference when chunking is enabled.
If not set, all chunks generated for the document are processed.

### Required updates

- Add `long_document_strategy` and `max_chunks_per_doc` to the `service_settings` object of the [PUT inference API documentation](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-elasticsearch)

- Update the [Configuring chunking documentation](https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#infer-chunking-config) to mention that chunking must be enabled by setting:

`"long_document_strategy": "chunk"`

- Update the description of the `chunking_settings` object to clarify that enabling chunking requires setting the `long_document_strategy parameter` to `chunk` in the rerank inference endpoint configuration.

PR #5476, opened 2025-10-13
Adds new parameters to the elasticsearch inference API for the rerank task type

kosabogi:
This PR adds the new `long_document_strategy` and `max_chunks_per_doc` parameters to the `service_settings` object of the [Create an Elasticsearch inference endpoint](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-elasticsearch) documentation.

It also updates the description of the `chunking_settings` object to clarify that this setting is only applicable for the `sparse_embeddings` and `text_embeddings` task types.

Related issue: https://github.com/elastic/elasticsearch-specification/issues/5451

### DIFF
--- a/.github/workflows/validate-apis.yml
+++ b/.github/workflows/validate-apis.yml
@@ -41,14 +41,14 @@ jobs:
         with:
           node-version: 22
 
-      - name: Cache JSON report
-        id: cache-json-report
-        uses: actions/cache@v4
+      - name: Restore JSON report
+        if: github.event_name == 'pull_request'
+        uses: actions/cache/restore@v4
         with:
           path: ./clients-flight-recorder/recordings/types-validation/types-validation.json
-          key: types-validation-json-${{ github.event_name == 'pull_request' && format('{0}-{1}', github.base_ref, github.event.pull_request.base.sha) || format('{0}-{1}', github.ref_name, github.sha) }}
+          key: types-validation-json-${{ format('{0}-{1}', github.base_ref, github.event.pull_request.base.sha) }}
           restore-keys: |
-            types-validation-json-${{ github.event_name == 'pull_request' && github.base_ref || github.ref_name }}-
+            types-validation-json-${{ github.base_ref }}-
 
       - name: Install deps 1/2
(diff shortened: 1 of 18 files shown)


specification/inference/_types/CommonTypes.ts
```diff
@@ -1306,6 +1306,23 @@ export class ElasticsearchServiceSettings {
    * The maximum value is 32.
    */
   num_threads: integer
+  /**
+   * Only for the `rerank` task type.
```
dan-rubinstein: A quick clarification. For 9.2, these two values are only configurable for rerank endpoints using the elastic reranker model.

specification/inference/put_elasticsearch/PutElasticsearchRequest.ts
```diff
@@ -78,10 +78,10 @@ export interface Request extends RequestBase {
   }
   body: {
     /**
-     * The chunking configuration object.
+     * The chunking configuration object. For the `rerank` task type, you can enable chunking by setting the `long_document_strategy` parameter to `chunk` in the `service_settings` object.
```
dan-rubinstein: I'm not sure if we need to be more specific about this anywhere but for this new method of chunking the user can not set `chunking_settings` the way that they would for embeddings. We handle building the chunking settings for them. If we want to clarify how we build the chunking settings somewhere we can.

specification/inference/_types/CommonTypes.ts
```diff
@@ -1306,6 +1306,23 @@ export class ElasticsearchServiceSettings {
...
+   * Controls the strategy used for processing long documents during inference.
+   *
+   * Possible values:
+   * - `truncate` (default): Processes only the beginning of each document.
+   * - `chunk`: Splits long documents into smaller parts (chunks) before inference.
```
dan-rubinstein, line 1315: I'm not sure where it's best to clarify this but with chunking enabled we will return to the user a single score per document (same as we do for truncating) with the score correlating to the highest score of any chunk. I just want to make it clear that the structure of the response to the user will not change, only the rerank relevance scores.

PR #5492, opened 2025-10-15
Clarifies that the docs.count field in the _cat/indices API includes hidden nested documents.

mridula-s109:
## Summary
Clarifies that the `docs.count` field in the `_cat/indices` API includes hidden nested documents, which can lead to higher counts than expected when using field types like `semantic_text`.

## Changes
- Updated the documentation for `docs.count` in `CatIndicesRequest.ts` to explicitly state that it includes nested documents
- Added guidance to use `_count` or `_cat/count` APIs for logical document count

## Related Issues
Fixes #127354

## Related PRs
- elasticsearch [#136329 ](https://github.com/elastic/elasticsearch/pull/136329)(semantic_text documentation update)

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -48540,7 +48540,7 @@
             "type": "string"
           },
           "docs.count": {
-            "description": "available docs",
+            "description": "The number of documents in the index, including hidden nested documents.\nFor indices with `semantic_text` fields or other nested field types,\nthis count includes the internal nested documents.\nTo get the logical document count (excluding nested documents), use\nthe `_count` API or `_cat/count` API instead.",
             "oneOf": [
               {
                 "type": "string"


lcawl:
I generated the output and confirmed that it appears in a docs preview:

image

output/openapi/elasticsearch-openapi.json
```diff
@@ -72819,7 +72819,7 @@
             "type": "string"
           },
           "docs.count": {
-            "description": "available docs",
+            "description": "The number of documents in the index, including hidden nested documents.\nFor indices with `semantic_text` fields or other nested field types,\nthis count includes the internal nested documents.\nTo get the logical document count (excluding nested documents), use\nthe `_count` API or `_cat/count` API instead.",
```
Mikep86, line 72822: Nit: Do we want/need newlines in the JSON like this? Honest question, I don't know how they affect downstream rendering of the spec.

leemthompo: This is automatic, don't think need to be concerned about this. Lisa's [preview](https://github.com/elastic/elasticsearch-specification/pull/5492#issuecomment-3407410829) LGTM.

PR #5502, opened 2025-10-16
Adds `weight` parameter type definition for RRF retrievers.

mridula-s109:
## Changes

- `RRFRetriever.retrievers`: `RetrieverContainer[]` → `RRFRetrieverContainer[]`
- New [RRFRetrieverContainer](cci:2://file:///Users/mrids/git_tree/elasticsearch-specification/specification/_types/Retriever.ts:149:0-163:1) extends [RetrieverContainer](cci:2://file:///Users/mrids/git_tree/elasticsearch-specification/specification/_types/Retriever.ts:31:0-51:1) with optional `weight?: float`
- Defaults to 1.0, must be non-negative

Related: elastic/elasticsearch#130658, elastic/elasticsearch#136477


leemthompo:
You'll need to backport this to 9.2 right?

mridula-s109:
> You'll need to backport this to 9.2 right?

Thanks for the reminder! I've added the `backport-9.2` label but the automation isn't triggering. Does the backport job only run after the PR is merged, or is there a specific way to trigger it manually?

leemthompo:
The backport job only runs after the PR is merged (if there are conflicts might need to use the manual backport tool to clean those up)

specification/_types/Retriever.ts
```diff
@@ -144,6 +144,15 @@ export class RRFRetriever extends RetrieverBase {
   fields?: string[]
 }
 
+/**
+ * RRF retriever container supporting both direct and weighted formats.
+ * @variants container
+ */
+export class RRFRetrieverContainer extends RetrieverContainer {
+  /** Weight multiplier for this retriever's contribution to the RRF score. Higher values increase influence. Defaults to 1.0 if not specified. Must be non-negative. */
+  weight?: float
```
pmpailis: I think that we also have to include the `RetrieverContainer` definition here as we have:
```json
retrievers: [
   { 
         "retriever": {},
         "weight": 1
]
```

Maybe we could also combine with the nested structure of `LinearRetriever` which uses an `InnerRetriever` component.

flobernd:
Thanks for the PR @mridula-s109 ! 

From looking at the JSON example provided in https://github.com/elastic/elasticsearch/pull/130658, @pmpailis seems to be right that the structure is not correct this way.

```json
"retrievers": [ // <- RRFRetriever.retrievers
  {
    "retriever": { // <- RRFRetrieverComponent.retriever
      "standard": { // <- RetrieverContainer variant
        "query": { "match": { "description": "pizza" } } 
      }
    },
    "weight": 0.7 // <- RRFRetrieverComponent.weigth
  },
```

I think it should rather look like this:

```ts
export class RRFRetrieverComponent { // name taken from the server code
  retriever: RetrieverContainer
  weigth?: float
}
```

and

```ts
// In RRFRetriever:
retrievers: RRFRetrieverComponent[]
```

Btw.: I'm a little bit surprised that we ship this breaking change in a minor release. Is this expected?

Side note: For `container` variants, it's important to mark all fields that are not actual variants with the `@container_property` tag (see [here](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#container)).

pmpailis:
> Btw.: I'm a little bit surprised that we ship this breaking change in a minor release. Is this expected?

I think that this is an extension, i.e. an alternative format that we support, and not a breaking change. The previous format has not gone away (which also implies some adjustments in this PR)

flobernd:
@pmpailis 

> I think that this is an extension, i.e. an alternative format that we support, and not a breaking change. The previous format has not gone away

Ah right, I just saw the corresponding test cases 👍 

> which also implies some adjustments in this PR

Unfortunately. This kind of extension will result in an untagged/undiscriminated union in the specification - which is the worst possible construct to have for all statically typed clients 😐

Anaethelion:
I have removed the `9.2` backport label.
This sadly comes too late for us to include in the release of clients, especially with an untagged union which causes a breaking change in the API.
Please for now merge to `main`, thank you!

flobernd:
@mridula-s109 Are you going to update the PR? Happy to assist, if you need help.

I would suggest something like this:

```ts
export class RRFRetrieverComponent { // name taken from the server code
  retriever: RetrieverContainer
  weigth?: float
}

export type RRFRetrieverDescriptor = RetrieverContainer | RRFRetrieverComponent;
// ^ naming is hard. Open for suggestions.
```

```ts
// In RRFRetriever:
retrievers: RRFRetrieverDescriptor[]
```

mridula-s109:
> @mridula-s109 Are you going to update the PR? Happy to assist, if you need help.
> 
> I would suggest something like this:
> 
> ```ts
> export class RRFRetrieverComponent { // name taken from the server code
>   retriever: RetrieverContainer
>   weigth?: float
> }
> 
> export type RRFRetrieverDescriptor = RetrieverContainer | RRFRetrieverComponent;
> // ^ naming is hard. Open for suggestions.
> ```
> 
> ```ts
> // In RRFRetriever:
> retrievers: RRFRetrieverDescriptor[]
> ```

Thanks for the suggestion @flobernd, Yes i will update the PR accordingly today and rerequest review on the same.

specification/_types/Retriever.ts
```diff
@@ -133,9 +133,25 @@ export class KnnRetriever extends RetrieverBase {
...
+  retriever: RetrieverContainer
+  /** Weight multiplier for this retriever's contribution to the RRF score. Higher values increase influence. Defaults to 1.0 if not specified. Must be non-negative. */
+  weight?: float
+}
+
+/**
+ * Either a direct RetrieverContainer (backward compatible) or an RRFRetrieverComponent with weight.
+ * @codegen_names retriever, weighted
+ */
+export type RRFRetrieverEntry = RetrieverContainer | RRFRetrieverComponent
```
flobernd, line 155: 👍 that name fits better

specification/_types/Retriever.ts
```diff
@@ -133,9 +133,25 @@ export class KnnRetriever extends RetrieverBase {
   rescore_vector?: RescoreVector
 }
 
+/**
+ * Wraps a retriever with an optional weight for RRF scoring.
+ */
+export class RRFRetrieverComponent {
+  /** The nested retriever configuration. */
+  retriever: RetrieverContainer
+  /** Weight multiplier for this retriever's contribution to the RRF score. Higher values increase influence. Defaults to 1.0 if not specified. Must be non-negative. */
```
flobernd: /nit: we should also add the @server_default tag as a machine readable alternative to mentioning the default value in the doc string.

mridula-s109: Have done the same.

PR #5514, opened 2025-10-20
Introduce `UserDefinedObject`

flobernd:
The specification uses `Dictionary<string, UserDefinedValue>` in many places to indicate that the server returns an "arbitrary" JSON object.

This PR introduces a new alias type `UserDefinedObject` that should instead be used for this purpose, because:

1. The descriptive name ensures that the semantics are clear
2. The alias type allows generators to replace the type with a more natural one, if needed (see e.g. https://github.com/elastic/elasticsearch-net/issues/8758)


qn895 (approved):
ML UI: LGTM 🎉

l-trotta (approved):

flobernd:
@l-trotta @Anaethelion Would you be fine with backporting this to 8.19 and potentially 9.2?

l-trotta:
yep java resolves all aliases so this changes nothing for me

Anaethelion:
Nope, we need proper handling for this first, this breaks the API badly.
I'll add an issue in the generator to handle this.

PR #5523, opened 2025-10-20
Adding batch size to custom service

jonathan-buttner:
This PR updates the docs to include `batch_size` which I forgot to add.

The original ES PR is here: https://github.com/elastic/elasticsearch/pull/129558

### DIFF
--- a/specification/inference/_types/CommonTypes.ts
+++ b/specification/inference/_types/CommonTypes.ts
@@ -948,6 +948,9 @@ export class CohereTaskSettings {
 export class CustomServiceSettings {
   /**
    * Specifies the batch size used for the semantic_text field. If the field is not provided, the default is 10.
+   * The batch size is the maximum number of inputs in a single request to the upstream service.
+   * The chunk within the batch are controlled by the selected chunking strategy for the semantic_text field.
+   * @ext_doc_id sematic-text-chunking
    */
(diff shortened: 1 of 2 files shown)


specification/inference/_types/CommonTypes.ts
```diff
@@ -946,6 +946,10 @@ export class CohereTaskSettings {
 }
 
 export class CustomServiceSettings {
+  /**
+   * Specifies the batch size used for the semantic_text field. If the field is not provided, the default is 10.
```
prwhelan, line 950: Do we need to expand on this to explain what the batch size is? Something like "the batch size is the maximum number of inputs based on the selected chunking strategy for the semantic_text field" and then link out to the semantic text field? Or is that too much information for this location?

Separately, can we add a link to the semantic text page? it'd be an entry in `specification/_doc_ids/table.csv` for `https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/semantic-text#auto-text-chunking` and then here we'd add the id for `@ext_doc_id`, looks like: https://github.com/elastic/elasticsearch-specification/blob/main/specification/transform/_types/Transform.ts#L139

jonathan-buttner: Yeah good idea.

PR #5552, opened 2025-10-28
Allow single percent in PercentilesAggregation

pquentin:
I don't know if we actually want to do that, but that's possible.

First, `percents` is parsed in [PercentilesAggregationBuilder](https://github.com/elastic/elasticsearch/blob/60406a6315bb9b1fc847e614175899a9161b2e82/server/src/main/java/org/elasticsearch/search/aggregations/metrics/PercentilesAggregationBuilder.java#L38-L46) but I don't understand where a single value is allowed

However there's a [YAML test](https://github.com/elastic/elasticsearch/blob/60406a6315bb9b1fc847e614175899a9161b2e82/modules/aggregations/src/yamlRestTest/resources/rest-api-spec/test/aggregations/empty_field_metric.yml#L84-L87) that uses this form, the response looks like this:

```json
              "tdigest_percentiles#percentiles_agg": {
                "values": {
                  "50.0": null
                }
              },
```

However, since it's undocumented, I'm happy to ignore that test.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -46410,10 +46410,17 @@
               },
               "percents": {
                 "description": "The percentiles to calculate.",
-                "type": "array",
-                "items": {
-                  "type": "number"
-                }
+                "oneOf": [
+                  {
+                    "type": "number"
+                  },
+                  {
+                    "type": "array",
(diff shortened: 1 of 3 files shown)


flobernd:
Would be fine for me. The "single or many" construct shouldn't even be breaking for the statically typing clients - so backport might be possible.

PR #5589, opened 2025-10-31
Revert "Allow single function in FunctionScoreQuery.functions"

pquentin:
Reverts elastic/elasticsearch-specification#5576

Turns out this is not supported by Elasticsearch! See https://github.com/elastic/elasticsearch/pull/137421. Sorry.

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -37763,17 +37763,10 @@
               },
               "functions": {
                 "description": "One or more functions that compute a new score for each document returned by the query.",
-                "oneOf": [
-                  {
-                    "$ref": "#/components/schemas/_types.query_dsl.FunctionScoreContainer"
-                  },
-                  {
-                    "type": "array",
-                    "items": {
-                      "$ref": "#/components/schemas/_types.query_dsl.FunctionScoreContainer"
-                    }
-                  }
-                ]
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/_types.query_dsl.FunctionScoreContainer"
+                }
               },
               "max_boost": {
                 "description": "Restricts the new score to not exceed the provided limit.",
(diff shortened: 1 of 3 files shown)


pquentin:
To clarify what is going on, there are two ways to use function_score, either with a shortcut (that we don't support):

```
GET /_search
{
  "query": {
    "function_score": {
      "query": { "match_all": {} },
      "boost": "5",
      "random_score": {},
      "boost_mode": "multiply"
    }
  }
}
```

Or with a functions array (that we support):

```
GET /_search
{
  "query": {
    "function_score": {
      "query": { "match_all": {} },
      "boost": "5",
      "functions": [
        {
          "filter": { "match": { "test": "bar" } },
          "random_score": {},
          "weight": 23
        },
        {
          "filter": { "match": { "test": "cat" } },
          "weight": 42
        }
      ],
      "max_boost": 42,
      "score_mode": "max",
      "boost_mode": "multiply",
      "min_score": 42
    }
  }
}
```

However, those are complex examples; in many YAML tests, we just see this:

```
GET /_search
{
  "query": {
    "function_score": {
      "random_score": {},
    }
  }
}
```

Which passes validation on the main branch, thanks to the current definition of FunctionScoreQuery. Here are the relevant bits:

```ts
/**
 * @shortcut_property functions
 */
export class FunctionScoreQuery extends QueryBase {
  /**
   * One or more functions that compute a new score for each document returned by the query.
   */
  functions?: FunctionScoreContainer | FunctionScoreContainer[]
}
```

The fact that 1/ we allow FunctionScoreContainer in functions and 2/ shortcut_property is still defined on functions makes all YAML tests about function_score pass, even though it is technically wrong. I think we have two options:

1. Remove the shortcut_property and define functions as FunctionScoreContainer[] - clarifying that we don't support the shortcut form at all.
2. Keep both the shortcut_property and keep defining functions as FunctionScoreContainer | FunctionScoreContainer[], which makes the easy case work while allowing wrong things.

PR #5583, opened 2025-10-30
Fix indices.put_data_lifecycle API

pquentin:
It looks like there are more issues with `DataStreamLifecycleDownsampling`, but let's just focus on `indices.put_data_lifecycle`. Since it's going to change soon (given https://github.com/elastic/elasticsearch/pull/137023 and https://github.com/elastic/elasticsearch-specification/pull/5563), I'd like the existing tests to pass.

PR #5593, opened 2025-10-31
Fix data stream lifecycle type, to correctly handle downsampling.

gmarouli:
Similar to https://github.com/elastic/elasticsearch-specification/pull/5583, the downsampling type in the data stream lifecycle does not reflect how the API is structured.

### DIFF
--- a/specification/indices/_types/DataStreamLifecycle.ts
+++ b/specification/indices/_types/DataStreamLifecycle.ts
@@ -35,7 +35,7 @@ export class DataStreamLifecycle {
   /**
    * The list of downsampling rounds to execute as part of this downsampling configuration
    */
-  downsampling: DownsamplingRound[]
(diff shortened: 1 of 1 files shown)


specification/indices/_types/DataStreamLifecycle.ts
```diff
@@ -33,9 +33,9 @@ export class DataStreamLifecycle {
    */
   data_retention?: Duration
   /**
-   * The downsampling configuration to execute for the managed backing index after rollover.
+   * The list of downsampling rounds to execute as part of this downsampling configuration
    */
-  downsampling?: DataStreamLifecycleDownsampling
+  downsampling: DownsamplingRound[]
```
nielsbauman: Actually, I just noticed that you removed the `?` here (which you add back in #5594).
```suggestion
  downsampling?: DownsamplingRound[]
```

PR #5597, opened 2025-10-31
Async search error handling

l-trotta:
As discussed in the latest spec meeting, adding exception handling to async search. [server code](https://github.com/elastic/elasticsearch/blob/84d04a620f7b9e3befaf2e1705176c91571cc726/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/search/action/AsyncSearchResponse.java#L42)

To reproduce:

```
POST /non-existing-index/_async_search
```


pquentin:
Sorry, I have no idea how to get to this result from the server code. Did you consider creating an `AsyncSearchException` class since the content appears to be the same in both cases?

l-trotta:
@pquentin thanks, updated and added reproducible

flobernd (approved):

Issue #4538, opened 2025-06-12
[validation] Native typescript collection types should not be used


l-trotta:
es-lint should only allow our aliases (e.g. Dictionary instead of Map)

JoshMock:
Based on the [modeling guide](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md), I believe this breaks down into a couple rules:

- Error: Use `Dictionary` instead of `Record`
- Error: use `Dictionary` instead of `Map`

PR #5598, opened 2025-10-31
Forbid native TypeScript collection types

margaretjgu:
Closes https://github.com/elastic/elasticsearch-specification/issues/4538

### DIFF
--- a/validator/rules/no-native-types.js
+++ b/validator/rules/no-native-types.js
@@ -30,7 +30,6 @@ const TYPE_SUGGESTIONS = {
   'Set': 'Use an array type instead (e.g., string[])',
   'WeakMap': 'Use Dictionary instead',
   'WeakSet': 'Use an array type instead',
-  'Array': 'Use array syntax instead (e.g., string[])',
 };
 
 export default createRule({
--- a/validator/test/no-native-types.test.js
+++ b/validator/test/no-native-types.test.js
@@ -34,7 +34,6 @@ ruleTester.run('no-native-types', rule, {
   valid: [
     `type MyDict = Dictionary<string, object>`,
     `type MyMapping = Dictionary<string, any>`,
-    `type MyArray = string[]`,
     `type MyType = { field: string }`,
     `class MyClass { prop: integer }`,
   ],
@@ -77,10 +76,6 @@ ruleTester.run('no-native-types', rule, {
     },
     {
       code: `class MyClass { items: Map<string, number> }`,
-      errors: [{ messageId: 'noNativeType' }]
-    },
-    {
-      code: `type MyList = Array<integer>`,
       errors: [{ messageId: 'noNativeType' }]
     },
   ],


validator/test/no-native-types.test.js
```diff
@@ -32,32 +32,53 @@ const ruleTester = new RuleTester({
 
 ruleTester.run('no-native-types', rule, {
   valid: [
-    `type MyRecord = Record<string, object>`,
-    `type MyPart = Partial<Record>`,
-    `type MyReq = Required<string>`,
-    `type MyPick Pick<integer,"something">`,
-    `type MyOmit = Omit<Record, "something">`,
+    `type MyDict = Dictionary<string, object>`,
```
margaretjgu, line 35: The original test file had a bug...it included native types like `Record`, `Partial`, etc. in BOTH the `valid` and `invalid` arrays. This was contradictory and causing test failures.

I corrected the tests to properly reflect the rule's intent:
- **`valid`**: Now uses spec-defined aliases (`Dictionary`) and allowed types (arrays, custom classes)
- **`invalid`**: Contains native utility types (`Record`, `Partial`, etc.) and collection types (`Map`, `Set`, etc.)

also added test coverage for the new collection types (`Map`, `Set`, `WeakMap`, `WeakSet`).

validator/rules/no-native-types.js
```diff
@@ -20,25 +20,45 @@ import { ESLintUtils } from '@typescript-eslint/utils';
 
 const createRule = ESLintUtils.RuleCreator(name => `https://example.com/rule/${name}`)
 
-const TYPES_TO_AVOID = ['Record', 'Partial', 'Required', 'Pick', 'Omit'];
+const UTILITY_TYPES = ['Record', 'Partial', 'Required', 'Pick', 'Omit'];
+
+const COLLECTION_TYPES = ['Map', 'Set', 'WeakMap', 'WeakSet'];
```
pquentin: Can we please add `Array` to the list? It could also be a follow-up, as it generates 106 errors.

validator/rules/no-native-types.js
```diff
@@ -20,25 +20,45 @@ import { ESLintUtils } from '@typescript-eslint/utils';
...
+
+const TYPES_TO_AVOID = [...UTILITY_TYPES, ...COLLECTION_TYPES];
+
+const TYPE_SUGGESTIONS = {
+  'Record': 'Use Dictionary instead',
+  'Map': 'Use Dictionary instead',
+  'Set': 'Use an array type instead (e.g., string[])',
+  'WeakMap': 'Use Dictionary instead',
+  'WeakSet': 'Use an array type instead',
+};
```
pquentin: When trying to add `Array` to the list, I added it to `TYPE_SUGGESTIONS` and realized it had no effect. Can we please stop using `TYPES_TO_AVOID` and `TYPE_SUGGESTIONS` and only rely on `TYPE_SUGGESTIONS`? Perhaps that will enable us to provide more helpful suggestions than "Use spec-defined aliases instead". (There's no equivalent to `Omit` or `Pick`, as far as I know.)

pquentin (commented):
Sorry, I realize I approved, but I think the code could be made a bit more maintainable. The `Array` suggestion can wait, but we should create an issue for it if we don't do it today.

validator/rules/no-native-types.js
margaretjgu: I will do this in a follow up pr

Issue #4737, opened 2025-07-01
[validation] Union related ESLint rules


flobernd:
## Introduction

Most statically typed languages do not provide first class support for generic union types.

Besides missing language support, unions are often hard to disambiguate during deserialization.

### 1. Prefer tagged variants

**Condition**: Use of union `A | B | ...`  for a field/property when all members are `class` types.
**Severity**: 🔴 error
**Message**: Use tagged variants instead (`external` or `internal`).
**Exceptions**: Exceptions only allowed for legacy APIs. Never suppress for new APIs.

Use `internal` tagged variants to model a form of polymorphism that is as well broadly supported by JSON serializer libraries.

Use `external` tagged variants (containers), if the individual union variants are not directly related.

### 2. Do not use inline unions

**Condition**: Use of union `A | B | ...` for a field/property.
**Severity**: 🔴 error
**Message**: Use an alias type instead.
**Exceptions**: None.

Defining an alias type `export type MyUnion = A | B | ...` and using this as the field/property type instead of the inline union, allows to hand-craft the type in some statically typed languages which greatly improves usability or even is the only way to correctly some union types at all.

cc @pquentin @l-trotta @Anaethelion

PR #5603, opened 2025-11-03
Implement union type validation for static language support

margaretjgu:
Closes https://github.com/elastic/elasticsearch-specification/issues/4737

Implements two new validation rules to improve code generation for statically-typed languages:
1. `no-inline-unions` - Forbids inline union types in properties
2. `prefer-tagged-variants` - Enforces tagged variants for unions of class types

Total existing violations: 91
- `no-inline-unions`: 37
- `prefer-tagged-variants`: 54

The rules **automatically skips** the following patterns: 
1. `Type | null` / `Type | undefined`
2. `Type | Type[]`  patterns

### DIFF
--- a/validator/eslint-plugin-es-spec.js
+++ b/validator/eslint-plugin-es-spec.js
@@ -37,6 +37,5 @@ export default {
     'no-variants-on-responses': noVariantsOnResponses,
     'no-inline-unions': noInlineUnions,
     'prefer-tagged-variants': preferTaggedVariants,
-
   }
 }


specification/_global/search/SearchRequest.ts
```diff
@@ -463,6 +463,7 @@ export interface Request extends RequestBase {
      * @availability serverless
      * @ext_doc_id knn-approximate
      */
+    // eslint-disable-next-line es-spec-validator/prefer-tagged-variants -- TODO: use tagged variant
     knn?: KnnSearch | KnnSearch[]
```
pquentin, line 466: Should this really raise a warning?

specification/indices/_types/IndexSettings.ts
```diff
@@ -84,11 +84,13 @@ export class IndexSettings
    * @server_default 1
    * @availability stack
    * */
+  // eslint-disable-next-line es-spec-validator/no-inline-unions -- TODO: create named alias
   number_of_shards?: integer | string // TODO: should be only int
```
pquentin, line 88: For the record, this a known issue with index settings that they can be passed as strings too.

specification/indices/_types/IndexRouting.ts
```diff
@@ -60,5 +60,6 @@ export class IndexRoutingAllocationInitialRecovery {
 
 // ES: DiskThresholdSettings
 export class IndexRoutingAllocationDisk {
+  // eslint-disable-next-line es-spec-validator/no-inline-unions -- TODO: create named alias
   threshold_enabled?: boolean | string
```
pquentin, line 64: I'm wondering about all those `X | string` unions. Should we have a specific rule to ask using `Stringified` instead?

specification/indices/_types/IndexSettings.ts
flobernd: This should actually be Stringyfied<int> I think.

specification/indices/_types/IndexRouting.ts
flobernd: Not always. There are cases like DateTime where the string does not encode the scalar value, but has a different meaning.

specification/_global/search/SearchRequest.ts
flobernd: Noticed multiple "single or many" constructs in the PR that should not emit a warning. Let's double check 🙂

margaretjgu: Thanks for flagging these, I've updated the rule and removed all of them with a script.

specification/_global/search/SearchRequest.ts
```diff
@@ -463,6 +463,7 @@ export interface Request extends RequestBase {
      * @availability serverless
      * @ext_doc_id knn-approximate
      */
+
```
pquentin: Can you please remove the empty lines too?

flobernd:
Let's backport all the nice validations to 8.19 LTS branch as well, if they don't have incompatible changes 🙂

specification/security/_types/RoleDescriptor.ts
```diff
@@ -106,6 +107,7 @@ export class RoleDescriptorRead implements OverloadOf<RoleDescriptor> {
   /**
    * An object defining global privileges. A global privilege is a form of cluster privilege that is request-aware. Support for global privileges is currently limited to the management of application privileges.
    */
+  // eslint-disable-next-line es-spec-validator/no-inline-unions, es-spec-validator/prefer-tagged-variants -- TODO: use tagged variant
   global?: GlobalPrivilege[] | GlobalPrivilege
```
pquentin: This is valid. We have two options here:

1. Change this to `GlobalPrivilege | GlobalPrivilege[]`
2. Change the logic do detect this case

specification/inference/text_embedding/TextEmbeddingRequest.ts
```diff
@@ -54,6 +54,7 @@ export interface Request extends RequestBase {
      * Inference input.
      * Either a string or an array of strings.
      */
+    // eslint-disable-next-line es-spec-validator/no-inline-unions -- TODO: create named alias
     input: string | Array<string>
```
pquentin: Note that `string | Array<string>` is the same as `string | string[]`. I've pushed [#5603 (6e8c9affd)](https://github.com/elastic/elasticsearch-specification/pull/5603/commits/6e8c9affda9f3c8c26611dcd9670ed8c06814625) to fix it.

pquentin:
> Let's backport all the nice validations to 8.19 LTS branch as well, if they don't have incompatible changes 🙂

None of the validation work exists in 8.19, so this is out of scope.

specification/security/_types/RoleDescriptor.ts
flobernd: Vote for 2.! The validation should not care about the order.

specification/inference/text_embedding/TextEmbeddingRequest.ts
flobernd: The other validation about using native types should catch and prevent this case in the future, I guess.

PR #5604, opened 2025-11-03
Fix reindex-from-remote source host constraints and availability

PeteGillinElastic:
The reindex documentation previously referenced the node setting needed to whitelist remote hosts. This does not apply to serverless. This change clarifies this, and documents the hosts allowed in serverless.

Also, reindex-from-remote will be a tech preview in serverless, so this annotates it as such. (The fact that it was previously not available at all in serverless appears not to have been reflected in the spec.)


PeteGillinElastic:
Skipping backport as the change only really affects serverless.

PeteGillinElastic:
Hoping to get this reviewed and ready to review once the change goes live.

PeteGillinElastic:
> LGTM! I assume this doc change will be merged after the whitelist setting is added for production.

Yeah, exactly.

pete-naylor (approved):
Thanks @PeteGillinElastic - I think it is good to go!

PR #5595, opened 2025-10-31
Allow triggering validate-apis CI

pquentin:
It can be useful to update the cache.

PR #5611, opened 2025-11-05
Run PR validation on triggered builds

pquentin:
Turns out https://github.com/elastic/elasticsearch-specification/pull/5595 was not enough. The goal is really to refresh the cache.


pquentin:
That didn't work, because cache entries are immutable. I really need to stop with the caching idea and instead run base validation in parallel.

flobernd:
@pquentin It's another workaround, but you can enumerate, remove cache entries using the gh CLI.

Issue #778, opened 2021-09-22
Incorrect use of `@variants container` on response


swallez:
`ml.evaluate_data_frame` response is a container, which isn't supported in `metamodel.ts` (`variants` is only defined on `Interface`).

https://github.com/elastic/elasticsearch-specification/blob/bc30f1aa4fca2ac7d14e32f4026044e9e26ca21c/specification/ml/evaluate_data_frame/MlEvaluateDataFrameResponse.ts#L26-L33

We should either
- formally support it for `properties` requests and responses and update `metamodel.ts`
- or change this class to use a `value_body` pointing to a container class and forbid `@variants` on requests and responses.

PR #5622, opened 2025-11-05
Fix variants on Response class - use value_body pattern

margaretjgu:
Closes https://github.com/elastic/elasticsearch-specification/issues/778
Original code defined response as an object with no reusable type name, and cannot put `@variant container` on this object
Refactored to use the `value_body` pattern (same as [SearchResponse](https://github.com/elastic/elasticsearch-specification/blob/ff912c0ee91b20284290a6a3d91c901bc23ce9a4/specification/_global/search/SearchResponse.ts#L34-L39))

**Does this bug warrant a linting rule?** -  its a rare and obvious mistake. Maybe a rule is not necessary


pquentin:
> Happy to merge this as is for now. We can improve later :-)

Merging, then! @margaretjgu Can you please capture the improvements that Florian identified in a new spec validation issue?

margaretjgu:
> Let's add a rule that explicitly checks if a variant is valid for a given type, e.g.:
> 
> Request > none Response > none Interface > Container Alias > Internal, External, Untagged ...
> 
> In this particular case the compiler should also throw an exception and refuse emitting the schema.json since this creates a situation that is not mappable to the metamodel. We do this in other situations when a tag is incorrectly used, so this is probably an oversight.

Created new issue https://github.com/elastic/elasticsearch-specification/issues/5642

l-trotta:
manually backported to 9.2 and 9.1

PR #5627, opened 2025-11-06
Switch from string to enum for metrics

pquentin:
I'm not including descriptions for now, this will be next step. I'm hopeful that the Docs team can help me for that.

Is the way I've added enums specific to an API in the Request files OK?

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -1992,7 +1992,7 @@
             "in": "query",
             "name": "local",
             "description": "If true, the request retrieves information from the local node only.\nDefaults to false, which means information is retrieved from the master node.",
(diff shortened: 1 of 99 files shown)


specification/_types/common.ts
```diff
@@ -391,3 +388,30 @@ export enum ClusterInfoTarget {
...
+  request_cache,
+  recovery,
+  bulk,
+  shard_stats,
+  mappings,
+  dense_vector,
+  sparse_vector
+}
+
+export type CommonStatsFlags = CommonStatsFlag | CommonStatsFlag[]
```
flobernd, line 417: Could we please introduce an alias like this in all the other places as well instead of inlining the "single or many"?

This would allow me (and potential other consumers) to translate this construct into a native C# `Flags` enum for improved usability.

PR #5635, opened 2025-11-07
add enum values for Metric and IndexMetrics path params

ketkee-aryamane:
Linked to [#360](https://github.com/elastic/docs-content-internal/issues/360)
adds enums for Metrics and IndexMetrics for [node stats endpoint](https://www.elastic.co/docs/api/doc/elasticsearch/v9/operation/operation-nodes-stats)

Added the enum values by referring the [8.19 docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.19/cluster-nodes-stats.html)

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -431,3 +431,30 @@ export enum ClusterInfoTarget {
 }
 
 export type ClusterInfoTargets = ClusterInfoTarget | ClusterInfoTarget[]
+
+export enum CommonStatsFlag {
+  _all,
+  store,
+  indexing,
+  get,
+  search,
+  merge,
+  flush,
+  refresh,
+  query_cache,
+  fielddata,
+  docs,
+  warmer,
+  completion,
+  segments,
+  translog,
+  request_cache,
+  recovery,
+  bulk,
+  shard_stats,
+  mappings,
+  dense_vector,
+  sparse_vector
+}
+
+export type CommonStatsFlags = CommonStatsFlag | CommonStatsFlag[]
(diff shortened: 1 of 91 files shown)


flobernd:
Wondering if this case is similar to @pquentin recent related PR in a way that the server accepts combinations of multiple of these enum values (flags).

We should define an alias:
SomeMetricFlags = SomeMetricEnum | SomeMetricEnum[]

if that's the case.

specification/_types/common.ts
```diff
@@ -253,6 +254,48 @@ export enum NodeStatsLevel {
...
+  indexing_pressure,
+  indices,
+  ingest,
+  jvm,
+  os,
+  process,
+  repositories,
+  thread_pool,
+  transport
+}
```
pquentin, line 273: This enum is not used. But I already defined it as NodeStatsMetric in specification/nodes/stats/NodesStatsRequest.ts anyway, with a few values that were missing in the 8.x docs so far.

specification/nodes/stats/NodesStatsRequest.ts
```diff
@@ -70,7 +71,7 @@ export interface Request extends RequestBase {
     /*+ Limits the information returned to the specific metrics. */
     metric?: NodeStatsMetrics
     /** Limit the information returned for indices metric to the specific index metrics. It can be used only if indices (or all) metric is specified.*/
-    index_metric?: CommonStatsFlags
+    index_metric?: IndexMetrics
```
pquentin, line 74: `CommonStatsFlag` is the name of the enum in the server code, so that was already correct. (Your version is also missing `_all` that is accepted by the server too.)

pquentin:
Closing for now, but happy to review another PR that adds a description or adds enum to more APIs.

PR #5647, opened 2025-11-10
chore: add eslint rule `no-duplicate-class-names`

MattDevy:
## Description

Adding class or enum declarations in the specification is a breaking change for some languages where we must flatten the types into one file.

This change:
- adds an eslint check which will fail the build if duplicate types are found in the specification
  - uses caching and lazy type lookup logic to avoid performance issues where possible
  - allows configuration options:
    - allowedNames: a list of type names to skip (`Request`, `Response`, `ResponseBase`)
    - existingDuplicates: a dictionary of type names and their existing conflict, these are ignored by the lint rule; new conflicts although will be flagged

## Example output

```shell
make lint

> specification@0.1.0 lint
> eslint


/Users/mdevy/Documents/Github/elasticsearch-specification/specification/_global/search/SearchResponse.ts
  39:14  error  Duplicate: "ResponseBody" is already declared in 1 other location:
        /Users/mdevy/Documents/Github/elasticsearch-specification/specification/ml/evaluate_data_frame/MlEvaluateDataFrameResponse.ts:32:14
Rename this type to avoid conflicts with other types in the specification  es-spec-validator/no-duplicate-type-names

/Users/mdevy/Documents/Github/elasticsearch-specification/specification/ml/evaluate_data_frame/MlEvaluateDataFrameResponse.ts
  32:14  error  Duplicate: "ResponseBody" is already declared in 1 other location:
        /Users/mdevy/Documents/Github/elasticsearch-specification/specification/_global/search/SearchResponse.ts:39:14
Rename this type to avoid conflicts with other types in the specification  es-spec-validator/no-duplicate-type-names

✖ 2 problems (2 errors, 0 warnings)

make: *** [lint] Error 1
```

Closes: #5646

PR #5648, opened 2025-11-10
fix: revert changes to MlEvaluateDataFrameResponse from #5622

MattDevy:
* removes duplicated `ResponseBody`
* moves body back to `Response`
* removes `@variants: container` from Response

Split from #5647

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -6513,6 +6513,11 @@
                     "description": "When set to `true` and performing a cross-cluster/cross-project query, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.",
                     "default": false,
                     "type": "boolean"
+                  },
+                  "include_execution_metadata": {
+                    "description": "When set to `true`, the response will include an extra `_clusters`\nobject with information about the clusters that participated in the search along with info such as shards\ncount.\nThis is similar to `include_ccs_metadata`, but it also returns metadata when the query is not CCS/CPS",
+                    "default": false,
+                    "type": "boolean"
                   }
                 },
                 "required": [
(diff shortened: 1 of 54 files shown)


pquentin:
The intention is to avoid having two classes with the same name, which breaks the Go client generator.

That said, we don't want to revert; we want to change the name of the new class.

flobernd:
> That said, we don't want to revert; we want to change the name of the new class.

Ok good! That would have been my suggestion as well.

MattDevy:
@flobernd @pquentin thank you for the feedback! Let's discuss this in the spec meeting and I'll amend the PR accordingly :)

MattDevy:
@flobernd we reviewed this today with @pquentin and @l-trotta .
We agreed that this is the correct approach, as `@variants container` doesn't make sense on a Response type, since we use it mainly to allow client libs to use polymorphism on Request types. 
However, responses from the server don't need this. 
Since `@variants container` isn't needed (and make no difference) on Response types. Instead of the changes in #5622 
```diff
- /** @variants container */
- export class Response {
-   body: {
-     /**
-      * Evaluation results for a classification analysis.
-      * It outputs a prediction that identifies to which of the classes each document belongs.
-      */
-     classification?: DataframeClassificationSummary
-     /**
-      * Evaluation results for an outlier detection analysis.
-      * It outputs the probability that each document is an outlier.
-      */
-     outlier_detection?: DataframeOutlierDetectionSummary
-     /**
-      * Evaluation results for a regression analysis which outputs a prediction of values.
-      */
-     regression?: DataframeRegressionSummary
-   }
+   /** @codegen_name result */
+   body: ResponseBody
+ }
+ 
+ /** @variants container */
+ export class ResponseBody {
+   /**
+    * Evaluation results for a classification analysis.
+    * It outputs a prediction that identifies to which of the classes each document belongs.
+    */
+   classification?: DataframeClassificationSummary
+   /**
+    * Evaluation results for an outlier detection analysis.
+    * It outputs the probability that each document is an outlier.
+    */
+   outlier_detection?: DataframeOutlierDetectionSummary
+   /**
+    * Evaluation results for a regression analysis which outputs a prediction of values.
+    */
+   regression?: DataframeRegressionSummary
```
It makes sense to simply remove the `/** @variants container */`

```diff
- /** @variants container */
  export class Response {
```

PR #5654, opened 2025-11-11
chore: add eslint rule `no-property-overload` to disallow property shadowing in interfaces/classes

### LINKED ISSUE #4943: Handle the `human` parameter overloading in the xpack.info API
## 🚀 Feature Proposal

All Elasticsearch APIs accept a `human` parameter, which defaults to false. This is encoded in the [CommonQueryParameters](https://github.com/elastic/elasticsearch-specification/blob/aa8474480381d247e053564587a3ee63a4d0b18c/specification/_spec_utils/behaviors.ts#L41-L75) interface. However, the XPack Info API is special because [human defaults to true there](https://github.com/elastic/elasticsearch-specification/blob/aa8474480381d247e053564587a3ee63a4d0b18c/specification/xpack/info/XPackInfoRequest.ts#L49-L54), which [is an error](https://github.com/elastic/elasticsearch-specification/blob/aa8474480381d247e053564587a3ee63a4d0b18c/compiler/src/steps/validate-model.ts#L823-L825).

It's actually fine in practice, and the output is correct. But we don't want any validation errors, so that we can fail CI if there are errors. Moving this check to eslint would allow us to selectively ignore it using something like `eslint-disable-line`.

MattDevy:
When classes and interfaces inherit properties via `extends` or `implements` chains, this linter rule will raise an error.
This can be ignored with `// eslint-disable-next-line es-spec-validator/no-property-overload`

Closes #4943

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -13839,20 +13839,110 @@
                   "description": "Run `PUT _inference/text_embedding/google_vertex_ai_embeddings` to create an inference endpoint to perform a `text_embedding` task type.",
                   "value": "{\n    \"service\": \"googlevertexai\",\n    \"service_settings\": {\n        \"service_account_json\": \"service-account-json\",\n        \"model_id\": \"model-id\",\n        \"location\": \"location\",\n        \"project_id\": \"project-id\"\n    }\n}"
                 },
+                "PutGoogleVertexAiRequestExample10": {
+                  "summary": "A chat_completion task for Google Model Garden Meta shared endpoint with single streaming URL provided",
+                  "description": "Run `PUT _inference/chat_completion/google_model_garden_meta_chat_completion` to create an inference endpoint to perform a `chat_completion` task using Meta's model hosted on Google Model Garden shared endpoint with single streaming URL provided. See the endpoint's `Sample request` page for the variable values used in the URL.",
+                  "value": "{\n    \"service\": \"googlevertexai\",\n    \"service_settings\": {\n        \"provider\": \"meta\",\n        \"service_account_json\": \"service-account-json\",\n        \"streaming_url\": \"https://%LOCATION_ID%-aiplatform.googleapis.com/v1/projects/%PROJECT_ID%/locations/%LOCATION_ID%/endpoints/%ENDPOINT_ID%/chat/completions\"\n    }\n}"
+                },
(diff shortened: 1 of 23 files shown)


validator/rules/no-property-overload.js
```diff
@@ -0,0 +1,116 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { ESLintUtils, TSESTree } from '@typescript-eslint/utils'
+
+const IMPLEMENTS_KEYWORD = 119;
+const EXTENDS_KEYWORD = 96;
```
pquentin, line 23: There are enums in the typescript-eslint to avoid using magic numbers like this, can we please use them? Maybe with `AST_NODE_TYPES`? Not sure. https://github.com/typescript-eslint/typescript-eslint/blob/1d15881d73cf120ba1aeaa763597a521d403d036/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts#L226

pquentin (commented):
> Closes https://github.com/elastic/elasticsearch-specification/issues/4943

This does not close the issue! Not all `spec validation` issues are about moving checks to ESLint. The issue here is that we don't have a clear plan to handle `human` in `xpack.info` which has a different default and description. Should it be done at the language generator level? Or at the spec level somehow? But how?

Additionally, the check is actually incorrect as it does not detect the `human` problem with `xpack.info`: https://github.com/elastic/elasticsearch-specification/blob/1d91d746259be2d26a59e1efbeddf6c871ac4b2b/specification/xpack/info/XPackInfoRequest.ts#L52-L57

We would need to add CommonQueryParameters and CommonCatQueryParameters (for CAT APIs) to the list of overloads. Honestly, this is a perfect example of a rule that is indeed achievable with ESLint, but it needs 600 lines of code, and is not correct yet. It's much easier to do within the compiler which has the correct representation to do that in a few lines of code. Having it in the compiler means we have a single place where to change that logic, too.

Note that I did ignore this specific error in the compiler with https://github.com/elastic/elasticsearch-specification/pull/4944, which should eventually be reverted along with the check removal if we do decide to use ESLint here.

pquentin:
Closing per offline discussion, we'll keep this check in the compiler as this is a rare case so the complexity isn't worth it here. Thank you for your work on this!

PR #5655, opened 2025-11-12
Update specs for ES|QL include_execution_metadata

luigidellaquila:
Updating ES|QL spec after changes introduced with https://github.com/elastic/elasticsearch/pull/137641

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -35676,6 +35676,13 @@
                 "$ref": "#/components/schemas/_types.EpochTimeUnitMillis"
               }
             ]
+          },
+          "error": {
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/_types.ErrorCause"
+              }
+            ]
           }
         },
         "required": [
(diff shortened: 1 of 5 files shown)


compiler/package-lock.json
```diff
@@ -567,6 +567,7 @@
       "resolved": "https://registry.npmjs.org/@types/node/-/node-22.15.31.tgz",
       "integrity": "sha512-jnVe5ULKl6tijxUhvQeNbQG/84fHfg+yMak02cT8QVhBx/F05rAVxCGBYYTh2EKz22D6JF5ktXuNwdx7b9iEGw==",
       "license": "MIT",
+      "peer": true,
```
idegtiarenko: This seems unrelated to `include_execution_metadata`.
Could you please confirm this is intended?

idegtiarenko (approved):
Changes around `include_execution_metadata` looks good to me 👍

compiler/package-lock.json
luigidellaquila: I have no idea what it does, it was generated by `make contrib`.

@pquentin can you confirm that it's expected?

flobernd: Good catch. Might be related to the local NPM version.

Since you didn't change anything on the JS tooling itself, we should probably remove all the package-lock changes from this PR to be on the safe side.

luigidellaquila: Done, thanks!

flobernd (approved):

compiler/package-lock.json
pquentin: > @pquentin can you confirm that it's expected?

As long as CI passes, I don't think that was harmful! We often have small random changes like this that don't make any difference in practice.

pquentin:
Thank you for the contribution! Appreciate it.

PR #5662, opened 2025-11-12
Add OpenShift AI integration specifications

Jan-Kazlouski-elastic:
This PR adds changes to specification caused by https://github.com/elastic/elasticsearch/pull/136624

Corresponding PR for elasticsearch repo that has to be merged first:
https://github.com/elastic/elasticsearch/pull/138025

**Additional actions**

- [x] Signed the CLA
- [x] Executed make contrib

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -14637,7 +14637,7 @@
                     "externalDocs": {
                       "url": "https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#infer-chunking-config"
                     },
-                    "description": "The chunking configuration object.",
+                    "description": "The chunking configuration object.\nApplies only to the `text_embedding` task type.\nNot applicable to the `rerank`, `completion`, or `chat_completion` task types.",
                     "allOf": [
                       {
                         "$ref": "#/components/schemas/inference._types.InferenceChunkingSettings"
@@ -14661,7 +14661,7 @@
                     ]
                   },
                   "task_settings": {
-                    "description": "Settings to configure the inference task.\nThese settings are specific to the task type you specified.",
+                    "description": "Settings to configure the inference task.\nApplies only to the `rerank` task type.\nNot applicable to the `text_embedding`, `completion`, or `chat_completion` task types.\nThese settings are specific to the task type you specified.",
                     "allOf": [
                       {
                         "$ref": "#/components/schemas/inference._types.OpenShiftAiTaskSettings"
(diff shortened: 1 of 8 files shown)


specification/_doc_ids/table.csv
```diff
@@ -387,6 +387,7 @@ inference-api-put-jinaai,https://www.elastic.co/docs/api/doc/elasticsearch/opera
 inference-api-put-llama,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-llama,,
 inference-api-put-mistral,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-mistral,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/infer-service-mistral.html,
 inference-api-put-openai,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-openai,https://www.elastic.co/guide/en/elasticsearch/reference/8.18/infer-service-openai.html,
+inference-api-put-openshift-ai,https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-openshift-ai,,
```
JoshMock, line 390: This URL does not exist yet, which will cause some clients' docs builds to fail. Will this page be available soon?

Jan-Kazlouski-elastic: I believe this will be added by merging of https://github.com/elastic/elasticsearch/pull/138025 PR.
And in order to merge - overall changes to specification must be reviewed. Any change to mentioned PR - will result in changes to this PR.

Please correct me if my understanding is not valid.

JoshMock: You're correct! Just making sure any broken URLs are temporary.

Jan-Kazlouski-elastic: > You're correct! Just making sure any broken URLs are temporary.

Sure thing! Thank you for your vigilance.

Jan-Kazlouski-elastic:
Hi @DonalEvans 
I'd like to know more about what is meant by `one response example` in your comment. It is common practice for the integrations not to provide any response examples for operations. And none was added. So I'm a bit confused by what response example you're referring to.

But still - I have added XML response for OpenShift AI for every request example. That might be good practice to establish for our next integrations if you see that suitable.
Type `string` is replaced `enum` as well as in [other PR](https://github.com/elastic/elasticsearch/pull/138025)

DonalEvans:
> It is common practice for the integrations not to provide any response examples for operations.

Thanks for this context, I wasn't aware that that was the common practice in the past. Thank you also for adding response examples for each of the task types, I think it improves the usefulness of the docs for users.

PR #5696, opened 2025-11-18
Fix type of CategoryId

pquentin:
This type is only used in a single API, `ml.get_categories`.

Server code: https://github.com/elastic/elasticsearch/blob/f86f58fbf3ebfb4e54fd711acae8290e3669e835/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/rest/results/RestGetCategoriesAction.java#L53-L55

### DIFF
--- a/specification/_types/common.ts
+++ b/specification/_types/common.ts
@@ -49,7 +49,7 @@ export type Uri = string
 export type ScrollId = string
 export type ScrollIds = ScrollId | ScrollId[]
 
-export type CategoryId = string
+export type CategoryId = long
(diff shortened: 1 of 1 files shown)


pquentin:
The failure is caused by one wrong YAML test: https://github.com/elastic/elasticsearch/blob/ab55d97b078ddbb020f4db740f982d4312b5c974/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/ml/jobs_get_result_categories.yml#L127-L135

I opened https://github.com/elastic/elasticsearch/pull/138220 to fix it.

PR #5705, opened 2025-11-19
Fix find_stucture format enum

pquentin:


specification/text_structure/find_structure/FindStructureRequest.ts
```diff
@@ -207,3 +207,10 @@ export interface Request<TJsonDocument> {
    * @codegen_name text_files */
   body: Array<TJsonDocument>
 }
+
+export enum FindStructureFormat {
+  ndjson,
+  xml,
+  delimited,
+  semi_structured_text
+}
```
pquentin, line 216: https://github.com/elastic/elasticsearch/blob/3b7ac32eb68f42f006bc6187c977c73fc238479a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/textstructure/structurefinder/TextStructure.java#L37-L42

PR #5728, opened 2025-11-24
Add new options for generic vector formats to dense_vector

thecoop:
Update the spec for changes in https://github.com/elastic/elasticsearch/pull/138492 in 9.3


specification/_types/mapping/DenseVectorProperty.ts
```diff
@@ -166,8 +171,8 @@ export class DenseVectorIndexOptions {
   /**
    * `true` if vector rescoring should be done on-disk
    *
-   * Only applicable to `bbq_hnsw`
-   * @server_default false
```
flobernd, line 170: Should we keep the `@server_default` or does that default value is no longer a constant?

thecoop: Oh, it should still have the default.

PR #5730, opened 2025-11-24
 Add project_routing param to _project/tags

smalyshev:
See https://github.com/elastic/elasticsearch-serverless/pull/4868

### DIFF
--- a/specification/_json_spec/project.tags.json
+++ b/specification/_json_spec/project.tags.json
@@ -20,7 +20,7 @@
     "params": {
       "project_routing": {
         "type": "string",
-        "description": "A Lucene query using project metadata tags to limit which projects to search, such as _alias:_origin or _alias:*pr*."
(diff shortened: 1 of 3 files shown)


specification/_json_spec/project.tags.json
```diff
@@ -13,9 +13,15 @@
...
           "path": "/_project/tags",
-          "methods": ["GET"]
+          "methods": ["GET", "POST"]
         }
       ]
+    },
+    "params": {
+      "project_routing": {
+        "type": "string",
+        "description": "A Lucene query using project metadata tags to limit which projects to search, such as _alias:_origin or _alias:*pr*."
```
quux00: I think for this endpoint, this description (which you probably grabbed from the other endpoints) is misleading, since this usage is different and doesn't actually search linked projects. 

So I would like something (for this endpoint only) more like:

> A Lucene query using project metadata tags used to filter which projects are returned in the response, such as ...

smalyshev: Updated

quux00 (approved):
LGTM

PR #4276, opened 2025-04-14
Remove knn_search API

pquentin:
As it was only ever experimental and then deprecated, it isn't supported in 9.0, and only works with `compatible-with=8` is set.

PR #5737, opened 2025-11-25
Add back private knn_search API for rest-api-spec generation

pquentin:
This reverts commit #4276 since _knn_search is still needed in rest-api-spec to run YAML tests against it, as the API is
still available with v8 compatibility mode. However, we make the API private so it does not get back into clients, as it would fail due to v9 compatibility.

Additionally, we fix the compiler to ignore `UpdateForV10` in all JSDoc blocks, given this is only meaningful at the ESLint level.


pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.1|[](https://github.com/elastic/elasticsearch-specification/pull/5770)|

### Questions ?

PR #5779, opened 2025-12-02
Add media types

### LINKED ISSUE #297: Track content type directly in the input spec
Currently, the input specification does not contain any information about the body content type, we assume that is a plain JSON. In some cases, it should be ndjson, which we normally represent as an array of objects.

For example, here's the bulk API:

https://github.com/elastic/elastic-client-generator/blob/ba32e0ee31ba6d2054903f10c6dd586689d7e7e9/specification/specs/document/multiple/bulk/BulkRequest.ts#L43

The only way for language generators for knowing this information is to read the `contentType` field of the `Endpoint` definition, which we are gathering from the rest-api-spec. For example:

```jsoni
"contentType": [
   "application/x-ndjson"
]
```

## 🚀 Feature Proposal

Create a new js doc tag to track this information directly in the input spec.

## Motivation

The rest-api-spec will not be there forever, and overtime we'll move most of the rest-api-spec information directly in the input spec. Furthermore I think strongly type languages will benefit from having this information easily accessible, but let's discuss more about it.

## Example

```diff
 /**
  * @rest_spec_name bulk
  * @since 0.0.0
  * @stability stable
+ * @content_type ndjson
  */
 interface BulkRequest<TSource> extends RequestBase {
   path_parts?: {
     index?: IndexName
     type?: Type
   }
   query_parameters?: {
     pipeline?: string
     refresh?: Refresh
     routing?: Routing
     _source?: boolean
     _source_excludes?: Fields
     _source_includes?: Fields
     timeout?: Time
     type_query_string?: string
     wait_for_active_shards?: WaitForActiveShards
     require_alias?: boolean
   }
   body?: Array<BulkOperationContainer | TSource>
 }
```

pquentin:
As decided offline:

 * Even though this is an endpoint-level property, there's no endpoint-level file, so we use request types and `request_media_type`/`response_media_type` keys, as done for URLs
 * We use a union of enums and may introduce enums that refer to specific combinations
 * We don’t change the metamodel, which stays as string[]
 * As a first step, we replicate what is in rest-api-spec. We can fix things later. In particular, for ES|QL and EQL.

This works, as shown by the unchanged schema.json. However, the way I'm getting the `request_media_type` and `response_media_type` values is ugly, and I would appreciate a better solution here.

Closes #297


lcawl:
It would be great to ultimately have those `text/plain` variations appear in the OpenAPI documents too. It's supported per https://swagger.io/docs/specification/v3_0/describing-responses/#response-media-types

pquentin:
@lcawl Agreed! However, this one fixes https://github.com/elastic/elasticsearch-specification/issues/297, not https://github.com/elastic/elasticsearch-specification/issues/4081.

JoshMock:
> However, the way I'm getting the `request_media_type` and `response_media_type` values is ugly, and I would appreciate a better solution here.

Do you mean `mediaTypeToStringArray`? If so, that's a pretty straightforward solution. If we had a lot more media types it could feel a bit verbose, but I think it's fine as is.

pquentin:
Right, parsing unions as strings feels wrong. I would rather use the AST if possible. :shrug:

compiler/src/model/utils.ts
```diff
@@ -1506,3 +1506,36 @@ export function sortTypeDefinitions (types: model.TypeDefinition[]): void {
...
+  } else {
+    enumTypeList = [mediaType.trim()]
+  }
+
+  const mediaTypeList: string[] = []
+  for (const enumType of enumTypeList) {
+    mediaTypeList.push(mediaTypeToString(enumType))
+  }
+  return mediaTypeList
+}
```
Anaethelion:
```suggestion
export function mediaTypeToStringArray (mediaType: string, allEnums: EnumDeclaration[]): string[] {
  const mediaTypeEnum = allEnums.find(e => e.getName() === 'MediaType')

  // Handle strings separated by a pipe and return multiple media types
  let enumTypeList: string[]
  if (mediaType.includes('|')) {
    enumTypeList = mediaType.split('|').map(mt => mt.trim())
  } else {
    enumTypeList = [mediaType.trim()]
  }

  const mediaTypeList: string[] = []
  for (const enumType of enumTypeList) {
    let memberName = enumType.split('.').pop()
    let value = mediaTypeEnum?.getMembers().find(m => m.getName() === memberName)?.getValue() as string
    mediaTypeList.push(value)
  }
  return mediaTypeList
}
```

Anaethelion:
I took a jab at using more the AST.
This however raise a validation error on the codestyle that we should take care of if we decide to move forward with this.

JoshMock (approved):

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.2|[](https://github.com/elastic/elasticsearch-specification/pull/5824)|
|✅|9.1|[](https://github.com/elastic/elasticsearch-specification/pull/5829)|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/5830)|

### Questions ?

PR #5781, opened 2025-12-03
Add externalDocs for multiple tags

stefnestor:
👋 howdy, team! This adds `externalDocs` links for multiple `tags` so that users can cross-reference from API docs back to Elasticserch documentation.

Side: it is my first time contributing & just wanted to note [CONTRIBUTING.md](https://github.com/elastic/elasticsearch-specification/blob/main/CONTRIBUTING.md#send-your-pull-request-from-a-branch) requests sending "a pull request from a branch, **not a fork**" but that's permission denied.

PR #5783, opened 2025-12-03
Add externalDocs for multiple tags

stefnestor:
Porting https://github.com/elastic/elasticsearch-specification/pull/5781 to branch-not-fork by request now that I have permissions, cheers.

This adds externalDocs links for multiple tags so that users can cross-reference from API docs back to Elasticserch documentation.

### DIFF
--- a/docs/overlays/elasticsearch-shared-overlays.yaml
+++ b/docs/overlays/elasticsearch-shared-overlays.yaml
@@ -112,7 +112,7 @@ actions:
           description: >
             The feature APIs enable you to introspect and manage features provided by Elasticsearch and Elasticsearch plugins.
           externalDocs:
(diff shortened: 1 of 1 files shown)


docs/overlays/elasticsearch-shared-overlays.yaml
```diff
@@ -99,10 +111,16 @@ actions:
           x-displayName: Features
           description: >
             The feature APIs enable you to introspect and manage features provided by Elasticsearch and Elasticsearch plugins.
+          externalDocs:
+            url: https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore
```
lcawl: Could link right down to the relevant section:
```suggestion
            url: https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#feature-state
```

PR #5800, opened 2025-12-05
Codegen exclude decorator

Anaethelion:
This adds a `@codegen_exclude` decorator only for requests that will translate into an optional boolean in the metamodel's Endpoint.

This also comes with a new eslint rule to avoid using the decorator outside of requests.

### DIFF
--- a/typescript-generator/src/metamodel.ts
+++ b/typescript-generator/src/metamodel.ts
@@ -469,6 +469,8 @@ export class Endpoint {
     index?: string[]
     cluster?: string[]
   }
+
+  codegenExclude?: boolean
 }
 
 export class UrlTemplate {


JoshMock (approved):
LGTM!

Should we add a tracking conformance issue to ensure all clients respect the new decorator?

Anaethelion:
> LGTM!
> 
> Should we add a tracking conformance issue to ensure all clients respect the new decorator?

That will be the topic of a future discussion!

MattDevy:
> LGTM!
> 
> Should we add a tracking conformance issue to ensure all clients respect the new decorator?

Done for Go generator :) https://github.com/elastic/elastic-client-generator-go/commit/575420ee6db8ebdc05b1aa45ab215ed54d4a05cd

PR #5835, opened 2025-12-11
Drop `project_routing` from query parameters

pawankartik-elastic:
ES-13725: We previously assumed that `project_routing` could appear either as a query parameter or within the request body. Now that we restrict it to just the request body, this PR makes the suitable adjustments within the specs.

### DIFF
--- a/specification/eslint.config.js
+++ b/specification/eslint.config.js
@@ -36,6 +36,7 @@ export default defineConfig([
       'es-spec-validator/single-key-dictionary-key-is-string': 'error',
       'es-spec-validator/dictionary-key-is-string': 'error',
       'es-spec-validator/no-native-types': 'error',
+      'es-spec-validator/no-same-name-as-enclosing-type': 'error',
       'es-spec-validator/invalid-node-types': 'error',
       'es-spec-validator/no-generic-number': 'error',
       'es-spec-validator/codegen-exclude-on-request-only': 'error',
--- a/specification/logstash/_types/Pipeline.ts
+++ b/specification/logstash/_types/Pipeline.ts
@@ -53,6 +53,7 @@ export class PipelineSettings {
    */
   'queue.checkpoint.writes': integer
 }
+// eslint-disable-next-line es-spec-validator/no-same-name-as-enclosing-type
 export class Pipeline {
   /**
    * A description of the pipeline.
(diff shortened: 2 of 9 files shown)


output/schema/validation-errors.json
```diff
@@ -7,11 +7,59 @@
       ],
       "response": []
     },
+    "async_search.submit": {
+      "request": [
+        "Request: missing json spec query parameter 'project_routing'"
```
quux00: I don't understand what this file is reporting here.

specification/eql/search/EqlSearchRequest.ts
```diff
@@ -98,18 +92,6 @@ export interface Request extends RequestBase {
...
-     * metadata tags in a subset of Lucene query syntax.
-     * Allowed Lucene queries: the _alias tag and a single value (possibly wildcarded).
-     * Examples:
-     *  _alias:my-project
-     *  _alias:_origin
-     *  _alias:*pr*
-     * Supported in serverless only.
-     * @availability serverless stability=stable visibility=feature_flag feature_flag=serverless.cross_project.enabled
-     */
-    project_routing?: ProjectRouting
```
quux00, line 112: I don't see `project_routing` in the body for EQL or SQL, but I'm pretty sure support for them was added in the code. @luigidellaquila is that right?

If yes, @pawankartik-elastic, can you add those here?

quux00 (changes requested):
Left a few suggestions.

output/schema/validation-errors.json
pawankartik-elastic: This is exactly why I asked you what commands you ran. I see similar changes in your PR as well, and I wasn't really sure what's all this about.

specification/eql/search/EqlSearchRequest.ts
pawankartik-elastic: Done. I've added them.

output/schema/validation-errors.json
pawankartik-elastic: This file no longer shows up in the diff after my recent branch update and commits. I believe it's all ok now? I re-ran the commands we use to check if I forgot to check it in after any changes, and that's not the case — all the files are checked in and up-to-date.

quux00 (approved):
LGTM

PR #5851, opened 2025-12-16
add docUrlServerless field to endpoints

### LINKED ISSUE #5188: Add serverless docURL in schema.json for Console
## 🚀 Feature Proposal

Per https://github.com/elastic/kibana/issues/163329 the API documentation links that appear in the Dev Tools > Console are derived as follows:

"The script specifically reads the docUrl property from [output/schema/schema.json](https://raw.githubusercontent.com/elastic/elasticsearch-specification/refs/heads/main/output/schema/schema.json). To support Serverless, we’d also need something like docUrlServerless in that schema so console can consume the correct link. Do you think that would be doable?"

This request does not affect the OpenAPI output, so as a writer my concern is just making it as easy as possible to maintain an accurate list of URLs for each API that exists in either stack or serverless API docs (or both). It seems likely that this will either mean a serverless-specific variation of https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv or else another column in that file to accommodate any doc IDs that have serverless-specific variations (i.e. all the API doc URLs and maybe some of the narrative doc URLs). Alternatively if there's some better way for Console code to pull from the OpenAPI files instead of the `schema.json`, for example, that's worth pondering as long as we end up with a way for them to resolve the unique set of URLs for each context.

## Motivation

The documentation links that exist within the Dev Tools console in Serverless projects are not pointing to the appropriate API docs. Rather, they just go to the landing page for all API docs.
If we get to the point were we have a consolidated OpenAPI document that covers both Stack and Serverless contexts there will no longer be a need for context-aware URLs but AFAIK that's not a near-term goal.

## Example

Image

In Serverless projects that link targets https://www.elastic.co/docs/api/ for all Elasticsearch APIs. In the non-Serverless use case that link targets the appropriate page for the API you've selected (e.g. https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-mapping)

miguelgrinberg:
Fixes #5188  (partially)

This change adds a column to `table.csv` for documentation URLs that apply to serverless. The new column is inserted in the 3rd position, right after `doc-url`. The value can be a URL, empty when there is no serverless equivalent, or the string `[use-doc-url]` to use the same URL for both stack and serverless.

The new documentation URLs are included in schema.json for all endpoints with the name `docUrlServerless`.

This change adds the support for generating the URLs. The table.csv entries are all set to `[use-doc-url]` for now. The docs team needs to fill URLs that are different as needed.


miguelgrinberg:
@lcawl This PR is waiting for you now. The 3rd column in table.csv is now the serverless doc url. If you leave it set to `[use-doc-url]` then the same URL is used for both. This is what I put in all the entries for now. You can replace this with a real URL, or with an empty string when there is no URL. Thanks!

pquentin:
Even though this is only for serverless, I added the `backport 9.1` and `backport 9.2` labels in the interest of reducing merge conflicts down the road, as the `table.csv` file is a big source of merge conflicts. Case in point, we have a conflict on this file right now.

lcawl:
I rebased this PR to catch the three PRs that had added rows in the `table.csv` file. I then used an AI script to update the file as follows: 271 rows were updated with valid serverless URLs, and 264 rows that contained Elasticsearch API URLs were skipped because their serverless URLs returned 404. I have left the latter as `[use-doc-url]` for now.

[Edit] On second thought, I set those 264 rows to an empty value instead, to more accurately indicate that they don't exist.

lcawl (approved):
This does seem to address my original issue, thanks! It also addresses the situation where we have serverless-only APIs like https://github.com/elastic/elasticsearch-specification/pull/5874

It might be cool to use that new serverless-specific column for all URL resolution in the serverless-specific OpenAPI output, but that can be discussed as a follow-up.

miguelgrinberg:
@lcawl Do you want to give this PR a last look before we merge? I had to do another merge of the main branch, and that came with one new row in table.csv, the one for `cef-processor`.

PR #5860, opened 2025-12-17
Add spec for project routing CRUD REST API endpoints

alexey-ivanov-es:
Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/139634


specification/project_routing/_types/RoutingExpression.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export class ProjectRoutingExpression {
```
flobernd, line 24: Is this class named the same in the Elasticsearch code base? I personally would have used `ProjectRouting` as the class name since the expression is only a part of it, but if it's the same in Elasticsearch, let's keep it as is 🙂

specification/project_routing/_types/RoutingExpression.ts
```diff
@@ -0,0 +1,22 @@
...
+  expression: string
```
flobernd: It _might_ make sense to use an alias type like `export type RoutingExpression = string` here instead of `string` directly. This allows upstream consumers of the spec to implement custom handling. Certain clients could e.g. implement DSL functionality and in the docs we could have a dedicated documentation for this expression.

specification/project_routing/create_single/CreateRoutingRequest.ts
```diff
@@ -0,0 +1,52 @@
...
+import { RequestBase } from '@_types/Base'
+import { MediaType } from '@_types/common'
+import { ProjectRoutingExpression } from '../_types/RoutingExpression'
+
+/**
+ * Create of update a single named project routing expression.
+ *
+ * Create of update a single named project routing expression.
+ * @doc_id project-routing
+ * @rest_spec_name project_routing.create_single
```
flobernd: Let's just use `.get`, `.create`, `.update`, etc. for the "single" version to keep it simple. The "many" version can still be called `create_multiple` (or `create_many`?).

flobernd (commented):
Looks good, thank you 🙂 Just a few minor questions/comments.

specification/project_routing/_types/RoutingExpression.ts
l-trotta: I'm looking at the server code (in serverless) and it seems like every class there is actually called `SomethingProjectRoutingExpression`, so I'd actually argue all other endpoint classes should be named as closely as possible to what's in the server code

alexey-ivanov-es: In the server code it is `SavedProjectRoutingExpression`, but I think `Saved` prefix would be confusing in API

specification/project_routing/_types/RoutingExpression.ts
alexey-ivanov-es: Done

specification/project_routing/create_single/CreateRoutingRequest.ts
alexey-ivanov-es: Done

specification/project_routing/_types/RoutingExpression.ts
l-trotta: agree for `Saved`, I'd keep `Expression` though

lcawl:
FYI These APIs are now documented here: https://www.elastic.co/docs/api/doc/elasticsearch-serverless/group/endpoint-project

PR #5861, opened 2025-12-17
Document the append processor's media_type

joegallo:
The `set` processor's `media_type` didn't document that there's a default value of `"application/json"`, which I've fixed. More than that, though, the `append` processor also has a `media_type` that behaves the same way.

Neither of these things are new, they've both been this way more or less forever -- I don't know how backports and whatnot are handled on this repo, though. Can you help me with that, @pquentin?

### DIFF
--- a/compiler/package-lock.json
+++ b/compiler/package-lock.json
@@ -553,6 +553,7 @@
       "integrity": "sha512-jnVe5ULKl6tijxUhvQeNbQG/84fHfg+yMak02cT8QVhBx/F05rAVxCGBYYTh2EKz22D6JF5ktXuNwdx7b9iEGw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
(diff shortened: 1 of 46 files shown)


joegallo:
I added some `backport` labels to cover all 9.x branches and the last 8.x branch. I don't know if what's ordinary for changes like this on this repo, though, so please tell me if we're should do something else.

pquentin:
> Neither of these things are new, they've both been this way more or less forever -- I don't know how backports and whatnot are handled on this repo, though. Can you help me with that, @pquentin?

Thanks for asking! Changes often break clients in surprising ways. In this case:

 * Server defaults are [only useful for docs](https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#server_default), so they can be backported.
 * Optional fields are also not breaking, so backporting is fine too.

Our backporting policy is mentioned at [the end of CONTRIBUTING.md](https://github.com/elastic/elasticsearch-specification/blob/main/CONTRIBUTING.md#send-your-pull-request-from-a-branch).

pquentin (approved):
Thanks! LGTM.

Looks like we don't have YAML tests for this, so it can't be validated.

(By the way, this would be nicer as an enum, but I think that would be breaking.)

pquentin:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.2|[](https://github.com/elastic/elasticsearch-specification/pull/5880)|
|✅|9.1|[](https://github.com/elastic/elasticsearch-specification/pull/5881)|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/5882)|

### Questions ?

PR #5780, opened 2025-12-02
Updated java examples

l-trotta:
Regenerated examples after bugfixes applied to the java request converter. 
Also fixed 5 broken examples: 
- SynonymsPutRequestExample1:  missing the required body
- ClusterPutComponentTemplateRequestExample2: indented wrongly
- PutAi21RequestExample2: wrong enum (chat-completion instead of chat_completion)
- PutLlamaRequestExample3: same as above
- InferenceUpdateExample1: missing required property "service"

PR #5865, opened 2025-12-18
update examples with latest request-converter

miguelgrinberg:
This PR refreshes all the language code examples. To ensure that always the latest request-converter is used for this task, I've also added an `npm update @elastic/request-converter` to the `make generate-language-examples` task.

### DIFF
--- a/docs/examples/generate-language-examples.js
+++ b/docs/examples/generate-language-examples.js
@@ -58,7 +58,7 @@ async function generateLanguages(example) {
     const partialRequest = await parseRequest(request);
     const java = new JavaCaller({
       minimumJavaVersion: 21,
-      jar: "path/to/converter/jar/java-es-request-converter-1.0-SNAPSHOT.jar",
+      jar: process.env.JAVA_ES_REQUEST_CONVERTER_JAR,
     });
 
     let correctParams = getCodeGenParamNames(partialRequest.params, partialRequest.request);
--- a/docs/examples/languageExamples.json
+++ b/docs/examples/languageExamples.json
@@ -18645,6 +18645,10 @@
     {
       "language": "curl",
       "code": "curl -X PUT -H \"Authorization: ApiKey $ELASTIC_API_KEY\" -H \"Content-Type: application/json\" -d '{\"service\":\"nvidia\",\"service_settings\":{\"url\":\"nvidia-completion-url\",\"api_key\":\"nvidia-completion-token\",\"model_id\":\"microsoft/phi-3-mini-128k-instruct\"}}' \"$ELASTICSEARCH_URL/_inference/completion/nvidia-completion\""
+    },
+    {
+      "language": "Java",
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -78,6 +78,7 @@ overlay-docs: ## Apply overlays to OpenAPI documents
 	rm output/openapi/elasticsearch-openapi-docs.tmp*.json
 
 generate-language-examples:
+	@npm update --prefix docs/examples @elastic/request-converter
```
pquentin, line 81: Should we pin the version in package.json/package-lock.json instead of always updating it?

l-trotta:
could we merge https://github.com/elastic/elasticsearch-specification/pull/5780 before? I've fixed some examples

Makefile
miguelgrinberg: The version is set to the latest 9.x.x in `docs/example/package.json`, so the `npm update` just makes sure the latest release is pulled. Not sure I understand your suggestion. Are you saying we should edit package.json to have an exact version specified each time this command runs?

pquentin: Disclaimer: I'm not super familiar with npm.

> The version is set to the latest 9.x.x in docs/example/package.json

It's set to the latest `9.1.x` (which I think is what `^9.1.2` does). You probably wanted something like `^9.x` instead.

> so the npm update just makes sure the latest release is pulled.

It does indeed update to the latest `9.x.x` (`9.3.0` right now), even though the package.json file forbids it.

> Are you saying we should edit package.json to have an exact version specified each time this command runs?

No, but I'm saying we should record the version we're using in the package-lock.json file, like we do for all other dependencies. This would:

 * allow us to use `npm clean-install` for reproducible builds and offline installs
 * properly record the actual version we're using
 * avoid surprises when trying to figure out the version that was used in CI
 * give us a single mechanism for dependencies versions

miguelgrinberg: The caret means "compatible with". When you say `^9.1.2` you are saying `>= 9.1.2, < 10.0.0`. So package.json is correct for 9.3.0 and all future updates until we go to v10. [semver docs](https://github.com/npm/node-semver?tab=readme-ov-file#caret-ranges-123-025-004)

> but I'm saying we should record the version we're using in the package-lock.json file

I agree in principle, but we do not run this command, the docs team use it. This isn't part of any automated workflows, it is pretty much one of the docs team members running this locally on their laptops to generate a fresh openapi file to upload to the docs site. If we want to enforce package-lock.json to be up to date then it will have to be them that commit it each time they use this command to generate an openapi update.

miguelgrinberg:
This PR is now updated. I noticed that at some point the examples were re-generated with an older request converter, because a bunch of them are incorrect. These are all fixed in this PR as well.

docs/examples/generate-language-examples.js
```diff
@@ -58,7 +58,7 @@ async function generateLanguages(example) {
     const partialRequest = await parseRequest(request);
     const java = new JavaCaller({
       minimumJavaVersion: 21,
-      jar: "path/to/converter/jar/java-es-request-converter-1.0-SNAPSHOT.jar",
+      jar: process.env.JAVA_ES_REQUEST_CONVERTER_JAR,
```
miguelgrinberg, line 61: Once we merge the Java PR in the request converter we will be able to remove the JavaCaller code and run the Java conversions in the same way as the other languages.

l-trotta:
@miguelgrinberg LGTM, but are we ready to merge the other PR?

PR #5866, opened 2025-12-19
Adds close_job param to the stop datafeed request

edsavage:
Adds the specification for the new `close_job` parameter to the specification for the stop datafeed request.

Added in https://github.com/elastic/elasticsearch/pull/138634


pquentin (approved):
Thanks! LGTM.

PR #5868, opened 2025-12-19
Support sort in ExtendedSearchUsage

pquentin:
Added by @davidkyle in https://github.com/elastic/elasticsearch/pull/139530. I initially wanted to do this instead:

```ts
export enum SortType {
  _doc,
  _geo_distance,
  _score,
  _script,
  field_sort,
}

sort?: Dictionary<SortType, long>
```

However, typescript-generator converts that to a `Record` which expects all keys to be present:

https://github.com/elastic/elasticsearch-specification/blob/696f1cd1e6cbb75d67c67a605ad8da3d53350de8/typescript-generator/src/index.ts#L129-L132

Interestingly, `SingleKeyDictionary` uses `Partial` which makes validation pass, but that's wrong. Since search usage is highly dynamic instead, I've opted for `string`.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -10404,11 +10404,6 @@ export interface ClusterStatsExtendedRetrieversSearchUsage {
 
 export interface ClusterStatsExtendedSearchUsage {
   retrievers?: ClusterStatsExtendedRetrieversSearchUsage
-  section?: ClusterStatsExtendedSectionSearchUsage
-}
-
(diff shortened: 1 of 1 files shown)


flobernd:
> However, typescript-generator converts that to a Record which expects all keys to be present:

Mhh, I wonder if this is intended and if we should really prefer `string` vs. a proper enum just because of that technical constraint?

PR #5887, opened 2026-01-07
[9.3] Revert "Mark chunk_rescorer as GA (#5871)"

pquentin:
This reverts commit 21c964435531ad230e8076c1ab2ce3f27119fd39. chunk_rescorer is only GA in 9.4, but we did not have a 9.3 branch when this commit was merged.


pquentin:
No, your commit will stay on `main` and end up in 9.4 clients and docs. I'm only reverting it for 9.3. And it's not your fault, we should have created the branches earlier!

PR #5889, opened 2026-01-07
Adds 'late_chunking' parameter to the JinaAI inference endpoint 

kosabogi:
This PR adds the `late_chunking` parameter to JinaAI Inference endpoint's task settings object.

Related issue: https://github.com/elastic/docs-content/issues/4264


specification/inference/_types/CommonTypes.ts
```diff
@@ -1711,6 +1711,16 @@ export class JinaAITaskSettings {
    * * `search`: Use it for storing embeddings of search queries run against a vector database to find relevant documents.
    */
   task?: JinaAITextEmbeddingTask
+  /**
+   * For a `text_embedding` task, controls when text is split into chunks.
+   *
+   * When set to `true`, Elasticsearch sends each document individually to Jina AI and embeds the full document before splitting it into chunks. This allows chunk embeddings to retain context from the entire document, improving semantic quality.
```
jonathan-buttner: Hmm I think we want to say something like:

_When set to `true`, a request from Elasticsearch will only contain chunks relating to a single document. Instead of batching chunks across documents, Elasticsearch puts them into separate requests. This allows chunk embeddings to retain context from the entire document, improving semantic quality._

kosabogi: Thanks for your review and suggestion! I’ve updated this part in my [latest commit](https://github.com/elastic/elasticsearch-specification/pull/5889/commits/1f00b476b26e7f7e5f26f4dcdaacc2d49eaf3147).

PR #5892, opened 2026-01-08
Fix permissions in validate-apis workflow

pquentin:

### DIFF
--- a/.github/workflows/validate-apis.yml
+++ b/.github/workflows/validate-apis.yml
@@ -8,7 +8,7 @@ on:
 
 permissions:
   pull-requests: write  # To create/update PR comments
-  permissions.id-token: write  # To use the ephemeral token
+  id-token: write  # To use the ephemeral token
 
 jobs:
   generate-validation:


.github/workflows/validate-apis.yml
```diff
@@ -7,7 +7,8 @@ on:
       - '[0-9]+.[0-9]+'
 
 permissions:
-  pull-requests: write    # To create/update PR comments
+  pull-requests: write  # To create/update PR comments
+  permissions.id-token: write  # To use the ephemeral token
```
MattDevy:
```suggestion
  id-token: write  # To use the ephemeral token
```

MattDevy: See: https://github.com/elastic/go-elasticsearch/blob/14ed4dfab541f30eb04f75dc76d4e01c8927124b/.github/workflows/release-please.yml#L12-L23 for example

MattDevy: Thinking about it `vault-role: "token-policy-6a6580bb0e7d"` may not be needed in the `Fetch ephemeral GitHub token` step

PR #5900, opened 2026-01-08
Use ephemeral GitHub token to update output files

pquentin:

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -31,6 +31,11 @@ jobs:
     # Avoid running on forks and prevent infinite loop
     if: github.repository_owner == 'elastic' && github.actor != 'elasticmachine'
     steps:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/generate.yml
```diff
@@ -22,17 +22,24 @@ defaults:
...
 
+permissions:
+  id-token: write # To use the ephemeral token
+
 jobs:
   build:
     runs-on: ubuntu-latest
-    # The actor check currently does not work. The `secrets.PAT` belongs to `delvedor` and not
-    # `elasticmachine`. We should make sure to change this in the future.
-    if: github.repository_owner == 'elastic' && github.actor != 'elasticmachine'
```
pquentin, line 30: I don't think this check is actually needed. Since commits made with GitHub Actions don't trigger other actions, there's no risk of an infinite loop.

PR #5931, opened 2026-01-14
Disallow deprecated on everything but requests

l-trotta:
Closes https://github.com/elastic/elasticsearch-specification/issues/2643.
Removes the deprecation annotation for every base type that is not a request.

### DIFF
--- a/compiler/src/model/build-model.ts
+++ b/compiler/src/model/build-model.ts
@@ -157,7 +157,7 @@ export function compileSpecification (specsFolder: string, outputFolder: string)
   // Remove deprecation from everything other than request
   for (const type of model.types) {
     if (type.kind !== 'request') {
-      type.deprecation = undefined
+      delete type.deprecation
     }
   }
 


compiler/src/model/build-model.ts
```diff
@@ -154,6 +154,13 @@ export function compileSpecification (specsFolder: string, outputFolder: string)
     model.endpoints.push(endpointMappings[key])
   }
 
+  // Remove deprecation from everything other than request
+  for (var type of model.types) {
```
MattDevy:
```suggestion
  for (const type of model.types) {
```

compiler/src/model/build-model.ts
```diff
@@ -154,6 +154,13 @@ export function compileSpecification (specsFolder: string, outputFolder: string)
...
+    if (type.kind !== 'request') {
+      type.deprecation = undefined
```
MattDevy: Should this be a delete? To remove the key entirely from the output? Or do we want a null here

```suggestion
      delete type.deprecation
```

compiler-rs/compiler-wasm-lib/pkg/compiler_wasm_lib_bg.wasm
MattDevy, line 1: Should this be here? Was it accidentally committed, or if this intentional?

l-trotta: we need to update the wasm every time we update the rust openapi converter, I don't think this is part of ci..?

PR #5932, opened 2026-01-14
Fix project APIs

pquentin:
* Moves APIs from `project_routing` to the existing `project` namespace. Should `expression` be in the name? For example,`project.create_many_routing_expressions` vs. `project.create_many_routings`. It seems like the API names are already long enough.
* Adds missing request media types.
* Removes descriptions that duplicate the summary line.

Relates https://github.com/elastic/elasticsearch/pull/140647


alexey-ivanov-es:
>Should expression be in the name? For example,project.create_many_routing_expressions vs. project.create_many_routings. It seems like the API names are already long enough.

I think `expression` shouldn't be in the name. However, it probably should be `project_routing` (`create_many_project_routings`), because in other endpoints it's called `project_routing`

pquentin:
Thanks! Fixed.

pquentin:
The build is passing but status are "delayed": https://www.githubstatus.com/incidents/5ccghcfrkv39

PR #5933, opened 2026-01-14
Add missing similarity field to PUT OpenAI inference API

szabosteve:
## Overview

Relates to https://github.com/elastic/enhancements/issues/22952

This PR adds the `similarity` service setting to the PUT OpenAI inference API docs.

### DIFF
--- a/specification/inference/_types/CommonTypes.ts
+++ b/specification/inference/_types/CommonTypes.ts
@@ -1958,7 +1958,7 @@ export class OpenAIServiceSettings {
    */
   rate_limit?: RateLimitSetting
   /**
-   * For a `text_embedding` task, the similarity measure. One of cosine, dot_product, l2_norm.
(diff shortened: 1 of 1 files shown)


specification/inference/_types/CommonTypes.ts
```diff
@@ -1957,6 +1957,10 @@ export class OpenAIServiceSettings {
    * For `completion`, it is set to `500`.
    */
   rate_limit?: RateLimitSetting
+  /**
+   * For a `text_embedding` task, the similarity measure. One of cosine, dot_product, l2_norm.
```
jonathan-buttner: Let's also add that if it's null it defaults to `dot_product`.

szabosteve:
```suggestion
   * For a `text_embedding` task, the similarity measure. One of cosine, dot_product, l2_norm. Defaults to `dot_product`.
```

PR #5934, opened 2026-01-15
Add embedding inference task specification

DonalEvans:
Related Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/140696

This commit adds documentation and updates the API spec to include the new embedding inference task, which has been implemented for the JinaAI service.

Additional changes in this commit:
- Rename TextEmbedding\*Result classes to DenseEmbedding\*Result since they're used for both text-only and multimodal embeddings


output/schema/validation-errors.json
```diff
@@ -60,6 +60,7 @@
     "Dangling type 'inference._types:AmazonSageMakerElementType'",
     "Dangling type 'inference._types:AmazonSageMakerOpenAiTaskSettings'",
     "Dangling type 'inference._types:AmazonSageMakerSimilarity'",
+    "Dangling type 'inference._types:JinaAIElementType'",
```
pquentin: Can you please remove this now unused type?

DonalEvans: Ah, thanks for catching that. I was originally planning to fix some other issues with the documentation (such as not documenting the element type field for JinaAI) but decided to split those changes out into a separate PR to make backporting easier.

specification/package-lock.json
```diff
@@ -516,6 +516,7 @@
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.15.0.tgz",
       "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
       "license": "MIT",
+      "peer": true,
```
jonathan-buttner, line 519: We do need these changes?

DonalEvans: I don't think they're strictly needed in the PR, but they get generated when preparing the docs using [the steps here](https://www.elastic.co/docs/contribute-docs/api-docs/elasticsearch-api-docs-quickstart#quickstart), so it doesn't feel wrong to include them I guess? The next person to run the commands in that quickstart document will generate the same changes anyway, I assume.

PR #5949, opened 2026-01-16
Fix errors in Inference API specification

DonalEvans:
- Remove comments about the API key not being able to be updated. The _update operation allows API keys to be updated.
- Make model_id field required in JinaAIServiceSettings
- Include dimensions and element_type fields in JinaAIServiceSettings
- Include is_truncated field in SparseEmbeddingResult
- Fix SparseEmbeddingResponseExample1.yaml to use correct format
- Fix TextEmbeddingResponseExample1.yaml to use correct format


jonathan-buttner (approved):
Thanks for doing the clean up!

PR #5956, opened 2026-01-21
ESQL: Timezone API docs

ivancea:
Main ES docs PR: https://github.com/elastic/elasticsearch/pull/140986

Current API docs: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-query

Time_zone is moving from snapshot-only to tech-preview:
- Add the `time_zone` parameter in the ESQL query request
- Fix the `locale` parameter in there too, that was missing an asterisk, and had no documentation

PR #5963, opened 2026-01-21
ESQL named parameters

cimequinox:
@ivancea identified a gap in our specifications in #5956 regarding ESQL named parameters. Named parameters are provided as a json list of objects of the form `{ "name": value }` which the specification does not allow. This change amends the specification to allow these.

See also [rest-api-spec example](https://github.com/elastic/elasticsearch/pull/138051/files#diff-9e4a38fbf1ed2267447b4ac4c492670d5b2cc4186b316475f141a0eeccee5375R648) and our [slack discussion from 1/21](https://docs.google.com/document/d/1v5BJQPAkGjZxqpZ_wdLBXsMHRzvEwFRXTmfPgZls2tQ/edit?usp=sharing
)

### DIFF
--- a/specification/esql/_types/types.ts
+++ b/specification/esql/_types/types.ts
@@ -18,10 +18,10 @@
  */
 
 import { FieldValue } from '@_types/common'
-import { Dictionary } from '@spec_utils/Dictionary'
+import { SingleKeyDictionary } from '@spec_utils/Dictionary'
 
-export type ESQLParam = SingleOrMultiValue | NamedValue
-export type NamedValue = Dictionary<string, SingleOrMultiValue>
+export type ESQLParams = SingleOrMultiValue[] | NamedValue[]
+export type NamedValue = SingleKeyDictionary<string, SingleOrMultiValue>
 export type SingleOrMultiValue = FieldValue | FieldValue[]
 
 /**
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -6564,10 +6564,11 @@
                   },
                   "params": {
                     "description": "To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.",
-                    "type": "array",
-                    "items": {
-                      "$ref": "#/components/schemas/esql._types.ESQLParam"
-                    }
+                    "allOf": [
+                      {
+                        "$ref": "#/components/schemas/esql._types.ESQLParams"
+                      }
+                    ]
                   },
                   "profile": {
                     "description": "If provided and `true` the response will include an extra `profile` object\nwith information on how the query was executed. This information is for human debugging\nand its format can change at any time but it can give some insight into the performance\nof each part of the query.",
@@ -64458,13 +64459,19 @@
           "arrow"
         ]
       },
-      "esql._types.ESQLParam": {
+      "esql._types.ESQLParams": {
         "oneOf": [
           {
-            "$ref": "#/components/schemas/esql._types.SingleOrMultiValue"
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/esql._types.SingleOrMultiValue"
+            }
           },
           {
-            "$ref": "#/components/schemas/esql._types.NamedValue"
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/esql._types.NamedValue"
+            }
           }
         ]
       },
@@ -64485,7 +64492,9 @@
         "type": "object",
         "additionalProperties": {
           "$ref": "#/components/schemas/esql._types.SingleOrMultiValue"
-        }
+        },
+        "minProperties": 1,
+        "maxProperties": 1
       },
       "esql._types.TableValuesContainer": {
         "type": "object",
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -11410,7 +11410,7 @@ export interface EsqlAsyncEsqlResult extends EsqlEsqlResult {
   is_running: boolean
 }
 
-export type EsqlESQLParam = EsqlSingleOrMultiValue | EsqlNamedValue
+export type EsqlESQLParams = EsqlSingleOrMultiValue[] | EsqlNamedValue[]
 
 export interface EsqlESQLView {
   name: string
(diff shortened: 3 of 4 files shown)


specification/esql/_types/types.ts
```diff
@@ -18,8 +18,10 @@
  */
 
 import { FieldValue } from '@_types/common'
+import { Dictionary } from '@spec_utils/Dictionary'
 
-export type ESQLParam = FieldValue | FieldValue[]
+export type ESQLParam = FieldValue | FieldValue[] | NamedValue
+export type NamedValue = Dictionary<string, FieldValue | FieldValue[]>
```
ivancea: As to organize and explain the types a bit, I would maybe extract `FieldValue | FieldValue[]` into a type like `SingleOrMultiValue`

cimequinox: Done

specification/esql/_types/types.ts
```diff
@@ -18,8 +18,11 @@
  */
 
 import { FieldValue } from '@_types/common'
+import { Dictionary } from '@spec_utils/Dictionary'
 
-export type ESQLParam = FieldValue | FieldValue[]
+export type ESQLParam = FieldValue | FieldValue[] | NamedValue
```
ivancea:
```suggestion
export type ESQLParam = SingleOrMultiValue | NamedValue
```

specification/esql/_types/types.ts
ivancea, line 1: Also, somewhat unrelated, given only either named or non-named params area llowed (right?), maybe we should convert the request type from:
`ESQLParam[]`
To:
`SingleOrMultiValue[] | NamedValue[]`

As it's either one or the other

ivancea: Just checked, and indeed. I don't know how will this be rendered, but unless it looks _bad_, this would be the correct type

specification/esql/_types/types.ts
cimequinox: Done

specification/esql/_types/types.ts
cimequinox: We currently use an Array in [params](https://github.com/elastic/elasticsearch-specification/blob/main/specification/esql/query/QueryRequest.ts#L95) 
```ts
    /**
     * To avoid any attempts of hacking or code injection, extract the values in a ...
     * @doc_id esql-query-params
     */
    params?: Array<ESQLParam>
```
which did confuse me a bit yesterday but I'm new to this repo.  
I was expecting something like `params?: ESQLParams` with the Array 
inside ESQLParams e.g. `export type ESQLParams = Array<...>`.
but `% rg ': Array'` shows the current use appears in ~100 other places in this repository
so I assume that's the convention here.

cimequinox: Hmm.. looking at other definitions in common.ts e.g. [IndexPatterns](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/common.ts#L64) shows we occasionally keep the Array in the type so I like this variation of your suggestion
```ts
export type ESQLParams = SingleOrMultiValue[] | NamedValue[]
```

l-trotta:
hey @cimequinox, should this be also applied to the async query api?

cimequinox:
@l-trotta in principle yes, but looking the docs I see a slight difference in how params are specified
* [Run an ES|QL query - params](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-query#operation-esql-query-body-application-json-params) 
 
  > params array[number | string | boolean | null | array] | array[object]
  > To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.
  > One of:
[array-1 array[number | string | boolean | null | array]](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-query#operation-esql-query-body-application-json-params)
[array-2 array[object]](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-query#operation-esql-query-body-application-json-array-2-params-array-object)

* [Run an async ES|QL query - params](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-async-query#operation-esql-async-query-body-application-json-params)
   
   > params array[number | string | boolean | null]
   > To avoid any attempts of hacking or code injection, extract the values in a separate list of parameters. Use question mark placeholders (?) in the query string for each of the parameters.

It looks like an oversight.

cimequinox:
Looking closer I see this change updated the [spec for /_query](https://github.com/elastic/elasticsearch-specification/blob/main/specification/esql/query/QueryRequest.ts#L102) but did not update the [spec for /_query/async](https://github.com/elastic/elasticsearch-specification/blob/main/specification/esql/async_query/AsyncQueryRequest.ts#L108).

l-trotta:
@cimequinox yeah the API docs are currently generated from this specification, so they're technically correct. should we port the changes to async query as well?

cimequinox:
I want to look at our existing async api tests and perhaps add one 
that uses named parameters if we don't already have one.
Once I've confirmed that I'll create a PR to update the spec for `/_query/async`.
I think the change is just `params?: Array<FieldValue>` ─▶ `params?: ESQLParams`.
I should probably also see if there any other `params` hiding somewhere as well.

flobernd:
@cimequinox Sounds like a plan, thank you!

AFAIK the async query request should be a superset of the regular query request.

@l-trotta Not sure if this would be to implicit/complicated, but looking at the recent OverloadOf improvements, we might also want to support that on requests (affecting only the body). OR adding an annotation that forces "value bodies" to get inlined as "properties body". We have a few cases that would benefit from this.

flobernd:
Hi @cimequinox ,

I'm currently blocked by the incorrect specification of the async query parameters and did some research. It seems like both the regular [query action](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/RestEsqlQueryAction.java) and the [async query action](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/RestEsqlAsyncQueryAction.java#L55) are using the same [`EsqlQueryRequest`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/EsqlQueryRequest.java#L206) type (async forwards to the "regular" [processing method](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/RestEsqlQueryAction.java#L59)):

I created #6141 to update the `params` field for the async query request.

PR #5969, opened 2026-01-22
Avoid deprecation availability in types lint rule

l-trotta:
As titled, also fixing the 2 errors it caught (which were redundant because the annotations were already correctly present at the field level)

### DIFF
--- a/validator/rules/no-deprecated-availability-on-types.js
+++ b/validator/rules/no-deprecated-availability-on-types.js
@@ -28,8 +28,17 @@ export default createRule({
     const sourceCode = context.sourceCode || context.getSourceCode()
 
     return {
-      'TSEnumDeclaration, TSTypeAliasDeclaration, ClassDeclaration'(node) {
+      'TSInterfaceDeclaration, TSEnumDeclaration, TSTypeAliasDeclaration, ClassDeclaration'(node) {
         const className = node.id?.name
+
+        if (node.type === 'TSInterfaceDeclaration') {
+          if (className === 'Request') return
+          const extendsRequestBase = node.extends?.some(
+            (e) => e.expression?.type === 'Identifier' && e.expression.name === 'RequestBase'
+          )
+          if (extendsRequestBase) return
+        }
+
         const targetNode =
             node.parent?.type === 'ExportNamedDeclaration' ? node.parent : node
         const comments = sourceCode.getCommentsBefore(targetNode)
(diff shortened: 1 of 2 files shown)


flobernd:
Thank you!

What do think about making this more generic? Like maintaining a list of allowed tags per type in the metamodel (alias, interface, ..., property, enum-member, ..., request, ...).

The compiler already does these kind of checks internally, but it might be useful for minting as well.

We can do this in a follow up task, but we should chose a more generic name for the linter rule/check right from the start, if we want to go that route 🙂

validator/rules/no-deprecated-availability-on-types.js
```diff
@@ -0,0 +1,95 @@
...
+            }
+          })
+        }
+      }
+    }
+  },
+  meta: {
+    docs: {
+      description:
+        'AAAA'
```
MattDevy: This looks like a placeholder?

validator/rules/no-deprecated-availability-on-types.js
```diff
@@ -0,0 +1,95 @@
...
+  (name) => `https://example.com/rule/${name}`
+)
+
+export default createRule({
+  name: 'no-deprecated-availability-on-types.js',
+  create(context) {
+    const sourceCode = context.sourceCode || context.getSourceCode()
+
+    return {
+      'TSEnumDeclaration, TSTypeAliasDeclaration, ClassDeclaration'(node) {
```
MattDevy: Are there Interfaces in the spec that we also want to disallow availability and deprecated tags on? If so we should add `TSInterfaceDeclaration` and exclude any named `Request` or that extend `RequestBase`

validator/rules/no-deprecated-availability-on-types.js
```diff
@@ -0,0 +1,95 @@
...
+ * under the License.
+ */
+import { ESLintUtils } from '@typescript-eslint/utils'
+
+const createRule = ESLintUtils.RuleCreator(
+  (name) => `https://example.com/rule/${name}`
+)
+
+export default createRule({
+  name: 'no-deprecated-availability-on-types.js',
```
MattDevy:
```suggestion
  name: 'no-deprecated-availability-on-types',
```

MattDevy (commented):
A few minor suggestions here. I agree with @flobernd , it would be nice to make this more generic so it can be extended in the future by simply adding more tags that should not be present

PR #5982, opened 2026-02-03
added _name to knn search

msimalunsal:
[Closes elastic/elasticsearch-java#1156](https://github.com/elastic/elasticsearch-java/issues/1156)

Closes #5987 

This PR adds the missing optional `_name` field to the `Knn Search` query definition in the API specification.

Currently the `_name` parameter is not present in the common specification, which makes it impossible for Java Client to define named knn queries. Although Elasticsearch supports named queries in general, this field is not yet exposed for knn queries through the generated client APIs.

This change updates the specification so that client generators can properly support the `_name` parameter for knn queries.

Implementation details:
- Added `_name?: string` to the `Knn.ts` type definition.
- The field is annotated using `/** @codegen_name query_name */`, following the same convention used in `function_score` queries for named queries.

This update aligns the knn query definition with the existing named query mechanism used across other query types.

Regarding contribution workflow:
The repository guidelines mention sending pull requests from a branch rather than a fork. However, as an external contributor I do not have write access to the `elastic/elasticsearch-specification` repository. Attempting to push a branch directly resulted in the following error:
`ERROR: Permission to elastic/elasticsearch-specification.git denied`
For this reason, the change is submitted via a fork. 

Thanks!

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -40499,6 +40499,9 @@
                 "$ref": "#/components/schemas/_types.RescoreVector"
               }
             ]
+          },
+          "_name": {
+            "type": "string"
           }
         },
         "required": [
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -2678,6 +2678,7 @@ export interface KnnSearch {
   similarity?: float
   inner_hits?: SearchInnerHits
   rescore_vector?: RescoreVector
+  _name?: string
 }
 
 export interface LatLonGeoLocation {
--- a/specification/_types/Knn.ts
+++ b/specification/_types/Knn.ts
@@ -66,6 +66,8 @@ export interface KnnSearch {
    * @availability serverless
    */
   rescore_vector?: RescoreVector
+  /** @codegen_name query_name */
+  _name?: string
 }
 
 /**


pquentin:
Thanks!

~~We just have one small issue here: Elasticsearch does not implement `_name` in [KnnSearch](https://github.com/elastic/elasticsearch/blob/929406ed7583ce591a52e27d2ac3574d2af8cf5b/server/src/main/java/org/elasticsearch/search/vectors/KnnSearchRequestParser.java#L198-L204), it's a more general concept implemented in [AbstractQueryBuilder](https://github.com/elastic/elasticsearch/blob/929406ed7583ce591a52e27d2ac3574d2af8cf5b/server/src/main/java/org/elasticsearch/index/query/AbstractQueryBuilder.java#L55-L56).~~

~~The correct fix is to extend `QueryBase`, but that also means removing `boost` from `KnnSearch`. I'll need to check if that breaks language clients, including the Java one.~~


Nevermind, `KnnSearch` is not a query, and boost/name were added "manually" in https://github.com/elastic/elasticsearch/pull/107645, so this change looks good to me.

msimalunsal:
Thank you for the fast review and validation. Really appreciate the support!

I have one more question regarding compatibility:  
Once this change is available in elasticsearch-java 9.x.x, will it be possible to use this feature while still running Elasticsearch server 8.x? Additionally, will the includeNamedQueriesScore feature also work when using the 9.x Java client against an 8.x Elasticsearch cluster?

Thanks in advance!

l-trotta:
Hey @msimalunsal, the java client is [forward compatible](https://github.com/elastic/elasticsearch-java?tab=readme-ov-file#compatibility), so using a version 9 client to connect to a version 8 server is not allowed, but the change will be backported to the next java client 8.19.x release, so you can keep using that!

l-trotta:
manually backported 8.19

erayarslan:
hi @l-trotta,
when will the backported version be released? I think 8.19.11 is the latest version, and we desperately need 8.19.12 🙏

l-trotta:
@erayarslan  8.19.12 will be out before the end of the month!

PR #5995, opened 2026-02-05
Adding new `_name` to function score objects

benwtrent:
Related: https://github.com/elastic/elasticsearch/pull/141950

### DIFF
--- a/specification/_types/query_dsl/compound.ts
+++ b/specification/_types/query_dsl/compound.ts
@@ -268,6 +268,7 @@ export class FunctionScoreContainer {
   /**
    * A name to identify which function matched and influenced the score.
    * @availability stack since=9.4.0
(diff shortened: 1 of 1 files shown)


specification/_types/query_dsl/compound.ts
```diff
@@ -265,6 +265,11 @@ export class FunctionScoreContainer {
   filter?: QueryContainer
   /** @variant container_property */
   weight?: double
+  /**
+   * A name to identify which function matched and influenced the score.
+   * @availability stack since=9.4.0
```
Anaethelion, line 270:
```suggestion
   * @availability stack since=9.4.0
   * @variant container_property
```

This should be labeled `container_property` otherwise it will be considered as an alternative member of the container by clients generators.

PR #5998, opened 2026-02-06
Add descriptions to node usage response

samxbr:
Adds descriptions for the response of the [node usage API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-nodes-usage). Specifically this adds a note to indicate the `rest_actions` names are __NOT__ considered stable.

The changes in the PR are made following these instructions: https://www.elastic.co/docs/contribute-docs/api-docs/elasticsearch-api-docs-quickstart

Related to https://github.com/elastic/elasticsearch/pull/140527

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   },
   "dependencies": {
     "@redocly/cli": "^1.34.6",
-    "bump-cli": "^2.9.11"
+    "bump-cli": "^2.9.10"
   },
   "version": "overlay"
 }
--- a/compiler/package-lock.json
+++ b/compiler/package-lock.json
@@ -553,7 +553,6 @@
       "integrity": "sha512-r0bBaXu5Swb05doFYO2kTWHMovJnNVbCsII0fhesM8bNRlLhXIuckley4a2DaD+vOdmm5G+zGkQZAPZsF80+YQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~7.16.0"
       }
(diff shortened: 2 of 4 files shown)


samxbr:
The [API spec](https://www.elastic.co/docs/api/doc/elasticsearch/v9/) have only v8 and v9, so I am only backporting to 8.19 and 9.3. Let me know if we should backport to other versions.

specification/nodes/usage/types.ts
```diff
@@ -23,8 +23,21 @@ import { Dictionary } from '@spec_utils/Dictionary'
...
+   */
   since: EpochTime<UnitMillis>
+  /**
+   * The timestamp for when these statistics were collected.
+   */
   timestamp: EpochTime<UnitMillis>
+  /**
+   * The total number of times search aggregations have been called on this node since the last restart.
+   */
   aggregations: Dictionary<string, UserDefinedValue>
```
samxbr, line 42: Highlight the "real" change, the rest are generated.

package.json
```diff
@@ -4,7 +4,7 @@
   },
   "dependencies": {
     "@redocly/cli": "^1.34.6",
-    "bump-cli": "^2.9.10"
+    "bump-cli": "^2.9.11"
```
samxbr: This is auto-included, which seems innocent. But I am happy to remove it there's concern.

PeteGillinElastic:
A couple of process nits:

I think we want to revert the `package-lock.json` changes. I don't know why the tooling changed those for you, but they shouldn't happen as part of the change.

And I was told earlier today that it's best not to include the `output` changes, because it makes backporting earlier. If you just commit the `specification` change then I think something will come along afterwards and catch up the generated code.

PeteGillinElastic:
> The [API spec](https://www.elastic.co/docs/api/doc/elasticsearch/v9/) have only v8 and v9, so I am only backporting to 8.19 and 9.3. Let me know if we should backport to other versions.

Yeah, as chance would have it, I was making a change earlier which would theoretically apply to all past versions going back years, and the advice I was given by the devtools folks was indeed to backport to 8.19 and 9.3.

PeteGillinElastic (approved):

samxbr:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/6005)|

### Questions ?

PR #6013, opened 2026-02-11
[DOCS]: Correct `_cat/count` method to `GET`

charlotte-hoblik:
Updates the documentation to correctly list the _cat/count endpoint as a GET request instead of POST.

Connected to: https://github.com/elastic/docs-content/issues/4598

Preview: 
Screenshot 2026-02-11 at 17 36 15

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -937,7 +937,7 @@
           "cat"
         ],
         "summary": "Get a document count",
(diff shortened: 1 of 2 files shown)


pquentin:
This isn't a mistake, `_cat/count` also supports `POST` since Elasticsearch 9.3.0: https://github.com/elastic/elasticsearch/pull/138369. I don't know if there's a way to specify that in the docs, though.

charlotte-hoblik:
Thanks @pquentin! To clarify, I didn't remove POST, I just swapped the order so the UI defaults to GET (since that's the main use case). Good call on the 9.3 update though. I'll add a note to the description clarifying that POST is supported for serverless routing.

PR #6019, opened 2026-02-13
Fix routing being used in bodies

l-trotta:
Followup of https://github.com/elastic/elasticsearch-specification/pull/5676. The change works fine for cases where `Routing` is used in path or query parameters, because clients serialize `string[]` as csv string. But when applied to fields in bodies, static clients simplify the `string | string[]` union in `string[]`, which gets serialized into an actual json array, which the server doesn't accept. Currently all the fields modified with this PR are broken in the static clients. 

Originally reported in https://github.com/elastic/elasticsearch-java/issues/1161

### DIFF
--- a/docs/add-new-api.md
+++ b/docs/add-new-api.md
@@ -161,15 +161,24 @@ And the generic will be used somewhere inside the definition.
 #### Path/Query parameters csv
 
 In case the server accepts either a string or comma separated values for path or query parameters, they can be mapped as:
+
 ```ts
+export type MyAlias = string | string[]
+
 path_parts: {
-    param_name?: string | string[]
+    param_name?: MyAlias
 }
 query_parameters: {
-    param_name?: string | string[]
+    param_name?: MyAlias
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/add-new-api.md
```diff
@@ -158,6 +158,19 @@ class Response<Generic> {
...
+```ts
+path_parts: {
+    param_name?: string | string[]
+}
+query_parameters: {
+    param_name?: string | string[]
+}
+```
+This is just an example, it's better to create an alias and use that instead of the unnamed union. Even better, the right alias might exist already, so always check [common.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/common.ts) first.  
+
```
flobernd, line 182:
````suggestion
In case the server accepts either a string or comma separated values for path or query parameters, they can be mapped as:

```ts
export type MyAlias = string | string[]

path_parts: {
    param_name?: MyAlias
}
query_parameters: {
    param_name?: MyAlias
}
```

The statically typed clients will take care of proper serialization.

The right alias might exist already, so always check [common.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/common.ts) first.

> [!WARNING]
> If the same `string | string[]` construct (or alias) is also used in body/interface properties, clients will serialize these values to a standard JSON array instead - which might or might not be desired.

````

PR #6021, opened 2026-02-13
Update inference api specification for mixedbread

Evgenii-Kazannik:


jonathan-buttner:
Closing because mixedbread is no longer supporting the apis and we reverted the ES code.

PR #6029, opened 2026-02-17
Fix default value for the size parameter of GetSnapshots API

ywangd:
The default is -1 for no limit instead of 0 which triggers validation error.


DaveCTurner (approved):
Good catch yes

pquentin (approved):

pquentin:
We no longer maintain the 9.0 and 9.1 branches, and docs are cumulative, so I removed the 9.0 and 8.19 backports.

PR #6037, opened 2026-02-20
Updates the Update documents API's max_docs parameter description

kosabogi:
The current `max_docs` documentation states that a scroll is not used when `max_docs` <= `scroll_size`. However, this is only true when `conflicts=abort`. When `conflicts=proceed`, a scroll may still be required to ensure up to `max_docs` successful updates. 

This PR updates the documentation to reflect the full condition.

Related issue: https://github.com/elastic/elasticsearch/issues/136067

### DIFF
--- a/compiler/package-lock.json
+++ b/compiler/package-lock.json
@@ -553,7 +553,6 @@
       "integrity": "sha512-r0bBaXu5Swb05doFYO2kTWHMovJnNVbCsII0fhesM8bNRlLhXIuckley4a2DaD+vOdmm5G+zGkQZAPZsF80+YQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~7.16.0"
       }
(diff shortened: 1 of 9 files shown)


specification/_global/update_by_query/UpdateByQueryRequest.ts
```diff
@@ -216,7 +216,7 @@ export interface Request extends RequestBase {
     /**
      * The maximum number of documents to process.
      * It defaults to all documents.
-     * When set to a value less then or equal to `scroll_size` then a scroll will not be used to retrieve the results for the operation.
+     * When set to a value less then or equal to `scroll_size` and `conflicts` is set to `abort`, a scroll will not be used to retrieve the results for the operation.
```
leemthompo:
```suggestion
     * When set to a value less than or equal to `scroll_size` and `conflicts` is set to `abort`, a scroll will not be used to retrieve the results for the operation.
```

leemthompo (approved):
🚀

PR #6059, opened 2026-02-24
Validation and docs for codegen_names

l-trotta:
Adding @codegen_names to the "variants" section in the modeling guide, and updating the "prefer-tagged-variants" lint rule so that it also considers codegen_names.

### DIFF
--- a/compiler/package-lock.json
+++ b/compiler/package-lock.json
@@ -200,9 +200,9 @@
       }
     },
     "node_modules/@eslint/eslintrc/node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.3.tgz",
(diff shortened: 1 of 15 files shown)


validator/rules/prefer-tagged-variants.js
```diff
@@ -102,23 +105,61 @@ export default createRule({
...
+                suggestion: 'Missing comment: use tagged variants with @variants internal or @variants container (external), or any other variant option. See modeling guide: https://github.com/elastic/elasticsearch-specification/blob/main/docs/modeling-guide.md#variants'
+              }
+            })
+            return
+          }
+
+          const blockComment = jsdoc.value
+
+          const hasCodegenNamesTag =
+              /@codegen_names\s/.test(
```
MattDevy: I'm not sure if this would be caught by another linting rule, but if the user simply puts `@codegen_names ` in the jsdoc, and fails to put actual names after, should we not catch this? Maybe this could be a follow up?

validator/rules/prefer-tagged-variants.js
```diff
@@ -102,23 +105,61 @@ export default createRule({
...
+                  blockComment
+              )
+
+          const hasVariantsTag =
+              /@variants\s/.test(
```
MattDevy: Similar to above, do we already catch if a user puts variants without any parameters?

MattDevy (approved):
Some minor non-blocking suggestions, but otherwise LGTM

validator/rules/prefer-tagged-variants.js
l-trotta: good point, for now I'll just check if there's some text after the annotation, then we can discuss if we want to check for syntax (which could get complex)

flobernd:
The compiler will emit an error for wrong tag syntax if I'm not mistaken. Having it directly in the linter would be cool, but for now we should be fine.

PR #6070, opened 2026-02-27
Expand info about deprecation API usage 

leemthompo:
Contributes to https://github.com/elastic/elasticsearch/issues/143234

Updated documentation to clarify the use of the deprecation API.

### DIFF
--- a/docs/overlays/elasticsearch-shared-overlays.yaml
+++ b/docs/overlays/elasticsearch-shared-overlays.yaml
@@ -312,6 +312,13 @@ actions:
           externalDocs:
             url: https://www.elastic.co/docs/explore-analyze/query-filter/languages/sql
             description: Check out the overview and tutorials for the Elasticsearch SQL features.
+        - name: streams
+          x-displayName: Streams
+          # description: >
+          #   The streams APIs enable you to manage streams.
+          # externalDocs:
+          #   url: https://www.elastic.co/docs/solutions/observability/streams/streams
+          #   description: Learn more about streams.
         - name: synonyms
           x-displayName: Synonyms
           description: >
--- a/specification/streams/logs_disable/StreamsDisableRequest.ts
+++ b/specification/streams/logs_disable/StreamsDisableRequest.ts
@@ -30,6 +30,7 @@ import { StreamType } from '@streams/_types/StreamType'
  * @availability stack since=9.1.0 stability=experimental visibility=public
  * @cluster_privileges manage
  * @doc_id streams-logs-disable
+ * @doc_tag streams
  */
 export interface Request extends RequestBase {
   urls: [
--- a/specification/streams/logs_enable/StreamsEnableRequest.ts
+++ b/specification/streams/logs_enable/StreamsEnableRequest.ts
@@ -35,6 +35,7 @@ import { StreamType } from '@streams/_types/StreamType'
  * @availability stack since=9.1.0 stability=experimental visibility=public
  * @cluster_privileges manage
  * @doc_id streams-logs-enable
+ * @doc_tag streams
  */
 export interface Request extends RequestBase {
   urls: [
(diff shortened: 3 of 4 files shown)


DaveCTurner (commented):
I think this API is also useful in its own right - although it is required for customers to address all deprecations before upgrading to v10, the more diligent ones might want to migrate away from their usage of deprecated features even today, and this API will let them do so.

The thing I think we're missing here is a clearer statement that "deprecated" imples "will continue to work, and is fully supported, throughout the whole 9.x series".

leemthompo:
had another crack, I guess could be more explicit about 9.x and 10.x 🤔

DaveCTurner (commented):
How about something like this?

```diff
diff --git a/specification/migration/deprecations/DeprecationInfoRequest.ts b/specification/migration/deprecations/DeprecationInfoRequest.ts
index cfe9fdb34..458ee3201 100644
--- a/specification/migration/deprecations/DeprecationInfoRequest.ts
+++ b/specification/migration/deprecations/DeprecationInfoRequest.ts
@@ -23,13 +23,17 @@ import { IndexName, MediaType } from '@_types/common'
 /**
  * Get deprecation information.
  *
- * Get information about different cluster, node, and index level settings that use deprecated features that will be removed or changed in the next major version.
- *
- * Deprecated features are fully supported and continue to work throughout the current major version. However, they will be removed or changed in the next major version.
+ * Returns information about deprecated features which are in use in the cluster.
+ * The reported features includes cluster, node, and index level settings that will be removed or changed in the next major version.
+ * You must address the reported issues before upgrading to the next major version.
+ * However, no action is required when upgrading within the current major version.
+ * Deprecated features remain fully supported and will continue to work in the current version, and when upgrading to a newer minor or patch release in the same major version.
  * Use this API to review your usage of these features and migrate away from them at your own pace, before upgrading to a new major version.
  *
  * > info
- * > This API is designed for indirect use by the Upgrade Assistant. Using the [Upgrade Assistant](https://www.elastic.co/docs/deploy-manage/upgrade/prepare-to-upgrade/upgrade-assistant) is strongly recommended.
+ * > This API is designed for indirect use by the [Upgrade Assistant](https://www.elastic.co/docs/deploy-manage/upgrade/prepare-to-upgrade/upgrade-assistant).
+ * > Prefer to access information about deprecated features using the Upgrade Assistant rather than calling this API directly.
+ *
  * @rest_spec_name migration.deprecations
  * @availability stack since=6.1.0 stability=stable
  * @cluster_privileges manage
  *
```

leemthompo:
@DaveCTurner nice 👌  —I'd ship that with some tiny tweaks. Sorry I really shouldn't even be involved in this, but I thought it would be a quick fix 😄 `#friday-sidequest`

Tweaks for consideration

## 1

>  * However, no action is required when upgrading within the current major version.

possibly superfluous considering the following sentence?

## 2


> + * Deprecated features remain fully supported and will continue to work in the current version, and when upgrading to a newer minor or patch release in the same major version.

could possibly be trimmed a bit?

Deprecated features remain fully supported within the current major version, including all minor and patch releases.

## 3

Final sentence alternative candidate

> * > Prefer to access information about deprecated features using the Upgrade Assistant rather than calling this API directly.


We recommend learning about deprecated features using the Upgrade Assistant instead of calling this API directly.

DaveCTurner:
> possibly superfluous considering the following sentence

I think it's worth saying in both ways. No action is required, and the things are fully supported, and neither of these quite implies the other.

> Deprecated features remain fully supported within the current major version, including all minor and patch releases.

The customer in question was particularly concerned about _upgrading_ to a newer 9.x release while deprecation warnings existed, and IMO they are right to consider upgrading as a slightly riskier process than just keeping on with the same version. Again, I think it's worth spelling out that we don't just mean "it's ok to stay where you are in 9.x, whatever minor/patch release you're on".

> We recommend learning about deprecated features using the Upgrade Assistant instead of calling this API directly.

Yep that's good with me.

leemthompo:
sounds good explicit > implicit I guess 👍

PR #6086, opened 2026-03-03
[Inference API] Add fireworks ai docs

jonathan-buttner:
These docs correspond to the PRs in elasticsearch here:
Fireworks Chat Completions - https://github.com/elastic/elasticsearch/pull/142664
Fireworks Embeddings - https://github.com/elastic/elasticsearch/pull/137130


jonathan-buttner:
It looks like the `specification/_json_spec` files aren't needed any more 🤔 or at least I can't find them anymore. I added one over in Elasticsearch in this PR though: https://github.com/elastic/elasticsearch/pull/143527

specification/inference/put_fireworksai/examples/response/PutFireworksAiResponseExample2.yaml
```diff
@@ -0,0 +1,23 @@
+summary: A text embedding task with dimensions
```
DonalEvans: Do we need a separate example just for specifying `dimensions`? We don't have other examples for setting `similarity` or `rate_limit`, for example. Maybe instead we could have one minimal example that only specifies the required fields, and one which explicitly sets all the fields?

specification/inference/put_fireworksai/examples/request/PutFireworksAiRequestExample1.yaml
```diff
@@ -0,0 +1,12 @@
+summary: A text embedding task
```
DonalEvans: These summary lines are potentially a little confusing, since they all show endpoint creation rather than actually performing an inference task. Maybe they'd be better as "Create an endpoint for text embedding tasks" or something along those lines.

specification/inference/put_fireworksai/examples/request/PutFireworksAiRequestExample1.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: Run `PUT _inference/text_embedding/my-fireworks-embeddings` to create an inference endpoint that performs a `text_embedding` task.
+method_request: 'PUT _inference/text_embedding/my-fireworks-embeddings'
+# type: "request"
```
DonalEvans: I think these commented-out lines can be removed. They seem to have been copy-pasted a lot so they show up in lots of files, but I don't think they actually do anything.

specification/inference/_types/CommonTypes.ts
```diff
@@ -2051,6 +2051,83 @@ export enum NvidiaInputType {
...
+   * IMPORTANT: You need to provide the API key only once, during the inference model creation.
+   * The get inference endpoint API does not retrieve your API key.
+   * @ext_doc_id fireworksai-api-keys
+   */
+  api_key: string
+  /**
+   * The name of the model to use for the inference task.
+   * Refer to the Fireworks AI documentation for the list of available models for chat completion, completion, and text embedding.
+   * For text embedding, supported models include the Qwen3 embedding family (e.g. `fireworks/qwen3-embedding-8b`) and other models in the Fireworks model library.
+   * @ext_doc_id fireworksai-embeddings
```
DonalEvans: The external link here is for embedding models specifically. Is there a link we could use that covers both completion and text embedding models? Maybe https://docs.fireworks.ai/guides/recommended-models?

specification/inference/put_fireworksai/examples/response/PutFireworksAiResponseExample2.yaml
jonathan-buttner: Actually I think I'll just remove this file. These are the responses so they'll pretty much all look the same for a single task type.

PR #6087, opened 2026-03-04
Update SearchRequest Documentation

joshua-adams-1:
Updates the `SearchRequest` slicing documentation to make it clearer now we have both scroll-based and point-in-time search.

Relates: https://github.com/elastic/elasticsearch-team/issues/2088

### DIFF
--- a/specification/_global/search/SearchRequest.ts
+++ b/specification/_global/search/SearchRequest.ts
@@ -66,13 +66,12 @@ import { SourceConfig, SourceConfigParam } from './_types/SourceFilter'
  *
  * When paging through a large number of documents, it can be helpful to split the search into multiple slices to consume them independently with the `slice` and `pit` properties.
  * By default the splitting is done first on the shards, then locally on each shard.
- * The local splitting partitions the shard into contiguous ranges based on Lucene document IDs.
  *
  * For instance if the number of shards is equal to 2 and you request 4 slices, the slices 0 and 2 are assigned to the first shard and the slices 1 and 3 are assigned to the second shard.
  *
  * IMPORTANT: The same point-in-time ID should be used for all slices.
  * If different PIT IDs are used, slices can overlap and miss documents.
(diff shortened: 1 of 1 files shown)


PeteGillinElastic:
LGTM, but let's see if someone from Search Foundations wants to review this.

specification/_global/search/SearchRequest.ts
```diff
@@ -66,13 +66,12 @@ import { SourceConfig, SourceConfigParam } from './_types/SourceFilter'
  *
  * When paging through a large number of documents, it can be helpful to split the search into multiple slices to consume them independently with the `slice` and `pit` properties.
  * By default the splitting is done first on the shards, then locally on each shard.
- * The local splitting partitions the shard into contiguous ranges based on Lucene document IDs.
```
PeteGillinElastic, line 69: For the benefit of other reviewers, since @joshua-adams-1 and I discussed this elsewhere: The idea with removing this is that (a) it's misleading stating it here without qualification, since this is written in a context where we're talking about both scroll and PIT, but it's only true for PIT, and (b) we say the same thing in a more helpful context below.

specification/_global/search/SearchRequest.ts
```diff
@@ -66,13 +66,12 @@ import { SourceConfig, SourceConfigParam } from './_types/SourceFilter'
...
  *
  * For instance if the number of shards is equal to 2 and you request 4 slices, the slices 0 and 2 are assigned to the first shard and the slices 1 and 3 are assigned to the second shard.
  *
  * IMPORTANT: The same point-in-time ID should be used for all slices.
  * If different PIT IDs are used, slices can overlap and miss documents.
- * This situation can occur because the splitting criterion is based on Lucene document IDs, which are not stable across changes to the index.
```
PeteGillinElastic, line 75: For the benefit of other reviewers: This is just a nit-pick since the user can specify a different slicing field.

spinscale (approved):

PR #6090, opened 2026-03-06
Add missing num_failure_retries setting to Transform API specification

### LINKED ISSUE #5801: Missing num_failure_retries setting in Transform API specification
## 🐛 Missing type

**API name:** Transform API (operation-transform-put-transform)

### Issue

The API for Elasticsearch transform supports a setting called `num_failure_retries`, which allows the transform to retry indefinitely for recoverable failures when set to `-1`. However, this option is missing from both the official Transform API documentation ([reference](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-transform-put-transform)) and the elasticsearch-specification.

Calling the API with this option works, but it is not in the API spec.

### Definition

> `num_failure_retries`:
> (Optional, integer) Defines the number of retries for a recoverable failure before the transform task is marked as `failed`. The minimum value is `-1`, which indicates infinite retries, and the maximum value is `100`. When set to `-1`, the transform never stops retrying a recoverable failure. If unset, the transform uses the cluster-level setting `num_transform_failure_retries`.
>
> This setting cannot be specified when `unattended` is `true`, because unattended transforms always retry indefinitely.

References:

- https://github.com/elastic/elasticsearch/blob/760ffd44b50664221d0618f10d534d1f10a531bc/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/transform/transforms/SettingsConfig.java
- https://github.com/elastic/elasticsearch/pull/87361
- https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_transform#num_failure_retries-3
- [Transform API reference](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-transform-put-transform)
- [Package spec validation](https://github.com/elastic/package-spec/blob/main/spec/integration/elasticsearch/transform/transform.spec.yml#L151-L190) (this incorporates the elasticsearch-specification of transforms and will be updated after this issue is addressed)

andrewkroh:
Adds the `num_failure_retries` optional integer field to the transform `Settings` type. This setting defines the number of retries on a recoverable failure before the transform task is marked as failed, with -1 indicating infinite retries. Available since 8.4.0.

Closes #5801


l-trotta:
hey @andrewkroh, thanks for the fix! If you want to quickly solve the conflict and make this much easier to backport could you please push the changes without the output? It will be regenerated automatically once pushed to the branches.

andrewkroh:
Done. It now only contains the Transform.ts change.

PR #6173, opened 2026-04-01
Add Cursor agent skill for pre-commit contrib workflow

Jan-Kazlouski-elastic:
Adds a project-local Cursor skill that mirrors [CONTRIBUTING.md](CONTRIBUTING.md): run `make contrib` before opening a PR when changes affect the specification or committed generated output. Documents how `nvm install`, `make setup`, and `make contrib` relate, consistent with [README.md](README.md) and the [Makefile](Makefile) `contrib` target.

No changes to the specification or generated artifacts.

### DIFF
--- a/.cursor/skills/pre-commit-contrib/SKILL.md
+++ b/.cursor/skills/pre-commit-contrib/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: pre-commit-contrib
-description: Runs the Elasticsearch specification pre-commit pipeline (Node check, dependencies, compile, generate, license, format, OpenAPI, doc lint). Use before committing specification changes, when validating the full schema output, or when the user asks for contrib/setup validation.
+description: Runs the Elasticsearch specification contrib pre-commit pipeline (lint, generate, license, format, OpenAPI, doc lint). Use before committing specification changes, when validating the full schema output, or when the user asks for contrib/setup validation.
 ---
 
 # Pre-commit validation (elasticsearch-specification)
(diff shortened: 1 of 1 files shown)


MattDevy:
Hey @Jan-Kazlouski-elastic, thanks for putting this together! The skill content itself is solid, and I like the idea of giving AI agents guidance on running `make contrib` before committing.

That said, I think the approach could go further. Right now this only benefits Cursor users, but the underlying problem (contributors forgetting to run `make contrib`) applies to everyone: other AI tools like Claude Code, Copilot, Codex, and human developers too.

A couple of ideas worth considering:

- **`AGENTS.md` at repo root** (with symlinks for `CLAUDE.md`, `.cursorrules`, etc.): This is becoming a standard way to give all AI coding tools the same project guidance in one place. One file, many consumers.
- **[pre-commit](https://pre-commit.com/) hooks**: A `.pre-commit-config.yaml` that wraps the `make contrib` steps would enforce the checks deterministically at the git layer, for anyone (or anything) that calls `git commit`. No reliance on the agent reading instructions first.

Together, `AGENTS.md` handles the guidance and `pre-commit` handles the enforcement, so you get coverage across the whole contributor base rather than just one tool.

Happy to discuss if you want to iterate on this!

Jan-Kazlouski-elastic:
Hello @MattDevy 
Thank you for showing interest in my humble attempts to introduce common skills in `elasticsearch-specification` codebase. This is a raw draft and I agree this can and must go further than that. This initiative is currently moved to the side by some of my other tasks. When I get back to it - I'll be sure to include your suggestions.
Thanks again!

PR #6177, opened 2026-04-02
docs: update GET / to explain Serverless version reporting

leemthompo:
## Contributes to: https://github.com/elastic/docs-content-internal/issues/934

The root API now returns the next upcoming minor release in Serverless, and is no longer pinned to 8.11. Updated the API description and response example to explain this and direct users to use build_flavor: serverless instead of gating on version.

## Preview of API docs 

Screenshot 2026-04-07 at 10 34 25

### DIFF
--- a/docs/examples/languageExamples.json
+++ b/docs/examples/languageExamples.json
@@ -18672,23 +18672,23 @@
   "specification/inference/embedding/examples/request/EmbeddingRequestExample1.yaml": [
     {
       "language": "Python",
-      "code": "resp = client.inference.embedding(\n    inference_id=\"my-multimodal-endpoint\",\n    embedding={\n        \"input\": [\n            {\n                \"content\": {\n                    \"type\": \"image\",\n                    \"format\": \"base64\",\n                    \"value\": \"data:image/jpg;base64,...\"\n                }\n            },\n            {\n                \"content\": {\n                    \"type\": \"text\",\n                    \"value\": \"Some text to create an embedding\"\n                }\n            }\n        ]\n    },\n)"
+      "code": "resp = client.inference.embedding(\n    inference_id=\"my-multimodal-endpoint\",\n    embedding={\n        \"input\": [\n            {\n                \"content\": {\n                    \"type\": \"image\",\n                    \"format\": \"base64\",\n                    \"value\": \"data:image/jpeg;base64,...\"\n                }\n            },\n            {\n                \"content\": {\n                    \"type\": \"text\",\n                    \"value\": \"Some text to create an embedding\"\n                }\n            }\n        ]\n    },\n)"
     },
     {
       "language": "JavaScript",
-      "code": "const response = await client.inference.embedding({\n  inference_id: \"my-multimodal-endpoint\",\n  embedding: {\n    input: [\n      {\n        content: {\n          type: \"image\",\n          format: \"base64\",\n          value: \"data:image/jpg;base64,...\",\n        },\n      },\n      {\n        content: {\n          type: \"text\",\n          value: \"Some text to create an embedding\",\n        },\n      },\n    ],\n  },\n});"
+      "code": "const response = await client.inference.embedding({\n  inference_id: \"my-multimodal-endpoint\",\n  embedding: {\n    input: [\n      {\n        content: {\n          type: \"image\",\n          format: \"base64\",\n          value: \"data:image/jpeg;base64,...\",\n        },\n      },\n      {\n        content: {\n          type: \"text\",\n          value: \"Some text to create an embedding\",\n        },\n      },\n    ],\n  },\n});"
     },
     {
       "language": "Ruby",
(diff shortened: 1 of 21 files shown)


pete-naylor:
Thanks @leemthompo. I wanted to suggest this alternative to the wording:

Serverless is always up to date with the latest in Elasticsearch development and does not track to a traditional release versioning model. To represent this while maintaining compatibility with existing clients, the version.number reported by Serverless is the next target Elasticsearch release version. Clients calling this API to assess feature availability and API coverage should detect a Serverless Elasticsearch endpoint by looking for a version.build_flavor of “serverless” (in this case, the version.number value is not meaningful).

leemthompo:
thanks @pete-naylor! I'll do an update next week when back to keyboard, sorry I meant to tag you for review on the PR initially too :)

kunisen:
Thank you @leemthompo for your work!
From support perspective, it looks great about the change on API docs. 

However, may we please have two more suggestions please?

**[1]**
Could we please also put this into Serverless normal docs like https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings please? 


**[2]**

Could we please also have a section describing "version", and also mention that serverless is always "versionless", means version will be auto-upgraded to next minor ahead of stateful stack version.

**Side note**

Doc specific talk we might want to have a new environment variable, if not available yet, to represent the serverless latest version, which is always 1 minor ahead of stack. For now, 9.4 (stateful stack latest: 9.3). Next, 9.5, and so on...

---

Thanks again!

shainaraskas:
>Doc specific talk we might want to have a new environment variable, if not available yet, to represent the serverless latest version, which is always 1 minor ahead of stack. For now, 9.4 (stateful stack latest: 9.3). Next, 9.5, and so on...

Ultimately, serverless is still versionless, so I don't think we want to do this. I can't see a situation where we would want to use it. The only reason why we are using a version number here is because this field only accepts semver, and can't be blank.

kunisen:
> Ultimately, serverless is still versionless, so I don't think we want to do this. I can't see a situation where we would want to use it. The only reason why we are using a version number here is because this field only accepts semver, and can't be blank.

Thank you @shainaraskas for the comment. 

I think we are on the same page, and I 100% agree `serverless is still versionless`. 

It's just the fact that the version existence itself causes confusion:
* Previously it was 8.11 - we got lots of inquiries saying why serverless uses such an old version
* Now it's marked as 9.4 - we get (internal) feedbacks saying why serverless uses a version that is not released (*) yet

(*) Where `released` actually means the normal (stateful) Elastic Stack version, but not for serverless. 

---

So you can see that, we must clarify this in our doc, otherwise, we will have the same problem - it just changes the context from why version bing too old to version being too new, and will not likely reduce the incoming questions from customers.

---

Based on that, our opinion from support perspective would be:
* A) The key is to make the concept "_serverless is versionless_" clear. We don't have to have an env var to represent version, also we don't have to represent the version in our API/any other public doc, if that's against doc team's direction.
* B) However, since we show 9.4 in `GET /` response (and later it will increase to 9.5, 9.6...), which is not clear enough "it's versionless". We may need to clarify the connection between this **semver version value 9.4** and **the concept of serverless is versionless**.

IMHO, the best is to take off the version number from `GET /` and mention it's versionless, however, I understand the tech difficulties. Therefore, as a workaround, I proposed that we make the doc clear. Thanks for taking that into consideration!

---

Last note, if @shainaraskas you disagree with using env var to represent this, it's totally fine. As long as we can achieve this by adding clarities into our APIs, and public docs, that would be enough. IIUC, this version number will increase and will not stay at 9.4 forever - @rjernst could correct me if this is wrong - so we may need to have something to represent the version for serverless.

---

Hope the above is clear and thank you again!

leemthompo:
@kunisen 

> However, may we please have two more suggestions please?

> 
> [1]
> Could we please also put this into Serverless normal docs like https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings please?
> 
> [2]
> 
> Could we please also have a section describing "version", and also mention that serverless is always "versionless", means version will be auto-upgraded to next minor ahead of stateful stack version.
> 
> 

This is already tracked in the parent issue: https://github.com/elastic/docs-content-internal/issues/934

This is the PR for the API docs :)

> Doc specific talk we might want to have a new environment variable, if not available yet, to represent the serverless latest version, which is always 1 minor ahead of stack. For now, 9.4 (stateful stack latest: 9.3). Next, 9.5, and so on...


I think the question is moot for the API docs because we don't have env var functionality in the API docs as they're hosted by a third-party

But I don't think it's a bad idea for the normal docs

kunisen:
Thank you @leemthompo a lot for the detailed and helpful response! 🙏

kunisen (approved):
It looks great from support perspective by adding the description of "versionless", and the relation between 9.4 and versionless. Thanks!

leemthompo:
> Ultimately, serverless is still versionless, so I don't think we want to do this. I can't see a situation where we would want to use it. The only reason why we are using a version number here is because this field only accepts semver, and can't be blank.

I think because the API will always report that future version, then it would at least make things less confusing if the docs stayed up-to-date with that (in narrative docs where we mention the GET / API of course, I wouldn't mention it outside of a specific note about the version returned by the API). It is confusion mitigation though, for sure, not elimination :)

leemthompo:
@pete-naylor https://github.com/elastic/elasticsearch-specification/pull/6177/commits/12beb770f9a319a6718a5284a7aef4e5e951eb5b tries to incorporate your feedback 🙏

pete-naylor:
LGTM @leemthompo - thanks!

shainaraskas (approved):

PR #6179, opened 2026-04-03
[Transform] Defer on Schedule Now

prwhelan:
Schedule Now has a defer parameter, defaulting to false, that will schedule transform after the configured delay to allow users to ingest new documents then immediately trigger the transform.

Relate https://github.com/elastic/elasticsearch/pull/145693


edsavage (approved):
LGTM

PR #6183, opened 2026-04-08
Update reindex-from-remote spec for serverless

PeteGillinElastic:
We are changing the whitelisting for reindex-from-remote in serverless to allow serverless hosts (in addition to ECH, which is all we allow now). This updates the spec to reflect that.

ES-14414


joshua-adams-1 (approved):
LGTM!

PR #6206, opened 2026-04-14
Add follow_relocations request param to GET /_tasks/{task_id}

samxbr:
Adds an internal only param `follow_relocations` to `GET /_tasks/{task_id}`.

Relates to https://github.com/elastic/elasticsearch/pull/146161
Relates to https://github.com/elastic/elasticsearch-team/issues/2588

### DIFF
--- a/specification/tasks/get/GetTaskRequest.ts
+++ b/specification/tasks/get/GetTaskRequest.ts
@@ -64,6 +64,7 @@ export interface Request extends RequestBase {
     wait_for_completion?: boolean
     /**
      * Internal use only
+     * @availability stack since=9.5.0 stability=experimental visibility=private
      * @server_default true
      */
     follow_relocations?: boolean
--- a/specification/tasks/get/GetTaskRequest.ts
+++ b/specification/tasks/get/GetTaskRequest.ts
@@ -62,5 +62,10 @@ export interface Request extends RequestBase {
      * @server_default false
      */
     wait_for_completion?: boolean
+    /**
+     * Internal use only
+     * @server_default true
+     */
+    follow_relocations?: boolean
   }
 }


specification/tasks/get/GetTaskRequest.ts
```diff
@@ -62,5 +62,10 @@ export interface Request extends RequestBase {
      * @server_default false
      */
     wait_for_completion?: boolean
+    /**
+     * Internal use only
```
szybia, line 66: two suggestions
1. should we give ourselves a more explicit get-out-of-jail card and say that this might change?
2. should we still document what this does, for people who do use it (internal or not)

```suggestion
     * Whether to follow relocations when getting the task, set to false to see what the original task result was.
     * Internal use only, for debugging, do not use in production as the parameter might change or be removed.
```

szybia (approved):

specification/tasks/get/GetTaskRequest.ts
samxbr: I am hesitant to mention relocations here just yet. We can always come back and update this if we decide to make this param more usable to users.

PeteGillinElastic: Agreed. We can revisit this when we come to do the docs we do when we launch. It doesn't make sense for this to be the first place that we mention the concept of relocation.

l-trotta: should we add some availability info? I'd document this as:
`@availability stack since=9.5.0 stability=experimental visibility=private`

PeteGillinElastic: That sounds like a good idea, thanks for point it out Laura!

PR #6214, opened 2026-04-15
Add note about repo service disruptions

DaveCTurner:
Adds a note to spell out that repo analysis will fail if the repo
service is disrupted, and thus such failures can be ignored as long as
they can be ascribed to disruptions and occur only rarely.


DaveCTurner:
Ugh forgot about the lack of support for forks here - closing in favour of https://github.com/elastic/elasticsearch-specification/pull/6215

PR #6215, opened 2026-04-17
Add note about repo service disruptions

DaveCTurner:
Adds a note to spell out that repo analysis will fail if the repo
service is disrupted, and thus such failures can be ignored as long as
they can be ascribed to disruptions and occur only rarely.


ywangd (approved):
LGTM

PR #6235, opened 2026-04-24
Mark downsampling API as stable

gmarouli:
The downsampling API has been GA since the time series data streams were also GA in `8.7.0`, somehow we missed this in the documentation, this should fix it.


gmarouli:
Do we backport documentation changes? ~I added the label `skip-backport`, but I will change it if you think we need to backport it.~

I decided to add it so the clients reflect that this is GA.

l-trotta (approved):

l-trotta:
@gmarouli thanks! just an fyi, documentation will be updated from the backports, but for clients there won't ever be any other 8.x releases except for 8.19

gmarouli:
Thank you @l-trotta ! That is more than enough :)

PR #6263, opened 2026-05-12
Remapping SourceOnlyRepositorySettings

### LINKED ISSUE #6260: SourceOnlyRepositorySettings is incomplete
In [SnapshotRepository](https://github.com/elastic/elasticsearch-specification/blob/62092179e10298e6ff3b6dbfdbe68af8990119a3/specification/snapshot/_types/SnapshotRepository.ts), SourceOnlyRepository is defined as a wrapper over the other types of repositories (from the [source code](https://github.com/elastic/elasticsearch/blob/dd8dc59ced8532dac258bbc8abcbce1debe1134b/x-pack/plugin/core/src/main/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotRepository.java#L79): "Snapshots restored from source only snapshots are minimal indices that are read-only and only allow match_all scroll searches in order to reindex the data.").

Users need to choose the actual underlying repository type in `SourceOnlyRepositorySettings` by setting `delegate_type`. First thing here: `delegate_type` should be an Enum, since it can only be a defined set of values. 
After choosing the type, the type's setting become available for `SourceOnlyRepository`, for example (from the [docs](https://www.elastic.co/docs/reference/elasticsearch/configuration-reference/source-repository-settings)):

```	
PUT _snapshot/my_src_only_repository					
{
  "type": "source",
  "settings": {
    "delegate_type": "fs",
    "location": "my_backup_repository"
  }
}
```

The delegate type is `fs`, meaning the mandatory `location` field must be set. This behavior is not defined in the specification, in practice `SourceOnlyRepositorySettings` should have every other setting's field available, as non mandatory. Its definition should only have `delegate_type`, every other setting is inherited, right now the two other field it has (`max_number_of_snapshots` and `read_only`) are actually from other settings (also `read_only` is wrong, it should be `readonly`). 

I'm not sure how to fix this in a way that's easily understandable in docs. In theory the most fitting solution would be having some setting field that is a union of all settings, but the flat json structure of the request doesn't allow this. 

In summary:

- `SourceOnlyRepositorySettings` is missing a way to access all settings from all other repo types
- `SourceOnlyRepositorySettings` only defines one field, `delegate_type`, all others are inherited (`max_number_of_snapshots` and `read_only` should be removed)
- `delegate_type` should be of type Enum, with a limited set of options being the other repo types names

l-trotta:
Thanks @flobernd for suggesting the structure, let me know if this works for the other clients too! @MattDevy 

Closes https://github.com/elastic/elasticsearch-specification/issues/6260

### DIFF
--- a/specification/snapshot/_types/SnapshotRepository.ts
+++ b/specification/snapshot/_types/SnapshotRepository.ts
@@ -413,6 +413,9 @@ export class ReadOnlyUrlRepositorySettings extends RepositorySettingsBase {
 }
 
 /**
+ * The delegated repository type.
+ * Source repositories can use `settings` properties for its delegated repository type.
+ *
  * @variants internal tag='delegate_type'
  */
 export type SourceOnlyRepositorySettings =
--- a/specification/snapshot/_types/SnapshotRepository.ts
+++ b/specification/snapshot/_types/SnapshotRepository.ts
@@ -20,6 +20,7 @@
 import { ByteSize, Uuid } from '@_types/common'
 import { integer } from '@_types/Numeric'
 import { Duration } from '@_types/Time'
+import { OverloadOf } from '@spec_utils/behaviors'
 
 /**
  * @variants internal tag='type'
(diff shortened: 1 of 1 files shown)


flobernd (approved):

MattDevy (approved):
LGTM, verified it works in the Go generator, it's a breaking change - but it didn't work before so worth it :)

PR #6268, opened 2026-05-12
Fix esql validation

l-trotta:
Breaking change, but acceptable IMO. Server PR: https://github.com/elastic/elasticsearch/pull/145816

### DIFF
--- a/specification/esql/delete_view/DeleteViewRequest.ts
+++ b/specification/esql/delete_view/DeleteViewRequest.ts
@@ -18,7 +18,7 @@
  */
 
 import { RequestBase } from '@_types/Base'
-import { Id, MediaType } from '@_types/common'
+import { Ids, MediaType } from '@_types/common'
 
 /**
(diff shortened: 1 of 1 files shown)


specification/esql/delete_view/DeleteViewRequest.ts
```diff
@@ -40,7 +40,7 @@ export interface Request extends RequestBase {
   ]
   path_parts: {
     /** The view name to remove. */
-    name: Id
+    name: Ids
```
flobernd, line 43: I wonder if this should be `Id` in the first place. We also have `Name`/`Names` - but all of this is quite arbitrary.

flobernd (approved):
LGTM, but left a question.

specification/esql/delete_view/DeleteViewRequest.ts
l-trotta: used Ids to follow the other endpoints (get and create)

PR #6273, opened 2026-05-15
docs: clarify split index shard requirements

julius-ban:
The previous description only mentioned that the target shard count must be a multiple of the source shard count. However, the actual validation also requires that index.number_of_routing_shards of the source index is divisible by the target shard count with no remainder. Without this, users can encounter an IllegalStateException even when the multiple condition is satisfied.

Relates to: elastic/elasticsearch#75137

### DIFF
--- a/specification/indices/split/IndicesSplitRequest.ts
+++ b/specification/indices/split/IndicesSplitRequest.ts
@@ -57,7 +57,8 @@ import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
  *
  * * The target index must not exist.
  * * The source index must have fewer primary shards than the target index.
- * * The number of primary shards in the target index must be a multiple of the number of primary shards in the source index and must also be a divisor of the source index's index.number_of_routing_shards.
+ * * The number of primary shards in the target index must be a multiple of the number of primary shards in the source index.
+ * *  The number of primary shards in the target index must be a divisor of the source index's `index.number_of_routing_shards`.
  * * The node handling the split process must have sufficient free disk space to accommodate a second copy of the existing index.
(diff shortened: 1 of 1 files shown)


l-trotta:
hey @julius-ban, thanks for contributing! the file you modified is one of the output result of the spec compilation (it's found in the /output folder). the correct file to modify would be just [IndicesSplitRequest.ts ](https://github.com/elastic/elasticsearch-specification/blob/8260d402ec3ef0dfdf86a88736c190016f1adbad/specification/indices/split/IndicesSplitRequest.ts#L30), could you make the changes to the PR?

julius-ban:
Done! I've reverted the changes to the output file and kept only the 
modification to IndicesSplitRequest.ts. Thanks for the guidance!

specification/indices/split/IndicesSplitRequest.ts
```diff
@@ -57,7 +57,7 @@ import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
  *
  * * The target index must not exist.
  * * The source index must have fewer primary shards than the target index.
- * * The number of primary shards in the target index must be a multiple of the number of primary shards in the source index.
+ * * The number of primary shards in the target index must be a multiple of the number of primary shards in the source index and must also be a divisor of the source index's index.number_of_routing_shards.
```
l-trotta: just a little formatting and we're good to go :) 
```suggestion
 * * The number of primary shards in the target index must be a multiple of the number of primary shards in the source index.
 * *  The number of primary shards in the target index must be a divisor of the source index's `index.number_of_routing_shards`.
```

l-trotta (approved):

PR #6278, opened 2026-05-20
ci: analyze Kibana type check failures with Claude via GitHub Actions

margaretjgu:
## Summary

- Adds a daily GitHub Actions workflow that checks if the Buildkite [Kibana type checks](https://buildkite.com/elastic/kibana-type-checks) pipeline failed
- On failure, fetches the build log via Buildkite API, extracts tsc errors, and calls Claude (via Elastic's LiteLLM gateway) to classify each error as GENERATOR / SPEC / KIBANA / UNKNOWN
- Opens a GitHub issue with the analysis so failures are triaged automatically instead of requiring manual review

Requires two GitHub Actions secrets to be set on this repo:
- `LITELLM_API_KEY` - virtual key from elastic.litellm-prod.ai
- `BUILDKITE_API_TOKEN` - personal Buildkite API token with read access to pipelines, builds, and build logs

## Test plan

- [ ] Add `LITELLM_API_KEY` and `BUILDKITE_API_TOKEN` to repo Actions secrets
- [ ] Trigger workflow manually via workflow_dispatch to verify Buildkite API connectivity
- [ ] Confirm GitHub issue is created when the Buildkite build is in a failed state


.github/workflows/kibana-type-check-analysis.yml
JoshMock, line 1: Would it be possible to run this as part of the `.buildkite/kibana.yml` pipeline instead of a GitHub action? That way it's all guaranteed to run sequentially and ensures analysis results are tied to a specific run. It'll be slightly more friction to get the initial setup working, unfortunately, but more stable in the long run.

JoshMock (approved):
very cool. :clap:

PR #6335, opened 2026-05-29
Correct Jina inference service setting name

DonalEvans:
The name of the service setting used to specify the embedding element type for the Jina integration is "embedding_type", not "element_type". See [the service settings code](https://github.com/elastic/elasticsearch/blob/673de3bcb25426a0ac0cbc2cc118e1557864c2c0/x-pack/plugin/inference/src/main/java/org/elasticsearch/xpack/inference/services/jinaai/embeddings/BaseJinaAIEmbeddingsServiceSettings.java#L111). This corrects the error introduced in [this PR](https://github.com/elastic/elasticsearch-specification/pull/5949).

The only manual changes in this PR are in `CommonTypes.ts`, other changes were automatically generated as a result of running the docs build locally.

PR #6338, opened 2026-06-01
ci: use PAT for backport instead of broken Vault token

margaretjgu:
## Problem

The backport workflow has been **failing on every merged PR** since #6326 switched it to fetch an ephemeral GitHub token from Vault. The Vault role for this workflow was never provisioned, so the `fetch-github-token` step fails immediately:

```
##[error]failed to retrieve vault token. code: ERR_NON_2XX_3XX_RESPONSE,
Response code 400 (Bad Request),
vaultResponse: {"errors":["role \"token-policy-c34099963f9d\" could not be found"]}
```

Because the job errors out at the token step, the backport never runs and the bot never posts its "to backport manually..." comment — which is the comment the **AI Backport Resolver** listens for. So both the normal backport *and* the AI resolver are broken.

Example: #6335 (5 backport labels) produced 4 failed Backport runs and zero bot comments.

## Fix

Drop the Vault step and use the repo's existing `secrets.PAT` (already used by `generate.yml` to push). It's a real account token, so:
- backport works again with no Vault provisioning needed, and
- its failure comments still trigger the `issue_comment`-based AI Backport Resolver (the original goal of #6326).

## Test plan
- [ ] Merge, then re-apply a backport label on #6335 to verify the backport runs and (on conflict) posts the manual-backport comment that triggers the AI resolver.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -24,13 +24,9 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
-      id-token: write
     steps:
-      - name: Fetch ephemeral GitHub token
-        id: fetch-token
-        uses: elastic/ci-gh-actions/fetch-github-token@8a7604dfdd4e7fe21f969bfe9ff96e17635ea577 # v1.0.0
-        with:
-          vault-instance: "ci-prod"
+      # Use a real PAT (not GITHUB_TOKEN) so that the failure comment posted on
+      # conflicting backports triggers the issue_comment-based AI Backport Resolver.
       - uses: tibdex/backport@9565281eda0731b1d20c4025c43339fb0a23812e # v2.0.4
         with:
-          github_token: ${{ steps.fetch-token.outputs.token }}
+          github_token: ${{ secrets.PAT }}


flobernd:
@margaretjgu Did we already try to debug why the ephemeral token was not provisioned?

margaretjgu:
Closing in favor of the proper fix: elastic/catalog-info#3943 registers the missing Vault role (`token-policy-c34099963f9d`) for this repo's backport workflow, mirroring how the clients (`token-policy-elasticsearch-js-backport`, etc.) are provisioned.

That keeps `backport.yml` on the ephemeral `fetch-github-token` App token (consistent with the clients), and the failure comment is posted by `elastic-vault-github-plugin-prod[bot]`, which is already in the AI Backport Resolver's allow-list. Once #3943 merges and Backstage syncs, the backport — and the AI resolver — will work without a long-lived PAT.

Can reopen this if we need an immediate stopgap before the Backstage sync lands.

margaretjgu:
> @margaretjgu Did we already try to debug why the ephemeral token was not provisioned?

@flobernd yes missed your comment before closing - the policy was missing in backstage
https://github.com/elastic/catalog-info/pull/3943

PR #6357, opened 2026-06-05
Caveat the spec for `max_docs` with sliced reindex

PeteGillinElastic:
See https://github.com/elastic/elasticsearch/pull/150808 for the background.

### DIFF
--- a/specification/_global/reindex/ReindexRequest.ts
+++ b/specification/_global/reindex/ReindexRequest.ts
@@ -138,7 +138,7 @@ export interface Request extends RequestBase {
      * If it is a value less then or equal to `scroll_size`, a scroll will not be used to retrieve the results for the operation.
      *
      * If `conflicts` is set to `proceed`, the reindex operation could attempt to reindex more documents from the source than `max_docs` until it has successfully indexed `max_docs` documents into the target or it has gone through every document in the source query.
-     * 
+     *
      * If `slices` is set, the `max_docs` limit is split evenly across the slices.
      * If the number of documents in the source is equal to or slightly more than `max_docs`, this could result in slightly fewer than `max_docs` documents being reindexed, due to skew in the slicing.
      */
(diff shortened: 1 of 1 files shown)


specification/_global/reindex/ReindexRequest.ts
```diff
@@ -138,6 +138,9 @@ export interface Request extends RequestBase {
      * If it is a value less then or equal to `scroll_size`, a scroll will not be used to retrieve the results for the operation.
      *
      * If `conflicts` is set to `proceed`, the reindex operation could attempt to reindex more documents from the source than `max_docs` until it has successfully indexed `max_docs` documents into the target or it has gone through every document in the source query.
+     *
+     * If `slices` is set, the `max_docs` limit is split evenly across the slices.
+     * If the number of documents in the source is equal to or slightly more than `max_docs`, this could result in slightly fewer than `max_docs` documents being reindexed, due to skew in the slicing.
```
joshua-adams-1, line 143: Slightly is a weasel word - is there an exact number here, or formula, or is slightly the best we can do?

PeteGillinElastic: I mean, I could do some statistics to figure out the probability as a function of `max_docs`, `slices`, and the size of the source... But I don't really want to, and I definitely don't want to quote the formula here (it won't be pretty). I think most readers will realize that, if you're doing `max_docs=1000` and you have millions of docs, you're fine.

PeteGillinElastic: (I'm not even sure there is a solution in closed form...)

joshua-adams-1: Ahahah I didn't want to give you lots of work lol, I meant if it happened to be 1 or two more than `max_docs` then we be specific 👍

joshua-adams-1 (approved):

PR #6363, opened 2026-06-15
Add new stats field to count req, fix same in search req

l-trotta:
Introduced in https://github.com/elastic/elasticsearch/pull/140143, code is the same in SearchRequest, so needs to be updated there to support single value.

### DIFF
--- a/specification/_global/count/CountRequest.ts
+++ b/specification/_global/count/CountRequest.ts
@@ -144,6 +144,10 @@ export interface Request extends RequestBase {
      */
     routing?: Routing
     /**
+     * Specific `tag` of the request for logging and statistical purposes.
+     */
+    stats?: string[] | string
+    /**
      * The maximum number of documents to collect for each shard.
      * If a query reaches this limit, Elasticsearch terminates the query early.
      * Elasticsearch collects documents before sorting.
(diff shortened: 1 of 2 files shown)


Anaethelion (approved):
Verified against Elasticsearch server source pinned at [`b0394ba5`](https://github.com/elastic/elasticsearch/commit/b0394ba51fdfb23e4884fed8f1477a1fc9a4052e) (merge of ES #140143).

Both `stats` fields are query parameters, parsed identically — a single `request.param("stats")` string comma-split into a list:

| Change | Server code | Verdict |
|---|---|---|
| `CountRequest` add `stats?: string[] \| string` | [`RestCountAction.java` L95-97](https://github.com/elastic/elasticsearch/blob/b0394ba51fdfb23e4884fed8f1477a1fc9a4052e/server/src/main/java/org/elasticsearch/rest/action/search/RestCountAction.java#L95-L97) — `request.param("stats")` → `Strings.splitStringByCommaToArray` | ✅ new param added by #140143 |
| `SearchRequest` `string[]` → `string[] \| string` | [`RestSearchAction.java` L382-384](https://github.com/elastic/elasticsearch/blob/b0394ba51fdfb23e4884fed8f1477a1fc9a4052e/server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java#L382-L384) — identical parse | ✅ single scalar is valid |

The server splits a comma-separated string, so a single scalar is accepted at the wire — `string[] | string` is correct and makes both endpoints consistent. LGTM.

PR #6366, opened 2026-06-16
Add new esql datasource and dataset crud api

l-trotta:
Added in https://github.com/elastic/elasticsearch/pull/146600.

This is pretty straightforward, I just have a couple of doubts:

- Is there a requested format for the `resource` field? some examples we could add to the docs?
- `type` is a string (even in the server code) but it could be a non exhaustive enums if we had a list of the values that are currently accepted (similar to how we handle [Repository](https://github.com/elastic/elasticsearch-specification/blob/0af659ccc0171743ba0fe97cba210a94b3b3538f/specification/snapshot/_types/SnapshotRepository.ts#L29))

### DIFF
--- a/specification/esql/_types/types.ts
+++ b/specification/esql/_types/types.ts
@@ -17,8 +17,9 @@
  * under the License.
  */
 
-import { FieldValue } from '@_types/common'
-import { SingleKeyDictionary } from '@spec_utils/Dictionary'
+import { FieldValue, Name } from '@_types/common'
+import { Dictionary, SingleKeyDictionary } from '@spec_utils/Dictionary'
+import { UserDefinedValue } from '@spec_utils/UserDefinedValue'
 
 /**
  * @codegen_names value, named
@@ -37,4 +38,37 @@ export class ESQLView {
   name: string
   /** The ES|QL query */
   query: string
+}
+
+/**
+ * Represents a data source definition stored in cluster state. A data source holds
+ * connection settings (credentials, endpoints, auth) for an external data provider.
+ */
+export class ESQLDataSource {
+  /** The data source name. */
+  name: Name
+  /** The data source type. */
+  type: string
+  /** A free-text description. */
+  description?: string
+  /** Type-specific settings. */
+  settings?: Dictionary<string, UserDefinedValue>
+}
+
+/**
+ * Represents a dataset definition stored in cluster state. A dataset is a named reference
+ * to external data that participates in the index namespace alongside indices, aliases, and views.
+ * Datasets inherit credentials from their parent data source at query time.
+ */
+export class ESQLDataset {
+  /** The dataset name. */
+  name: Name
+  /** The name of the parent data source. */
+  data_source: Name
+  /** The data source-specific resource path */
+  resource: string
+  /** A free-text description. */
+  description?: string
+  /** Type-specific settings. */
+  settings?: Dictionary<string, UserDefinedValue>
 }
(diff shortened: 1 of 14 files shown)


swallez:
Some additional information to possibly bring more structure:

* Datasources (e.g. S3 bucket)
  * their configuration is defined in subclasses of [DataSourceConfiguration](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/datasources/spi/DataSourceConfiguration.java)

* Datasets (e.g. `/foo/bar/*.ndjson` in a datasource):
  * every format reader (e.g. Parquet, ndjson) has its specific settings. They're checked in every [`FormatReader.withConfigTrackingConsumedKeys`](https://github.com/elastic/elasticsearch/blob/b32e2c8e1efbcdbcdbb6d75a98f3cab694380453/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/datasources/spi/FormatReader.java#L186) implementation and commonly listed in `RECOGNIZED_KEYS` in implementation classes.
  * some settings may be added by decompression filters (e.g. gzip on top of ndjson) which are inferred from the datasource URL

leemthompo:
added myself as reviewer for the docs component once this gets eng review :)

leemthompo:
👋 I compared this with the locally verified examples and conceptual model in my docs drafts: elastic/elasticsearch#146945. A few details might help tighten the generated API documentation:

- `resource` is a URI rather than only a path. For S3, it can include glob patterns, for example `s3://bucket/logs/**/*.parquet`.
- Dataset `settings` are format- and parsing-specific, not type-specific. Accepted keys depend on the format reader; compression can be inferred from the resource URI.
- Could “parent data source” become “referenced data source”? A data source defines how to connect; a dataset identifies what to read.
- GET responses redact credentials as `::es_redacted::`. This should be stated on the response model.

The narrative docs include locally verified request and response examples for the [data source APIs](https://github.com/elastic/elasticsearch/blob/b45a4cd4e9defac19c6fe60dd2e423585d1f5903/docs/reference/query-languages/esql/esql-data-federation-sources.md#data-source-api) and [dataset APIs](https://github.com/elastic/elasticsearch/blob/b45a4cd4e9defac19c6fe60dd2e423585d1f5903/docs/reference/query-languages/esql/esql-data-federation-datasets.md#dataset-api), which could support the missing specification tests.

l-trotta:
thank you @swallez and @leemthompo, I've updated the comments!

specification/esql/_types/types.ts
```diff
@@ -38,3 +39,42 @@ export class ESQLView {
...
+ */
+export class ESQLDataSource {
+  /** The data source name. */
+  name: Name
+  /** The data source type. */
+  type: string
+  /** A free-text description. */
+  description?: string
+  /** Type-specific settings. */
+  settings?: Dictionary<string, UserDefinedValue>
```
Anaethelion: settings is always emitted for a data source — [GetDataSourceAction](https://github.com/elastic/elasticsearch/blob/6dbb42a3d14b958818242107189bb21bfaf00ebb/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/datasources/datasource/GetDataSourceAction.java#L89-L104) and [DataSource.toXContent](https://github.com/elastic/elasticsearch/blob/6dbb42a3d14b958818242107189bb21bfaf00ebb/server/src/main/java/org/elasticsearch/cluster/metadata/DataSource.java#L146-L160) both write it unconditionally. Should it be required?

Anaethelion (changes requested):

Anaethelion (approved):

PR #6371, opened 2026-06-18
Add delete backing index action to modify data stream API

jbaiera:
Adds specification details for the new "delete_backing_index" action in the modify data stream API.

See https://github.com/elastic/elasticsearch/pull/151137

### DIFF
--- a/specification/indices/modify_data_stream/types.ts
+++ b/specification/indices/modify_data_stream/types.ts
@@ -34,6 +34,12 @@ export class Action {
    * A data stream’s write index cannot be removed.
    */
   remove_backing_index?: IndexAndDataStreamAction
+  /**
+   * Deletes a backing index from a data stream.
+   * The deletion is atomically executed along with the other actions in this operation.
+   * A data stream’s write index cannot be deleted.
+   */
+  delete_backing_index?: IndexAndDataStreamAction
 }
 
 export class IndexAndDataStreamAction {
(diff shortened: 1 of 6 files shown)


specification/indices/modify_data_stream/types.ts
```diff
@@ -34,6 +34,12 @@ export class Action {
    * A data stream’s write index cannot be removed.
    */
   remove_backing_index?: IndexAndDataStreamAction
+  /**
+   * Deletes a backing index from a data stream.
```
leemthompo, line 39:
```suggestion
   * @availability stack since=9.5.0 stability=stable
   * Deletes a backing index from a data stream.
```

@pquentin I haven't worked in the spec in a while 😄 , but is this the correct way to mark this off as 9.5+?

pquentin: Yes, this seems correct! https://github.com/elastic/elasticsearch-specification/blob/98f4204c4b2792f72d530902637111b96fb3bafe/docs/modeling-guide.md#availability. I'm not sure if this will mark it as available on serverless or not, though.

l-trotta (approved):

specification/indices/modify_data_stream/types.ts
leemthompo: oh yeah good point, @jbaiera should add  `* @availability serverless` here too for completeness :)

l-trotta:
@jbaiera  ?

PR #6393, opened 2026-07-02
Clarify the three kinds of NDJSON lines in the bulk request body docs

leemthompo:
## Reviewer

- Can you please add the relevant backport tags, I'm not sure how far back we want to go with this one if it's valid
- Current docs (check the "One of" tabs): https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-bulk#operation-bulk-body-application-json

## Summary

The generated `_bulk` API docs rendered the request body union as `OperationContainer | UpdateAction | object-3`, with no explanation of which NDJSON line format goes with which action. Given how heavily `_bulk` is used, this page deserved better.

Full writeup with a pipeline diagram: https://gist.github.com/leemthompo/953912b4119b5167ffb02803c1618bb6

## For the reviewer

Two commits, matching where each problem lives:

1. **Spec source** (`specification/_global/bulk/`): adds class-level JSDoc to `OperationContainer` and `UpdateAction` describing their role as action and update source lines, plus regenerated output. Purely additive descriptions, no shape or type changes.
2. **Docs overlays** (`docs/overlays/elasticsearch-shared-overlays.yaml`): titles the three union members **Action and metadata** / **Update source** / **Document source** (the last one previously rendered as `object-3` because a bare generic `TDocument` becomes an anonymous object in OpenAPI), and adds a request body description explaining the three kinds of NDJSON lines.

Notes:

- The body description lives in the overlay because the spec compiler drops JSDoc descriptions on value-kind bodies (unions/arrays). The JSDoc is kept in `BulkRequest.ts` so the source stays self-documenting; the overlay becomes redundant if that compiler gap is ever fixed.
- `OperationContainer` and `UpdateAction` are referenced only by the bulk request body, so the docs-only retitling can't affect other endpoints. Client-facing OpenAPI files and codegen are untouched by the overlays.
- Verified in the docs-final JSON for both stack and serverless, and `make lint-docs` passes with only pre-existing unrelated warnings.

### Before
before

### After

after

🤖 Generated with [Claude Code](https://claude.com/claude-code)


georgewallace:
We will want to backpoet this though to the proper versions so we can ensure when we publish v8 as well that it has it.

l-trotta (approved):

PR #6394, opened 2026-07-02
Add spec for encryption reset API

jfreden:
Adds the specification for `POST /_encryption/_reset`, which was previously entirely undocumented in this repo.

The endpoint destroys the current project encryption key (PEK) and generates a new one — the recovery path for when the on-disk encrypted PEK becomes permanently inaccessible (disk failure, lost HSM access, etc.). It requires the `accept_data_loss` query parameter to be explicitly set to `true`, since all data encrypted under the destroyed key becomes permanently unrecoverable.

This spec is prompted by elastic/elasticsearch#152181, which changes how the ES|QL datasource `EncryptedDataHandler` behaves during this reset (preserving non-secret datasource config instead of dropping the whole entry), but the REST endpoint itself already existed and had no spec.

### Modeling notes

- New `encryption` namespace/`encryption.reset` rest_spec_name, since `/_encryption` didn't fit any existing namespace.
- `accept_data_loss` is modeled as a required (non-optional) boolean query parameter, matching `EncryptionResetRequest#validate()` in Elasticsearch, which rejects the request unless it's `true`.
- Marked `@availability stack since=9.5.0 stability=experimental` and `serverless ... visibility=private`: the action is restricted to operator users only (see `DefaultOperatorOnlyRegistry` in Elasticsearch), so it's modeled the same way as other operator-scoped endpoints (e.g. `cat.*`, `xpack.usage`) — usable on self-managed stack, hidden from serverless docs/clients.
- Added a placeholder `encryption-reset` entry to `_doc_ids/table.csv` (`#TODO` URL) since there's no published docs page yet, following the same pattern used for other recently-added APIs (e.g. `streams-logs-disable`).

### DIFF
--- a/specification/_doc_ids/table.csv
+++ b/specification/_doc_ids/table.csv
@@ -216,6 +216,7 @@ docvalue-fields,https://www.elastic.co/docs/reference/elasticsearch/rest-apis/re
 dot-expand-processor,https://www.elastic.co/docs/reference/enrich-processor/dot-expand-processor,[use-doc-url],,About the dot expander processor
 drop-processor,https://www.elastic.co/docs/reference/enrich-processor/drop-processor,[use-doc-url],,More about the drop processor
 eland-import,https://www.elastic.co/docs/explore-analyze/machine-learning/nlp/ml-nlp-import-model#ml-nlp-import-script,[use-doc-url],,About importing a trained model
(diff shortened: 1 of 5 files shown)


l-trotta (approved):
thank you!

PR #6401, opened 2026-07-06
CI: Add OpenAPI lint + S3 publish workflows

reakaleek:
## What
Adds a fork-safe PR lint check (`openapi-pr.yml`) and a publish-on-merge workflow (`openapi-publish.yml`) that build the docs-final OpenAPI specs and upload them to S3.

## Why
Rollout of RFC elastic/docs-eng-team#634 — replaces the hourly pull-based mirror (`open-api-mirror.yml` in docs-internal-workflows) with the producer repo pushing its own spec. Closes elastic/docs-eng-team#642.

## How
- Both workflows run the same build: `make setup` → `make generate` → `make transform-to-openapi-for-docs` → `make overlay-docs`, producing the two `*-docs-final.json` files.
- `openapi-pr.yml`: `pull_request`, paths-filtered to `specification/**`, `docs/overlays/**`, `docs/linters/**`, `Makefile`. No secrets — lints both specs with `openapi/lint`.
- `openapi-publish.yml`: `push` to `main` and `[0-9]+.[0-9]+`. Lints, then uploads via `openapi/upload` (OIDC): stateful spec on every publishable branch, serverless spec on `main` only (always-latest).
- Existing `generate.yml` / `publish-release-artifacts.yml` are untouched.

**Blocked / not yet mergeable:** actual S3 upload needs the bucket/OIDC roles from elastic/docs-infra (tracked in elastic/docs-eng-team#641). Draft until that lands.

## Test plan
- [x] docs-actions `openapi/lint` + `openapi/upload` released under `v1`; SHA pins swapped to `@v1`
- [ ] docs-infra bucket + OIDC roles provisioned (elastic/docs-eng-team#641)
- [x] `openapi-pr` run is green on this PR
- [x] Lint verified to catch real errors: it flagged a malformed doc URL (double `#`), fixed in this PR
- [ ] After merge to `main`, confirm both keys land at `s3://elastic-docs-openapi-specs/elastic/elasticsearch-specification/main/elasticsearch{,-serverless}.json`
- [ ] Confirm a push to a version branch (e.g. `9.4`) publishes the stateful spec only

## Notes
Labeled for backport to `9.4` and `8.19` via this repo's label-driven backport bot. 7.17 was considered but dropped: its Makefile lacks the docs-final build targets (`transform-to-openapi-for-docs`, `overlay-docs`), so it cannot produce the spec.


.github/workflows/openapi-lint.yml
```diff
@@ -0,0 +1,36 @@
+name: openapi-lint
+
+on:
+  pull_request:
```
l-trotta, line 4: should this really run on every PR?

.github/workflows/openapi-lint.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches:
+      - main
+      - '[0-9]+.[0-9]+'
```
l-trotta: let's use the same 8.x + interval
```suggestion
      - '8.19'
      - '9.[0-9]+'
```

.github/workflows/openapi-publish.yml
```diff
@@ -0,0 +1,54 @@
+name: openapi-publish
+
+on:
+  push:
+    branches:
+      - main
```
l-trotta, line 6: does it make sense to publish docs every time the spec is updated? could this be a daily task?

.github/workflows/openapi-lint.yml
reakaleek: Intentionally left without a `paths` filter. The docs OAS is derived from specification sources, compiler code, language-example generation, Makefile targets, and dependencies — a hand-maintained allowlist would drift and silently skip validation. Filtering on checked-in `output/openapi/*.json` would also miss most PRs, since `generate.yml` only commits those files after merge. The check currently completes in under two minutes, so running it on every PR is the safest coverage.

.github/workflows/openapi-publish.yml
reakaleek: Fixed in 5c79ef2144a20cf49bac37c516b3bf2ed4c38185 — added `push.paths` so publish runs only when the checked-in OAS files change (the `elasticmachine` generated-output commit after a spec merge), when request examples change, or when docs example tooling changes. That keeps it event-driven rather than daily, avoids publishing on unrelated merges, and still publishes immediately when the spec actually changes.

l-trotta (approved):
LGTM, thank you!

PR #6412, opened 2026-07-07
Add inference flag to field caps

Mikep86:
Updates the spec to add the `inference` flag and `non_inference_indices` array to field caps. These were added in 9.5 in this PR: https://github.com/elastic/elasticsearch/pull/152102

### DIFF
--- a/output/openapi/elasticsearch-serverless-openapi.json
+++ b/output/openapi/elasticsearch-serverless-openapi.json
@@ -66686,6 +66686,20 @@
           "type": {
             "type": "string"
           },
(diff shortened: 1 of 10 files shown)


ah89 (approved):
One minor cosmetic suggestion, if you’d like to apply it: `non_inference_indices` is placed near the type rather than grouped with the other `non_*_indices` fields.

LGTM

Mikep86:
@ah89 This is already inconsistent. `searchable` and `non_searchable_indices` are next to each other.

PR #6413, opened 2026-07-07
Add GeoPoint shape type for geo_shape query

### LINKED ISSUE #501: Builtin GeoPoint shape
### Description

Hi.

It is not possible to build a `geo_shape` query with the type of `point` with existing shape and query builders in elastic java client v8.6.0. There is `GeoLine` shape but it requires multiple lists of coordinates.

Desired search request is:
```
"geo_shape": {
  "field_name": {
    "shape": {
      "type": "point",
      "coordinates": [
        29.0,
        40.0
      ]
    }
  }
}
```

It can be achieved with a [custom GeoPoint shape](https://gist.github.com/ykozcan/fda2818f1ac88fbf0b26afaced1a8b21) class but I think it would be nice to have it along with the `GeoLine` in the client api.

m15flores:
Closes elastic/elasticsearch-java#501

Adds a `GeoPoint` type in `_types/Geo.ts`, following the same shape as the existing `GeoLine`, and references it from `GeoShapeFieldQuery.shape` as `GeoShape | GeoPoint`.

Tried adding `GeoPoint` standalone first (like `GeoLine`), but it gets dropped as a dangling type since nothing points to it (`GeoLine` only survives through the aggregation reference). Referencing `GeoPoint` from `shape` fixes that without changing `GeoShape` itself, so this isn't a breaking change: raw JSON shapes still work exactly as before, `GeoPoint` is just an additional typed option.

Tested against a live 8.15.0 cluster: querying a geo_shape field with `{"type": "point", "coordinates": [...]}` as the shape returns the expected results.

### DIFF
--- a/output/typescript/types.ts
+++ b/output/typescript/types.ts
@@ -2629,6 +2629,11 @@ export interface GeoLine {
 
 export type GeoLocation = LatLonGeoLocation | GeoHashLocation | double[] | string
 
+export interface GeoPoint {
+  type: string
+  coordinates: double[]
+}
+
 export type GeoShape = any
 
 export type GeoShapeRelation = 'intersects' | 'disjoint' | 'within' | 'contains'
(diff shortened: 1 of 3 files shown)


l-trotta:
Thanks @m15flores! Couple things: while this is not breaking for the java client, it could be for other clients that don't override to JsonData, so I'll have to check before merging; also, could you add a comment about this being a hackfix?

m15flores:
Hey! Added the hackfix comment. Let me know if there's anything I can help verify for the other clients.

PR #6414, opened 2026-07-07
Fix Serverless create index API examples

ankikuma:
## Summary
- Add Serverless-specific OpenAPI overlays for the create index API so examples no longer reference `number_of_shards`, `number_of_replicas` and `routing`.
- Preserve existing Stack examples and language snippets unchanged.
- Override the primary Console and client code samples on Serverless to use a minimal create-index request.

## Test plan
- [x] Run `make overlay-docs` and verify `output/openapi/elasticsearch-serverless-openapi-docs-final.json` no longer includes shard settings in create index examples.
- [x] Confirm Stack OpenAPI docs still show the original shard settings examples.
- [ ] After publish, verify https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-indices-create shows Serverless-compatible examples.

### DIFF
--- a/docs/overlays/elasticsearch-serverless-openapi-overlays.yaml
+++ b/docs/overlays/elasticsearch-serverless-openapi-overlays.yaml
@@ -54,6 +54,10 @@ actions:
     description: Use a Serverless-compatible create index with mappings example
     update:
       value: "{\n  \"mappings\": {\n    \"properties\": {\n      \"field1\": { \"type\": \"text\" }\n    }\n  }\n}"
+  - target: "$.paths['/{index}'].put.requestBody.content['application/json'].examples.indicesCreateRequestExample3"
+    description: Use a Serverless-compatible create index with aliases example
+    update:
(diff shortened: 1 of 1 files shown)


ankikuma:
@lcawl I see in [IndexSettings.ts](https://github.com/elastic/elasticsearch-specification/blob/main/specification/indices/_types/IndexSettings.ts) that `number_of_shards` and `number_of_replicas` are marked `@availability stack` but they still show up in examples. 

Is the way I fixed this in this PR the best way to fix it, or is there some better automation that can remove these from examples ?

Also, while working on this I realized that we refer to the following [GLOBALLY_REJECTED_PARAMETERS](https://github.com/elastic/elasticsearch-serverless/blob/1151d26c09aa2958189a116ca7444820f78d3ccd/modules/serverless-rest-controller/src/main/java/co/elastic/elasticsearch/serverless/rest/RestrictedRestParameters.java#L103-L113) in many places in the serverless specs:
`master_timout`
`routing`
`wait_for_active_shards`

Can we fix this by tagging them as `@availability stack` ?

lcawl:
I think your PR is a good case for fixing the broader issue, so I've created https://github.com/elastic/elasticsearch-specification/pull/6446 instead.

ankikuma:
Superceded by #6446

PR #6415, opened 2026-07-08
Fix NodeInfoDiscover.seed_providers type

MattDevy:
## Summary
- change `NodeInfoDiscover.seed_providers` to accept either `string` or `string[]`
- align the field with observed `nodes.info` server responses
- match the existing union used by `seed_hosts` in the same type

## Testing
- not run (spec-only change)

Relates to: https://github.com/elastic/go-elasticsearch/issues/1524

### DIFF
--- a/specification/nodes/info/types.ts
+++ b/specification/nodes/info/types.ts
@@ -189,7 +189,7 @@ export class NodeInfoDiscover
 {
   seed_hosts?: string[] | string
   type?: string
-  seed_providers?: string[]
+  seed_providers?: string[] | string
 }
 
 export class NodeInfoAction {


Anaethelion (approved):
Approving. Both fields are the same setting type in the Elasticsearch source, so `seed_providers` must have the same spec type as `seed_hosts`:

- `discovery.seed_hosts` — [`SettingsBasedSeedHostsProvider.java#L34`](https://github.com/elastic/elasticsearch/blob/d68ebc3d5d1ce343a6b182f4bbe721709f4d3bff/server/src/main/java/org/elasticsearch/discovery/SettingsBasedSeedHostsProvider.java#L34)
- `discovery.seed_providers` — [`DiscoveryModule.java#L77`](https://github.com/elastic/elasticsearch/blob/d68ebc3d5d1ce343a6b182f4bbe721709f4d3bff/server/src/main/java/org/elasticsearch/discovery/DiscoveryModule.java#L77)

Both are `Setting.stringListSetting(...)`, serialized as either a string or an array in [`Settings.toXContent`](https://github.com/elastic/elasticsearch/blob/d68ebc3d5d1ce343a6b182f4bbe721709f4d3bff/server/src/main/java/org/elasticsearch/common/settings/Settings.java#L710-L724), matching `string[] | string`.

PR #6416, opened 2026-07-08
Model allocations in nodes.stats response

MattDevy:
## Summary
- add the missing `allocations` response model for `nodes.stats`
- add `allocations` to `specification/nodes/_types/Stats.ts`

## Validation
- `npm run compile:specification --prefix compiler`
- `npm run lint --prefix specification -- specification/nodes/_types/Stats.ts` (fails locally because `eslint` is not installed in `specification/node_modules`)

Relates to: https://github.com/elastic/go-elasticsearch/issues/1526

### DIFF
--- a/specification/nodes/_types/Stats.ts
+++ b/specification/nodes/_types/Stats.ts
@@ -33,6 +33,10 @@ export class Stats {
    */
   adaptive_selection?: Dictionary<string, AdaptiveSelection>
   /**
+   * Statistics about shard allocations on the node.
+   */
+  allocations?: Allocations
+  /**
    * Statistics about the field data circuit breaker.
    */
   breakers?: Dictionary<string, Breaker>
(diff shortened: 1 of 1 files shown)


Anaethelion (approved):
Approving. The model matches the Elasticsearch source exactly — [`NodeAllocationStats.toXContent` (L52-59)](https://github.com/elastic/elasticsearch/blob/d68ebc3d5d1ce343a6b182f4bbe721709f4d3bff/server/src/main/java/org/elasticsearch/cluster/routing/allocation/NodeAllocationStats.java#L52-L59):

| Field | ES type |
|---|---|
| `shards` | `int` → `integer` |
| `undesired_shards` | `int` → `integer` |
| `forecasted_ingest_load` | `double` → `double` |
| `forecasted_disk_usage_in_bytes` | `long` (bytes) → `long` |
| `current_disk_usage_in_bytes` | `long` (bytes) → `long` |

The `*_in_bytes` fields come from `humanReadableField`, which emits the raw byte `long` by default — modeling them as `long` is correct and consistent with the other `_in_bytes` fields in this file. `?` optional is right since these are only populated under the desired-balance allocator.

PR #6417, opened 2026-07-08
Remove obsolete compress/compress_threshold properties from SourceField

### LINKED ISSUE #1657: SourceField properties compress and compress_threshold removed in 5.0
These properties are available in 2.0 but not in 5.0 and onwards:
- 2.0: https://github.com/elastic/elasticsearch/blob/2.0/core/src/main/java/org/elasticsearch/index/mapper/internal/SourceFieldMapper.java#L97-L99
- 5.0: https://github.com/elastic/elasticsearch/blob/5.0/core/src/main/java/org/elasticsearch/index/mapper/SourceFieldMapper.java

Thus we should remove them from the `SourceField` class. Probably were inherited via NEST and due to being optional didn't make any noise when they went away.

m15flores:
Closes #1657

Removes `compress` and `compress_threshold` from `SourceField`. Both were removed from `_source` field mapping in Elasticsearch 5.0 (see the diff between 2.0 vs 5.0 SourceFieldMapper sources), so they no longer belong here.

I've verified this against a live 8.15.0 cluster, trying to set `compress` on `_source` fails with `mapper_parsing_exception: unknown parameter [compress] on metadata field [_source]`.


l-trotta (approved):
thank you!

PR #6425, opened 2026-07-09
[Inference API] Updating custom service task settings to indicate overriding secret parameters is forbidden

jonathan-buttner:
This PR updates the CustomService task settings to include that the `task_settings.parameters` cannot override the `secret_parameters` keys.

Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/153309

PR #6429, opened 2026-07-10
[Backport 9.3] Updating custom service task settings to indicate overriding secret parameters is forbidden

jonathan-buttner:
Backport of https://github.com/elastic/elasticsearch-specification/pull/6425 for 9.3

### DIFF
--- a/output/openapi/elasticsearch-openapi-docs.redirects.csv
+++ b/output/openapi/elasticsearch-openapi-docs.redirects.csv
@@ -28,7 +28,6 @@ cluster-stats-1,cluster-stats
 connector-put-1,connector-put
 count-1,count
 enrich-get-policy-1,enrich-get-policy
-esql-get-view-1,esql-get-view
 field-caps-1,field-caps
 fleet-msearch-1,fleet-msearch
 health-report-1,health-report
--- a/docs/examples/languageExamples.json
+++ b/docs/examples/languageExamples.json
@@ -4038,7 +4038,7 @@
     },
     {
       "language": "curl",
-      "code": "curl -X GET -H \"Authorization: ApiKey $ELASTIC_API_KEY\" -H \"Content-Type: application/json\" -d '[{},{\"id\":\"my-search-template\",\"params\":{\"query_string\":\"hello world\",\"from\":0,\"size\":10}},{},{\"id\":\"my-other-search-template\",\"params\":{\"query_type\":\"match_all\"}}]' \"$ELASTICSEARCH_URL/my-index/_msearch/template\""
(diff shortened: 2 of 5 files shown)


output/schema/schema.json
```diff
@@ -161595,19 +161595,19 @@
         "IndicesDeleteSampleConfigurationRequest1": {
           "alternatives": [
             {
-              "code": "resp = client.indices.delete_sample_configuration(\n    index=\"my-index\",\n)",
+              "code": "resp = client.perform_request(\n    \"DELETE\",\n    \"/my-index/_sample/config\",\n)",
```
jonathan-buttner, line 161598: @l-trotta I just ran the commands here: https://www.elastic.co/docs/contribute-docs/api-docs/elasticsearch-api-docs-quickstart

Should I not do that? Or are these changes ok?

output/openapi/elasticsearch-openapi-docs.redirects.csv
```diff
@@ -28,7 +28,6 @@ cluster-stats-1,cluster-stats
 connector-put-1,connector-put
 count-1,count
 enrich-get-policy-1,enrich-get-policy
-esql-get-view-1,esql-get-view
```
jonathan-buttner, line 31: @l-trotta I'm not sure why this is being removed either 🤔

l-trotta:
hey @jonathan-buttner, it's very common to have backporting errors when the original PR included output changes, that's why usually we avoid pushing any output file (since it will be generated anyways). I see that the AI backporting tool also failed, so there might be some manual cherrypicking to be done :) I'll take care of this on monday.

output/openapi/elasticsearch-openapi-docs.redirects.csv
l-trotta: probably because it doesn't exist in 9.3, I suspect

jonathan-buttner:
> hey @jonathan-buttner, it's very common to have backporting errors when the original PR included output changes, that's why usually we avoid pushing any output file (since it will be generated anyways). I see that the AI backporting tool also failed, so there might be some manual cherrypicking to be done :) I'll take care of this on monday.

Ah ok thanks. Also if there's a specific set of `make` commands that you recommend running prior to putting up a PR, that'd be super helpful. I've just been assuming we need to follow: https://www.elastic.co/docs/contribute-docs/api-docs/elasticsearch-api-docs-quickstart

l-trotta:
that's definitely outdated, thanks for pointing it out! what I usually run is `make generate && make compile && make spec-format-fix && make contrib`, and then after a quick check to the output, I only push the non output changes.

PR #6459, opened 2026-07-14
Fix validation for get component template

l-trotta:
Addressing https://github.com/elastic/elasticsearch/pull/145376 and https://github.com/elastic/elasticsearch/pull/144912.
Relevant code: 

- temporality_field: [IndexSettings](https://github.com/elastic/elasticsearch/blob/ce5a29cf2d3401213b41dd05793e3bc0e604c164/server/src/main/java/org/elasticsearch/index/IndexSettings.java#L648)
- preserve_leaf_arrays: [FlattenedFieldMapper](https://github.com/elastic/elasticsearch/blob/030dbbbf5fc27f5b6e735760e7cf3318fbde840b/server/src/main/java/org/elasticsearch/index/mapper/flattened/FlattenedFieldMapper.java#L357)

### DIFF
--- a/specification/_types/mapping/complex.ts
+++ b/specification/_types/mapping/complex.ts
@@ -31,10 +31,21 @@ export class FlattenedProperty extends PropertyBase {
   index?: boolean
   index_options?: IndexOptions
   null_value?: string
+  /**
+   * When set to lossy (the default), leaf arrays are sorted, de-nulled, and deduplicated in the returned synthetic source.
+   * When set to exact, leaf arrays preserve order, nulls, and duplicates.
+   * @server_default lossy
+   */
+  preserve_leaf_arrays?: PreserveLeafArrays
   similarity?: string
   split_queries_on_whitespace?: boolean
   time_series_dimensions?: string[]
   type: 'flattened'
+}
+
+export enum PreserveLeafArrays {
+  lossy,
+  exact
 }
 
 export class NestedProperty extends CorePropertyBase {
--- a/specification/_types/mapping/complex.ts
+++ b/specification/_types/mapping/complex.ts
@@ -32,9 +32,9 @@ export class FlattenedProperty extends PropertyBase {
   index_options?: IndexOptions
   null_value?: string
   /**
-   * When set to lossy (the default), leaf arrays are sorted, de-nulled, and deduplicated in the returned synthetic source.
-   * When set to exact, leaf arrays preserve order, nulls, and duplicates.
-   * @server_default lossy
+   * How leaf arrays are represented in synthetic source.
+   * When set to `lossy`, leaf arrays are sorted, de-nulled, and deduplicated in the returned synthetic source.
+   * When set to `exact`, leaf arrays preserve order, nulls, and duplicates.
    */
   preserve_leaf_arrays?: PreserveLeafArrays
   similarity?: string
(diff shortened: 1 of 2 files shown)


Anaethelion (changes requested):
Two issues, both in the docs/annotations rather than the field shapes.

### 1. `feature_flag=time_series_temporality` no longer exists

The flag was real and named exactly that, but it was removed in [`86b789b3433f`](https://github.com/elastic/elasticsearch/commit/86b789b3433f6454c63fd5c376ca75547baf9f61) (elastic/elasticsearch#152091, merged 2026-06-25) — present in `9.5` and `main`, not in `9.4`. That commit replaced the gate with an unconditional registration:

- [`IndexScopedSettings#L273`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/common/settings/IndexScopedSettings.java#L273) — `settings.add(TIME_SERIES_TEMPORALITY_FIELD)`, no flag check
- [`IndexSettings#L648-L653`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/index/IndexSettings.java#L648-L653) — plain `Setting.simpleString` with `Property.ServerlessPublic`

A repo-wide grep for `time_series_temporality` now returns nothing, and #152091's changelog ships it as a notable GA feature. So `visibility=feature_flag` is stale on both the stack and serverless lines. The flag did gate it in 9.4, so `since=9.5.0` GA vs. `since=9.4.0` with the flag dropped is your call.

(The PR body pins `ce5a29cf...`, which predates the removal — likely the source of the drift.)

### 2. `@server_default lossy` is conditional

[`FlattenedFieldMapper#L357-L364`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/index/mapper/flattened/FlattenedFieldMapper.java#L357-L364) defaults to `LOSSY` only when `synthetic_source_keep == NONE && isStrictColumnar() == false`, else `EXACT`. Since [`synthetic_source_keep` defaults to `ARRAYS` for logsdb](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/index/mapper/Mapper.java#L87-L96), logsdb indices default to `exact`. ES docs call this out directly in [flattened.md#L532](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/docs/reference/elasticsearch/mapping-reference/flattened.md?plain=1#L532): "logsdb index mode defaults `index.mapping.synthetic_source_keep` to `arrays`, so flattened fields default to `exact` unless you override them." `@server_default` can't express that, so worth noting in the doc comment instead of asserting "lossy (the default)".

### Verified correct

- `lossy`/`exact` match the [enum's lowercasing `toString`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/index/mapper/flattened/FlattenedFieldMapper.java#L166-L174); sorted/de-nulled/deduplicated matches the docs
- `since=9.4.0` correct for both — elastic/elasticsearch#144912 and elastic/elasticsearch#145376 are both in `9.4`
- `temporality_field` docs accurate, incl. "keyword dimension field" — [`IndexMode#L179-L211`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/server/src/main/java/org/elasticsearch/index/IndexMode.java#L179-L211) enforces both `KeywordFieldType` and `isDimension()`; null resolution matches `LOAD_CONSTANT_NULLS` in `EsPhysicalOperationProviders`
- `PreserveLeafArrays` doesn't collide with any existing spec type

Anaethelion (approved):

PR #6461, opened 2026-07-14
Fix validation for ml info

l-trotta:
Addressing https://github.com/elastic/elasticsearch/pull/147496, specifically https://github.com/elastic/elasticsearch/blob/a6141b9dbec012eb147c8280ccc65f847b2956fa/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportMlInfoAction.java#L107. Not really an enum in the server code, but since it's a string with only 2 values I'm modeling it as such.


qn895 (approved):
ML UI LGTM

PR #6463, opened 2026-07-14
Add logging section to _xpack/usage

l-trotta:
Addressing https://github.com/elastic/elasticsearch/pull/148087, relevant code: https://github.com/elastic/elasticsearch/blob/4b594675f059e54283b89780ea04ab45e680be37/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/logging/LoggingFeatureSetUsage.java#L99-L133

### DIFF
--- a/specification/xpack/usage/XPackUsageResponse.ts
+++ b/specification/xpack/usage/XPackUsageResponse.ts
@@ -29,6 +29,7 @@ import {
   GpuVectorIndexing,
   HealthStatistics,
   Ilm,
+  Logging,
   MachineLearning,
   Monitoring,
   RuntimeFieldTypes,
@@ -65,6 +66,10 @@ export class Response {
     gpu_vector_indexing?: GpuVectorIndexing
     health_api?: HealthStatistics
     ilm: Ilm
+    /**
+     * @availability stack since=9.5.0
+     */
+    logging?: Logging
     logstash: Base
     ml: MachineLearning
     monitoring: Monitoring
--- a/specification/xpack/usage/types.ts
+++ b/specification/xpack/usage/types.ts
@@ -78,7 +78,7 @@ export class Datafeed {
 /**
  * Usage statistics about logging configuration reported by `_xpack/usage`.
  */
-export class Logging extends Base {
+export class Logging {
   /**
    * Search query log configuration.
    */
@@ -105,7 +105,7 @@ export class QueryLoggingConfig {
   /**
    * The configured logging threshold, if any.
    */
-  threshold?: string
+  threshold?: Duration
 }
 
 export class EsqlLoggingConfig {
@@ -120,7 +120,7 @@ export class EsqlLoggingConfig {
   /**
    * The configured logging thresholds, keyed by threshold name, if any.
    */
-  thresholds?: Dictionary<string, string>
+  thresholds?: Dictionary<string, Duration>
 }
 
 export class DataStreams extends Base {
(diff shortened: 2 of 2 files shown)


Anaethelion (changes requested):
### `Logging extends Base` — `available`/`enabled` are not emitted

[`XPackFeatureUsage#toXContent`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackFeatureUsage.java#L63-L73) is `final`, and its only emission path is `innerXContent` — which is also where the base class writes `available`/`enabled`. A subclass gets those two fields **only** by calling `super.innerXContent(...)`.

[`LoggingFeatureSetUsage#innerXContent`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/logging/LoggingFeatureSetUsage.java#L129-L133) doesn't:

```java
protected void innerXContent(XContentBuilder builder, Params params) throws IOException {
    builder.field("querylog", queryConfig);
    builder.field("esql", esqlConfig);
}
```

Compare [`VectorDBDocumentFeatureSetUsage#innerXContent`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/application/VectorDBDocumentFeatureSetUsage.java#L50-L54) (#6462), which calls `super.innerXContent(builder, params)` first — that's why `extends Base` is correct there and not here.

The REST test agrees: [`10_usage.yml`](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/logging/10_usage.yml) only ever asserts `logging.querylog.*` and `logging.esql.*`.

**This looks like an ES bug rather than a spec bug.** The constructor passes `available=true, enabled=true` and `writeTo` serializes both over the wire — they're just dropped from the XContent, and every other usage section exposes them. So either drop `extends Base` to match current behavior, or raise it with the logging team and keep it. Worth checking before changing the spec.

### Verified correct

- `querylog`/`esql` names and order match `innerXContent`
- `QueryLoggingConfig`: `enabled` + `user` from the shared [`LoggingConfig` record](https://github.com/elastic/elasticsearch/blob/89f9ea5f7d02b2739db56ceda428b2c94146e155/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/logging/LoggingFeatureSetUsage.java#L29-L46), plus `system`; `threshold` optional via the `Strings.isNotEmpty` guard
- `EsqlLoggingConfig`: `enabled` + `user`, `thresholds` optional via the `isEmpty() == false` guard; `Dictionary<string, string>` matches `Map<String, String>`
- `since=9.5.0` correct — `1685f3ce2ab7` (elastic/elasticsearch#148087) is in `9.5`
- No name collisions

### Nit (optional)

`threshold` and the `thresholds` values are durations (`50ms`, `0ms` in the tests). They're `String` server-side so `string` is defensible, but `Duration` would carry more meaning.

l-trotta:
good catch on the missing super in toXcontent, I really took for granted that it was going to be the same as the others. I'll create an issue in the server repo.

Anaethelion (approved):

l-trotta:
No issue needed, this is the intended behavior, confirmed by @smalyshev : Logging is always enabled and available, so it doesn't need those 2 fields. They are included in writeTo, since it's the correct definition of the java object, but they are excluded from json serialization (toXContent).

PR #6500, opened 2026-07-20
Codegen name for _slice

l-trotta:
Followup of https://github.com/elastic/elasticsearch-specification/pull/6427. The new `_slice` parameter clashes with the old `slice` in 3 classes (SearchRequest, DeleteByQueryRequest, UpdateByQueryRequest) because some of the language clients remove underscores, or convert snake case to camel case. Adding a `@codegen_name` annotation to rename it to `route_slice` in all its occurrences.


MattDevy (approved):
LGTM!

PR #6509, opened 2026-07-23
Generate C# language examples via the .NET request converter

flobernd:
## Summary

Wires C# into the language-examples generation, side by side with Python, JavaScript, Ruby, PHP, curl, and Java. `docs/examples/generate-language-examples.js` converts request examples to C# through `@elastic/request-converter`'s new `csharp` format (elastic/request-converter#107), backed by the `@elastic/request-converter-dotnet` WASM bundle (elastic/elasticsearch-net#8931). The regenerated `languageExamples.json` is deliberately not part of this PR; the C# entries land in a follow-up once the converter packages are published.

## Changes

- `'C#'` added to `LANGUAGES`. C# conversion is best-effort: a failed conversion skips that example's C# entry and is reported in an end-of-run summary instead of aborting, because the .NET converter has known endpoint coverage gaps. The other languages keep their fail-fast behavior.
- `@elastic/request-converter-dotnet` (`~9.5.0`) added to `docs/examples/package.json`; the Makefile's `generate-language-examples` target updates it together with `@elastic/request-converter`.

## Notes

- There is no CI automation for `languageExamples.json`: the `generate.yml` workflow rebuilds `output/` but never runs `generate-language-examples`, which remains a manual Makefile target. A full local dry run against locally built converter packages converted 649 of 761 request examples (112 skipped: 96 unsupported endpoints, 16 conversion issues); generated examples use fluent descriptor syntax with an illustrative `MyDocument` document type.
- `package-lock.json` is intentionally untouched: the dotnet package is published from elastic/elasticsearch-net#8931, and the lockfile refresh must follow the first publish.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ overlay-docs: ## Apply overlays to OpenAPI documents
 	rm output/openapi/elasticsearch-openapi-docs.tmp*.json
 
 generate-language-examples:
-	@npm update --prefix docs/examples @elastic/request-converter
+	@npm update --prefix docs/examples @elastic/request-converter @elastic/request-converter-dotnet
 	@node docs/examples/generate-language-examples.js
 	@npm run format:fix-examples --prefix compiler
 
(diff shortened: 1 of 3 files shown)


l-trotta (approved):
:tada:

PR #6510, opened 2026-07-27
Mark ES|QL data federation APIs as internal

leemthompo:
Marks the Stack ES|QL data source and dataset APIs as internal. These APIs are excluded from public Stack and all Serverless outputs.


l-trotta (approved):
thank you!

leemthompo:
@l-trotta I won't merge this immediately in case anything else changes in meantime 😉

lcawl (approved):
visiblity setting LGTM too

PR #6513, opened 2026-07-28
Tighten up definitions of snapshot status

DaveCTurner:
This API has (confusingly) always returned `SUCCESS` when the snapshot
is awaiting finalization. Fixing this would be a breaking change but we
can at least clarify the docs in this area.

### DIFF
--- a/output/openapi/elasticsearch-openapi.json
+++ b/output/openapi/elasticsearch-openapi.json
@@ -132488,7 +132488,7 @@
             "type": "string"
           },
           "state": {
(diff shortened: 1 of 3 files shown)


DaveCTurner:
If https://github.com/elastic/elasticsearch/pull/155241 lands I'll update this again in the affected branches, but until that happens we should apply this change anyway. We're definitely not going to fix this in any unmaintained versions.

joshua-adams-1 (approved):
LGTM!

PR #6528, opened 2026-07-31
Regenerate language examples

flobernd:
Runs `make generate-language-examples` on `main`, using `@elastic/request-converter` 9.4.2 and `@elastic/request-converter-dotnet` 9.4.0 (the versions declared after #6535).

## What changed

- **Adds C# examples for 638 requests.** C# generation was introduced in #6509, but the generated examples themselves were never committed, so `languageExamples.json` had no C# entries at all until now. 123 requests are still skipped because the .NET request converter does not support those endpoints yet (mostly `cat.*`, `watcher.*`, `connector.*`, `logstash.*`), which the generator treats as best-effort.
- **Each language is now rendered with its own converter policy.** Before elastic/request-converter#120 (shipped in 9.4.2), the exporters leaked their Handlebars helpers into one another, so in multi-language runs everything after the first request was rendered with the PHP exporter's supported-API and aliasing rules. With that fixed:
  - The ES|QL data federation APIs (`esql.get_data_source`, `esql.get_dataset`, ...), `indices.create_from`, the migrate-reindex APIs, `connector.last_sync`, and `security.delegate_pki` generate typed Python/JavaScript/Ruby calls instead of transport-level fallbacks; they were previously suppressed by the PHP exporter's policy.
  - Python examples use valid keyword arguments again: `from_` instead of the reserved word `from`, `security_tokens`/`security_profile` instead of `security-tokens`/`security-profile`, and `index_auto_expand_replicas` instead of `index.auto_expand_replicas`. The JavaScript `ingest.put_pipeline` example keeps `_meta` instead of `meta`.
  - Ruby examples for `rollup.*`, `autoscaling.*`, and `shutdown.*`, and Python examples for the serverless-only `project.*` APIs, are rendered as transport-level `perform_request(...)` calls: these endpoints are deliberately marked unsupported by the respective exporters and only appeared as typed calls before because of the leaked PHP policy.
- **Syncs `docs/examples/package-lock.json`** with the `~9.4.2` / `~9.4.0` ranges declared in `package.json`.

The Ruby `autoscaling.put_autoscaling_policy` fallback renders its path placeholder HTML-escaped (`/_autoscaling/policy/&lt;name&gt;`); that is a converter template bug tracked in elastic/request-converter#125.

Curl and Java examples are unaffected.

### DIFF
--- a/docs/examples/languageExamples.json
+++ b/docs/examples/languageExamples.json
@@ -21,6 +21,10 @@
       "code": "curl -X GET -H \"Authorization: ApiKey $ELASTIC_API_KEY\" \"$ELASTICSEARCH_URL/_xpack/usage\""
     },
     {
+      "language": "C#",
+      "code": "var response = await client.Xpack.UsageAsync();"
+    },
+    {
       "language": "Java",
       "code": "client.xpack().usage(u -> u);\n"
     }
@@ -45,6 +49,10 @@
     {
       "language": "curl",
       "code": "curl -X GET -H \"Authorization: ApiKey $ELASTIC_API_KEY\" \"$ELASTICSEARCH_URL/_xpack\""
+    },
+    {
+      "language": "C#",
+      "code": "var response = await client.Xpack.InfoAsync();"
     },
     {
       "language": "Java",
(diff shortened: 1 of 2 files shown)


l-trotta (approved):

l-trotta (approved):

flobernd:
Refreshed with a new generation: bumped `@elastic/request-converter-dotnet` to `0.0.0-main.g1cd4e5e34530` (current `latest-main`, pinned in the lock file; the package.json spec stays `latest-main`). C# examples now include the executing client invocation inline, wrapping multi-line calls onto their own method line (elastic/elasticsearch-net#8967).

PR #6531, opened 2026-07-31
Regenerate language examples

flobernd:
Runs `make generate-language-examples` on `8.19`, using `@elastic/request-converter` 8.19.3 and `@elastic/request-converter-dotnet` 8.19.0 (the versions declared after #6532). Companion to #6528 (`main`) and #6530 (`9.5`).

## What changed

- **Adds C# examples for 564 requests.** C# generation was backported in #6527, but the generated examples themselves were never committed, so `languageExamples.json` had no C# entries at all until now. 108 requests are still skipped because the .NET request converter does not support those endpoints yet (mostly `cat.*`, `watcher.*`, `connector.*`, `logstash.*`), which the generator treats as best-effort.
- **Refreshes stale examples in Python/JavaScript/Ruby/PHP/curl.** The lockfile on this branch was still resolving `@elastic/request-converter` **9.1.2** from a stale `^9.1.2` range, i.e. a 9.x converter was being used on the 8.19 branch, and some examples had not been regenerated since their request YAML changed. Notable fixes:
  - `_bulk` and `_msearch` curl examples now send `Content-Type: application/x-ndjson` with newline-delimited documents instead of a JSON array with `application/json`.
  - `indices/put_alias` example 1 now generates `client.indices.put_alias(...)` against `my-index-2099.05.06-000001`, matching the request in the YAML; it previously generated an unrelated `client.indices.update_aliases(...)` call.
- **Each language is now rendered with its own converter policy.** Before elastic/request-converter#120 (shipped in 8.19.3), the exporters leaked their Handlebars helpers into one another, so in multi-language runs everything after the first request was rendered with the PHP exporter's supported-API and aliasing rules. With that fixed:
  - `indices.create_from`, the migrate-reindex APIs, `snapshot.repository_verify_integrity`, `connector.last_sync`, and `security.delegate_pki` generate typed Python/JavaScript/Ruby calls instead of transport-level fallbacks; they were previously suppressed by the PHP exporter's policy.
  - Python examples use valid keyword arguments again: `from_` instead of the reserved word `from`, `security_tokens`/`security_profile` instead of `security-tokens`/`security-profile`, and `index_auto_expand_replicas` instead of `index.auto_expand_replicas`. The JavaScript `ingest.put_pipeline` example keeps `_meta` instead of `meta`.
  - Ruby examples for `rollup.*`, `autoscaling.*`, and `shutdown.*` are rendered as transport-level `perform_request(...)` calls: these endpoints are deliberately marked unsupported by the Ruby exporter and only appeared as typed calls before because of the leaked PHP policy.
- **Rewraps four example descriptions** in `inference/put_amazonsagemaker` and `inference/put_custom` to the generator's 132-column width.
- **Syncs `docs/examples/package-lock.json`** to the `~8.19.3` / `~8.19.0` ranges declared in `package.json`.

The Ruby `autoscaling.put_autoscaling_policy` fallback renders its path placeholder HTML-escaped (`/_autoscaling/policy/&lt;name&gt;`); that is a converter template bug tracked in elastic/request-converter#125.

Curl and Java examples are unaffected.

### DIFF
--- a/docs/examples/languageExamples.json
+++ b/docs/examples/languageExamples.json
@@ -21,6 +21,10 @@
       "code": "curl -X GET -H \"Authorization: ApiKey $ELASTIC_API_KEY\" \"$ELASTICSEARCH_URL/_xpack/usage\""
     },
     {
+      "language": "C#",
+      "code": "var response = await client.Xpack.UsageAsync();"
+    },
+    {
       "language": "Java",
       "code": "client.xpack().usage(u -> u);\n"
     }
@@ -45,6 +49,10 @@
     {
       "language": "curl",
       "code": "curl -X GET -H \"Authorization: ApiKey $ELASTIC_API_KEY\" \"$ELASTICSEARCH_URL/_xpack\""
+    },
+    {
+      "language": "C#",
+      "code": "var response = await client.Xpack.InfoAsync();"
     },
     {
       "language": "Java",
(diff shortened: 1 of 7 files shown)


l-trotta (approved):

l-trotta (approved):

flobernd:
Refreshed with a new generation: bumped `@elastic/request-converter-dotnet` to `~8.19.3`. C# examples now include the executing client invocation inline, wrapping multi-line calls onto their own method line (elastic/elasticsearch-net#8967, backported via elastic/elasticsearch-net#8968).

PR #6544, opened 2026-08-05
Document transform dest.aliases and move_on_creation in the API spec

### LINKED ISSUE #3755: [Feedback]: Document `move_on_creation` setting in the create transform API
When setting up a transform, it is possible to specify in the destination of the transform the `move_on_creation` setting.
````
"dest": {
    "index": "logs-ti_misp_latest.dest_threat_attributes-3",
    "aliases": [
      {
        "alias": "logs-ti_misp_latest.threat_attributes",
        "move_on_creation": true
      }
    ]
````
Such a setting is not documented anywhere on our documentation. Only here in Terraform https://registry.terraform.io/providers/elastic/elasticstack/0.11.14/docs/resources/elasticsearch_transform#:~:text=move_on_creation%20(Boolean)%20Whether%20the%20destination%20index%20should%20be%20the%20only%20index%20in%20this%20alias.%20Defaults%20to%20false
 > move_on_creation
(Optional, boolean) Whether or not the destination index should be the only index in this alias. If true, all the other indices will be removed from this alias before adding the destination index to this alias. Defaults to false.

This is the explanation I got from Dev. `move_on_creation: true` controls alias behavior, NOT index deletion. When the destination index is created, move the alias to point to this new index.

One Example Scenario: If the alias `logs-ti_misp_latest.threat_attributes` was pointing to an older index (e.g., -1 or -2), it will now point to -3

The old index data remains, but queries using the alias will now hit the new index.

charlotte-hoblik:
## Summary
- Document `dest.aliases` and `move_on_creation` on the transform Destination type so they appear in the Create/Update/Get transform API docs.
- Point get/update transform responses at the transform Destination type (instead of reindex’s).
- Add a create-transform example that uses destination aliases.

Fixes elastic/docs-content#3755

## Preview:
Screenshot 2026-08-05 at 11 20 30
Screenshot 2026-08-05 at 11 20 42
Screenshot 2026-08-05 at 11 33 15

### DIFF
--- a/specification/transform/_types/Transform.ts
+++ b/specification/transform/_types/Transform.ts
@@ -24,7 +24,13 @@ import {
   HistogramAggregation,
   TermsAggregation
 } from '@_types/aggregations/bucket'
-import { Field, IndexName, Indices, ProjectRouting } from '@_types/common'
+import {
+  Field,
+  IndexAlias,
+  IndexName,
+  Indices,
+  ProjectRouting
+} from '@_types/common'
 import { RuntimeFields } from '@_types/mapping/RuntimeFields'
 import { float, integer } from '@_types/Numeric'
 import { QueryContainer } from '@_types/query_dsl/abstractions'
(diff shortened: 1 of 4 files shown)


kosabogi (commented):
LGTM for the writing

l-trotta (approved):
LGTM, thanks!

Issue #6542, opened 2026-08-04
[OpenAPI] Fix simple missing defaults for query parameters


lcawl:
@georgewallace noted that we were missing default values in the API docs.
Above and beyond https://github.com/elastic/elasticsearch-specification/issues/5369, AI analysis shows there's another bug:

# Add server defaults to OpenAPI parameters

Fixes the primary gap behind [#5369](https://github.com/elastic/elasticsearch-specification/issues/5369): query/path parameters never copy `Property.server_default` into OpenAPI `schema.default`.

## Root cause

[`parameter_data`](compiler-rs/clients_schema_to_openapi/src/paths.rs) builds parameter schemas with only `convert_value_of(&prop.typ)` and never sets `default`:

```65:78:compiler-rs/clients_schema_to_openapi/src/paths.rs
    fn parameter_data(prop: &Property, in_path: bool, tac: &mut TypesAndComponents) -> anyhow::Result<ParameterData> {
        // ...
        format: ParameterSchemaOrContent::Schema(tac.convert_value_of(&prop.typ)?),
        // ...
    }
```

Body properties already work via [`convert_property`](compiler-rs/clients_schema_to_openapi/src/schemas.rs) + `fill_data_with_prop` (sets `data.default` from `server_default`, and wraps `$ref` in `allOf` for OAS 3.0).

```mermaid
flowchart LR
  ts["@server_default in TS"] --> schemaJson["schema.json serverDefault"]
  schemaJson --> bodyProps["convert_property: default set"]
  schemaJson --> params["parameter_data: default missing"]
  bodyProps --> openapiOk["OpenAPI schema.default present"]
  params --> openapiGap["OpenAPI parameter schema has no default"]
```

lcawl:
Fixed by https://github.com/elastic/elasticsearch-specification/pull/6548

PR #6548, opened 2026-08-05
[OpenAPI] Include default query parameter values

### LINKED ISSUE #5369: Default value missing from OpenAPI output
There seem to be cases where a default value specified in the typescript file is missing from the OpenAPI output. For example, in https://github.com/elastic/elasticsearch-specification/blob/main/specification/inference/text_embedding/TextEmbeddingRequest.ts

... there is a query parameter with a default value as follows:

```
  query_parameters: {
    /**
     * Specifies the amount of time to wait for the inference request to complete.
     * @server_default 30s
     */
    timeout?: Duration
```

However, that default is missing from both the docs-specific and generic output `elasticsearch-openapi.json` file:

```
          {
            "in": "query",
            "name": "timeout",
            "description": "Specifies the amount of time to wait for the inference request to complete.",
            "deprecated": false,
            "schema": {
              "$ref": "#/components/schemas/_types.Duration"
            },
            "style": "form"
          }

...
      "_types.Duration": {
        "description": "A duration. Units can be `nanos`, `micros`, `ms` (milliseconds), `s` (seconds), `m` (minutes), `h` (hours) and\n`d` (days). Also accepts \"0\" without a unit and \"-1\" to indicate an unspecified value.",
        "oneOf": [
          {
            "type": "string"
          },
          {
            "type": "string",
            "enum": [
              "-1"
            ]
          },
          {
            "type": "string",
            "enum": [
              "0"
            ]
          }
        ]
      },
```

This makes me wonder if for case where we're re-using a type but are trying to add additional information to it (as in this case with `@server_default`, whether the generator ought to consolidate that info rather than using a `$ref`. For example, create this output instead:

```
  {
            "in": "query",
            "name": "timeout",
            "description": "Specifies the amount of time to wait for the inference request to complete. A duration. Units can be `nanos`, `micros`, `ms` (milliseconds), `s` (seconds), `m` (minutes), `h` (hours) and\n`d` (days). Also accepts \"0\" without a unit and \"-1\" to indicate an unspecified value.",",
            "deprecated": false,
            "schema": {
                "default": "30s"
                "oneOf": [
                  {
                    "type": "string"
                  },
                 {
                    "type": "string",
                    "enum": [
                      "-1"
                    ]
                  },
                 {
                    "type": "string",
                    "enum": [
                      "0"
                    ]
                  }
                ]
              },
```

Maybe that's too much to ask of the generator, but I think it's nonetheless worth revisiting how we override these re-used types.

lcawl:
Fixes https://github.com/elastic/elasticsearch-specification/issues/5369 and https://github.com/elastic/elasticsearch-specification/issues/6542

Query/path parameters now get OpenAPI `default` from `@server_default`.

## AI summary

- Extracted `into_inline_schema` in `schemas.rs` (wraps `$ref` in `allOf` for OAS 3.0)
- Updated `parameter_data` in `paths.rs` to apply `prop.server_default` on the parameter schema
- Rebuilt `compiler-wasm-lib` WASM
- Regenerated stack/serverless OpenAPI (+ docs variants locally; docs JSON is gitignored)
- Verified
    - `create-include_source_on_error` → `"default": true`
    - `create-timeout` → `"default": "1m"` + `allOf`/`$ref`
    - `inference text_embedding timeout` → `"default": "30s"` + `allOf`/`$ref` (closes #5369)
    - `_types.Duration` unchanged (no shared default)
    - Params without `serverDefault` stay without `default`
    - `cargo test -p clients_schema_to_openapi` — 5 passed

## Tests

When I generate the OpenAPI output, I now see `default` values like this:

```json
...
          {
            "in": "query",
            "name": "size",
            "description": "Specifies a max number of results to get",
            "deprecated": false,
            "schema": {
              "default": 100.0,
              "type": "number"
            },
            "style": "form"
          },

...
          {
            "in": "query",
            "name": "master_timeout",
            "description": "Period to wait for a connection to the master node.",
            "deprecated": false,
            "schema": {
              "default": "30s",
              "allOf": [
                {
                  "$ref": "#/components/schemas/_types.Duration"
                }
              ]
            },
```

Thus it seems to work for both the simple inline and `allOf` scenarios.

## Screenshots

The query parameter defaults now work:

image

The request body defaults still work:

image


l-trotta:
@lcawl is it necessary to always wrap the fields, or could we just do it for fields with a default?

lcawl:
> @lcawl is it necessary to always wrap the fields, or could we just do it for fields with a default?

You're right, only necessary for the latter. I've pushed a commit so that `parameter_data` now wraps `$ref` in `allOf` only when a `server_default` is present.

l-trotta (approved):
thank you!

lcawl:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/elasticsearch-specification/pull/6556)|

### Questions ?

PR #6553, opened 2026-08-06
Add priority to recovery APIs

PeteGillinElastic:
This reflects the API changes in elastic/elasticsearch#156018.

### DIFF
--- a/output/schema/schema.json
+++ b/output/schema/schema.json
@@ -108534,7 +108534,7 @@
         "name": "CatAliasesColumns",
         "namespace": "cat._types"
       },
-      "specLocation": "cat/_types/CatBase.ts#L2529-L2529",
+      "specLocation": "cat/_types/CatBase.ts#L2536-L2536",
       "type": {
         "kind": "union_of",
         "items": [
@@ -108669,7 +108669,7 @@
         "name": "CatAllocationColumns",
         "namespace": "cat._types"
       },
-      "specLocation": "cat/_types/CatBase.ts#L2530-L2530",
+      "specLocation": "cat/_types/CatBase.ts#L2537-L2537",
       "type": {
         "kind": "union_of",
         "items": [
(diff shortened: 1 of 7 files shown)


PeteGillinElastic:
I think we need a flight recorder run with the ES PR in it (which should happen tomorrow?) to make the validation go green.

PeteGillinElastic:
> I think we need a flight recorder run with the ES PR in it (which should happen tomorrow?) to make the validation go green.

i.e. I think we need to wait for https://github.com/elastic/clients-flight-recordings/tree/main/elasticsearch/indices.recovery to be updated.

specification/indices/recovery/types.ts
```diff
@@ -143,13 +143,37 @@ export enum RecoveryType {
   SNAPSHOT
 }
 
+export enum RecoveryPriority {
```
samxbr, line 147: It may be worth mentioning the order of priority (eg. descending highest to lowest)

output/schema/schema.json
samxbr, line 1: I guess the files under `output` directory are updated by automation?

specification/cat/_types/CatBase.ts
```diff
@@ -2366,6 +2366,13 @@ export enum CatRecoveryColumn {
    * @aliases st
    */
   stage,
+  /**
+   * The recovery priority.
+   * @aliases pr
+   * @availability stack since=9.6.0
+   * @availability serverless
+   */
+  priority,
```
samxbr, line 2375: Does the `_cat` API already return the priorities, it comes for free?

output/schema/schema.json
PeteGillinElastic: Yes. The latest I heard, the advice is _not_ to include the `make contrib` changes in your PR, and let the automation take care of it. I think this is to reduce the chances of merge conflicts.

specification/cat/_types/CatBase.ts
PeteGillinElastic: It has them because of the changes to `RestCatRecoveryAction` in the change linked from the PR description here.

specification/indices/recovery/types.ts
PeteGillinElastic: Done.

samxbr (approved):
LGTM, thanks!

PR #6557, opened 2026-08-07
Add CI job to generate examples, including Java

l-trotta:
The PR does 2 things:

- Remove the `generate-language-examples-with-java` command, now `generate-language-examples` will also generate java examples if it finds `JAVA_ES_REQUEST_CONVERTER_JAR` as env variable.
- Adds a CI job that generates the jar file for the java request converter, and then runs example generation.


.github/workflows/generate-language-examples.yml
```diff
@@ -0,0 +1,78 @@
+name: Generate Language Examples
```
Anaethelion: I'd like us to figure out a less generic name for this.

Anaethelion (approved):
Just a nit on the job name, could be clearer to avoid confusion later.

.github/workflows/generate-language-examples.yml
l-trotta: generate language clients example requests for docs?

Anaethelion: LGTM

PR #6558, opened 2026-08-07
Updating semantic text default for the inference_id field

jonathan-buttner:
This PR updates the semantic text field's default value and description for the `inference_id` field.

The default is jina for 9.4 and greater.

I'll open a separate PR for 9.3 because it defaults to a different value: https://github.com/elastic/elasticsearch-specification/pull/6560


georgewallace (approved):
LGTM

PR #6582, opened 2026-08-28
Add 'columnar_stored' option to SourceFieldMode enum

pickypg:
Without setting this, strict clients reject templates (and likely index mappings) that try to use this value in 9.5.0+.


pickypg:
For reference: https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-source-field#columnar-stored

l-trotta (approved):
thank you!

PR #6583, opened 2026-08-28
Add missing bbq_disk index_options and fix InnerRetriever optionality

john-wagster:
## Summary

This PR brings the elasticsearch-specification up to date with the Elasticsearch server for `bbq_disk` index options and fixes an `InnerRetriever` optionality mismatch.

### Dense vector `index_options` — 5 missing `bbq_disk` properties

The following fields are fully implemented and GA in Elasticsearch but were absent from the specification:

| Field | Type | Default | GA Since |
|---|---|---|---|
| `cluster_size` | integer | 384 | 9.2.0 |
| `default_visit_percentage` | float | 0 | 9.2.0 |
| `bits` | integer | 1 | 9.4.0 |
| `precondition` | boolean | false | 9.4.0 |
| `auto_calibrate` | boolean | false | 9.5.0 |

Server-side source: `DenseVectorFieldMapper.java`, `BBQIVFIndexOptions` (lines 2113-2243).

### Additional fixes

- **`flat_index_threshold` docstring**: Updated to include `bbq_disk` in the list of applicable index types (the server parses it for `bbq_disk` at line 2139).
- **`confidence_interval` deprecation**: Added `@deprecated 9.5.0` annotation. The server emits a deprecation warning for this parameter since `IndexVersions.UPGRADE_TO_LUCENE_10_4_0`.

### `InnerRetriever` — `weight` and `normalizer` should be optional

The `InnerRetriever` class (used by `LinearRetriever.retrievers[]`) declared `weight` and `normalizer` as required, but the server parses both with `optionalConstructorArg()`:

- `weight` defaults to `1.0f` (`LinearRetrieverComponent.java:30`)
- `normalizer` falls back to the top-level `LinearRetriever.normalizer`, then to `none`/identity (`LinearRetrieverBuilder.java:129-137`)

Fixes both fields to optional with `@server_default` annotations.

PR #6585, opened 2026-08-28
Add missing bbq_disk cluster_size and default_visit_percentage

john-wagster:
## Summary

Add two missing `bbq_disk`-specific properties to `DenseVectorIndexOptions`:

| Field | Type | Default | GA Since |
|---|---|---|---|
| `cluster_size` | integer | 384 | 9.2.0 |
| `default_visit_percentage` | float | 0 | 9.2.0 |

Both have been in the server since `bbq_disk` was introduced in 9.2 but were missing from the specification.

Split from #6583 for targeted backporting per @margaretgu's feedback.


l-trotta (approved):

tteofili (approved):
LGTM

PR #6617, opened 2026-09-02
Describe cleanup in delete-snapshot API

DaveCTurner:
`DELETE _snapshot/${REPO}/${SNAP}?wait_for_completion=true` also waits
for the associated cleanup work to complete but this isn't documented
today. Moreover, it's unclear that different requests may do different
parts of the cleanup work before returning.

Relates elastic/elasticsearch#158333


specification/snapshot/delete/SnapshotDeleteRequest.ts
```diff
@@ -58,8 +58,9 @@ export interface Request extends RequestBase {
      */
     master_timeout?: Duration
     /**
-     * If `true`, the request returns a response when the matching snapshots are all deleted.
      * If `false`, the request returns a response as soon as the deletes are scheduled.
+     * If `true`, the request returns a response when the matching snapshots are all deleted, and the cleanup work associated with the request has completed.
+     * If you make several requests to the delete-snapshots API targetting overlapping collections of snapshots then some of those requests may perform different parts of the associated cleanup work, returning their responses at different times.
```
ywangd: I wonder whether we should provide an explicit example to say the 2nd request may not perform the cleanup work and hence returns before the 1st one?

ywangd (approved):
LGTM

specification/snapshot/delete/SnapshotDeleteRequest.ts
DaveCTurner: Yes good idea, see c3c605c37

PR #6618, opened 2026-09-02
Add `CREATED` recovery stage to recovery API

inespot:
Follows: https://github.com/elastic/elasticsearch/pull/158086, which introduced a new initial `CREATED` stage to `RecoveryState`. Adds `CREATED` as the first value in the `RecoveryStage` enum. The `active_only` parameter docs for both `_recovery` and `_cat/recovery` are updated to make clear that only completed recoveries (`DONE`) are excluded, so queued recoveries are included alongside in-flight ones. Time fields are also documented for this new stage.

Closes: elastic/elasticsearch-team#5041

### DIFF
--- a/docs/examples/package-lock.json
+++ b/docs/examples/package-lock.json
@@ -69,9 +69,9 @@
       }
     },
     "node_modules/@elastic/request-converter-dotnet": {
-      "version": "9.4.2",
-      "resolved": "https://registry.npmjs.org/@elastic/request-converter-dotnet/-/request-converter-dotnet-9.4.2.tgz",
-      "integrity": "sha512-52Wk5ng/GT1AQNXOC3fm2/bs4R8s0Dn949HzNJbkUUOSmmrLZobRsAEm/f2CESe0Ggb6/g+WImUcE3zu9coW3g==",
+      "version": "0.0.0-main.g9784f393693d",
+      "resolved": "https://registry.npmjs.org/@elastic/request-converter-dotnet/-/request-converter-dotnet-0.0.0-main.g9784f393693d.tgz",
+      "integrity": "sha512-mUrAw9DT98KdCFtxUh5ESTH+X600YunzIfUBTWVSVho8JfGrL5QAhTYBuIkciL9SxDBWCYT3v3mRdo9z83gk/g==",
       "license": "Apache-2.0"
     },
     "node_modules/@emotion/is-prop-valid": {
--- a/docs/examples/package-lock.json
+++ b/docs/examples/package-lock.json
@@ -69,9 +69,9 @@
       }
     },
     "node_modules/@elastic/request-converter-dotnet": {
-      "version": "0.0.0-main.g9784f393693d",
-      "resolved": "https://registry.npmjs.org/@elastic/request-converter-dotnet/-/request-converter-dotnet-0.0.0-main.g9784f393693d.tgz",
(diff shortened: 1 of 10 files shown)


docs/examples/package-lock.json
PeteGillinElastic, line 1: This change should probably be reverted. I don't think it belongs in this PR.

output/openapi/elasticsearch-openapi.json
PeteGillinElastic, line 1: All the changes under `output` should be reverted, I think. The latest advice I heard was that we _don't_ include them in our PRs, we just include the `specification` changes and let the automation that runs periodically generate the `output`. The reason is that it reduces the chances of merge conflicts.

specification/cat/recovery/types.ts
```diff
@@ -34,26 +34,31 @@ export class RecoveryRecord {
...
   'start_time'?: DateTime
   /**
    * The recovery start time in epoch milliseconds.
+   * For recoveries in the `created` stage (not yet started), this value is 0.
    * @aliases start_millis
    */
   'start_time_millis'?: EpochTime<UnitMillis>
   /**
    * The recovery stop time.
+   * Only present for completed recoveries (`done` stage).
```
PeteGillinElastic: I don't think this is true, is it? The `_cat` API has tabular output and doesn't support optional fields. So I think this would report the epoch, just like `start_time`.

specification/cat/recovery/types.ts
```diff
@@ -34,26 +34,31 @@ export class RecoveryRecord {
...
    * @aliases stop
    */
   'stop_time'?: DateTime
   /**
    * The recovery stop time in epoch milliseconds.
+   * Only present for completed recoveries (`done` stage).
```
PeteGillinElastic: Same as above (except it's 0 of course).

package-lock.json
PeteGillinElastic, line 1: Revert, I think.

package.json
PeteGillinElastic, line 1: Revert, I think.

inespot: I was not sure if this should be included. I sort of imitated this one: https://github.com/elastic/elasticsearch-specification/pull/6553 and left the autogenerated file but happy to revert them if this is not correct

specification/cat/recovery/types.ts
inespot: Ah, you're right, thanks for catching! Will remove this

output/openapi/elasticsearch-openapi.json
inespot: Oki, thanks for the explanation! Will revert

specification/cat/recovery/types.ts
inespot: Removed in [3c38abe](https://github.com/elastic/elasticsearch-specification/pull/6618/commits/3c38abe18ce5781acfe2db0c2e6f4fd12b6a14b2)

specification/cat/recovery/types.ts
inespot: Removed in [3c38abe](https://github.com/elastic/elasticsearch-specification/pull/6618/commits/3c38abe18ce5781acfe2db0c2e6f4fd12b6a14b2)

package.json
PeteGillinElastic: Yeah, I think it was a mistake to include those in my PR 🙂. (I sort of thought I hadn't done...)

PeteGillinElastic (approved):
LGTM.

PR #6620, opened 2026-09-03
Fix global privilege type in RoleDescriptor

jfreden:
The server parser ([ConfigurableClusterPrivileges.parse](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ConfigurableClusterPrivileges.java#L133)) unconditionally expects a `START_OBJECT` token, so passing an array throws an `XContentParseException`. Remove the erroneous array variant.


l-trotta (approved):
LGTM, thanks!

PR #6630, opened 2026-09-11
User managed service accounts

ebarlas:
Mirrors the API surface added in elastic/elasticsearch#157818, which introduces service accounts that a caller creates, alongside the built-in accounts that ship with Elasticsearch.

Adds two endpoints on `/_security/service/{namespace}/{service}`, both authorized by `manage_security` rather than `manage_service_account`: a `PUT` that creates or wholly replaces an account from a list of role names and an enabled flag, and a `DELETE` with a `force` parameter deciding whether an account that still has service tokens may be removed.

Changes the get service accounts response from a bare role descriptor wrapper to `ServiceAccountInfo`, an internally tagged union on a new required type field. Built-in accounts keep their role descriptor; user-managed accounts report role names and whether they are enabled. This renames the generated `RoleDescriptorWrapper` type. A `type` query parameter filters the listing by kind; omitting it returns built-in accounts only for a whole-cluster listing, preserving that response, and both kinds for a request scoped to a namespace.

Documents the privilege split on the create and delete service token routes, which serve both kinds of account from one URL but dispatch to separate transport actions so that `manage_service_account` no longer reaches tokens of a user-managed account.


l-trotta:
Thanks @ebarlas! Couple of things:

- The `@availability` of `PutUserManagedServiceAccountRequest` has `stability=stable`, but the [json rest api spec](https://github.com/elastic/elasticsearch/blob/0f47181d6fb204617b5bb87434f9593ed78c9374/rest-api-spec/src/main/resources/rest-api-spec/api/security.put_user_managed_service_account.json#L7-L8) still shows experimental and private. Which one is correct? 
- I'd rewrite the comment above `ServiceAccountInfo`, we usually don't include in the docs information about the meta model (like "this is modeled as a union") .
- The validation failure is due to a pre existing bug, if it's okay with you I'll fix it after this PR is merged and ask you for a review :)

Everything else LGTM

ebarlas:
Thanks, @l-trotta!

> The `@availability` of `PutUserManagedServiceAccountRequest` has `stability=stable`, but the [json rest api spec](https://github.com/elastic/elasticsearch/blob/0f47181d6fb204617b5bb87434f9593ed78c9374/rest-api-spec/src/main/resources/rest-api-spec/api/security.put_user_managed_service_account.json#L7-L8) still shows experimental and private. Which one is correct?

This `elastic/elasticsearch-specification` PR is correct. I wasn't sure about the ordering. I was planning to update the ES repo with adjusted values along with the `documentation.url` after this PR lands, since I believe that value is required for `stable`/`public`. How is this typically done?

> I'd rewrite the comment above `ServiceAccountInfo`, we usually don't include in the docs information about the meta model (like "this is modeled as a union") .

Applied in https://github.com/elastic/elasticsearch-specification/pull/6630/changes/ea94127093cfc4bce381a4a635b41b6a33241087

> The validation failure is due to a pre existing bug, if it's okay with you I'll fix it after this PR is merged and ask you for a review :)

I noticed that as well. And yes, I would be happy to.

l-trotta:
> I wasn't sure about the ordering
nevermind then just wanted to be sure :)

l-trotta (approved):
LGTM

PR #6635, opened 2026-09-16
Clarify updating PIT ids with concurrent requests

DaveCTurner:
Today the docs say that you must always use the latest PIT id for
subsequent requests, but this only makes sense for sequential requests.
If making requests concurrently, you can receive several different PIT
ids in their responses, with no clear ordering relationship allowing you
to identify the unique latest value in all cases. This commit clarifies
that you can safely pick any of them in that situation: there is no need
to implement any strongly-consistent coordination of PIT ids in the
client application layer.


l-trotta (approved):

cbuescher (approved):
LGTM

DaveCTurner:
Thanks both!

PR #6642, opened 2026-09-18
[DOCS] Note rest_spec_name implications on CLI command

lcawl:
Relates to https://github.com/elastic/kibana/pull/284350

This PR adds some tips equivalent to what @florent-leborgne added in https://www.elastic.co/docs/extend/kibana/tutorials/generating-oas-for-http-apis#set-an-operation-id

@l-trotta if there are more considerations that ought to be covered re `endpoint.name` or you prefer a different location for this info, lmk!


l-trotta (dismissed):

docs/modeling-guide.md
```diff
@@ -854,3 +854,12 @@ There are a few benefits of using JSDoc instead of a simple TODO comment:
 
  * With the allowlist of tags in our eslint config, we can't make typos
  * We enforce a proper description to explain why the breaking change needs to be made
+
+### `@rest_spec_name` implications
+
+The TypeScript compiler sets `endpoint.name` from `@rest_spec_name`.
+The `endpoint.name` is used to generate an `operationID` in the OpenAPI documents (with `_` and `.` replaced by `-`).
+
+If the same endpoint has more than one path/method, later operations get a suffix (for example, `cancel-reindex-1`, and so on).
```
l-trotta: should we use an actual existing example, so that it's searchable in the openapi output? like bulk with  bulk-1, bulk-2, and bulk-3.

l-trotta (changes requested):

docs/modeling-guide.md
lcawl: That's funny, I swear I had that in an earlier draft. Thanks for catching!
```suggestion
If the same endpoint has more than one path/method, later operations get a suffix (for example, `bulk-1`, `bulk-2`, and so on).
```

lcawl: Fixed in cf96dfd5fa8c99a022083dcaa143d0bc95fb7f7f

l-trotta (approved):
thank you :)

PR #6644, opened 2026-09-18
[Inference API] Add docs for non-streaming chat completion

jonathan-buttner:
This PR adds the docs for non-streaming chat completion.

This corresponds to the Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/158022


specification/inference/stream_chat_completion_unified/examples/request/PostChatCompletionRequestExample1.yaml
```diff
@@ -0,0 +1,16 @@
+summary: A chat completion task
```
jonathan-buttner, line 1: `specification/inference/stream_chat_completion_unified` was basically a git mv from the old directory `chat_completion_unified`. I changed a bit of wording in the API description. So most of these files aren't "new".

specification/inference/stream_chat_completion_unified/UnifiedRequest.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Id, MediaType } from '@_types/common'
+import { Duration } from '@_types/Time'
+import { RequestChatCompletion } from '@inference/_types/CommonTypes'
+/**
+ * Perform streaming chat completion inference on the service.
+ *
+ * The chat completion inference API enables real-time responses for chat completion tasks by delivering answers incrementally, reducing response times during computation.
+ * It only works with the `chat_completion` task type.
+ *
+ * NOTE: The `chat_completion` task type supports both streaming and non-streaming.
```
jonathan-buttner, line 30: I changed some of the wording in this comment block

specification/inference/chat_completion_unified/examples/request/PostChatCompletionRequestExample1.yaml
```diff
@@ -1,8 +1,8 @@
-summary: A chat completion task
+summary: A non-streaming chat completion task
```
jonathan-buttner, line 1: Treat everything in `specification/inference/chat_completion_unified` as new files

specification/inference/chat_completion_unified/UnifiedRequest.ts
```diff
@@ -22,24 +22,23 @@ import { Id, MediaType } from '@_types/common'
...
- * NOTE: The `chat_completion` task type is only available within the _stream API and only supports streaming.
- * The Chat completion inference API and the Stream inference API differ in their response structure and capabilities.
+ * NOTE: The `chat_completion` task type supports both streaming and non-streaming.
  * The Chat completion inference API provides more comprehensive customization options through more fields and function calling support.
  * To determine whether a given inference service supports this task type, please see the page for that service.
  * @rest_spec_name inference.chat_completion_unified
- * @availability stack since=8.18.0 stability=stable visibility=public
- * @availability serverless stability=stable visibility=public
- * @doc_id inference-api-chat-completion
+ * @availability stack since=9.6.0 stability=experimental visibility=public
```
jonathan-buttner, line 34: I assume using `experimental` is probably good idea initially? If we have issues with timeouts it'd be good to iron that out before marking the route as `stable`.

specification/inference/non_streaming_chat_completion/UnifiedRequest.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Perform non-streaming chat completion inference on the service.
+ *
+ * The chat completion inference API enables rich responses for chat completion tasks.
+ * It only works with the `chat_completion` task type.
+ *
+ * NOTE: The `chat_completion` task type supports both streaming and non-streaming.
+ * The Chat completion inference API provides more comprehensive customization options through more fields and function calling support.
+ * To determine whether a given inference service supports this task type, please see the page for that service.
+ * @rest_spec_name inference.non_streaming_chat_completion
```
jonathan-buttner, line 49: My understanding is that the value here results in the function exposed in the clients. Ideally `chat_completion` would be what we used here and we'd drop the `non_streaming` part. The problem is that we already use `inference.chat_completion_unified` for the `streaming` one here: https://github.com/elastic/elasticsearch-specification/blob/main/specification/inference/chat_completion_unified/UnifiedRequest.ts#L34

@l-trotta correct me if I'm wrong but we can't change the `inference.chat_completion_unified` without having a breaking change right? Ideally that one would be called `inference.streaming_chat_completion` and we'd call this one `inference.chat_completion`.

specification/inference/non_streaming_chat_completion/UnifiedRequest.ts
```diff
@@ -0,0 +1,78 @@
...
+ * - IBMWatsonX
+ * - Llama
+ * - Mistral
+ * - NVIDIA
+ * - OpenAI
+ * - OpenShiftAI
+ * 
+ * @rest_spec_name inference.non_streaming_chat_completion
+ * @availability stack since=9.6.0 stability=experimental visibility=public
+ * @availability serverless stability=experimental visibility=public
```
jonathan-buttner, line 51: I assume using `experimental` is probably good idea initially? If we have issues with timeouts it'd be good to iron that out before marking the route as `stable`.

specification/inference/_types/Results.ts
```diff
@@ -128,6 +129,173 @@ export class DeleteInferenceEndpointResult extends AcknowledgedResponseBase {
...
+   */
+  id?: string
+  /**
+   * The function that the model called.
+   */
+  function?: ChatCompletionToolCallFunction
+  /**
+   * The type of the tool call. May be `null` when the provider did not return one.
+   */
+  type?: string
```
l-trotta: if it's null, but always returned then you can use:

```suggestion
  type: string | null
```

specification/inference/non_streaming_chat_completion/UnifiedRequest.ts
l-trotta: Yes this would be a breaking change for clients, CLI, docs... I'd avoid that for a stable API from 8.x

specification/inference/non_streaming_chat_completion/UnifiedRequest.ts
l-trotta: Usually we match what the server code declares, is there nothing there?

l-trotta: not the server code sorry, the rest-api-spec JSON

specification/inference/_types/Results.ts
jonathan-buttner: Hmm, actually I think this should be `type: string`. I'll fix this 👍

specification/inference/non_streaming_chat_completion/UnifiedRequest.ts
jonathan-buttner: I chatted with our team and we decided to go ahead and make it GA. That should match the rest-api-spec json too.

PR #6646, opened 2026-09-21
Add max_consecutive_extraction_failures to ML datafeed specs

edsavage:
## Summary

Documents the new optional `max_consecutive_extraction_failures` datafeed setting introduced in elastic/elasticsearch#159554. The setting makes a datafeed stop itself cleanly (leaving the associated job open) after too many *consecutive* real-time extraction failures. Requested by @valeriy42 in his review of that PR ("remember to update the specs in elasticsearch-specification").

The field is added to the shared `DatafeedConfig` type and to the `put_datafeed` / `update_datafeed` request and response bodies, mirroring the existing `max_empty_searches`:

- `specification/ml/_types/Datafeed.ts` (`DatafeedConfig`)
- `specification/ml/put_datafeed/MlPutDatafeedRequest.ts` / `MlPutDatafeedResponse.ts`
- `specification/ml/update_datafeed/MlUpdateDatafeedRequest.ts` / `MlUpdateDatafeedResponse.ts`

Behaviour (per the ES PR): unset ⇒ on by default with a threshold proportional to `frequency` (~one day of searches, floored at 1); positive `N` ⇒ stop after `N`; `-1` ⇒ disabled (retry indefinitely); `0` or `< -1` ⇒ rejected.

Regenerated `output/schema`, `output/typescript`, and `output/openapi` on Node 24.

## ⚠️ Note for reviewers: `output/schema/schema-serverless.json`

I trimmed the `schema-serverless.json` diff to **only** the 5 new field entries (+58 lines). A full `make filter-for-serverless` regeneration would otherwise sweep in **~2,400 lines of unrelated drift**, because this file is chronically stale, for a fixable reason:

The `filter-for-serverless` Makefile target's `--output` path was broken in #4426 (Jun 2025):

```diff
- --output ../output/schema/schema-serverless.json
+ --output ../output/output/openapi/elasticsearch-serverless-openapi.json
```

Consequences I verified locally:
- The auto-regen CI (`generate.yml`) runs this broken target, so the bot **never** updates `schema-serverless.json` (0 of the last 20 "Update specification output" bot commits touched it).
- Since it was last correctly regenerated (#6611, Sep 1), **10 commits** changed `schema.json` but **0** propagated to `schema-serverless.json`.
- A correct regen therefore dumps all that accumulated drift at once (cf. #6611, which swept ~156k lines for the same reason).

[...]

PR #6611, opened 2026-08-31
Add project_routing to ML datafeed and transform API specifications

valeriy42:
## Summary

Document optional `project_routing` on ML datafeed shared types (`Datafeed`, `DatafeedConfig`) and on `put_datafeed` / `update_datafeed` request and response bodies. Elasticsearch already accepts and returns this field for serverless CPS datafeeds (elastic/elasticsearch#141828); this is API contract catch-up for OpenAPI and generated clients.

Transform `project_routing` and `cat.transforms` `project_routing` column (`@aliases pr, projectRouting`) are already covered on upstream/main; this branch merges main and keeps the ML datafeed additions.

Datafeed docstrings use public cross-project search scope language (no internal terminology).

Regenerated `output/schema`, `output/typescript`, and `output/openapi` artifacts after merging upstream/main.

PR #4426, opened 2025-05-29
[OpenAPI] Add a CLI to configure export parameters

swallez:
The work on the API reference docs requires a specialized version of the OpenAPI export (e.g. merging operations in PR #4415). This PR adds CLI arguments to the OpenAPI export so that we can easily produce several versions.

The native (`main.rs`) and the wasm (`compiler-wasm-lib/lib.rs`) main entry points now use the same CLI parameters. The `clap` library isn't compatible with wasm and has been replaced by `argh`, which is also smaller.

PR #4415 is dependent on this one to add the "merge operations" CLI flag.

PR #6647, opened 2026-09-21
Fix filter-for-serverless output path

edsavage:
## Summary

The `filter-for-serverless` Makefile target's `--output` path was changed to a wrong value in #4426 (Jun 2025):

```diff
- --output ../output/schema/schema-serverless.json
+ --output ../output/output/openapi/elasticsearch-serverless-openapi.json
```

Since then, `make filter-for-serverless` (and therefore the auto-regen CI in `.github/workflows/generate.yml`) has not regenerated `output/schema/schema-serverless.json` at all — it writes to a stray `output/output/...` path instead. Verified symptoms:

- 0 of the last 20 "Update specification output" bot commits touched `schema-serverless.json`.
- The file only changes when a contributor regenerates it by hand, at which point it dumps all the accumulated drift at once (e.g. #6611 swept ~156k lines; #6646 would have swept ~2.4k).

This restores the correct output path so the serverless schema is regenerated in place by the normal tooling and CI.

## Follow-up

Once this merges, the next push to `main` will let the auto-regen workflow bring `schema-serverless.json` back in sync with `schema.json` in a single bot "Update specification output" commit, clearing the backlog of drift.

## Test plan

- [x] `make filter-for-serverless` writes to `output/schema/schema-serverless.json` (confirmed the equivalent `filter-by-availability` invocation regenerates that file)


l-trotta (approved):
nice catch, thanks!

